PR #23, opened 2022-01-26
Add link to lobsters discussion.

pbsds:
Not has high profile as hackernews, but definitely a worthy read.


max-sixty:
Thanks @pbsds

Issue #26, opened 2022-01-27
Starting on the parser


max-sixty:
A couple of people reached out asking whether they could help writing the parser, and potentially there are others who would be interested in contributing.

My plan was to use [nom](https://github.com/Geal/nom), because I'd heard good things, had a good first experience, and it's easy to adapt it if we decide to make it more hand-written or use a PEG grammar.

My initial plan was to try and get some of the initial simple examples parsing. This will be mostly activation energy — it's not the most difficult language to parse. So if we can parse this into an AST, that would be a great success:

```elm
from employees
filter country = "USA"                           # Each line transforms the previous result.
derive [                                         # This adds columns / variables.
  gross_salary: salary + payroll_tax,
  gross_cost:   gross_salary + benefits_cost     # Variables can use other variables.
]           
filter gross_cost > 0
aggregate by:[title, country] [                  # `by` are the columns to group by.
    average salary,                              # These are aggregation calcs run on each group.
    sum     salary,
    average gross_salary,
    sum     gross_salary,
    average gross_cost,
    sum_gross_cost: sum gross_cost,
    count,
]
sort sum_gross_cost
filter count > 200
take 20
```

If anyone wants to have a go, I would suggest:
- Take one function (e.g. `derive`)
- Put a PR in after you have a few lines of code — then people won't do the same thing (though I'm not sure if that will actually be a problem!)
- Don't worry about making it complete (e.g. it'd be fine to start on `derive` only for anonymous columns), add comments where stuff is missing
- Do add a few tests. You can see the very early functions I added have tests

If parts are unclear, please feel free to reply here.

iafisher:
I'm going to take a stab at parsing `aggregate` clauses if that's okay.

mormahr:
I've just written a high level language parser (for FormCalc, including later proprietary extensions) with nom and had performance issues. If my understanding is right, nom is focused on parsing flat formats like network protocols. Implementing deeply recursive grammars with operator precedence doesn't seem to be nom's speciality. I think it's because if a parse is attempted and then discarded because it didn't match, all the work is lost and individual sub parts aren't memoized. Maybe some of this can be worked around, but I decided to switch to an LR(1) parser (I used [LALRPOP](https://github.com/lalrpop/lalrpop)) and optimized lexer ([Logos](https://github.com/maciejhirsz/logos)). The learning curve was a little higher, but the code actually got cleaner and more maintainable. Performance is very good.

[...]

PR #30, opened 2022-01-27
Parse basic `filter` clauses (#26)

iafisher:


max-sixty:
Excellent start, thanks @iafisher !

I'll tag #26 here

PR #36, opened 2022-01-28
Parse basic `aggregate` clauses (#26)

iafisher:
The current implementation of `parse_list` can only handle lists of
single words, so that's what the test case has although it is not
syntactically valid PRQL.

### DIFF
--- a/src/parser.rs
+++ b/src/parser.rs
@@ -205,6 +205,11 @@ pub fn parse_aggregate(input: &str) -> IResult<&str, Transformation> {
 #[test]
 fn test_parse_aggregate() {
     assert_eq!(
+        // TODO: The current implementation of parse_list can only handle lists of
+        // single words, so that's what the test case has although it is not
+        // syntactically valid PRQL.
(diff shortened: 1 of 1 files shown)


src/parser.rs
```diff
@@ -184,3 +184,34 @@ fn test_parse_filter() {
...
+            by: by,
+            calcs: calcs,
+        }),
+    ))
+}
+
+#[test]
+fn test_parse_aggregate() {
+    assert_eq!(
+        parse_aggregate("aggregate by:[title, country] [average, sum]"),
```
max-sixty, line 213:
```suggestion
        // TODO: The current implementation of parse_list can only handle lists of
        // single words, so that's what the test case has although it is not
        // syntactically valid PRQL.
        // TODO: allow for `by` as an optional arg, in either position (either specifically in `aggregate` or a
        // more general parsing function)
        parse_aggregate("aggregate by:[title, country] [average, sum]"),
```

max-sixty:
Great! Let's leave comments in the code so it's obvious what's unfinished, so we can be more permissive about merging partial progress. 

Thanks @iafisher !

PR #46, opened 2022-01-31
Fully parse first example

max-sixty:


max-sixty:
I can try and make the test output terser... 
image

PR #52, opened 2022-02-03
Set up structure of parsing into our own structs

max-sixty:
Not the prettiest code, and still more structs to parse, but I think
it's a reasonable structure.

If anyone has a clearer view of we should do (at this stage,
taking into account the tradeoffs of speed vs elegance), I'd very much
welcome the feedback...


max-sixty:
(even if the diff is +700, -850 LOC, most of that is test outputs — the actual parser is currently 100 LOC)

PR #63, opened 2022-02-07
Parse simple functions

iafisher:


max-sixty:
As discussed on discord, this is great, thank you @iafisher .

PR #78, opened 2022-02-12
Functions to transform AST to CTEs

max-sixty:
Not finished (though may merge a partially working version)


max-sixty:
This is not perfect code, but it's in a good enough state to merge. The next steps somewhat depend on https://github.com/max-sixty/prql/issues/97

PR #85, opened 2022-02-13
feat: join clauses

shota-kizawa:
ref: #26 

Implemented the join clause.
I will implement tests later.

### DIFF
--- a/src/parser.rs
+++ b/src/parser.rs
@@ -423,6 +423,26 @@ take 20
         select a"#,
             Rule::COMMENT
         ));
+        assert_debug_snapshot!(parse_to_pest_tree(
+            r"join country [id=employee_id]",
+            Rule::transformation
+        ));
+        assert_debug_snapshot!(parse_to_pest_tree(
+            r"inner_join country [id=employee_id]",
+            Rule::transformation
+        ));
+        assert_debug_snapshot!(parse_to_pest_tree(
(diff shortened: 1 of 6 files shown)


max-sixty:
Thanks @kiibo382 ! Great to have you as a contributor!

I would be fine to merge this (with a couple of quick tests, would be great, let me know if you need a hand — we're using https://github.com/mitsuhiko/insta so it should be very easy once you're familiar with it).

I'm less confident on the semantics of joins than other parts of the language, and obviously it's a really important part of SQL. There's been some discussion on discord. I would suggest we merge something and then keep iterating. One immediate choice is whether we have:


```elm
employees
right_join payroll
```

or


```elm
employees
join side:right payroll
```
...probably defaulting to `inner`. The advantage of this is it lets us add more types later, like equi / semi / anti joins, without increasing the number of keywords.

I'd be fine with either at the moment — your call. (But I hope it's OK to say that we may change it later based on consensus)

max-sixty:
Ah, you already added tests, faster than I could comment :) 

I will merge this and we can iterate further at your discretion

shota-kizawa:
Thank you!

> I'm less confident on the semantics of joins than other parts of the language, and obviously it's a really important part of SQL. There's been some discussion on discord. I would suggest we merge something and then keep iterating. One immediate choice is whether we have:

Let's discuss it in discord!

PR #86, opened 2022-02-13
feat: table clauses

shota-kizawa:
#26 

I implemented table clause.
Please review.


src/parser.rs
```diff
@@ -72,6 +73,7 @@ impl From<&str> for TransformationType {
             "aggregate" => TransformationType::Aggregate,
             "sort" => TransformationType::Sort,
             "take" => TransformationType::Take,
+            "table" => TransformationType::Table,
```
max-sixty: I think we should implement `table` more like how we implement `func` (https://github.com/max-sixty/prql/pull/63) — it can't be part of an existing pipeline — it's defining a new entity.

So it would be a new variant of `Item` like `func` here: https://github.com/max-sixty/prql/pull/63/files#diff-4a04259da480a6b794a2e947e4cc03eff4d1aa9330836f5b91cac68c5398193fR40

Do you agree @kiibo382 ?

max-sixty (changes requested):

src/parser.rs
```diff
@@ -424,6 +426,11 @@ take 20
             Rule::COMMENT
         ));
         assert_debug_snapshot!(parse_to_pest_tree(
+            r#"table newest_employees = (
```
max-sixty: This is a good test!

src/parser.rs
shota-kizawa: Oh,,, sorry,, OK!

I wil fix it.

max-sixty:
This looks good!

I think we could convert the body to the existing `Pipeline` type (an alias of vec of transformations) rather than making a new one. That means looping over the objects after the first one (the name) and converting each to a transformation.

Does that make sense? Let me know if not and I can clarify.

src/parser.rs
```diff
@@ -178,6 +185,22 @@ pub fn parse(pairs: Pairs<Rule>) -> Result<Items, Error<Rule>> {
...
+                Rule::table => {
+                    let mut items = parse(pair.into_inner())?.into_iter();
+                    let name = if let Item::Ident(ident) = items.next().unwrap() {
+                        ident
+                    } else {
+                        unreachable!()
+                    };
+
+                    let body = if let Item::Pipeline(pipline) = items.next().unwrap() {
+                        pipline
```
max-sixty:
```suggestion
                        pipeline
```

shota-kizawa:
> I think we could convert the body to the existing Pipeline type (an alias of vec of transformations) rather than making a new one. That means looping over the objects after the first one (the name) and converting each to a transformation.

@max-sixty 
I think this is the best way to convert the type. (Maybe I'm not understanding what you said well enough.)
If there is another way, please let me know at the code level. 
Or please change it. (Maybe it would be faster.)

shota-kizawa:
Perhaps you mean that it would be better to use the code below?

```
Rule::table => {
      let mut items = parse(pair.into_inner())?.into_iter();
      let name = if let Item::Ident(ident) = items.next().unwrap() {
          ident
      } else {
          unreachable!()
      };

      Item::Table(Table {
          name,
          body: items.collect(),
      })
  }
```

src/prql.pest
```diff
@@ -21,6 +21,8 @@ transformation = { WHITESPACE* ~ ident ~ expr }
 
 function = { "func" ~ idents ~ "=" ~ items }
 
+table = { "table" ~ ident ~ "=" ~ "(" ~ NEWLINE* ~ pipeline ~ NEWLINE* ~ ")" }
```
max-sixty, line 24: Would this work? (the current version is fine otherwise)
```suggestion
table = { "table" ~ ident ~ "=" ~  inline_pipeline }
```

src/parser.rs
```diff
@@ -178,6 +185,22 @@ pub fn parse(pairs: Pairs<Rule>) -> Result<Items, Error<Rule>> {
                         body,
                     })
                 }
+                Rule::table => {
+                    let mut items = parse(pair.into_inner())?.into_iter();
```
max-sixty, line 189: This is fine, but a small rust-point: I think we can use something like https://github.com/max-sixty/prql/blob/906b951ab1be39e6fda9df69bf163d700c216674/src/parser.rs#L126 to make it more concise, WDYT?

But it's also OK to merge and we can come back to refine the rust if you prefer

max-sixty:
> > I think we could convert the body to the existing Pipeline type (an alias of vec of transformations) rather than making a new one. That means looping over the objects after the first one (the name) and converting each to a transformation.
> 
> @max-sixty I think this is the best way to convert the type. (Maybe I'm not understanding what you said well enough.) If there is another way, please let me know at the code level. Or please change it. (Maybe it would be faster.)

@kiibo382 my apologies — I think you made these changes by the time I submitted my review! So we were thinking the same thing!

max-sixty (approved):

src/parser.rs
```diff
@@ -66,6 +67,12 @@ pub struct Function {
     pub body: Items,
 }
 
+#[derive(Debug, PartialEq, Clone, Serialize, Deserialize)]
+pub struct Table {
+    pub name: Ident,
+    pub body: Pipeline,
```
max-sixty: (possibly `pipeline` would be more consistent, but your call)

src/prql.pest
shota-kizawa: If you use `inline_pipline`, it will stop working when a new line is inserted.
So, I decided to write it this way. (There may be a better way to write it, though.)

src/parser.rs
shota-kizawa: Thanks! OK!

src/.parser.rs.pending-snap
```diff
@@ -0,0 +1,43 @@
+{"run_id":"1644797319-508675000","line":389,"new":{"module_name":"prql__parser__test","snapshot_name":"parse_table","metadata":{"source":"src/parser.rs","assertion_line":389,"expression":"parse(parse_to_pest_tree(r#\"table newest_employees = ( from employees )\"#,\n                         Rule::table).unwrap()).unwrap()"},"snapshot":"---\n- Table:\n    name: newest_employees\n    body:\n      - Pipeline:\n          - name: From\n            args:\n              - Ident: employees\n            named_args: []\n"},"old":{"module_name":"prql__parser__test","metadata":{},"snapshot":"---\n- Table:\n    name: newest_employees\n    body:\n      - name: From\n        args:\n          - Ident: employees\n        named_args: []"}}
```
max-sixty: I think this is an unapproved snapshot file

src/prql.pest
max-sixty: Ah OK. I am working on parentheses anyway, so let's keep this like you have it and we can iterate on it later. Thanks

max-sixty:
Looks good — feel free to merge if you can remove the snapshot file. Thanks @kiibo382 !

src/parser.rs
```diff
@@ -178,6 +185,19 @@ pub fn parse(pairs: Pairs<Rule>) -> Result<Items, Error<Rule>> {
                         body,
                     })
                 }
+                Rule::table => {
+                    let mut items = parse(pair.into_inner())?.into_iter();
+                    let name = if let Item::Ident(ident) = items.next().unwrap() {
+                        ident
+                    } else {
+                        panic!("Expected Table, got {:?}", items)
```
max-sixty: I think these should be unreachable — it shouldn't be possible to produce a `table` where the first item isn't an `ident`, or the second isn't a `pipeline`.

Though the practical difference isn't that high, please feel free to merge!

src/parser.rs
```diff
@@ -178,6 +185,19 @@ pub fn parse(pairs: Pairs<Rule>) -> Result<Items, Error<Rule>> {
...
+                    };
+                    if let Item::Pipeline(pipline) = items.next().unwrap() {
```
max-sixty:
```suggestion
                    if let Item::Pipeline(pipeline) = items.next().unwrap() {
```

:)

shota-kizawa:
Thanks so much for the thoughtful review!

src/.parser.rs.pending-snap
shota-kizawa: Thank you!

src/parser.rs
shota-kizawa: OK! Thanks!

src/parser.rs
shota-kizawa: Thanks!

PR #92, opened 2022-02-14
Refine some of the rust in parser.rs

max-sixty:


max-sixty:
@kiibo382 @iafisher I hope it's OK to refine the rust in the `parse` function a bit, as folks are now using existing code as a blueprint for new code. 

TBC, getting stuff to work is hard, making the rust nice is easy, so your contributions are predominantly all of the value... 

Lmk any objections!

shota-kizawa:
@max-sixty 
I agree!
I will also fix any code that can be improved!

PR #100, opened 2022-02-18
Refactor Transformation, move AST to ast.rs

max-sixty:
Some of this rust is not great, and there is still lots to do on the
`writer.rs`. But it makes the rs-sqlparser branch much more tractable.


max-sixty:
This also made a small change to the initial example; I realize that this was using an implicit name: https://github.com/max-sixty/prql/pull/100/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R75

I think it could be fine, but better to be initially explicit & strict and then we can relax it later.

Issue #97, opened 2022-02-16
Using sqlparser-rs for SQL AST


max-sixty:
We're getting to the point where we're converting PRQL AST to SQL AST, and then we'll need to convert the SQL AST to text.

[sqlparser-rs](https://github.com/sqlparser-rs/sqlparser-rs) is predominantly for parsing SQL — i.e. going the other direction.

But it has:
- A very complete AST: https://github.com/sqlparser-rs/sqlparser-rs/tree/main/src/ast
- It seems to have a complete `Display` implemented for this AST, which may do the writing for us; e.g. https://github.com/sqlparser-rs/sqlparser-rs/blob/main/src/ast/query.rs#L152-L195

It looks excellent quality, something I think we'd love to build on.

This would be "huge if true" — so adding this issue to socialize the idea. I'll optimistically tag @alamb as they've been the main contributor to the repo very recently, in case they have a view.

alamb:
👋  hi @max-sixty  -- I (biasedly) think sqlparser is high quality and fairly stable and likely to incrementally add support for new SQL features from the community

I think many projects use sqlparser successfully, both directly like [datafusion](https://github.com/apache/arrow-datafusion) as well as fork sqlparser for their own use internally and occasionally contribute patches back upstream. The license is the liberal Apache license.

The maintainers only have limited time to work on the project however, so changes in the core repo will likely remain on rounding out the SQL parsing support

max-sixty:
Thanks @alamb ! Can I ask a follow-up — we are trying to go the _other direction_ — i.e. construct SQL queries from an AST (rather than parsing SQL) — do you think sqlparser is a good fit for that? Does running `Display` for a sqlparser AST make a SQL query? It looks like it does, which would be awesome...

alamb:
> Thanks @alamb ! Can I ask a follow-up — we are trying to go the other direction — i.e. construct SQL queries from an AST (rather than parsing SQL) — do you think sqlparser is a good fit for that? 

Yes I think it is a good fit

> Does running Display for a sqlparser AST make a SQL query? It looks like it does, which would be awesome...

As long as the AST is an actual query then yes :) In fact many of the tests in sqlparser verify that parsing a sql string, and then `Display`ing it produces the original SQL string

max-sixty:
That's great news! Thanks!

PR #103, opened 2022-02-20
Implement initial conversion to SQL

max-sixty:
Lots of this doesn't produce correct SQL, and some of the rust is not pretty. 

But it mostly works, and there's lots of space for building on the framework.


max-sixty:
Closes #97

PR #123, opened 2022-02-24
fix: use nightly rust toolchain

kwigley:
I was unable to compile using stable channel of rustc, this uses a toolchain file to configure what toolchain to use for the project. No worries if this is not how you'd like to configure this! Calling this out in the README probably achieves the same thing.

Failed to compile with the following:
```sh
❯ rustc --version
rustc 1.58.1 (db9d1b20b 2022-01-20)
```
successfully compiled with the following:
```sh
❯ rustc --version
rustc 1.61.0-nightly (532d3cda9 2022-02-23)
```


max-sixty:
Thanks — I think beta should work, which is less likely to break. I'll add a commit to your PR if that's OK

max-sixty:
Thanks @kwigley !

max-sixty:
Ironically it now supports stable given today's 1.59 release. But let's leave on beta, at least for a while.

Issue #104, opened 2022-02-20
Add CLI


max-sixty:
It would be good to have a CLI which can take PRQL and do things like:
- Produce SQL, using `writer.rs` (most important, it'd be fine if this is all it does tbh)
- Produce an AST, using `parser.rs`
- Produce a parse tree, using `ast.rs`

This will provide the basis of the tool for producing SQL, and provide a simple entry point for people to see intermediate data.

If anyone wants to take this on, please feel free to start with a very small PR.

danjones1618:
There probably should be a `--flavour` flag which allows you to specify which DBMS (Postgress, MariaDB, etc) you're targeting. This would supercede the flavour value given in a PRQL file.

Something to consider is wether we want to be able to automatically query a DBMS. This means there would be two ways to interact with PRQL
1. `prql --flavour psql --host localhost --user user --password strongpass query.prql`
2. `prql -o query.sql query.prql`

Where the second workflow is specifying a file to output the generated SQL file to.

Another option may be to print the generated SQL to `stdout` if no `-o` or `--host` is set.
This would be the easiest route for v1 if the other options were put on the roadmap as this would allow piping into the `sql` program.

I'm not sure "flavour" is the right name for the flag and I'm open to suggestions.

max-sixty:
Yes, I think we can start with something fairly basic that gives an interface to what's currently available — fine to use stdin / stdout, with the option of supplying file names too.

I agree we could add dialects / flavors (and even direct querying) later.

Let me know if you want to take a swing at this @danjones1618

qharlie:
I think direct querying could be a game changer!  That might just be the killer app for PRQL.

PR #128, opened 2022-02-25
feat: adding initial cli and organizing package into binary and library

kwigley:
A starting place for #104.

- takes a file or stdin (denoted by `-`) as input
- writes to output file or stdout (default and denoted by `-`)
- format option to determine the output format (just added as an example cli option! this could probably used to set what intermediate representation to output)


kwigley:
happy to merge as is and anyone can actually implement something useful or wait til I get around to it 😎

src/parser.rs
```diff
@@ -112,12 +115,13 @@ pub fn parse(pairs: Pairs<Rule>) -> Result<Items> {
...
                 }),
                 Rule::operator | Rule::number => Item::Raw(pair.as_str().to_owned()),
-                _ => (Item::TODO(pair.as_str().to_owned())),
+                _ => (Item::Todo(pair.as_str().to_owned())),
             })
         })
         .collect()
 }
 
+#[cfg(test)]
```
max-sixty, line 124: How come these are marked for tests? Is it because they're not reachable from the CLI? And so we remove them as we make things reachable from there?

src/cli.rs
```diff
@@ -0,0 +1,153 @@
...
+        match self {
+            Output::Stdout(_) => write!(fmt, "-"),
+            Output::Pipe(path, _) => write!(fmt, "{:?}", path),
+            Output::File(path, _) => write!(fmt, "{:?}", path),
+        }
+    }
+}
+
+#[derive(Copy, Clone, PartialEq, Eq, PartialOrd, Ord, ArgEnum)]
+enum Format {
```
max-sixty:
```suggestion
enum Dialect {
```
WDYT?

src/cli.rs
```diff
@@ -0,0 +1,153 @@
+use anyhow::Error;
```
max-sixty, line 3: Very nice file!

src/ast.rs
```diff
@@ -29,7 +29,7 @@ pub enum Item {
     Table(Table),
     SString(Vec<SStringItem>),
     // Anything not yet implemented.
-    TODO(String),
+    Todo(String),
```
max-sixty, line 32: Thanks for catching this

max-sixty (approved):
This looks super! Thanks for starting it @kwigley !

max-sixty:
> happy to merge as is and anyone can actually implement something useful or wait til I get around to it 😎

Yes always feel free to merge things like this, it doesn't have to be complete to make progress. If you're confident, you don't need to wait for someone to approve — we can easily revert and retry if there's disagreement

danjones1618:
> takes ~a file or~ stdin (denoted by -) as input

Could we automatically detect stdin is a pipe as apposed to indicating it?
I guess this depend on if the cli lib requires positional arguments to be non-optional?

src/parser.rs
kwigley: Ya, right on! I'm sure there are a million ways to hush the lint warning, but this seemed reasonable. The compile will force you to remove these configuration conditions once an entrypoint uses these functions. I think it will help tease out a reasonable API.

kwigley:
> > takes ~a file or~ stdin (denoted by -) as input
> 
> Could we automatically detect stdin is a pipe as apposed to indicating it? I guess this depend on if the cli lib requires positional arguments to be non-optional?
 
Yes, I think we can use a default arg to achieve this. I'll try this out!

max-sixty:
Thanks @kwigley ! This is a great framework to build from.

If anyone wants to add commands to return a parse tree (i.e. `parse_to_pest_tree`) or an AST (i.e. `parse`), those would be great to have. 

Having actual SQL will be most helpful eventually — we could use the `to_select` function, but it's still fairly nascent.

PR #142, opened 2022-02-28
expose prql as lib + integration test

kwigley:
…yaml

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -19,4 +19,4 @@ repos:
         files: \.md$
 ci:
   # Currently network access isn't supported
-  skip: [fmt, check, clippy, markdown-link-check]
+  skip: [fmt, clippy, markdown-link-check]
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -11,7 +11,6 @@ repos:
     rev: v1.0.1
     hooks:
       - id: fmt
-      - id: check
       - id: clippy
   - repo: https://github.com/tcort/markdown-link-check
     rev: v3.9.3
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -1,12 +1,12 @@
-use std::io::{Read, Write};
-
+use crate::parser::{ast_of_string, Rule};
 use anyhow::Error;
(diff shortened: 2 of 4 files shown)


.pre-commit-config.yaml
```diff
@@ -11,7 +11,6 @@ repos:
     rev: v1.0.1
     hooks:
       - id: fmt
-      - id: check
       - id: clippy
```
kwigley, line 14: clippy runs `check` under the hood 👍

tests/test_parser.rs
```diff
@@ -0,0 +1,15 @@
+use prql::{ast_of_string, Result, Rule};
```
kwigley, line 1: test prql as a lib

src/cli.rs
```diff
@@ -27,11 +27,9 @@ impl Cli {
         let mut source = String::new();
         self.input.read_to_string(&mut source)?;
         match self.format {
-            Dialect::Sql => writeln!(
-                self.output,
-                "TODO! do something useful with source:\n{}",
-                &source
-            )?,
+            Dialect::Ast => self
```
max-sixty, line 30: Oh clever, I had been thinking of it as a different subcommand but this is better!

tests/test_parser.rs
```diff
@@ -0,0 +1,15 @@
...
+
+#[test]
+fn parse_simple_string_to_ast() -> Result<()> {
+    assert_eq!(
+        serde_yaml::to_string(&ast_of_string("select 1", Rule::query)?)?,
```
max-sixty, line 6: Very nice!

FYI if we use `insta`, we get the value added in "for free" (but no need to change)

max-sixty (approved):
Great, thanks @kwigley !

max-sixty:
Very cool!

```
$ echo "from employees
    select x
    sort y" |
      ./target/debug/prql
---
Query:
  - Pipeline:
      - From:
          - Ident: employees
      - Select:
          - Ident: x
      - Sort:
          - Ident: y
```

The SQL output got _worse_ this weekend given I've been changing how functions run, so I'll finish that off this week and then we'll have this running for SQL, at least for some limited set of queries — the first two examples is my goal

PR #145, opened 2022-03-03
Run functions with arguments

max-sixty:
Also includes lots of changes around `Items`/ `Expr`. I'm not sure
the code is yet in a great place — feels like a lot of whack-a-mole
once a test fails. I'll continue iterating on it.

### DIFF
--- a/src/ast.rs
+++ b/src/ast.rs
@@ -122,38 +122,71 @@ impl Item {
     /// Either provide a Vec with the contents of List / Item, or puts a scalar
     /// into a Vec. This is useful when we either have a scalar or a list, and
     /// want to only have to handle a single type.
-    pub fn into_items(self) -> Vec<Item> {
+    pub fn into_inner_items(self) -> Vec<Item> {
         match self {
-            Item::List(items) | Item::Items(items) | Item::Expr(items) => items,
+            Item::Items(items) | Item::Expr(items) => items,
             _ => vec![self],
         }
     }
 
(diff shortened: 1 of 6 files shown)


src/ast.rs
```diff
@@ -122,38 +122,71 @@ impl Item {
...
+            Item::Items(items) | Item::Expr(items) => items,
             _ => vec![self],
         }
     }
 
-    /// The scalar version of `into_items`. It's recursive, so will return the
-    /// lowest possible single item.
-    pub fn into_item(self) -> Item {
-        self.as_item().clone()
+    /// Wrap in an Items unless it already holds multiple items.
```
max-sixty:
```suggestion
    /// Wrap in Items unless it's already an Items.
```

src/parser.rs
```diff
@@ -167,14 +170,15 @@ impl TryFrom<Vec<Item>> for Transformation {
...
             "filter" => Ok(Transformation::Filter(Filter(args))),
             "derive" => {
-                let assigns = args
+                let assigns = (args)
                     .into_only()
                     .context("Expected at least one argument")?
-                    .into_unnested()
-                    .as_inner_items()?
-                    .iter()
+                    // Possibly these two is be an `unnest_list` method?
```
max-sixty:
```suggestion
                    // Possibly these two should be an `unnest_list` method?
```

max-sixty:
Some `Expr` / `List` stuff to clean up, but the functions work!

PR #146, opened 2022-03-03
Use EnumAsInner, rather than defining our own methods

max-sixty:


src/parser.rs
```diff
@@ -87,14 +91,17 @@ fn ast_of_parse_tree(pairs: Pairs<Rule>) -> Result<Items> {
...
+                    let [name, pipeline]: [Item; 2] = parsed
+                        .try_into()
+                        .map_err(|e| anyhow!("Expected two items; {:?}", e))?;
+                    Item::Table(Table {
+                        name: name
+                            .into_ident()
+                            .map_err(|e| anyhow!("Expected Ident; {:?}", e))?,
+                        pipeline: pipeline
+                            .into_pipeline()
+                            .map_err(|e| anyhow!("Expected Pipeline; {:?}", e))?,
```
max-sixty, line 103: @kwigley do you know whether the TODO you added around deriving our own Error type would make these terser? There are lots of them now... (I'm less familiar with these "newer" error-handling crates in rust...)

max-sixty: I think this is possibly a different issue: https://github.com/bluejekyll/enum-as-inner/issues/84

PR #170, opened 2022-03-09
Add Query type

max-sixty:
Start refactoring to_sql, to support Tables


max-sixty:
Unfortunately this creates large diffs in the snapshots because of the additional indentation

PR #178, opened 2022-03-12
Single quoted strings , for literals and s-strings

qharlie:
Added these new rules:

single_quoted_string ( matches string )
single_quoted_string_literal( matches string_literal )
single_quoted_s_string ( matches s_string ) 
single_quoted_s_string_string ( matches s_string_string ) 

And tests:
test_parse_single_quoted_string

However I cannot figure how to duplicate test_parse_s_string for single quoted strings, I'm still learning how to test in Rust , and can't figure out what its saying about new snapshots.

Please review I'm new to Rust.


max-sixty:
@qorrect thanks a lot, this is a great first rust PR!

I think we could parse both single and double quoted strings to the same parse tree. That would be a fairly small change from here — we could make a silent rule for each type of quoting, and then a standard rule which takes either.

Silent rules start with `_`, and to take either we do `{ single_quoted | double_quoted }`.

WDYT? As ever, I'm very open to whatever you decide is best, so feel free to not do this!

max-sixty:
(on the test failure, just run `rust fmt`!)

qharlie:
OK!  Agreed, I would rather not have all those extra rules but was struggling with the PEST syntax, let me take another stab at it.

max-sixty:
Yes, I think the easiest way is just to make a bunch more rules — have a `double_quoted_string_literal‘, ‘double_quoted_string`, `single_quoted_string_literal‘, etc. It won't be the tersest approach, but we'll have a working and tested version and can compress it later if we want.

max-sixty:
A terser approach would be to define a quote as `quote = { "\"" | "'" }` and use that in a single set of rules. If that works, that's probably the best approach (but it's fine to merge something that works in and then iterate later)

max-sixty:
I made the final changes as discussed with @qorrect on Discord. Thanks @qorrect !

max-sixty:
Sorry, actually I moved too fast and my approach parses `"hello'` as a string. I will redo.

Issue #181, opened 2022-03-13
Add subcommand to CLI


max-sixty:
Currently the `prql` binary has no subcommands — it always writes SQL (or AST depending on the arguments). This limits us from doing other things in the future, like querying a DB, or opening a console like PyPrql does.

Let's add a subcommand like `write-sql`, even if it's initially the only one at first. (Open to other names, though having all subcommands be verbs is nice for consistency).

PR #206, opened 2022-03-19
Add subcommand to cli

aljazerzen:
implements #181 

I've named the command `compile` with the reasoning that this is a common verb used in software. Also, other transpilers  such as Typescript use `compile`. It means "compile to SQL".

### DIFF
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -1,6 +1,6 @@
 use crate::*;
 use anyhow::Error;
-use clap::{ArgEnum, Parser};
+use clap::{ArgEnum, Args, Parser};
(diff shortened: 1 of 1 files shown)


max-sixty:
@aljazerzen thank you very much!

Do you prefer `compile` to `transpile`? Probably compile is more common? We could also have something like `to-sql`.

Merging, and we can change later depending on consensus.

aljazerzen:
Yea, i prefer compile to transpile. Its shorter and more common.

I tried to add alias "c" and "to-sql" but couldnt find a simple way with clap.

PR #207, opened 2022-03-19
Join parse and translate (but not materialize)

aljazerzen:
So, I've took a stab at `join`s :grin: 

I don't know if what I did is in line with code style and architecture of the project... I've simplified function `atomic_pipelines_of_pipeline`, take a look if you like it better.

I did not add any tests (because I don't know how these snapshots work), but it does work with my test cases. I will post them into original issue.

### DIFF
--- a/src/ast.rs
+++ b/src/ast.rs
@@ -75,13 +75,17 @@ pub enum Transformation {
     },
     Sort(Items),
     Take(i64),
-    Join(Items),
+    Join {
+        side: JoinSide,
+        with: Ident,
+        on: Vec<Item>,
+    },
     Func(FuncCall),
 }
 
 impl Transformation {
     /// Returns the name of the transformation.
-    pub fn name(&self) -> &str {
+    pub fn name(&self) -> &'static str {
         match self {
             Transformation::From(_) => "from",
             Transformation::Select(_) => "select",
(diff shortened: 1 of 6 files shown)


max-sixty:
@aljazerzen  this is extremely impressive! I hadn't expected someone to navigate the existing code as well as this.

The snapshot tests are actually very convenient; `cargo install insta` and then `cargo insta test` will cause snapshots to be created; either in a separate file, or inline if there is a `@""` as the second argument to the `assert_snapshot` macro. Do you want me to add some as an example?

(Also if you prefer a different form of tests then that's fine)

Thanks!

aljazerzen:
Yea, sure add the tests. You will probably find some other cases that don't work as expected.

max-sixty:
Great, test added! Let's merge now and we can keep iterating.

I really appreciate this — it's the "deepest" contribution to the project so far. If you have any feedback about the code you built on, please let me know; I saw you already made some small improvements to its quality.

For the test, I just added:

```rust
#[test]
fn transpile_joins() -> Result<()> {

    assert_snapshot!(transpile(r#"
    from employees
join side:left salaries [salaries.emp_no = employees.emp_no]
aggregate by:[employees.emp_no] [
  emp_salary: average salary
]
join side:left titles [titles.emp_no = emp_no]
aggregate by:[dept_emp.dept_no, titles.title] [
  avg_salary: average emp_salary
]
join side:left departments [departments.dept_no = dept_no]
select [dept_name, title, avg_salary]
"#)?, @"");
```

...and then ran `cargo insta test --accept`, and it fills it in. I sometimes write more explicit tests, but at this stage it's quite useful for fast iteration, particularly where we expect the results to change soon.

PR #208, opened 2022-03-19
Fix how `aggregate` impacts SQL `select`

max-sixty:
h/t @aljazerzen for spotting the bug, in (1) in https://github.com/max-sixty/prql/issues/194


src/translator.rs
```diff
@@ -162,6 +162,8 @@ fn select_columns_of_pipeline(pipeline: &Pipeline) -> Result<Vec<SelectItem>> {
                     .collect::<Result<Vec<SelectItem>>>()?,
             ),
             Transformation::Aggregate { calcs, assigns, .. } => {
+                is_inclusive = false;
+                selects.clear();
```
max-sixty, line 166: This the semantic change

PR #212, opened 2022-03-20
Materialize named args

max-sixty:


max-sixty:
Closes https://github.com/max-sixty/prql/issues/197

PR #218, opened 2022-03-20
Fix scoping of variables following `aggregate`

### LINKED ISSUE #213: Scope variables within splits
We're currently getting `AVG(AVG(salary)) AS avg_salary,` in https://github.com/max-sixty/prql/issues/194:

```sql
WITH table_0 AS (
  SELECT
    AVG(salary) AS emp_salary,
    *
  FROM
    employees
    LEFT JOIN salaries ON salaries.emp_no = employees.emp_no
  GROUP BY
    employees.emp_no
),
table_1 AS (
  SELECT
    AVG(AVG(salary)) AS avg_salary,        # <- whoops
    *
  FROM
    table_0
```

Unfortunately this may require some rethinking of the structure — currently `materialize` materializes all variables & functions, without knowing anything about the splits. But here, because we have a new split / CTE, we _can_ use the previous variable name.

Possibly we can have a rule of retaining the name in `aggregate` clauses — it'll work in the following CTEs, and it'll work in `having` & `order by` clauses (at least in most DBs)...

max-sixty:
This ended up being much simpler than I expected. I'm not sure if there
are corner cases — it relies on SQL respecting variables after any
`GROUP BY` — I'm fairly confident this holds for `HAVING` & `ORDER BY`
clauses. But possibly I'm missing something.

Closes #213

### DIFF
--- a/src/materializer.rs
+++ b/src/materializer.rs
@@ -64,6 +64,19 @@ impl AstFold for ReplaceVariables {
         }
         .into_unnested())
     }
+    // Once we get to an Aggregate, we want to run the replacement, but then we
+    // want to remove the variable, because SQL can support it from then on. If
+    // we don't do this, we get errors like `AVG(AVG(x))` in later CTEs; see #213.
+    fn fold_transformation(&mut self, transformation: Transformation) -> Result<Transformation> {
+        let out = fold_transformation(self, transformation.clone());
+
+        if let Transformation::Aggregate { assigns, .. } = transformation {
+            assigns.iter().for_each(|assign| {
+                self.variables.remove(&assign.lvalue);
+            });
+        }
+        out
+    }
 }
 
 #[derive(Debug)]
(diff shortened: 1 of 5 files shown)


max-sixty:
CC @aljazerzen

aljazerzen:
I thinks there is still an issue:

```elm
from dept_manager
aggregate by:emp_no [
  count: count *
]
derive [
  mng_no: dept_manager.emp_no
]
join employees [employees.emp_no = mng_no]
```

produces:

```SQL

WITH table_0 AS (
  SELECT
    COUNT(*) AS count,
    dept_manager.emp_no AS mng_no
  FROM
    dept_manager
  GROUP BY
    emp_no
),
table_1 AS (
  SELECT
    *
  FROM
    table_0
    JOIN employees ON employees.emp_no = dept_manager.emp_no
)
SELECT
  *
FROM
  table_1
```

JOIN employees should use SQL alias `mng_no` instead of materializing into `dept_manager.emp_no`.

I believe that this requires that we know the scope of each alias during materialization and the scope is bound to its atomic pipeline. Which means that splitting the pipeline into atomic pipelines must be done during materialization.

Please tell me that I'm wrong :cry:

aljazerzen:
Wait, why are we even materializing the aliases? Could always we just left the `mng_no` unresolved?

PR #227, opened 2022-03-21
fix: readme

shota-kizawa:
Fixed a query that could not be parsed.


max-sixty:
Whoops! Thanks a lot for spotting @kiibo382

PR #236, opened 2022-03-21
add fold pattern for `table`

### LINKED ISSUE #235: `pipeline` in `table` is not correctly translated to sql.
For example, this query
```
table average_salaries = (
  from salaries
  aggregate by:country [
    average_country_salary: average salary
  ]
)
```
is translated as follows.
```
WITH average_salaries AS (
  SELECT
    average salary AS average_country_salary
  FROM
    salaries
  GROUP BY
    country
)
```

However, it should be translated as follows.
```
WITH average_salaries AS (
  SELECT
    avg(salary) as average_country_salary
  FROM
    salaries
  GROUP BY
    country
)
```

shota-kizawa:
close: #235

Add fold pattern for `table` and unit test.

### DIFF
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -1,4 +1,4 @@
-use crate::*;
+use crate::{compile, materialize, parse};
 use anyhow::Error;
 use clap::{ArgEnum, Args, Parser};
 use clio::{Input, Output};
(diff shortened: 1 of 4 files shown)


src/cli.rs
```diff
@@ -1,4 +1,4 @@
-use crate::*;
+use crate::{compile, materialize, parse};
```
shota-kizawa, line 1: In addition, I refactor it.

max-sixty (approved):
Thanks @kiibo382 !

PR #246, opened 2022-03-22
Aggregate also selects by columns

aljazerzen:
Addition to #208

If you do:

```
from salaries
aggregate by:[emp_no] [
  emp_salary: average salary
]
```

You get:
```
SELECT
  AVG(salary) AS emp_salary
FROM
  salaries
GROUP BY
  emp_no
```

Result does not contain `emp_no` in select, which means that you cannot do subsequent:

```
```

You can add additional `select [emp_no]`, but I would argue that the desired result in 99% of cases is to include all `by` columns in the select. Only `emp_salary` does not have much use (except for distributions of the salary) if you don't know who it belongs to. 

In the case where you don't want that you can still use `select [only_columns_that_I_want]`.

### DIFF
--- a/examples/employees/0_average-title-salary.prql
+++ b/examples/employees/0_average-title-salary.prql
@@ -1,0 +1,23 @@
+# Task: rank the employee titles according to the average salary for each department.
+
+# My solution:
+# - for each employee, find their average salary,
+# - join employees with their departments and titles (duplicating employees for each of their titles and departments)
+# - group by department and title, aggregating average salary
+# - join with department to get department name
+
+from employees
+join side:left salaries [salaries.emp_no = employees.emp_no]
+aggregate by:[employees.emp_no] [
+  emp_salary: average salary
+]
+join side:left titles [titles.emp_no = table_0.emp_no]
+join dept_emp [dept_emp.emp_no = table_0.emp_no]
+aggregate by:[dept_emp.dept_no, titles.title] [
+  avg_salary: average emp_salary
+]
+join side:left departments [departments.dept_no = table_1.dept_no]
+select [dept_name, title, avg_salary]
+
+# Now writing about this I realize that I would have needed to select from employees table at all.
+# I could have only aggregated salaries table by emp_no. Oh well.
--- a/examples/employees/1_gender-by-departments.prql
+++ b/examples/employees/1_gender-by-departments.prql
@@ -1,0 +1,15 @@
+# Task: The company active pursues gender equality.
+# Prepare an analysis based on salaries and gender distribution by departments.
+
+from employees
+join salaries [salaries.emp_no = employees.emp_no]
+aggregate by:[employees.emp_no, gender] [
+  emp_salary: average salary
+]
+join dept_emp [dept_emp.emp_no = table_0.emp_no]
+aggregate by:[dept_emp.dept_no, gender] [
+  salary_avg: average emp_salary,
+  salary_sd: stddev emp_salary,
+]
+join departments [departments.dept_no = table_1.dept_no]
+select [dept_name, gender, salary_avg, salary_sd]
--- a/examples/employees/README.md
+++ b/examples/employees/README.md
@@ -1,0 +1,15 @@
+Example queries using [employees database](https://github.com/vrajmohan/pgsql-sample-data.git).
+
+Clone and init the database (requires a local PostgreSQL instance):
+ 
+    $ psql -U postgres -c 'CREATE DATABASE employees;'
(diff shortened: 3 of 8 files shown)


aljazerzen:
I've also added my PRQL files I've been testing with in `examples/employees`

max-sixty:
Your idea is excellent @aljazerzen , let's do it!

Are you OK if I make some comments on the examples? Please feel free to merge so we get the code in, and we can iterate on the examples later?

examples/employees/1_gender-by-departments.prql
```diff
@@ -0,0 +1,15 @@
+# Task: The company active pursues gender equality.
+# Prepare an analysis based on salaries and gender distribution by departments.
+
+from employees
+join salaries [salaries.emp_no = employees.emp_no]
```
max-sixty, line 5: I think this is nicer, and then the rest of the query also gets simpler — WDYT?

```suggestion
join salaries [emp_no]
```

examples/employees/1_gender-by-departments.prql
```diff
@@ -0,0 +1,15 @@
...
+aggregate by:[employees.emp_no, gender] [
```
max-sixty, line 6:
```suggestion
aggregate by:[emp_no, gender] [
```

examples/employees/1_gender-by-departments.prql
```diff
@@ -0,0 +1,15 @@
...
+  emp_salary: average salary
+]
+join dept_emp [dept_emp.emp_no = table_0.emp_no]
```
max-sixty, line 9:
```suggestion
join dept_emp [emp_no]
```

examples/employees/1_gender-by-departments.prql
```diff
@@ -0,0 +1,15 @@
...
+aggregate by:[dept_emp.dept_no, gender] [
```
max-sixty, line 10:
```suggestion
aggregate by:[dept_no, gender] [
```

examples/employees/1_gender-by-departments.prql
```diff
@@ -0,0 +1,15 @@
...
+  salary_avg: average emp_salary,
+  salary_sd: stddev emp_salary,
+]
+join departments [departments.dept_no = table_1.dept_no]
```
max-sixty, line 14:
```suggestion
join departments [dept_no]
```

examples/employees/README.md
```diff
@@ -0,0 +1,15 @@
+Example queries using [employees database](https://github.com/vrajmohan/pgsql-sample-data.git).
```
max-sixty, line 1: This is awesome! We could even turn this into a test later, ref https://github.com/max-sixty/prql/issues/219

examples/employees/0_average-title-salary.prql
```diff
@@ -0,0 +1,23 @@
...
+join side:left titles [titles.emp_no = table_0.emp_no]
+join dept_emp [dept_emp.emp_no = table_0.emp_no]
+aggregate by:[dept_emp.dept_no, titles.title] [
+  avg_salary: average emp_salary
+]
+join side:left departments [departments.dept_no = table_1.dept_no]
+select [dept_name, title, avg_salary]
+
+# Now writing about this I realize that I would have needed to select from employees table at all.
+# I could have only aggregated salaries table by emp_no. Oh well.
```
max-sixty, line 23: Should we put this one in a separate PR and iterate? I think we should try and have the examples path full of really good quality queries — the others are great — and maybe this one isn't as efficient as those?

max-sixty (approved):
Re the examples:

One issue with GitHub is that the syntax highlighting isn't good for PRQL, and so for examples — which are often something people who are completely new look at — it's important for them to be approachable.

I think we should:
- Keep these as `.prql` files, because they're "runnable", unlike the existing examples.
- Later, write a script to copy them into `.md` files and add the SQL, like in the other examples
- Later, add an intergration test to run these on postgres

examples/employees/1_gender-by-departments.prql
aljazerzen: Now that we have USING, of course :D

examples/employees/0_average-title-salary.prql
aljazerzen: Well, PRQL should also support developers writing inefficient queries. The mistake I did is not something you could detect without access to the database.

This particular mistake is a common with SQL queries, so maybe databases already optimize it out. But if they don't, we can still have this here and put up a notice on common pitfall?

PR #249, opened 2022-03-22
Aggregate also selects by columns

aljazerzen:
This is re-push of #246, I did a blunder...

### DIFF
--- a/examples/employees-0.md
+++ b/examples/employees-0.md
@@ -7,9 +7,8 @@
 # - group by department and title, aggregating average salary
 # - join with department to get department name
 
-from employees
-join side:left salaries [emp_no]
-aggregate by:[employees.emp_no] [
+from salaries
+aggregate by:[emp_no] [
   emp_salary: average salary
 ]
 join side:left titles [emp_no]
@@ -27,13 +26,12 @@ select [dept_name, title, avg_salary]
 ```sql
 WITH table_0 AS (
   SELECT
-    employees.emp_no,
+    emp_no,
     AVG(salary) AS emp_salary
   FROM
-    employees
-    LEFT JOIN salaries USING(emp_no)
+    salaries
   GROUP BY
-    employees.emp_no
+    emp_no
 ),
 table_1 AS (
   SELECT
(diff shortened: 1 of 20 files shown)


examples/prql/generate-md.sh
```diff
@@ -0,0 +1,19 @@
+#!/bin/bash
```
max-sixty, line 1: Amazing, this is great!!

examples/employees-0.md
```diff
@@ -0,0 +1,63 @@
...
+]
+join side:left titles [emp_no]
+join dept_emp [emp_no]
+aggregate by:[dept_emp.dept_no, titles.title] [
+  avg_salary: average emp_salary
+]
+join side:left departments [dept_no]
+select [dept_name, title, avg_salary]
+
+# Note that this solution contains a common pitfall in the first few lines:
```
max-sixty: I hear your point about PRQL handling imperfect queries well — I'm just not sure we want the first example to not be the most efficient way of representing the goal — does it communicate the strength of PRQL as well as a more efficient query?

But overall this PR is excellent, I don't want to be dogmatic on the small things — I'm very happy to merge and we can reflect and iterate on these in the future.

max-sixty (approved):

examples/sql-parameters.md
```diff
@@ -1,10 +1,9 @@
...
-
 ```elm
+# This doesn't yet work
+
 from mytable
 filter id = $1    # We should be able to pass parameters straight through.
 ```
 
-```sql
-SELECT * FROM mytable WHERE id = $1
```
max-sixty: If possible, it would be nice to keep our desired output. But that might be difficult with the current script — fine to merge and we can come back to this

examples/variables-1.md
```diff
@@ -20,20 +20,26 @@ filter ct > 200
...
+WHERE
+  country = 'USA'
+  and salary + payroll_tax + benefits_cost > 0
+GROUP BY
+  title,
+  country
+HAVING
+  ct > 200
+ORDER BY
+  sum_gross_cost```
```
max-sixty: The lack of a newline causes it to have the closing backticks in the visible markdown:
image

aljazerzen: Whoops

examples/sql-parameters.md
aljazerzen: Yeah, took care to make sure that the files do not get overwritten if the compilation does not succeed.

... but then forgot to restore this file.

max-sixty:
Thanks @aljazerzen !

Issue #248, opened 2022-03-22
Table aliases


aljazerzen:
So as SQL does, we can support table aliases. This will come useful when doing joins because of possible duplicate columns names.

To be consistent with derive, I would suggest:
```
from e:employees
select [e.emp_no, e.gender]
```

With joins:
```
from e:employees
join s:salaries side:left [s.emp_no = e.emp_no]
select [e.gender, s.salary]
```

Problem here is how to parse function parameters. If you don't use aliases, from and join have 1 and 2 positional parameters, respectfully. If you do, they have 0 and 1. This is not consistent with current rules around function parameters.

Any other suggestions?

Also, do you think we even need aliases?

max-sixty:
> Also, do you think we even need aliases?

I'll start with the bigger question!

I have a personal opinion, but with less confidence than other ideas. So I would be fine to go ahead with this if people would like it.

My opinion is that we should try and reduce the need for aliases. Their role is to make queries terser when we have identically named columns in multiple tables, and:
1. Having dozens of highly normalized tables is much less common since the growth of analytical, columnar DBs. It's now much more common to have fewer, wider tables. 
    - We should be making PRQL _accessible_ to everyone, but the _focus_ should be on people who are likely to use it soon, which is likely to be people who are also using other newer tools.
    - (though these can also be used for nested columns, which are more common)
3. The pipeline approach means it's easier to select a canonical representation of a column and use that for the remainder of the query. e.g adding a `derive` is easy: `from x | join y [x.id=y.id] side:left | derive id: x.id`. (But doesn't work if you actually need access to both columns later).
4. Terseness is often overrated — much of the time, it's easier to read a query when it has the actual table names than `e` and `et` and `te` etc...

---

That said, as above, if people want this, I'm fine adding it in. 

I think @aljazerzen 's syntax is great. Probably we would add a space so it's like:

```diff
-from e:employees
+from e: employees
-join s:salaries side:left [s.emp_no = e.emp_no]
+join s: salaries side:left [s.emp_no = e.emp_no]
select [e.gender, s.salary]
```

...and then it's treated as an `assign` rather than a `named_arg` (the syntax differences between these aren't ideal, but I think a reasonable tradeoff, as discussed in https://github.com/max-sixty/prql/issues/53, and still open to iterating on those)

> Problem here is how to parse function parameters. If you don't use aliases, from and join have 1 and 2 positional parameters

[...]

Issue #154, opened 2022-03-05
`count` vs `count *`


max-sixty:
In the [example functions](https://github.com/max-sixty/prql/blob/main/src/stdlib.prql#L15-L18), there are two functions with the same name:

```
func count = s"COUNT(*)"
func count column = s"COUNT({column})"
```

We need to decide how to handle these.

Options:
- Don't allow this — `count *` would be required rather than just `count`
  - This would be tempting to use in the short term, but has the disadvantage that we'd need to special-case the `*`, since it's currently an operator (i.e. `foo * 3`). I don't think it's a CS PhD to implement, but it's the sort of special-casing that causes issues later, both in the code and our ability to provide good error messages.
  - We could find a different symbol than `*`, but that's also unfamiliar.
- Multiple dispatch — i.e. the compiler allows for multiple functions with the same name as long as they have different signatures, and it selects the function that matches the signature. In this case the number of positional args defines the signature. So the function definitions above would be correct.
- Use a named argument — so `func count column="*" = s"COUNT({column})"`. But then people have to write `count column:salary`, which isn't friendly — and either we do it everywhere, or it's unfamiliar in the `count` case (there's no `sum *`).
- Allow optional positional args, and find a different way of specifying named args. As a reminder, [this is our current function specification](https://github.com/max-sixty/prql#functions), and has the advantage that it's extremely simple — only two types of arguments, with named arguments defined by having a default. [This is python's](https://docs.python.org/3/reference/compound_stmts.html#function-definitions).

qharlie:
I vote multi dispatch, based on the number of arguments.  Nothing fancy, just match the number of arguments supplied to the number defined.

max-sixty:
As discussed on discord, one downside of multiple dispatch which I forgot to mention, is it preludes partials / currying; e.g.:

```
func add a b = a + b
func plus_one = add 1
# Does `add 1` curry `add a b`? Or refer to a one-arg version of `add`?
```

(though I don't think it's that important, and definitely not in the short term)

max-sixty:
This is the last item in https://github.com/max-sixty/prql/milestone/1 and I can't think of sufficiently general solution to the problem. I was leaning to `count_all`, but it's just awkward. I'm fine sacrificing currying, but adding multiple dispatch isn't trivial and doesn't seem to serve any purpose beyond this specific case.

I think I'm going to hack the parser so `count *` is allowed, and then we can iterate from there.

max-sixty:
[...]

Issue #53, opened 2022-02-03
Named args vs assignments


max-sixty:
One issue https://github.com/max-sixty/prql/pull/52 raises is how assignments are parsed; specifically:

```elm
derive gross_salary: salary + payroll_tax,
```

Should parse into:
- func: derive
- assign:
  - lvalue: gross_salary
  - rvalue: salary + payroll_tax

But named args' syntax conflict with this. For example:

```elm
aggregate by:[title] [sum salary]
```

should parse into:
- func: aggregate
- assign:
  - lvalue: by
  - rvalue: [title]
- arg: [sum salary]

...and _not_
- func: aggregate
- assign:
  - lvalue: by
  - rvalue:[title] [sum salary]

...which is what the assign implies.

Options:
- change syntax so only one of these uses `:`
  - Though trying to avoid using `=` so we can keep the SQL (and human!) use of this
- force the assign to always take one arg, so `gross_salary: salary + payroll_tax` needs to be `gross_salary: (salary + payroll_tax)`
- restore the rules around whitespace, such that passing named args disallows whitespace (i.e. `by:[title]` and not `by: [title]`). But this makes semantics very dependent on small changes in whitespace, which could get confusing. An autoformatter isn't going to be able to deal with this

References:
- [OCaml approach](https://ocaml.org/manual/lablexamples.html)
- [F# approach](https://docs.microsoft.com/en-us/dotnet/fsharp/language-reference/parameters-and-arguments#named-arguments) (broadly similar)
- [Swift approach](https://docs.swift.org/swift-book/LanguageGuide/Functions.html) (broadly similar)

max-sixty:
Currently going with "restore the rules around whitespace, such that passing named args disallows whitespace" in https://github.com/max-sixty/prql/pull/52, though liable to change

qharlie:
I only just now realized 'by' was a named argument to aggregate.  Makes sense, I like your conclusion.

So named arguments will be supported for every function, both built in and user generated ?

max-sixty:
> So named arguments will be supported for every function, both built in and user generated ?

Yes exactly! The built-in ones are fairly limited though (e.g. `by` for `aggregate`, `order` for `sort`, etc)

aljazerzen:
[...]

PR #250, opened 2022-03-22
Table alias

aljazerzen:
Implements #248 , but uses named args instead of assigns. Does not cover JOINs

### DIFF
--- a/prql/src/ast.rs
+++ b/prql/src/ast.rs
@@ -1,13 +1,8 @@
-use super::utils::*;
-use crate::ast_fold::AstFold;
-use anyhow::{anyhow, Result};
-use itertools::Itertools;
+use anyhow::{anyhow, bail, Result};
 use serde::{Deserialize, Serialize};
 
 // Idents are generally columns
 pub type Ident = String;
-pub type Items = Vec<Item>;
-pub type Idents = Vec<Ident>;
 pub type Pipeline = Vec<Transformation>;
 
 use enum_as_inner::EnumAsInner;
@@ -18,24 +13,20 @@ pub enum Item {
     Ident(Ident),
     String(String),
     Raw(String),
-    Assign(Assign),
+    NamedExpr(NamedExpr),
     NamedArg(NamedArg),
     Query(Query),
     Pipeline(Pipeline),
     // Currently this is separate from `Pipeline`, but we could unify them at
     // some point. We'll need to relax the constraints on `Pipeline` to allow it
     // to start with a simple expression.
-    InlinePipeline(Items),
+    InlinePipeline(InlinePipeline),
     // Similar to holding an Expr, but we strongly type it so the parsing can be more strict.
     List(Vec<ListItem>),
-    // Holds "Terms", not including separators like `+`. Unnesting this (i.e.
-    // Terms([Item]) -> Item) does not change its semantics. (More detail in
-    // `prql.pest`)
-    Terms(Items),
     // Holds any Items. Unnesting _can_ change semantics.
-    Expr(Items),
-    Idents(Idents),
-    Function(Function),
+    Expr(Vec<Item>),
+    FuncDef(FuncDef),
+    FuncCall(FuncCall),
     Table(Table),
     SString(Vec<SStringItem>),
     // Anything not yet implemented.
@@ -45,14 +36,14 @@ pub enum Item {
 #[derive(Debug, PartialEq, Clone, Serialize, Deserialize)]
 pub struct Query {
     // TODO: Add dialect & prql version onto Query.
-    pub items: Items,
+    pub items: Vec<Item>,
 }
 
 #[derive(Debug, PartialEq, Clone, Serialize, Deserialize)]
-pub struct ListItem(pub Items);
+pub struct ListItem(pub NamedExpr);
 
 impl ListItem {
-    pub fn into_inner(self) -> Items {
+    pub fn into_inner(self) -> NamedExpr {
         self.0
     }
 }
@@ -64,23 +55,21 @@ impl ListItem {
 // We probably want to implement some of these as Structs rather than just
 // `Items`
 pub enum Transformation {
-    From(Ident),
-    Select(Items),
+    From(TableRef),
+    Select(Vec<NamedExpr>),
     Filter(Filter),
-    Derive(Vec<Assign>),
+    Derive(Vec<NamedExpr>),
     Aggregate {
         by: Vec<Item>,
-        calcs: Vec<Item>,
-        assigns: Vec<Assign>,
+        select: Vec<NamedExpr>,
     },
-    Sort(Items),
+    Sort(Vec<Item>),
     Take(i64),
     Join {
         side: JoinSide,
         with: Ident,
         on: Vec<Item>,
     },
-    Func(FuncCall),
 }
 
 impl Transformation {
@@ -95,29 +84,31 @@ impl Transformation {
             Transformation::Sort(_) => "sort",
             Transformation::Take(_) => "take",
             Transformation::Join { .. } => "join",
-            // Currently this is unused, since we don't encode function calls as
-            // anything more than Idents at the moment. We may want to change
-            // that in the future.
-            Transformation::Func(_) => "func",
         }
     }
 }
 
 /// Function definition.
 #[derive(Debug, PartialEq, Clone, Serialize, Deserialize)]
-pub struct Function {
+pub struct FuncDef {
     pub name: Ident,
     pub positional_params: Vec<Ident>,
     pub named_params: Vec<NamedArg>,
-    pub body: Items,
+    pub body: Box<Item>,
 }
 
 /// Function call.
 #[derive(Debug, PartialEq, Clone, Serialize, Deserialize)]
 pub struct FuncCall {
     pub name: String,
-    pub args: Items,
+    pub args: Vec<Item>,
     pub named_args: Vec<NamedArg>,
+}
+
+#[derive(Debug, PartialEq, Clone, Serialize, Deserialize)]
+pub struct InlinePipeline {
+    pub value: Box<Item>,
+    pub functions: Vec<FuncCall>,
 }
 
 #[derive(Debug, PartialEq, Clone, Serialize, Deserialize)]
@@ -136,9 +127,50 @@ pub struct NamedArg {
 }
 
 #[derive(Debug, PartialEq, Clone, Serialize, Deserialize)]
-pub struct Assign {
-    pub lvalue: Ident,
-    pub rvalue: Box<Item>,
+pub struct NamedExpr {
+    pub alias: Option<Ident>,
+    pub expr: Box<Item>,
+}
+
+impl NamedExpr {
+    pub fn unnamed(expr: Item) -> Self {
+        NamedExpr {
+            alias: None,
+            expr: Box::new(expr),
+        }
+    }
+
+    /// Almost all function parameters and list items are wrapped into NamedExpr, which
+    /// can be passed either to this function or to into_list_with_names.
+    pub fn discard_name(self) -> Result<Item> {
+        if self.alias.is_some() {
+            bail!("Cannot use alias for: {self:?}")
+        }
+        Ok(*self.expr)
+    }
+
+    pub fn map<F>(self, mut f: F) -> Self
+    where
+        F: FnMut(Item) -> Item,
+    {
+        NamedExpr {
+            alias: self.alias,
+            expr: Box::from(f(*self.expr)),
+        }
+    }
+
+    /// Often we don't care whether a List or single item is passed; e.g.
+    /// `select x` vs `select [x, y]`. This equalizes them both to a vec of
+    /// expression, including unnesting any ListItems.
+    pub fn coerce_to_named_list(self) -> Vec<NamedExpr> {
+        match self {
+            NamedExpr { alias: None, expr } => match *expr {
+                Item::List(items) => items.into_iter().map(|x| x.into_inner()).collect(),
+                x => vec![NamedExpr::unnamed(x)],
+            },
+            n_expr => vec![n_expr],
+        }
+    }
 }
 
 #[derive(Debug, PartialEq, Clone, Serialize, Deserialize)]
@@ -148,7 +180,7 @@ pub enum SStringItem {
 }
 
 #[derive(Debug, PartialEq, Clone, Serialize, Deserialize)]
-pub struct Filter(pub Items);
+pub struct Filter(pub Vec<Item>);
 
 #[derive(Debug, PartialEq, Clone, Serialize, Deserialize)]
 pub enum JoinSide {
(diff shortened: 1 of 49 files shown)


aljazerzen:
I've also split some of the tests into smaller functions, because they were failing even after I've run them once and did `insta review`. This was because the first assert in function failed and did not even run the second one.

max-sixty:
Thanks for working this through @aljazerzen .

Let me try and write down why having this as a named arg doesn't seem consistent. I'm happy to discuss more on Discord if that's better:
- Named args are generally arguments supplied to a pre-specified list of parameters
- Here we're using a _value_ supplied as a parameter — i.e. generally `e:employees` would be passing an argument of `employees` to an `e` parameter. I would almost prefer we had `employees:e` if we had to do a named arg.
- I understand that `e: employees` reads less well than `e:employees` — these should bind quite tightly, so `join e: employees side:left` doesn't read well because it's unclear the `e`&`employees` are bound. I'm very open to alternatives to try and improve this — e.g. we could have a lint / auto-format that put parentheses around them, like `join (e: employees) side:left`.
- We could also try and find a completely different syntax, either specific to aliases or generally for assignments like `derive`. (Though it's difficult to use `=` because it's valid SQL).

I hope this is reasonable and I so appreciate your contributions over the past few days. Let me know your thoughts. Thank you.

aljazerzen:
Well, I am also against this current implementation - parsing has to be changed because its inconsistent.

I had written the code before posting the issue :D But the AST changes as agnostic of the parsing so they can wait here until we decide on what syntax to use.

While I was working on this, I tried to have `employees:e`, but its just the opposite of the assign op (which is hard to distinguish from named arg, especially for newcomers. See #53 for my opinions.)

In both cases we break definition of function that "has fixed number of positional required args and pre-defined named optional args with default values". What I did to `from` is add an equivalent to python **kwargs (which is [reverse of dictionary unpacking](https://stackoverflow.com/a/1769478/2619989)). Rules there are:

- all args are named,
- args with defaults must come after args without,
- when calling, named args must come after positional
- if there is an `*args` it will contain a list of all non-named args
- if there is an `**kwargs` it will contain a dict all non-declared named args

My point here is that we non-declared named args are a thing in some languages and not such a big inconsistency.

So my opinion is to:
- combine assign and named param into one production, see #53
- extend function definition to allow passing non-declared named args. This would not make sense for user-defined functions, so it would only apply to pipeline transformations. So this would be only a change in how *we* think of functions.

aljazerzen:
Parse expression and function calls

Internal compiler changes:

1. PEST now also parses function calls, they are not constructed only in materialization 
    - `Item::Function` is now `Item::FuncDef`, similar to `Item::FuncCall`
2. PEST now also parses expressions as a tree, not just terms
    - solves #154 (and its uber-hack)
    - there is no need for `Item::Terms` or `Item:Idents`
    - there is no need for unnesting
3. materialized AST does not contain function definitions 
4. `NamedExpr` (prev. `Assign`) now wraps all expressions from `Transformation` or `List`
5. I've simplified quite a few chunks of code (`coerce_to_list`, `RunFunctions`, transformation parsing). This was possible mostly because of offloaded work to PEST.

Language changes:

- pipe allows multiple line breaks
- (conseq of 1.)  `[a b]` now parses as `List([FuncCall(name: a, args: [Ident(b)])])`,
- (conseq of 1.) functions must have at least one paramater,
- (conseq of 2.) transformations arguments cannot be direct function calls (see example below),
- (conseq of 4.) aliases for tables and derives,

Stylistic changes:

- Type aliases such as `Items` and `Terms` are confusing. It does not take much more time to type `Vec<Item>`, but it is much easier to understand for someone new to the codebase.
- I don't like `to_owned()`, because it hides what it really goes on under the hood: cloning.


### Function call changes

`filter upper country = "USA"` now parses as 
```
Transformation:
  Filter:
    - Terms:
        - Expr:
            - Ident: upper
    - Terms:
        - Ident: country
        - Raw: "="
        - String: USA
```
... which fails, because `filter` takes only one positional argument.

`filter (upper country) = "USA"` *still* parses as:
```
Transformation:
  Filter:
    - Terms:
        - Expr:
            - FuncCall:
                name: upper
                args:
                  - Expr:
                      - Ident: country
                named_args: []
        - Raw: "="
        - String: USA
```

prql/src/ast.rs
```diff
@@ -1,13 +1,8 @@
...
-use crate::ast_fold::AstFold;
-use anyhow::{anyhow, Result};
-use itertools::Itertools;
+use anyhow::{anyhow, bail, Result};
 use serde::{Deserialize, Serialize};
 
 // Idents are generally columns
 pub type Ident = String;
-pub type Items = Vec<Item>;
-pub type Idents = Vec<Ident>;
```
max-sixty, line 10: 👍

prql/src/ast.rs
```diff
@@ -95,31 +84,33 @@ impl Transformation {
...
-            // that in the future.
-            Transformation::Func(_) => "func",
         }
     }
 }
 
 /// Function definition.
 #[derive(Debug, PartialEq, Clone, Serialize, Deserialize)]
-pub struct Function {
+pub struct FuncDef {
```
max-sixty, line 93: 👍

prql/src/ast.rs
```diff
@@ -95,31 +84,33 @@ impl Transformation {
...
     pub name: Ident,
     pub positional_params: Vec<Ident>,
     pub named_params: Vec<NamedArg>,
-    pub body: Items,
+    pub body: Box<Item>,
```
max-sixty, line 97: What are your thoughts on this as `Item` rather than `Vec<Item>`? It's basically always an `Expr` — does it reduce the nesting to have a `Vec` here?

max-sixty:
This is quite the Herculean effort! I'm looking through now. Thanks a lot for covering so much ground.

prql/src/translator.rs
```diff
@@ -10,18 +10,18 @@
...
 use super::ast::*;
 use super::utils::*;
-use anyhow::{anyhow, ensure, Result};
+use anyhow::{anyhow, bail, ensure, Result};
 use itertools::Itertools;
 use sqlformat::{format, FormatOptions, QueryParams};
 use sqlparser::ast::{
-    Expr, Join, JoinConstraint, JoinOperator, ObjectName, OrderByExpr, Select, SelectItem, SetExpr,
-    TableFactor, TableWithJoins, Top,
+    self as sql_ast, Expr, Join, JoinConstraint, JoinOperator, ObjectName, OrderByExpr, Select,
```
max-sixty, line 17: 👍

prql/src/ast.rs
```diff
@@ -158,115 +190,61 @@ pub enum JoinSide {
...
-    pub fn into_items_from_maybe_list(self) -> Items {
-        self.coerce_to_list()
-            .into_inner_list_items()
-            .unwrap()
-            .into_iter()
-            .map(Item::Terms)
-            .map(|x| x.into_unnested())
-            .collect()
+    /// Item-ss.
+    pub fn coerce_to_list(self) -> Vec<Item> {
```
max-sixty: Is this the correct name? It's a bit surprising it's `to_list` but doesn't return an `Item::List`?

(I notice there's a few of these)

prql/src/ast.rs
```diff
@@ -158,115 +190,61 @@ pub enum JoinSide {
...
-                        .clone()
-                        .into_only()
-                        .unwrap_or(Item::Terms(inner_terms)),
-                )
-            }
-            _ => fold_item(self, item),
+impl IntoExpr for Vec<Item> {
+    fn into_expr(self) -> Item {
+        if self.len() == 1 {
+            self[0].clone()
```
max-sixty: This avoids the clone (though there may be easier ways?).

```suggestion
            self.into_iter().next().unwrap()
```

or (with `use crate::utils::*;`)

```suggestion
            self.into_only().unwrap()
```

Or maybe you know better than me — is this optimized away by the compiler just as well as the move?

max-sixty:
Are we sure that the `*` is being handled correctly? This was one of my concerns about adopting the `*`. It looks like this doesn't quite work:

```
echo "from employees | aggregate [c~ count *] | derive [y~ (x * 3)]" | ./target/debug/prql compile
Unknown function x
```

prql/src/ast.rs
```diff
@@ -158,115 +190,61 @@ pub enum JoinSide {
...
-impl IntoUnnested for Item {
-    /// Transitively unnest the whole tree, traversing even parents with more
-    /// than one child. This is more unnesting that `as_scalar' does. Only
-    /// removes `Terms` (not `Expr` or `List`), though it does walk all the
-    /// containers.
-    fn into_unnested(self) -> Self {
-        Unnest.fold_item(self).unwrap()
-    }
+/// Unnest Expr([x]) into x.
+pub trait IntoExpr {
```
max-sixty, line 198: At first glance, this looks like quite an improvement over my existing code...

prql/src/ast.rs
```diff
@@ -136,9 +127,50 @@ pub struct NamedArg {
...
+    pub fn unnamed(expr: Item) -> Self {
+        NamedExpr {
+            alias: None,
+            expr: Box::new(expr),
+        }
+    }
+
+    /// Almost all function parameters and list items are wrapped into NamedExpr, which
+    /// can be passed either to this function or to into_list_with_names.
+    pub fn discard_name(self) -> Result<Item> {
```
max-sixty: What were your thoughts about having everything be a `NamedExpr`, rather than having an `Item::NamedExpr` type? Both could work; I realize lots of things can have aliases, though is it a simpler construction to have _everything_ be potentially aliased?

prql/src/ast.rs
```diff
@@ -136,9 +127,50 @@ pub struct NamedArg {
...
+        if self.alias.is_some() {
+            bail!("Cannot use alias for: {self:?}")
+        }
+        Ok(*self.expr)
+    }
+
+    pub fn map<F>(self, mut f: F) -> Self
```
max-sixty: This is reasonable; FYI this is somewhat what the `AstFold` is for — letting this sort of operation run over any type (it could be called `AstMap` tbh).

max-sixty (commented):
I have to head out for a bit so I'm submitting the comments I had; a couple are quite minor. Overall this looks awesome. I wasn't so positive about `~` over `:` but if you are, then I will acquiesce :).

Obv it's easier to review a larger number of smaller changes; but I understand that sometimes you start working on something and then keep making changes..!

max-sixty:
I do think that parsing functions is probably better than leaving them as `Ident`s. 

One of the reasons I had left them as `Ident`s was that we don't know whether an `Ident` is a function with no arguments. (see note below on this)

(There's also an elegance that a "function with no arguments" had exactly the same semantics as "a variable"; I don't think that changes in this construction.)

prql/src/parser.rs
```diff
@@ -93,63 +70,89 @@ fn ast_of_parse_tree(pairs: Pairs<Rule>) -> Result<Items> {
...
+                    Item::FuncDef(FuncDef {
+                        name,
+                        positional_params,
+                        named_params,
+                        body: Box::from(body.into_only().cloned()?),
+                    })
                 }
-                Rule::function_params => Item::Expr(ast_of_parse_tree(pair.into_inner())?),
+                Rule::func_def_params => Item::Expr(ast_of_parse_tree(pair.into_inner())?),
+                Rule::func_call | Rule::func_curry  => {
```
max-sixty: Changing `Terms` into the actual `FuncCall` is such an improvement; and makes other parts (e.g. all the unnesting) much simpler. Great idea!

max-sixty:
It might not matter relative to the other benefits of this; but I'm not sure functions without args are working:

```
echo "
  func x = 3

  from a
  derive [b ~ x]
  " | ./target/debug/prql compile


SELECT
  *,
  x AS b
FROM
  a
```

---
Edit: you said this above!
> (conseq of 1.) functions must have at least one paramater,

prql/src/parser.rs
```diff
@@ -190,89 +190,78 @@ fn ast_of_parse_tree(pairs: Pairs<Rule>) -> Result<Items> {
...
+        let name = name.clone().into_ident()?;
+        match name.as_str() {
+            "from" => {
+                let named_expr = positional.into_only()
+                    .context("Expected from to have exactly one argument.\n  Hint: you can pass it as `from table_name` or `from table_name ~ my_alias`")?;
 
-        let func_call: FuncCall = [vec![name.clone()], terms].concat().try_into()?;
+                let table_ref = TableRef {
+                    name: named_expr.expr.into_ident()
+                        .map_err(|_| anyhow!("From does not support inline expression. You can only pass table name."))?,
```
max-sixty:
```suggestion
                        .map_err(|_| anyhow!("From does not support inline expression. You can only pass a table name."))?,
```

max-sixty:
After looking through this more, here's my summary of where we're at:
- Overall it's a huge improvement. Even aside from Table aliases, removing the acrobatics around `Terms` makes the code _both_ simpler and more explicit. Much code that I wrote has been replaced by better code. Thank you!
- Before we merge we should:
  - Resolve the `x * 3` issue
  - I don't quite understand why `filter upper country = "USA"` needs to fail. Could we parse the `upper country` as a function call, since it's between the transformation name `filter` and the operator `=`?
  - Add TODOs in areas that we don't address now which you agree we should follow-up on (or add an Issue for any bigger ones)
- After we merge we should:
  - Decide whether functions can have no arg. It's a bit odd that they can't. But it has little practical value; we could address the issue in the future. It's equivalent to a variable (I had hoped before that we could unify "variable" and "function call with no params" into the same thing).
  - Standardize function names — `list` should refer to a `List`; if it's a `Vec` we can call it `items`? And should structs hold `Box<Item>` and generally contain an `Expr`, or should they hold `Vec<Item>`?
  - Address the `~` vs `:` question! I would lean towards the `:`, but it seems like you prefer the `~`. If so, let's live with that for a while and see how it goes. 
  - Decide whether we should parse into `NamedExpr` first and then into `Expr` (if I'm reading the code correctly — that's the current flow for some of the parsing). I'm still thinking through whether we can instead parse in `pest` into one or the other.

I'm happy to help with these follow-ups. I've been trying to harangue wasm recently and hopefully will have an MVP quite soon.

prql/src/ast.rs
aljazerzen: On parser level it's always `expr` and never multiple expressions. Expression in now actually, an expression, something that we (or SQL) can evaluate to. So I prefer AST to have "correctly" expressed types, thus function only having one expression it evaluates to.

And also, it is not always `Expr`. I've replaced `IntoUnnested` with `IntoExpr` that converts list of items into `Expr(list)`, except when there is only one item in the list, when it converts just to `list[0]`.

https://github.com/max-sixty/prql/blob/f90b1c15b919b6ec3210e9a533b722721f9d53fe/prql/src/parser.rs#L751-L752

prql/src/ast.rs
aljazerzen: Yes, this is not the best naming. I had actual lists here, but because we always needed `Vec<_>`, i've simplified.

prql/src/ast.rs
aljazerzen: True that :+1:

prql/src/ast.rs
aljazerzen: Well, we do have `Item::NamedExpr` (that contains `struct NamedExpr`).

The functions on the `NamedExpr` are here because of typing: they are only applicable to `NamedExpr`, so it would not make sense to pollute `Item::` namespace (and `match self {  Item::NamedExpr() }` in each function).

prql/src/ast.rs
aljazerzen: Oh, yes this can go - it was here for some other reason and it's not needed any more.

aljazerzen:
> It might not matter relative to the other benefits of this; but I'm not sure functions without args are working:
> 
> ```
> echo "
>   func x = 3
> 
>   from a
>   derive [b ~ x]
>   " | ./target/debug/prql compile
> 
> 
> SELECT
>   *,
>   x AS b
> FROM
>   a
> ```
> 
> Edit: you said this above!
> 
> > (conseq of 1.) functions must have at least one paramater,

*Function definitions work, but function calls don't.

If calls would work, the result here would be 
```
SELECT
  *,
  3 AS b
FROM
  a
```

aljazerzen:
`filter upper country = "USA"` needs to fail, because
```
join employees id
```
has to fail.

We can have only one of:
- function name and its arguments separated by space,
- function & transformation arguments separated by space

So I've changed function and transformation parameters so they cannot contain direct function calls.

Unfortunately, I think this cannot be fixed without changing the syntax (or reverting to evaluating only functions that exist).

aljazerzen:
I had an idea for functions without arguments: you said that this is basically a variable. So we can process it as variable! In `RunFunctions`, we also inspect idents and replace them with function def, if found!

This way there is no need for `count *` and you can just write `count`.

aljazerzen:
> Decide whether we should parse into NamedExpr first and then into Expr (if I'm reading the code correctly — that's the current flow for some of the parsing). I'm still thinking through whether we can instead parse in pest into one or the other.

I believe yes, sometimes (like in transformation arguments) we want named parsed first. And sometimes, we even don't want it (like in parenthesis).

prql/src/ast.rs
aljazerzen: Just now I've changed it back: it is much nicer not to have everything wrapped into `NamedExpr`. It does provide some nice typing, so you have to discard possible name, but AST should not have such useless wrappers.

aljazerzen:
Now I've realized that `NamedArgs` and `NamedExpr` can be merged into one! I've dropped `~` because we can consistently parse everything with `:`.

aljazerzen:
I believe I've addressed everything, so there is no need for:

> Add TODOs in areas that we don't address now which you agree we should follow-up on (or add an Issue for any bigger ones)

Language changes (now):

- pipe allows multiple line breaks
- (conseq of 1.) [a b] now parses as List([FuncCall(name: a, args: [Ident(b)])]),
- (conseq of 1.) functions with zero parameters have their own namespace (`count` and `count first_name` can coexist)
- (conseq of 2.) transformations arguments and function arguments cannot be direct function calls (filter example),
- (conseq of 4.) aliases for tables and derives are now the same thing as named args

prql/src/materializer.rs
```diff
@@ -127,8 +134,7 @@ impl RunFunctions {
             let value = func_call
                 .named_args
                 .iter()
-                // Quite inefficient — we could instead sort each list and join
-                // them. Is there a native way of doing that?
+                // Quite inefficient when number of arguments > 10
```
max-sixty: Could we leave the note about how we could improve; or at least "we could improve the implementation with a merge join"?

aljazerzen: Sure

prql/src/ast.rs
```diff
@@ -118,49 +117,15 @@ pub struct Table {
...
-    /// expression, including unnesting any ListItems.
-    pub fn coerce_to_named_exprs(self) -> Vec<NamedExpr> {
-        match self {
-            NamedExpr { alias: None, expr } => match *expr {
-                Item::List(items) => items.into_iter().map(|x| x.into_inner()).collect(),
-                x => vec![NamedExpr::unnamed(x)],
-            },
-            n_expr => vec![n_expr],
-        }
+    pub fn unnamed(expr: Item) -> Item {
```
max-sixty: Should this be a method? e.g. `pub fn unnamed(self) -> Item {`

(possibly `into_unnamed` given it's a move, but no stress)

max-sixty: I see we also have [`discard_name`](https://github.com/max-sixty/prql/pull/250/commits/69ec7a1d023f5c448b04eb99b029402022d9c6ab#diff-03d9ab37a5c3933dabd1520ee3aacc9f093aad4272ecc9abf7027f148ddeea23R127).

Maybe we don't need this; given `expr` is pub already, and they're equivalent. (but no need to block the merge, we can come back to these later)

aljazerzen: This is a leftover relic...

prql/src/parser.rs
```diff
@@ -47,47 +47,21 @@ fn ast_of_parse_tree(pairs: Pairs<Rule>) -> Result<Vec<Item>> {
...
-                        [item] => {
-                            (None, item)
-                        }
-                        _ => bail!("Expected named_expr to have an expression and some identifier. Got {items:?}")
-                    };
-                    Item::NamedExpr(NamedExpr {
-                        alias,
-                        expr: Box::new(expr.clone()),
-                    })
+                    // this borrow could be removed, but it becomes much less readable without match
```
max-sixty, line 56: 👍  on both points. Wish this were easier (One thing Ocaml is better at...)

prql/src/parser.rs
```diff
@@ -322,22 +295,28 @@ impl TryFrom<Vec<Item>> for Transformation {
...
 
 /// Compares expected and passed function parameters
-fn require_named_args<const COUNT: usize>(
-    passed: Vec<NamedArg>,
+/// Returns positional and named parameters.
+fn unpack_arguments<const COUNT: usize>(
+    passed: Vec<Item>,
     expected: [&str; COUNT],
-) -> Result<[Option<Item>; COUNT]> {
+) -> Result<(Vec<Item>, [Option<Item>; COUNT])> {
```
max-sixty, line 302: This is clever!

prql/src/prql.pest
```diff
@@ -99,7 +99,7 @@ string_literal = ${ PUSH(quote) ~ string ~ POP }
 number = ${ ( ASCII_DIGIT | "." )+ }
 
 // These separate idents from each other — i.e. `foo bar and baz` will parse into
-// a) `foo bar`, b) `and`, c) `baz`, where `foo bar` is presumably a function call.
+// [`foo bar`, `and`, `baz`], where `foo bar` is presumably a function call.
```
max-sixty, line 102: 👍

max-sixty:
Shall we merge? The marathon is complete...

max-sixty (approved):

max-sixty:
Congrats and thank you @aljazerzen !

aljazerzen:
It's been quite a sprint and a test of how much do I remember "Introduction to compilers: week 1" :wink:

Issue #232, opened 2022-03-21
Decent documentation


max-sixty:
I realize we've been heads down working on rust, and have let the documentation fall behind, and now we're at 0.1, it would be good to have something that people who don't use rust can engage with.

For hosting, we could use `docs.rs`, or make a separate site like Insta's: https://insta.rs/docs/cli/. When we have the wasm working (not a small task), then we would need a separate site.

Ideally this can have a set of examples, some explanations of features, etc.

max-sixty:
https://github.com/max-sixty/prql/pull/252 is a good contribution here!

max-sixty:
This is now done!! 

Any gaps in the docs — please raise a specific issue.

PR #251, opened 2022-03-22
add docs.rs to readme

shota-kizawa:
ref: #232

For now, just add the docs.rs URL to the Readme.

I agree with a separate site like Insta, but think it would be better after WASM is completed.


max-sixty (approved):
Thanks @kiibo382

PR #252, opened 2022-03-23
Add language reference book

aljazerzen:
I've migrated notes from `README.md` into a [mdBook](https://github.com/rust-lang/mdBook). I think this would be most readable format for the language reference (for users), while docs.rs would contain docs for developers who know Rust.

I've also added Highlight.js custom language settings, so it looks nice :smile: 

Now, we can add GH action to [publish this to GH pages](https://github.com/rust-lang/mdBook/wiki/Automated-Deployment%3A-GitHub-Actions#github-pages-deploy).

Also, I've introduces term "Verb" for the pipeline transformations. I think it more understandable and consice than "transformations".


max-sixty:
This is really fantastic @aljazerzen ! Thank you very much. This will be a big step forward in making PRQL accessible.

I agree "transformation" is a bit awkward. I'm happy to try "verb" for a while and see how it goes — it's concise, and consistent with APL. It's a bit less obvious to say "pipeline of verbs" to people who aren't familiar. I had thought of "transforms" but maybe that's too much [nominalization](https://en.wikipedia.org/wiki/Nominalization). We could just have "function".

max-sixty (approved):

aljazerzen:
Well, pipelines transformations *are* functions, only on the level of the parser. On the level of materialization, they are not. 

It a bit clunky to say "pipeline contains verbs", but it's a tad better if you use "verbs used in pipelines".

PR #254, opened 2022-03-24
Compile for wasm

max-sixty:
Currently failing on `sqlformat`; I think that's the only dependency
remaining. `sqlformat` requires nom which requires minimal_lexer. That
has a `no_std` option, but it's not offered by `sqlformat`.


max-sixty:
The end result was much easier than the journey — unfortunately I lost a lot of time trying to get `--target=wasm32-unknown-unknown` to work, when `wasm-pack` seems to handle this. I still don't have a good mental model for how that works.

Next up will be integrating it with some javascript.

PR #259, opened 2022-03-25
Run examples as a test

max-sixty:
- This doesn't actually test that they don't change; it only writes the
  output. People running locally will see that. It would be great to
  have this as an actual output.
- It would be even better to have the script in rust.
- The design shares a lot with `insta`, which we use extensively
  already, including wanting to check the output matches, but I'm not
  sure we can make it work easily, given we want the output to be
  readable. An alternative would be to pull out the output from
  generated `insta` files into a readable format.
- This also reorganizes the integration tests a bit, along with
  @matklad's advice; whose post is linked in the source.
- `join side:left [col]` doesn't work; I'll add an issue (possibly I
  broke it when adding equi-joins?)


max-sixty:
I'll merge but and if @aljazerzen has any feedback, I can address it in a follow-up

aljazerzen:
Honestly, this feels like a hack. I did not intend `gen-examples.sh` to be in a test suite... But let's keep it for now, we can rewrite it to rust sometimes later.

aljazerzen:
Re `join side:left [id]`: I believe this to be the issue: https://github.com/max-sixty/prql/blob/f2eec9bfb9c76f195e4fdb450f015d2d1174dc19/src/translator.rs#L214

max-sixty:
ref https://github.com/max-sixty/prql/issues/260 for the `[id]` join

max-sixty:
> Honestly, this feels like a hack. I did not intend `gen-examples.sh` to be in a test suite... But let's keep it for now, we can rewrite it to rust sometimes later.

💯

PR #261, opened 2022-03-26
Split tests

aljazerzen:
I've split some of the tests into smaller functions, because they were failing even after I've run them once and did insta review. This was because the first assert in function failed and did not even run the second one.


max-sixty:
Great, thanks! This is better.

FYI my dev loop (not for everyone!) is to leave on something like `RUST_BACKTRACE=1 watchexec -c -e rs,pest,toml -- cargo +nightly insta test --accept --` and have the snapshots update automatically, since this runs & accepts snapshots on every file change. I use `git` rather than `insta` to review changes. If I want to focus on one test (e.g. while adding some `dbg!` expressions), I add restart the command after adding its name at the end.

max-sixty:
And FYI even using `cargo insta test` rather than `cargo test` will mean all `assert_snapshot`s run through

PR #273, opened 2022-03-28
Non-inner joins with idents

### LINKED ISSUE #260: Non-inner joins with idents
From https://github.com/max-sixty/prql/pull/259, currently `from employees | join salaries side:left [id]` compiles to `... left join on id`.

It's quite unlikely anyone actually wants to do `on id` — while it's valid SQL, is it ever useful? `full join on true` seems like an awkward way to express a `cross join`.

A couple of options:
- Compile to `select * from employees left join salaries on employees.id = salaries.id`. But we don't always know where the column comes from. We could have a heuristic that it's always from the `from` table? Seems like it would get confusing if someone switches those and things break.
- Raise an error

aljazerzen:
Closes #260

### DIFF
--- a/examples/employees-0.md
+++ b/examples/employees-0.md
@@ -12,8 +12,7 @@ aggregate by:[emp_no] [
   emp_salary: average salary
 ]
 join titles [emp_no]
-# TODO: add a `left` join
-join dept_emp [emp_no]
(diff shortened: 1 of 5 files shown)


max-sixty:
Excellent, thanks!

PR #275, opened 2022-03-28
Pretty error messages

aljazerzen:


aljazerzen:
Works for PEST :D

image

max-sixty:
This is indeed much better! I'm not a great fan of the Pest errors — 'expected EOI, operator, or named_expr_simple` is not easy to understand. But that's an orthogonal problem —  the `ariadne` library is a big upgrade, and we can make our own errors look good too.

The `print_error` func is great!

aljazerzen:
More errors!

image

aljazerzen:
@max-sixty I think this is ready, but I want you opinion on how to construct helpers for converting `Node` into `Item::variant`. 

Before we used `item.into_expr()?` (where returned item got converted into a simple anyhow error).

Now I've added [unwrap](https://github.com/max-sixty/prql/blob/9e0fcf359a8b79880a5ff6106014d8da408f50fb/prql/src/ast.rs#L200-L212):

```
expr.unwrap(Item::into_ident, "ident")?
```

Could we make this more compact? Could we pass a reference to enum variant to a function?

I've also added a crate for print enum variant name, so it displays as:
```
Error:
    ╭─[examples/prql/employees-0.prql:13:20]
    │
 13 │ join t:titles side:[left] [emp_no]
    ·                    ───┬───
    ·                       ╰───── expected ident, but found List
────╯
```

max-sixty:
The error messages look so good! It's already much better than SQL engines.

Thanks for implementing the `Node` change, that was fast! It works well, I think.

> Could we make this more compact? Could we pass a reference to enum variant to a function?

Unfortunately with rust, I don't think it's possible, without a macro. I think your current approach is very reasonable (I might change the name — `unwrap` is convenient because it represents a similar concept, but possibly confusing, because it represent a similar concept :) ). 

A possible approach is to replace [this impl](https://github.com/max-sixty/prql/blob/47dffff44afe03ddecf876d343675f4f9751eb8f/prql/src/ast.rs) with code in the `.unwrap`. Currently that's the code that converts from the `enum-as-inner` Result type of `Result<inner_type, Item>` into `Result<inner_type, anyhow::Error>`; so we could possibly use it to do the full conversion to `Result<T, Error>`. 

I think I have 80% of the context so please tell me if I'm misunderstanding something.

aljazerzen:
Yeah, the name is not good. I had `coerce` and `try`, which read as `coerce(Item::into_expr, "expr")`. But this is not coercing...

Which impl do you mean? I don't understand that.

I'll merge this now, and we can iterate later.

max-sixty:
> Which impl do you mean? I don't understand that.

I'm not sure the link was working; it was intending to show the `impl From<Item> for anyhow::Error` func: https://github.com/max-sixty/prql/blob/47dffff44afe03ddecf876d343675f4f9751eb8f/prql/src/ast.rs#L204-L209. This impl is only used by the `into_X` functions that `enum-as-inner` creates — should that instead have the `.unwrap` code?

> I'll merge this now, and we can iterate later.

👍

aljazerzen:
Initially I had it here, but the problem is that here, you get `Item`, but for nice errors you need `Node`, which is one level up.

That's why I've moved `unwrap` to `Node`...

Issue #247, opened 2022-03-22
Reference to the current pipeline


aljazerzen:
## Problem 

So when writing PRQL, at each line you have access to scope of columns, some of which belong to actual tables, some are aliases and some are columns of CTEs. The problem is that you cannot reference the latter!

See the example:

```
from salaries
aggregate by:[emp_no] [
  emp_salary: average salary
]
join dept_emp [dept_emp.emp_no = emp_no]
```

produces:
```
WITH table_0 AS (
  SELECT
    emp_no,
    AVG(salary) AS emp_salary
  FROM
    salaries
  GROUP BY
    emp_no
),
table_1 AS (
  SELECT
    *
  FROM
    table_0
    JOIN dept_emp ON dept_emp.emp_no = emp_no
)
SELECT
  *
FROM
  table_1
```

Executing this query in PG returns:
```
ERROR:  column reference "emp_no" is ambiguous
LINE 15:     JOIN dept_emp ON dept_emp.emp_no = emp_no
                                                ^
```

`emp_no` could reference `table_0.emp_no` or `dept_emp.emp_no`. We want to reference `table_0`, which we currently can, if we use:

    join dept_emp [dept_emp.emp_no = table_0.emp_no]

but this is an inconsistent hack, that will not work once we prefix the query with some other pipelines or aggregation.

## Proposal

Instead I propose that we introduce `$`, a table reference that points to the *last CTE*. In the query would be:

```
join dept_emp [dept_emp.emp_no = $.emp_no]
```

Alternatively, we could just use `.emp_no`, but I am against that, because it is not explicit enough.

max-sixty:
Yes, I agree the current state isn't good.

I think we should try and do as much as possible here ourselves without asking the user to think about it. 

In this specific example adding `table_0` to the side that isn't specified would work. (We could also encourage using `USING`, but that wouldn't work for non-inner joins).

Are there cases which that approach wouldn't cover?

aljazerzen:
@max-sixty  `USING` does not work for inner joins? Are you sure? #231

max-sixty:
> @max-sixty `USING` does not work for inner joins? Are you sure? #231

"non-inner joins" :)

aljazerzen:
But yes, there are still cases not covered. 

```
 table: company         
---------------------
 id            
 name


 table: user         
---------------------
 username            
 company_id           


select * from user join company using(id); 
-- ERROR: column "id" specified in USING clause does not exist in left table

[...]

Issue #177, opened 2022-03-12
Test the CLI


max-sixty:
It would be great to add tests on the CLI. These could start out very simple indeed.

aljazerzen:
Branch `cli-tests` has a tests broken such that there is a variety of different error messages. These could be converted into CLI tests.

max-sixty:
We can use https://github.com/mitsuhiko/insta-cmd, just released

Ciantic:
Maybe also package the CLI as downloadable release? I don't want to use PRQL as language binding but from CLI only.

E.g. I have "supercomplex.prql" file and want to compile it to "supercomplex.sql".

This way I can use it with C# too. Secondly I commit both (SQL and PRQL) files to my repo so that I see if the queries change. It also gives me nice off-ramp if I don't want to use PRQL anymore, I just remove the file and I still have SQL file.

aljazerzen:
@Ciantic Well, you already can do that via `cargo install prql-compiler`. There is also a homebrew recipe, but I am not sure of its stability.

max-sixty:
Update from the future, I was reminded about this by https://github.com/PRQL/prql/pull/1963#discussion_r1118058455:
- We have some CLI tests, but these only test the inner methods — they don't test the full output, which is both more complete and I think a generally better test.
- We can use https://github.com/mitsuhiko/insta-cmd to get the full test
- This would still be a nice contribution!

eitsupi:
I tried using `assert_cmd` in #2488.

Like

```rust
    fn get_targets() {
        let n_targets = Target::names().len();

        assert_cmd::Command::cargo_bin("prqlc")
            .unwrap()
            .args(["get-targets"])
            .assert()
            .success()
            .stdout(
                predicates::str::is_match(r"sql\.[a-z]+\n")
                    .unwrap()
                    .count(n_targets),
            );
    }
```

It works well, but I removed the test because we needed to run `cargo build -p prqlc` before running the test.

max-sixty:
I don't get errors around being able to run the binary, but possibly I'm doing something different (sorry I missed the PR before, as I said on Discord I've been out of action for a few days, but am back now)

Here's my current diff:

```diff
diff --git a/prql-compiler/prqlc/Cargo.toml b/prql-compiler/prqlc/Cargo.toml
index df8fa177..8b9ac1f9 100644
--- a/prql-compiler/prqlc/Cargo.toml
+++ b/prql-compiler/prqlc/Cargo.toml
[...]

PR #278, opened 2022-03-29
semantic analyzer

aljazerzen:
Input:
```
func lag_day x = s"lag_day_todo({x})"
func ret x = x / (lag_day x) - 1 + dividend_return
func count = s"COUNT(*)"
func excess x = ((x - interest_rate) / 252)
func if_valid x = s"IF(is_valid_price, {x}, NULL)"

from prices
derive [
  return_total     : if_valid (ret prices_adj),
  return_usd       : if_valid (ret prices_usd count),
  return_excess    : excess return_total,
  return_usd_excess: excess return_usd,
]
derive [
  return_usd: return_usd + 1
]
select [
  date,
  prices.sec_id,
  return_total,
  a: return_usd,
  return_excess,
  return_usd_excess,
]
```

Output:
```
    ╭─[:1:1]
    │
  2 │ func ret x = x / (lag_day x) - 1 + dividend_return
    ·                   ────┬────
    ·                       ╰────── Function at line 1
    ·
  9 │   return_total     : if_valid (ret prices_adj),
    ·                      ────────────┬────┬───────
    ·                                  ╰────────────── Function at line 5
    ·                                       │
    ·                                       ╰───────── Function at line 2
 10 │   return_usd       : if_valid (ret prices_usd count),
    ·                      ───────────────┬────┬──────┬───
    ·                                     ╰───────────────── Function at line 5
    ·                                          │      │
    ·                                          ╰──────────── Function at line 2
    ·                                                 │
    ·                                                 ╰───── Function at line 3
 11 │   return_excess    : excess return_total,
    ·                      ─────────┬───┬─────
    ·                               ╰─────────── Function at line 4
    ·                                   │
    ·                                   ╰─────── Variable at line 9
 12 │   return_usd_excess: excess return_usd,
    ·                      ────────┬───┬────
    ·                              ╰────────── Function at line 4
    ·                                  │
    ·                                  ╰────── Variable at line 10
    ·
 15 │   return_usd: return_usd + 1
    ·               ─────┬────
    ·                    ╰────── Variable at line 10
    ·
 20 │   return_total,
    ·   ──────┬─────
    ·         ╰─────── Variable at line 9
 21 │   a: return_usd,
    ·      ─────┬────
    ·           ╰────── Variable at line 15
 22 │   return_excess,
    ·   ──────┬──────
    ·         ╰──────── Variable at line 11
 23 │   return_usd_excess,
    ·   ────────┬────────
    ·           ╰────────── Variable at line 12
────╯
```


prql/src/ast.rs
```diff
@@ -6,7 +6,7 @@ use strum_macros::Display;
 use crate::error::{Error, Reason, Span};
 use crate::utils::*;
 
-// Idents are generally columns
+/// A name. Generally columns, tables, functions, variables.
```
max-sixty, line 9: 👍

prql/src/parser.rs
```diff
@@ -207,13 +207,12 @@ fn ast_of_parse_tree(pairs: Pairs<Rule>) -> Result<Vec<Node>> {
...
             };
 
-            Ok(Node {
-                item,
-                span: Span {
-                    start: span.start(),
-                    end: span.end(),
-                },
-            })
+            let mut node = Node::from(item);
```
max-sixty, line 212: (FYI there's a way to do this natively in rust: https://doc.rust-lang.org/book/ch05-01-defining-structs.html#creating-instances-from-other-instances-with-struct-update-syntax
...but exactly the same result)

prql/src/semantic/analyzer.rs
```diff
@@ -0,0 +1,276 @@
...
+                position,
+                declaration: Box::from(node.clone()),
+                name: None,
+            }
+        };
+        self.declare_variable(var_dec, node.span);
+    }
+}
+
+impl AstFold for SemanticAnalyzer {
```
max-sixty: Very cool approach

prql/src/semantic/reporting.rs
```diff
@@ -0,0 +1,84 @@
...
+    //                 .with_message(dec.to_string()),
+    //         );
+    //     }
+    // }
+
+    report.finish().print((source_id, source)).unwrap();
+}
+
+/// Traverses AST and add labels for each of the idents and function calls
+struct Labeller<'a> {
```
max-sixty: Despite being from London, I think we should probably keep the American spelling :)

```suggestion
struct Labeler<'a> {
```

prql/src/semantic/analyzer.rs
```diff
@@ -0,0 +1,276 @@
...
+    // save functions declarations
+    fn fold_nodes(&mut self, items: Vec<Node>) -> Result<Vec<Node>> {
+        // We cut out function def, so we need to run it
+        // here rather than in `fold_func_def`.
```
max-sixty: Forgive me if I'm missing something — how is it cut out? Doesn't the default method (which seems to still exist) get run?

prql/src/semantic/analyzer.rs
```diff
@@ -0,0 +1,276 @@
...
+        let mut r = Vec::with_capacity(items.len());
+
+        for item in items {
```
max-sixty: FYI we can also structure as an Iterator & `.filter_map` like below in `fold_pipeline` but same result

prql/src/semantic/analyzer.rs
```diff
@@ -0,0 +1,276 @@
...
+/// Scope within which we can reference variables, functions and tables
+/// Provides fast lookups for different names.
+#[derive(Debug, Default)]
+pub struct Scope {
+    // current table (result of last pipeline)
+    table: Vec<usize>,
+
+    // For each namespace (table), a list of its variables (columns)
+    // "" is default namespace
+    // "%" is namespace of functions without arguments
```
max-sixty:
```suggestion
    // "%" is namespace of functions without parameters
```
(arguments are in the call, params in the defn)

prql/src/semantic/analyzer.rs
```diff
@@ -0,0 +1,276 @@
...
+    namespaces: HashMap<String, HashMap<String, usize>>,
+
+    // Functions with arguments (name is duplicated, but that's not much overhead)
+    functions: HashMap<String, usize>,
+}
+
+impl Scope {
+    fn new_open() -> Self {
```
max-sixty: Should this just be `new`?

prql/src/semantic/reporting.rs
```diff
@@ -0,0 +1,84 @@
+use std::ops::Range;
+
+use anyhow::{Ok, Result};
+use ariadne::{Color, Label, Report, ReportBuilder, ReportKind, Source};
+
+use super::{Declaration, SemanticAnalyzer, VarDec};
+use crate::ast::FuncDef;
+use crate::internals::{AstFold, Node};
+
+pub fn print(analyzer: &SemanticAnalyzer, source_id: String, source: String) {
```
max-sixty: It looks like this isn't possible with the ariadne API, but generally a function that returns a `String` is easier to test and easier to build on. Atm it's difficult to use in a library.

(but obv fine for now)

max-sixty (approved):
This looks very cool indeed, and is going to help with language development a lot.

I didn't look through how the `scope` works in detail, but I can do that post-merge.

Do you think we could add a snapshot test to ensure it stays good?

(I added some quite refine-y comments, nothing that needs immediate attention)

prql/src/semantic/analyzer.rs
aljazerzen: Hmm, the code is not obvious enough...

From the function I return vector `r` (which is my standard notation for variables that will be returned). I am then pushing elements into `r`, except if they are `FuncDec`. The fold method still gets run, but the result is saved into declarations instead of the main AST

prql/src/semantic/analyzer.rs
aljazerzen: I prefer for loops instead of iterators because they seem "native" and behave better with `?`. But yeah, here itearator would be more readable.

aljazerzen: nvm, you can do:
```
items
            .into_iter()
            .map(|item| {
                Ok(match item {
                    ... => {
                        ...
                        None
                    }
                    _ => Some(self.fold_node(item)?),
                })
            })
            .filter_map(|x| x.transpose())
            .try_collect()
```
which is better. Map has to return `Result` (to allow `?`) so it cannot be `filter_map`, but then you can just transpose `Option` and `Result`.

prql/src/semantic/reporting.rs
aljazerzen: This is not intended to be used, it's only a tool for debugging. 

Ariadne also does not guarantee the layout will stay exactly the same, which means it would be hard to test.

prql/src/semantic/analyzer.rs
max-sixty: Thanks for explaining in the code, I see now.

prql/src/semantic/analyzer.rs
max-sixty: It's very clear in the code now, thanks

aljazerzen:
Latest breakthrough:
```elm
from invoice
select price_one: total_price + 1
filter price_one > 10
sort price_one
take 10
derive price_two: price_one + 1
```


```SQL
WITH table_0 AS (
  SELECT
    TOP (10) total_price + 1 AS price_one
  FROM
    invoice
  WHERE
    total_price + 1 > 10  -- not materialized
  ORDER BY
    price_one   -- not materialized
)
SELECT
  *,
  price_one + 1 AS price_two   -- price_one not materialized
FROM
  table_0
```

TODO:
- refactor translator code
- go over materializer tests
- resolver tests

examples/variables-1.md
```diff
@@ -41,7 +41,7 @@ GROUP BY
   title,
   country
 HAVING
-  ct > 200
+  COUNT(*) > 200
 ORDER BY
-  sum_gross_cost
+  SUM(salary + payroll_tax + benefits_cost)
```
max-sixty: These changes are fine — they still work. I think the original ones are a bit easier to read, because they use the variable names.

(but this is 5% of the weight against the huge improvements, and is easy to leave to another day)

prql/src/semantic/scope.rs
```diff
@@ -0,0 +1,226 @@
...
+use strum_macros::Display;
+
+use crate::ast::*;
+use crate::error::Span;
+
+/// Scope within which we can reference variables, functions and tables
+/// Provides fast lookups for different names.
+#[derive(Debug, Default, Serialize, Deserialize, Clone)]
+pub struct Context {
+    // current table columns (result of last pipeline)
```
max-sixty: Great comments! We can make these into doc strings!

```suggestion
    /// current table columns (result of last pipeline)
```

prql/src/semantic/materializer.rs
```diff
@@ -0,0 +1,586 @@
...
+use super::scope::Context;
+use super::scope::TableColumn;
+
+pub struct SelectedColumns(pub Vec<Node>);
+
+/// Replaces all resolved functions and variables with their declarations.
+pub fn materialize(
+    nodes: Vec<Node>,
+    context: Context,
+) -> Result<(Vec<Node>, Context, SelectedColumns)> {
```
max-sixty, line 22: Something I'm thinking about, which doesn't need immediate action, is "how easy it is for a new contributor to understand the code and try out simple examples at a high level"

If we can keep an "easy case" so that someone can take a PRQL string, push it through some functions, and get the result, that would be great.

This code does a lot, which is awesome. One change this makes is to add a couple of steps to the pipeline; e.g. `materialize(ast)` is now:

```
        let (res, context) = resolve(ast.nodes, None)?;
        let (mat, _, _) = materialize(res.clone(), context)?;
```

Possibly there's some work we can do with Default structs, or accepting `None` and making a default in those cases.

Again — no need to change things now, let's keep it in our mind though. There's a natural progression towards making things more complicated as we add features and people become more accustomed to the code, so it's often an area we need to push ourselves to think about.

max-sixty:
This is looking really good @aljazerzen !

examples/variables-1.md
aljazerzen: Oh, that should not be so. I see that this test was skiped and did not overwrite .md file...

prql/src/semantic/materializer.rs
aljazerzen: I did exactly that, probably just as you were writing this :smile: : 

https://github.com/max-sixty/prql/blob/semantic-analyzer/prql/src/semantic/mod.rs#L15-L25

It's not as simple as before, but it's not that hard either. 

I thought a lot how to structure these calls to be most understandable and decided this was the best I know. That's because the problem is hard: query has to be resolved and materialized at few different "cuts" in the pipeline. These cuts are mostly edges between atomic pipelines, which belong to translator logic because they are specific to SQL.

aljazerzen:
So now, with tests added, new CLI format and code refactored I think this is done.

It closes #213, starts work on #177 and sets up framework for #247, as well as [database cohesion](https://github.com/max-sixty/prql#database-cohesion).

The effect of all this new code can most be visible at [this example](https://github.com/max-sixty/prql/blob/semantic-analyzer/examples/variables-2.md) or the new CLI output format: 

*the comments are added by CLI*
```
from initial_table
select [first: name, last: last_name, gender]         # [first, last, gender]
derive full_name: first + " " + last                  # [first, last, gender, full_name]
take 23
select [last + " " + first, full: full_name, gender]  # [?, full, gender]
sort full                                             # [?, full, gender]
```



So if all is in order, @max-sixty feel free to merge.

examples/variables-1.md
max-sixty: Awesome!

prql/src/semantic/materializer.rs
max-sixty: Awesome, thank you!

I definitely agree it's a hard problem — the existing ones were easy, and so it's easy to find good abstractions. But now we're getting to the parts where the abstractions break down a bit.

One tiny point, if that's OK: function names like `process` aren't that descriptive without context; so if we can name them something a bit more distinct, that makes it easier for to discuss the code and for new contributors to find their way around. The `semantic` module is a good example of this done well. Sometimes that leads us to ugly names (`materialize` 😬)!

max-sixty:
This is a huge upgrade @aljazerzen , thank you! It's another mammoth PR — so I haven't gone through each line, but the structure and the tests look really good. I'll continue looking over the next couple of days. And hopefully will make some good strides on the wasm piece this weekend and engage more deeply with the code then.

prql/src/semantic/materializer.rs
aljazerzen: True that. Do you have a suggestion for the name?

PR #283, opened 2022-03-30
Decrease number of SQL CTEs produced

aljazerzen:


prql/src/translator.rs
```diff
@@ -52,67 +52,49 @@ impl TryFrom<Query> for sql_ast::Query {
...
-            cluster_by: vec![],
-            distribute_by: vec![],
-        })),
+        });
+    }
+
+    Ok(main_query)
+}
+
+fn table_to_sql_cte(table: Table) -> Result<sql_ast::Cte> {
```
max-sixty, line 89: (I've been trying to use `{output}_of_{input}` as a standard naming convention. Not the most important part of the code though!)

max-sixty (commented):
Great!!

PR #287, opened 2022-03-31
Run examples as tests

### LINKED ISSUE #176: Test the documented examples
We have some [documented examples](https://github.com/max-sixty/prql/tree/main/examples) — some of which currently work and some of which don't.

It would be great to test our documented examples. The current format for docs seems good; I'd also be open to a different format, but there should be no rust in these examples — they're for users rather than developers of prql.

(TBC, we already have an order of magnitude more tests, mostly using `insta` — not tests need to be docs, but all docs should be tested)

Assuming we're happy with the format, a really useful tool would:
- Parse the `.md` files
- Extract the PRQL
- Run the PRQL
- Compare to SQL
- Add something like a `# Not yet implemented` comment if it raises an error
- Allows for copying & pasting the result if it works but the output isn't exactly the same

aljazerzen:
Closes #176

Added tests for #213


max-sixty:
Brilliant, thank you @aljazerzen !

PR #289, opened 2022-03-31
Run GHA on all platforms

max-sixty:
I think this may fail on Windows. It's not high priority, but it would be friendlier for developers if they can reliably run tests locally on their platform.


max-sixty:
OK, not the most efficient PR, as you can see by the number of commits — but we now have all tests running in 32s, and all platforms on main, or if we attach a label to the PR.

PR #304, opened 2022-04-03
Monaco Editor in web

max-sixty:
As discussed on Discord: I'm struggling to get this working. I have a Monaco Editor up, and the PRQL compiling, but I'm not sure how to hook up the `on_did_change_content` to re-run the PRQL `compile` function on each change.

This PR replaces `sycamore` with `yew` which looks to be well-supported, and has restarted development in the past year. It uses @siku2's https://github.com/siku2/rust-monaco, which looked excellent.

I suspect getting this to work is actually not that difficult and I'm missing something — I have done almost zero web / react / etc before.

I'll pause here for a bit since I've neglected other parts of the code in favor of Web for a while, but I'll come back to it. If anyone has ideas, please opine! Thanks

Screenshot (obv the SQL would be displayed differently in the eventual version)

image


max-sixty:
That was probably my personal best for time spent / line written for these 82 lines, but it's on its way to being pretty good:

Screen Cast 2022-04-03 at 12 46 43 PM

max-sixty:
Many thanks to @siku2 for helping me!

PR #307, opened 2022-04-04
Use LIMIT by default rather than TOP

max-sixty:
LIMIT seems more prevalent. Soon we'll implement independently by DB


max-sixty:
Ref https://github.com/max-sixty/prql/issues/223, thank for raising @Akaame

PR #308, opened 2022-04-04
Both PRQL & SQL in Monaco editors

max-sixty:
The SQL editor is read-only, which requires
https://github.com/siku2/rust-monaco/pull/21

I have not worked out how to put them side-by-side yet! Rust is much
easier than HTML.


max-sixty:
Screen Cast 2022-04-03 at 8 26 21 PM

PR #309, opened 2022-04-04
Better error reporting

aljazerzen:


prql/tests/integration/compile.rs
```diff
@@ -16,8 +16,8 @@ nodes:
 }
 
 #[test]
-fn transpile_variables() -> Result<()> {
-    assert_snapshot!(compile("select 1")?, @r###"
+fn transpile_variables() {
```
max-sixty, line 19: Do you prefer not to have `Result`s in tests? I rather like them because it saves lots of `.unwrap`s...

max-sixty (commented):
Excellent!

prql/tests/integration/compile.rs
aljazerzen: I would have to write `From<(String, Option<SourceLocation>)> for Error` for `?` to work.

That's why I removed results. There is also no ` -> Result<()>` and `Ok(())` for each function, so overall, that may be less characters of code.

max-sixty: Ah OK, so just for these functions

PR #323, opened 2022-04-06
Fix atomic table splitting

aljazerzen:
Changes translator, such that

```
from employees
take 10
select first_name
```
produces a single SELECT statement.


max-sixty:
Great! 

BTW I think PRQL's compilation to reasonable SQL is a big plus for people who are familiar with SQL (which my guess will be almost all the _early_ users). I wish Malloy (#26) would do better at this — their SQL is somewhat inscrutable.

PR #324, opened 2022-04-06
Resolve all variable names

### LINKED ISSUE #222: Function vs. identifier namespaces
Something PRQL does that's distinct from general purpose languages — and common with SQL — is allow identifiers to refer to data; e.g. PRQL's `from a | select [b, c]` would generally be something like `load("a")["b", "c"]`, with quotes around the data references.

One disadvantage of PRQL's approach is it's difficult to mix identifiers and function names. SQL does this quite well, at the cost of no abstractions; for example:

```sql
select
  limit_date as date,             # Create a column named `data`
  date('2021-01-01') as start,    # Use `date` as a function
  count(*) as count               # Create a column name `count` without interfering with the function
from data
group by 1
having count > 0
```

But PRQL gets confused by:

```elm
from data
aggregate [count: count foo] by:limit_date
filter count > 0                 # What does `count` refer to here?
```

`count` refers to the function, rather than the column, so this raises an error.

I think that's fairly reasonable, but it leads to a few priorities:
- Good error messages — if someone uses `count` like this, we should say "you have an assignment above named `count`, but it refers to a function here", and not "wrong number of arguments for `count`".
- Additions to the stdlib become backward-incompatible, since they might conflict with existing column names
- We should add documentation for avoiding this. One workaround is using an s-string; i.e. `s'count'` always compiles into `count` in SQL. Possibly there's a language addition that allows a more specific way of being explicit, without giving up our treasured approach of allowing identifier for data references.

### LINKED ISSUE #247: Reference to the current pipeline

aljazerzen:
Changes:

- Resolver will throw errors if there are variables it cannot resolve.
  (it will take into account unknown table fields, so `from employees | select first_name` will still work)

- Resolved AST will contain full table names. This is required because of this use case:

  ```
  from companies
  select [name, company_no]
  join employees [company_no]
  ```
  What one would expect here is:
  ```
  SELECT name, company_no, employees.* 
  FROM companies JOIN employees USING(company_no)
  ```
  ...but what if employees also have column `name`? Only reference `name` would be ambigious. Which means full-quailified table names are required:
  ```
  SELECT companies.name, company_no, employees.* 
  FROM companies JOIN employees USING(company_no)
  ```

  We could only add full tables names when there are multiple tables in scope, but due to non-sequential nature of SQL SELECTs, this is quite an effort.

- table names will be corrected between splits, which closes #247
- ambiguous references will throw an error, which closes #222
- `table.*` is now allowed. This is useful in such cases:
  ```
  from employees
  join salaries [emp_no]
  select [emplyees.*, salary]
  ```


prql/src/ast_fold.rs
```diff
@@ -31,7 +31,7 @@ pub trait AstFold {
     fn fold_pipeline(&mut self, pipeline: Vec<Transform>) -> Result<Vec<Transform>> {
         pipeline
             .into_iter()
-            .map(|t| self.fold_transformation(t))
+            .map(|t| self.fold_transform(t))
```
max-sixty, line 34: Shall we call these `transforms` everywhere? I'm up for it if you are...

prql/src/semantic/context.rs
```diff
@@ -3,33 +3,43 @@ use enum_as_inner::EnumAsInner;
...
 /// Scope within which we can reference variables, functions and tables
 /// Provides fast lookups for different names.
 #[derive(Debug, Default, Serialize, Deserialize, Clone)]
 pub struct Context {
     /// current table columns (result of last pipeline)
-    pub(super) table: Vec<TableColumn>,
+    pub(super) frame: Vec<TableColumn>,
+
+    /// For each namespace (table), a map from column names to their definitions
+    /// "$" is namespace of variables not belonging to any table (aliased, join using)
```
max-sixty, line 22: It'd also be fine to use an `Enum` for this, but this is ofc fine

prql/src/semantic/materializer.rs
```diff
@@ -357,10 +419,10 @@ aggregate [
         let ast = parse(
             r#"
 func lag_day x = s"lag_day_todo({x})"
-func ret x = x / (lag_day x) - 1 + dividend_return
+func ret x dividend_return = x / (lag_day x) - 1 + dividend_return
```
max-sixty, line 422: I don't quite understand this change — `dividend_return` was supposed to be a column, rather than a param. Tell me if I'm missing something

prql/src/ast_fold.rs
aljazerzen: Oh, forgot to tell you about that: I've changed that in a few places. It seemed better name, just because its shorter.

We should go over the project and refacator it everywhere

prql/src/semantic/materializer.rs
aljazerzen: Functions cannot reference columns, because they could be used in different context, where that column would not be available.

Eg:
```
from employees
select first_name
derive [a: ret 15]   # there is only first_name in scope here, cannot reference dividend_return!
```

prql/tests/integration/compile.rs
```diff
@@ -63,7 +63,7 @@ take 20
       title,
       country
     HAVING
-      ct > 200
+      COUNT(*) > 200
```
max-sixty, line 66: I had thought we were getting `COUNT(*)` before actually. We can use `ct` here, which is slightly better, but either is totally fine.

aljazerzen: No, I don't think either is fine.
For:
```sqlSELECT COUNT(*) AS ct
FROM employees
GROUP BY last_name
HAVING ct > 0
```
postgres returns `ERROR: column "ct" does not exist`

max-sixty:
A good example of this:

```elm
from companies
select [name, company_no]
join employees [company_no]
```

previously compiled to

```sql
SELECT
  name,
  company_no
FROM
  companies
  JOIN employees USING(company_no)
```

...which is obv incorrect.

Now it compiles to:

```sql
SELECT
  companies.name,
  employees.*,
  company_no
FROM
  companies
  JOIN employees USING(company_no)
```

prql/tests/integration/compile.rs
max-sixty: Oh! You're right. Other DBs allow variables in `HAVING` and `ORDER BY`; I'm surprised postgres doesn't...

prql/src/ast_fold.rs
max-sixty: I agree! 

https://en.wikipedia.org/wiki/Nominalization is impossible to stop!

prql/src/semantic/materializer.rs
max-sixty: What do you think about "late binding" them? i.e. as though someone copy-pasted the function definition. In the way I was thinking about it, having:

```elm
from a
select (b / (lag_day b) - 1 + dividend_return)
```

..._would_ be allowed, and so having `ret b` would be equivalent to that.

I had thought of this as quite an important feature — otherwise functions can't be used to do things like "add X and subtract Y and multiply by Z" without specifying X & Y & Z each time.

I recognize that this sort of late-binding isn't common in languages; it's more like a macro. But also we're using `dividend_return` as more like another language would use `"dividend_return"` — i.e. if it's not specified as a parameter, then it's a column name, and can be treated as a literal.

Does my point make sense? WDYT?

aljazerzen: Our functions are very similar to macros, which is a case for this feature. I see the use case.

But, this kinda uncommon for other languages... I'm hesitant about this mostly because it breaks current resolver/materialiazer architecture. But to be honest, this would be useful to have.

Currently it can be bypassed by using s-strings, though.

max-sixty: Yes true re s-strings. 

Is it OK if I add an issue so we can track it?

PR #335, opened 2022-04-09
Rename max-sixty to prql

max-sixty:


max-sixty:
As discussed on discord. For anyone not following — I'm still as committed as ever! This formalizes that the project is led by the community, of which I'm a part. I don't think I'm even the author of most of the code now, mostly thanks to @aljazerzen , with excellent contributions from others.

---

People should run this locally:

```
git remote set-url upstream https://github.com/prql/prql
```

PR #347, opened 2022-04-11
Parse dialect & version

aljazerzen:


prql/src/ast.rs
```diff
@@ -273,3 +303,9 @@ impl From<Item> for anyhow::Error {
         anyhow!("Failed to convert {item:?}")
     }
 }
+
+impl Dialect {
+    pub fn use_top(&self) -> bool {
+        matches!(self, Dialect::MsSql)
```
max-sixty, line 309: Possibly in the future we want to have each Dialect be a Struct that we implement a `Trait` for — for the moment this is excellent.

max-sixty (approved):
Very nicely implemented! I can't even find something to improve on...

max-sixty:
I will do the docs for this...

PR #350, opened 2022-04-13
Compile and display examples in documentation

max-sixty:
This auto-compiles documentation, extending the existing approach to test examples.

It's over-engineered (I've been on a run of these the past week...). And the formatting for wide PRQL statements isn't good. But it does give us nice & always correct documentation, and snapshots it in our repo:

image


max-sixty:
Tests will likely fail since we need to configure the pre-processor

aljazerzen:
Damn, this looks real nice. I also like how in .md files there is no compiled SQL, so it's less cluttered.

Do you think that live editor is still needed here if it is gonna be on the main page?

max-sixty:
> Do you think that live editor is still needed here if it is gonna be on the main page?

I'm not sure. It's fairly related to the language docs — i.e. it's for exploring the language as well as just a quick intro. When we get the site up & running we can remove it if it seems redundant though.

I was almost thinking we could make every example use it (rather than the table that this PR adds), but we'd probably need to do some work to simplify the monaco interface in that environment. (And possibly there are enough people who can't load the editor that it would be restrictive? not sure)

Issue #300, opened 2022-04-02
Orthogonal treatment of groups (nee Window functions)


max-sixty:
*Edit — this issue evolved into a much broader discussion of how we do groups. The original message is below, and I've tried to summarize the current state towards the end of the issue so it's easier for others to engage.*

---

More than half the people I speak with about PRQL tell me that window functions are painful in SQL and would be a great feature to add to PRQL. I put window funcitons in the examples in the Readme at first, since I had the same experience in SQL.

For clarity: a window / analytic / analytical function is a function that takes other rows as an input, but instead of aggregating over many rows, it creates a new value for every row. For example, a moving average or lag. In SQL it's represented by the `OVER` keyword.

I haven't thought about this enough, but wanted to post an issue so we could start thinking about how to do it. A couple of initial options:

- A separate pipeline, with the column at the end, and the function (`lag` in this case) in the pipeline:
```elm
from prices
derive price_yesterday: (
  window
  by sec_id
  sort date
  lag 1
  price
)
```
- A function with lots of named args, with the function `lag` as a named arg (but how would this work for functions with 0 or >1 parameters?):
```elm
from prices
derive price_yesterday: (window by:sec_id sort:date lag:1 price)
```
- Eliminate `window` — because we have `aggregate`, we can't get confused about `sum(foo) OVER` and `sum(foo)`, so possibly we don't need to have `window`:
```elm
from prices
derive price_yesterday: (price | lag rows:1 sort:date by:sec_id)
```

I don't think this is sufficiently thought through, and I'm sure these examples have inadequacies.

Here are some example queries we can use as cases: https://cloud.google.com/bigquery/docs/reference/standard-sql/analytic-function-concepts#get_the_most_popular_item_in_each_category

aljazerzen:
For anyone new to window functions (like I just was): [good tutorial with illustrations](https://www.sqltutorial.org/sql-window-functions/)

A nice observation that we could use in language design is that we have 3 types of functions:
- scalar functions, which map a single value to a single value (e.g. `FLOOR`). They can be used in select and are applied to the whole column, on each value.
- multi-variate functions, which map many values to a single value (e.g. `SUM` or `MIN`). They can be used in aggregate and reduce each partition to one value.
- vector functions, which map a many values to many values (eg. `RANK` or `LAG`)/ They can be used as window functions in select and map column to column.

The problem is that when writing code you don't think about that. Is it advantageous if we force users to think about it?

[...]

PR #351, opened 2022-04-14
Add sort to sementic context

aljazerzen:
Introduces sort syntax for `asc` and `desc`:

```elm
from Employees
sort [id]
sort [age, desc:last_name, asc:first_name]
```

Includes refactor in preparation for #300

### DIFF
--- a/prql/src/ast.rs
+++ b/prql/src/ast.rs
@@ -101,7 +101,7 @@ pub enum Transform {
         by: Vec<Node>,
         select: Vec<Node>,
     },
-    Sort(Vec<Node>),
+    Sort(Vec<ColumnSort<Node>>),
     Take(i64),
     Join {
         side: JoinSide,
@@ -131,8 +131,8 @@ impl Transform {
             Transform::Select(nodes)
             | Transform::Filter(Filter(nodes))
             | Transform::Derive(nodes)
-            | Transform::Aggregate { by: nodes, .. }
-            | Transform::Sort(nodes) => nodes.first(),
+            | Transform::Aggregate { by: nodes, .. } => nodes.first(),
+            Transform::Sort(columns) => columns.first().map(|c| &c.column),
             Transform::Join { filter, .. } => filter.nodes().first(),
             Transform::Take(_) => None,
         }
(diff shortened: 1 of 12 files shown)


max-sixty:
Very nice! Thanks!

max-sixty:
> ```elm
> sort [age, desc:last_name, asc:first_name]
> ```

I'm actually not so sure about this. 

Generally the `lvalue`s are where to put the `rvalue`s — e.g. `e:employees`: "put the `employees` table into the `e` namespace". Or `by:title`: "Put the `title` column in the `by` parameter".

They're not generally _modifiers_ of the `rvalue`s. But here they are. 

That said, I'm not sure how else to do it. We don't currently support (though we could):

```elm
sort [age, last_name order:desc, first_name order:asc]
```

I haven't really seen other languages do this well — either the specific issue of how to specify sort order, or optional boolean arguments / flags. 

Probably the one that most resonates with me for these is shell conventions, which allows an optional flag, e.g. `--desc` (and not `--desc=true`).

aljazerzen:
I don't like this `sort [age, last_name order:desc, first_name order:asc]` because it's hard to parse since it is very similar to a function call of `last_name` with argument `order`.

What about nested arrays:
```
sort [age, [last_name, order:desc], [first_name order:asc]]
sort age
sort [last_name, order:desc]
sort [age, first_name]
```

Or tuples:
```
sort [age, (last_name, order:desc), (first_name order:asc)]
sort age
sort (last_name, order:desc)
sort (age, first_name)
```

max-sixty:
> I don't like this `sort [age, last_name order:desc, first_name order:asc]` because it's hard to parse since it is very similar to a function call of `last_name` with argument `order`.

Good point

Yes, I agree none of those are ideal.

What do you think about `sort [age, last_name:desc, first_name:asc]` over the current approach?

I'm not sure there's enough space in the lang for "flags", but if there were, `sort [age, last_name --desc, first_name --asc]` would work, maybe as sugar for one of the other options you suggest.

aljazerzen:
There is one more option from dyplr:

```
[age, -first_name, last_name]
```

max-sixty:
Great idea! I quite like it; what do you think?

aljazerzen:
It's simple and concise, I am all in for it. 

On the note of what to do with NULLs, dplyr also has some advice:

> ### Missing values
> 
> Unlike base sorting with [sort()](https://rdrr.io/r/base/sort.html), NA are:
> - always sorted to the end for local data, even when wrapped with [desc()](https://dplyr.tidyverse.org/reference/desc.html).
> - treated differently for remote data, depending on the backend.

max-sixty:
Great! Thanks for suggesting the idea!

Re nulls — yeah — how much of that can we do from PRQL? There are some things I'd love to change about nulls, but I think many of them are ingrained in the DB / in SQL standards...

aljazerzen:
Well there is the `ORDER BY col DESC NULLS [FIRST/LAST]`, which we can choose to support or not. 

My suggestion was that for now, we always append `NULLS LAST`

Issue #301, opened 2022-04-02
0.2.0 Milestone


max-sixty:
I'm not sure if this is literally "0.2.0", or we should have intermediate releases, but this is the next big milestone!

My sense is that to get to the next level, PRQL needs to "get out there" — getting feedback from users, demonstrating that it can solve real pain points, exploring ways to hook into real workflows.

For that to happen I think we need:
- A website with live updating PRQL<>SQL — https://github.com/max-sixty/prql/issues/271. 
  - I worked on this last weekend (though wasn't that productive, took a long time to get to the current version)
- Much better language documentation https://github.com/max-sixty/prql/issues/232. I have been intending to address this for a few days now...
- Window functions https://github.com/max-sixty/prql/issues/300 — more than half the people I speak to about PRQL tells me that this is painful in SQL and would be a great feature to add. 
- More language features — there's lots of small annoyances in SQL I think we could solve without Herculean effort. Malloy have done this very well https://github.com/max-sixty/prql/issues/71. Some things that come to mind; feel free to open specific issues to discuss them:
  - f-strings — `CONCAT(foo, " ", bar)` -> `f"{foo} {bar}"`?
  - Regex — `REGEX_MATCH(foo, "\\w{3}")` -> `foo ~ r"\w{3}"`? Or -> `regex foo r"\w{3}"`?
  - Ranges — `BETWEEN 1 AND 3` -> `in 1..3`?
  - Dates — `"2021-01-01"` -> `@2021-01-01`? And `DATE_TRUNC(foo_date, YEAR)` -> `foo_date.year`? Or -> `foo_date | as year`? Or `foo_date | to year`?
  - Offsets — `DATE_ADD(DATE "2008-12-25", INTERVAL 5 DAY)` -> `@2008-12-25 - 5day`?
  - Case / when / then
  - Any others?
  - TBC, I don't think we have to do all of these — and getting more feedback from users will help us target the most important ones. And if we want to change something later, that's fine too — we're early, and we'll have versioning soon anyway.
- Building out the standard library a bit https://github.com/max-sixty/prql/issues/217 — if we had a systematic way of doing this by parsing some docs (e.g postgres) and selecting what we want to include / exclude, that could be very efficient. Otherwise gradually adding things could be a great way for a new contributor to make an impact.
- [Edit, based on @aljazerzen comment below]: Logo & org https://github.com/max-sixty/prql/issues/303

Other big things that I don't think we need to do, but were in my mind when thinking about this:
[...]

PR #352, opened 2022-04-14
Ranges and `between` function

aljazerzen:
Prepares for #300 
Checks a mark of #301  

I've added this:
```
from employees
filter (age | between 18..40)
```

I've chosen `between` for the function name, because `in` seems too concise. If anyone feels different, we can change it.

Currently ranges can currently contain only number literals, and not expressions. Should this constraint be removed?

TODO: docs

### DIFF
--- a/prql/src/ast_fold.rs
+++ b/prql/src/ast_fold.rs
@@ -96,6 +96,10 @@ pub fn fold_item<T: ?Sized + AstFold>(fold: &mut T, item: Item) -> Result<Item> 
                 .map(|x| fold.fold_node(x.into_inner()).map(ListItem))
                 .try_collect()?,
         ),
+        Item::Range(range) => Item::Range(Range {
+            start: fold_optional_box(fold, range.start)?,
+            end: fold_optional_box(fold, range.end)?,
+        }),
         Item::Query(query) => Item::Query(Query {
             nodes: fold.fold_nodes(query.nodes)?,
             ..query
@@ -131,6 +135,13 @@ pub fn fold_item<T: ?Sized + AstFold>(fold: &mut T, item: Item) -> Result<Item> 
         // them.
         Item::String(_) | Item::Raw(_) => item,
     })
+}
+
+pub fn fold_optional_box<T: ?Sized + AstFold>(
+    fold: &mut T,
+    opt: Option<Box<Node>>,
+) -> Result<Option<Box<Node>>> {
+    Ok(opt.map(|n| fold.fold_node(*n)).transpose()?.map(Box::from))
 }
 
 pub fn fold_interpolate_item<T: ?Sized + AstFold>(
--- a/prql/src/ast_fold.rs
+++ b/prql/src/ast_fold.rs
@@ -97,6 +97,9 @@ pub fn fold_item<T: ?Sized + AstFold>(fold: &mut T, item: Item) -> Result<Item> 
                 .try_collect()?,
         ),
         Item::Range(range) => Item::Range(Range {
+            // This aren't strictly in the hierarchy, so we don't need to
+            // have an assoc. function for `fold_optional_box` — we just
+            // call out to the function in this module
             start: fold_optional_box(fold, range.start)?,
             end: fold_optional_box(fold, range.end)?,
         }),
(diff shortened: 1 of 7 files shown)


prql/src/translator.rs
```diff
@@ -1202,4 +1215,35 @@ take 20
...
+          employees
+        WHERE
+          age BETWEEN 18
+          AND 40
+        "###);
+
+        let query: Query = parse(
+            r###"
+        from employees
+        filter (age | between ..40)
```
max-sixty: (A future version could allow this as "less than", a bit like python)

prql/src/semantic/context.rs
```diff
@@ -295,28 +298,38 @@ impl Context {
...
+            // this variable can be from a namespace that we don't know all columns of
+
+            match ns.len() {
+                0 => unreachable!("inverse index contains empty lists?"),
+
+                // single match, great!
+                1 => Ok(ns.iter().next()),
+
+                // don't report ambiguous variable, database may be able to resolve them
+                _ => Ok(None),
```
max-sixty, line 328: This is clever!

prql/src/prql.pest
```diff
@@ -93,7 +93,10 @@ quote = _{ "\"" | "'" }
 string = { ( !( "\\" | PEEK ) ~ ANY )+ }
 string_literal = ${ PUSH(quote) ~ string ~ POP }
 
-number = ${ ( ASCII_DIGIT | "." )+ }
+number = ${ ( ASCII_DIGIT )+ ~ ("." ~ ( ASCII_DIGIT )+)? }
+
+// this could be generalized to use term_simple
+range = ${ number? ~ ".." ~ number? }
```
max-sixty, line 99: (in the future) this could also allow dates etc

prql/src/ast_fold.rs
```diff
@@ -96,6 +96,10 @@ pub fn fold_item<T: ?Sized + AstFold>(fold: &mut T, item: Item) -> Result<Item>
                 .map(|x| fold.fold_node(x.into_inner()).map(ListItem))
                 .try_collect()?,
         ),
+        Item::Range(range) => Item::Range(Range {
+            start: fold_optional_box(fold, range.start)?,
```
max-sixty, line 103:
```suggestion
            // This aren't strictly in the hierarchy, so we don't need to
            // have an assoc. function for `fold_optional_box` — we just
            // call out to the function in this module
            start: fold_optional_box(fold, range.start)?,
```

max-sixty (approved):
This looks great! Thank you for doing these.

> I've chosen `between` for the function name, because `in` seems too concise. If anyone feels different, we can change it.

`in` is quite concise, maybe too much so.

Something like `in` on `contains` could have a broader definition; e.g. it could also be used for a string (e.g. "is `child` in `children`") or "day in month" (relative to "between start of month and end of month").

It possibly also relates to the _concept_ of a range — we're saying "Is A *in* range-B?", whereas the between doesn't have a range; it's saying "Is A between C and D?". But maybe this is overthinking it.

I don't have a strong view. I'm very happy to go with this. If people object, we can change later.

---

Edit: To further the point, I could imagine if we use `between` we might not need a Range object — just `between 0 40` would make sense

prql/src/translator.rs
aljazerzen: Thats true. 

I did not allow it mosty because I did not know how to implement it :D

aljazerzen:
Actually, you do make a point:

`value | between 18 50` does make a point while `value | between 18..50` does not (it would mean is value between a range?)

`value | in 18..50` is the best option, while `value | contains 18..50` does not read alright with pipelines.

I will go ahead and change this to `value | in 18..50` and then merge.

max-sixty:
Great! Thanks.

Another thing I've been thinking about is whether terms like `in` should be infix operators. I imagine that will be friendlier to the user who just wants to write a query. Possibly we can see how the language evolves and offer it as an option, hopefully without multiplying the ways of accomplish goals with the language.

aljazerzen:
Hmm, yes, that would be more user-friendly. One another way would be to allow functions to be defined as binary operator:

```
func binary:true in value range = s"{value} BETWEEN {range}"
```

Just an idea

max-sixty:
Brilliant! Thanks for adding this!

> func binary:true

(Interesting idea, it's probably more like "infix" or "dyadic", but those are quite technical)

aljazerzen:
Yeah, infix is better.

PR #358, opened 2022-04-14
Add windows build

max-sixty:
- Temporarily remove Windows build
- Add back windows build


reference/tests/snapshot.rs
```diff
@@ -93,11 +96,14 @@ fn write_reference_examples() -> Result<()> {
...
-        }
-
+        dbg!(&prql);
+        let sql = compile(&prql).unwrap_or_else(|e| format!("Failed to compile `{prql}`; {e}"));
+        assert_snapshot!(sql);
+    });
+    write_reference_examples();
+    glob!("examples/**/*.prql", |path| {
+        let prql = fs::read_to_string(path).unwrap();
+        dbg!(&prql);
```
max-sixty: The proximate cause is it failing on the snapshot test _after writing_ the files. So normal snapshots work — and if we disable the writing, then this works. But because we generate the `.prql` files from the docs as part of the test, it's somehow getting confused with line endings.

Possibly it's writing them with `CRLF`, rather than `LF`, and insta is failing to read those correctly? I'm still not sure at this level

We could disable the writing on Windows as a real hack? But then what's the point of testing on windows...

PR #370, opened 2022-04-16
Add (brief) docs on ranges

max-sixty:


max-sixty:
I removed this part, since I doesn't yet work on dates. I would be up for implementing dates as something more than implicit strings (!), like Malloy's approach of `@2020-01-01`, and then we could get this to work,

````md
This can be used with dates, too:

```prql_no_test
from employees
filter (dob | in '1980-01-01'..'1990-01-01')
```
````

PR #377, opened 2022-04-17
Literal dates & times

### LINKED ISSUE #373: Date, time & duration literals
In SQL, creating date objects is a mixture of:
- A string `'2020-01-01'`
- `DATE(2016, 12, 25)` — BQ
- `make_date(year int, month int, day int)` — Postgres

The strings are dynamically typed so they can be implicitly converted when, for example, they're compared to another date (some quick experimenting suggests this also happens when compared to other strings that might be a date!).

Some better options:
- Malloy uses `@2020-01-01`
- We could extend our use of number & string combinations from https://github.com/prql/prql/pull/369 (`5days`) to dates and use `d2020-01-01` or `D2020-01-01`. We can also apply this to timestamps / times / etc with different letters.
- Very open to other ideas

Any thoughts?

max-sixty:
These aren't yet implemented. They would be implemented in the same way
that Intervals are implemented.

Would close https://github.com/prql/prql/issues/373


max-sixty:
Another option for this would be `d"2022-12-31"` rather than `d2022-12-31` — a bit less slick, but it removes any ambiguity — e.g. if there were a column actually called `d2022-12-31` in a DB (?!)

max-sixty:
I realized that `T10:30` can't easily be parsed as a time rather than a named expr `T10`, `:`, `30`.

So some options:
- `T"10:30"` (or `t"10:30"`)
- `T10.30`, but then no longer consistent with standards
- Adding code to the named expr code to exclude `T\d{2}` prefixes. Which is OK in a practical sense, but the workaround is a symptom of an underlying issue, and makes the parser code messy and less modular

aljazerzen:
Having `d2022-12-31` and `t10:40` is really convenient, but I would say it's not worth it if it adds a new parsing pattern.

What about having `@` for all datetime types: `@2022-12-31T10:40`, `@2022-12-31`, `@10:40`, which compile to datetime, date and time, respectively?

It introduces a new parsing pattern, but it covers all date-time formats at once. It also leaves `x"text"` pattern for things that are more string-like.

max-sixty:
The issue with `@` is how to differentiate between similar but different types e.g. `datetime` and `timestamp`. Though possibly there are workarounds.

For example, if we think 90% of cases are covered by a standard `@`, and we want to expend punctuation on this, then people can use `as` to get to `datetime`.

max-sixty:
Ready to go!

aljazerzen (approved):

PR #389, opened 2022-04-20
Adds prql-python that will deploy to PyPi on a tag event

qharlie:
This exposes prql_compiler::compile in a [pyfunction] for pyo3 , which uses maturin in workflows/python-publish.yaml to push to pypi using the repository secret PYPI_API_TOKEN .  It does this when a branch is tagged.

The current code in lib.rs will throw a python Syntax Exception if there is an exception when compiling the prql.  The python signature is 

The python module name is prql-python but can be configured in Cargo.toml .

```python
def to_sql(query: str) -> str 
```


.github/workflows/test.yaml
```diff
@@ -41,7 +41,7 @@ jobs:
         uses: actions-rs/cargo@v1
         with:
           command: clippy
-          args: --all-targets --workspace --target=wasm32-unknown-unknown -- -D warnings
+          args: --all-targets -p prql-web -p prql-compiler -p prql-js --target=wasm32-unknown-unknown -- -D warnings
```
max-sixty, line 44: If we can do this with attributes, then it would also run locally. No need to add now though. Please could we add a TODO?

.github/workflows/publish-python.yml
```diff
@@ -0,0 +1,66 @@
+name: publish-prql-python
+
+on:
+  push:
+  pull_request:
```
max-sixty, line 5: Would this publish on every PR?

.github/workflows/publish-python.yml
```diff
@@ -0,0 +1,66 @@
+name: publish-prql-python
```
max-sixty, line 1: This workflow looks great!

prql-python/src/lib.rs
```diff
@@ -0,0 +1,27 @@
+use prql_compiler::compile;
```
max-sixty, line 1: Awesome, very impressed at how simple this is!

prql-python/src/lib.rs
```diff
@@ -0,0 +1,27 @@
...
+        ))),
+    }
+}
+
+#[pymodule]
+fn prql_python(_py: Python, m: &PyModule) -> PyResult<()> {
+    m.add_function(wrap_pyfunction!(to_sql, m)?)?;
+    Ok(())
+}
+// This test below is causing a linking error here https://github.com/qorrect/prql/runs/6099160429?check_suite_focus=true
```
max-sixty, line 21: Great for putting in the PR, let's merge and iterate. Maybe others can take a glance, I'm happy to take a look in a couple days otherwise.

.github/workflows/pull-request.yaml
```diff
@@ -93,7 +93,7 @@ jobs:
       - name: Run cargo-tarpaulin
         uses: actions-rs/tarpaulin@v0.1
         with:
-          args: "-- --test-threads 1"
+          args: "--workspace --exclude prql-python -- --test-threads 1"
```
max-sixty, line 96: Very low priority but if we can add a TODO to put this in a config file, that'd be great

max-sixty (approved):
Great job @qorrect !

.github/workflows/publish-python.yml
manning79: It builds on every push , it only publishes when a branch is tagged [in the if section here ](https://github.com/prql/prql/pull/389/files/f69a8b5919a4442827237065672117cbf3a66191#diff-40a8e37be878e2b8c149cbae07d713e9b42743acfaac10c4a266bd7ded764019R51) , we can update the ownership  [of the pypi project](https://pypi.org/project/prql-python/) by updating the PYPI_API_TOKEN to prql official account

.github/workflows/test.yaml
manning79: TODO as an Issue or in the  file ?

max-sixty: I was thinking file as it's small, but either is totally fine

PR #392, opened 2022-04-21
Allow ranges with any literals

max-sixty:
I also refactored the parsing code a bit — though I would accept feedback that it's too much of a single block now.

### DIFF
--- a/prql-compiler/src/prql.pest
+++ b/prql-compiler/src/prql.pest
@@ -68,7 +68,8 @@ func_call = { ident ~ (named_term_simple | term_simple)+ }
 func_curry = { ident ~ (named_term_simple | term_simple)* }
 
 term = _{ ( s_string | f_string | func_call | term_simple ) }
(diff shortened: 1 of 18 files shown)


prql-compiler/src/prql.pest
```diff
@@ -94,8 +95,11 @@ string_literal = ${ PUSH(quote) ~ string_inner ~ POP }
 
 number = ${ ( ASCII_DIGIT )+ ~ ("." ~ ( ASCII_DIGIT )+)? }
 
-// this could be generalized to use term_simple
-range = ${ number? ~ ".." ~ number? }
+range = ${ range_edge ~ ".." ~ range_edge }
+// We need `literal` separate from `term_simple` for things like range edges,
+// which would infinitely recurse otherwise, since it'll keep trying to parse
+// the whole span, not just the part before `..`.
```
aljazerzen, line 101: Ah great, I didn't manage to figure this out when I was adding range

aljazerzen (approved):

Issue #391, opened 2022-04-21
Get PR tests back to < 1 min


max-sixty:
I'm trying to keep the PR builds below 1 min; they've been just above 30 secs for a while.

The recent python builds from https://github.com/prql/prql/pull/389 take 4 min — we could try:
- Adding caching
- Potentially not testing on windows on all PRs — there's an example in the existing workflow of only running on `main` or if the `run-all-tests` label is added; for the bigger PRs

max-sixty:
Closed by #394, though not excellently — we skip tests on that path, which means if we introduce an incompatible change, it won't get caught until after the merge. If we can find some way of caching the build, that would be much better.

PR #394, opened 2022-04-22
Speed up GHA Python workflows

max-sixty:


max-sixty:
The diff isn't that clear, because my autoformatter changed the YAML, and I renamed the file to `python` from `publish-python` given it also acts as a test. But this basically adds a cache to the steps in an attempt to close https://github.com/prql/prql/issues/391

max-sixty:
Cache doesn't seem to work, so instead skipping runs where no files in `prql-python` are changed

.github/workflows/python.yml
```diff
@@ -0,0 +1,93 @@
...
+    if: "startsWith(github.ref, 'refs/tags/')"
+    needs: [macos, windows, linux]
+    steps:
+      - uses: actions/download-artifact@v2
+        with:
+          name: wheels
+      - name: Publish to PyPI
+        uses: messense/maturin-action@v1
+        env:
+          MATURIN_PYPI_TOKEN: ${{ secrets.PYPI_API_TOKEN }}
```
max-sixty, line 90: @charlie-sanders does this exist?

manning79: @max-sixty It doesn't on this repo no.  On my fork I went to Settings to add the Secret, but I don't see a settings tab in the organizational repo ?

We can also move this under the prql/ umbrella now, if you want to create a PyPi account for prql and use that

Issue #130, opened 2022-02-25
Transpiling from SQL to PRQL


max-sixty:
This is a longer-term idea — but now we're using sqlparser to write SQL, we could also use it to _read_ SQL and produce PRQL.

We'd need to write:
1. a fairly simple compiler from sqlparser AST to PRQL AST
2. `Display` implementations which convert PRQL AST to PRQL code

This would be useful for people transitioning from SQL. At least initially, the main users are likely to be engineers who are frustrated by SQL because they've written a lot of it.

We wouldn't need to support all of SQL for it to be useful — there's a fairly small subset that's used a lot — and possibly a larger subset which we could implement with s-strings.

We also wouldn't need to automatically do things like variable "creation" (e.g. replacing `salary + payroll_tax + benefits_cost` with `gross_cost`), which would be much more difficult (but great if we could!).

aljazerzen:
Update: this is currently not possible yet, but we are not far off. 

The `Display` trait for prql's AST `Node` is done, we just need to parse SQL and convert it to AST, which should't be a major problem.

ryanrussell:
A SQL -> PRQL transpiler would  **REALLY** ease adoption for PRQL.

Even if it had some disclaimers, it lowers a lot of barriers if a new user is trying to learn/adopt PRQL and could copy/paste a bunch of SQL queries.

florent-martineau:
Using Apache Calcite for parsing queries and turning them into an AST could be an idea.

Calcite is written in Java, though, and not in Rust.

max-sixty:
We'd likely use sqlparser-rs, which we currently use to represent SQL AST, and it builds the SQL string.

PR #398, opened 2022-04-22
Pipelines, groups, prql fmt and semantic rewrite

aljazerzen:
I started to work on #300, broke most of the tests and then repeatably found out I needed *just one more* refactor/feature before tests start working again. Result is this branch with 3441 insertions(+) and 2785 deletions(-).

Work done:

- Parse `group` transforms, their pipelines and infer context within these pipelines #300,
- Unify pipeline and inline-pipeline. This includes parsing transforms as function calls, that are determined to be pipeline transforms only during resolving. Before that, they are FunCall nodes in AST. 
- Do resolving in a single pass. This decouples the code a bit and will be beneficial for other uses such as polars backend and language server. It was a major rewrite of `::semantic`, but I'm happier with it now. It's not exacly readable yet, but it's a step closer.
- For debugging reasons, I added `Display` trait to `Item`, which means that we can print PRQL queries. Combined with parsing, this is a formatter and a step toward #130 . Can be used as `prql compile -f fmt` (now I see this should be a separate command).
- Refactor `ast` and `sql` modules (including `ast::dialect`). Just splitting ~1000 lines into ~3x300 lines.

Language changes:
- aggregate must have two positional args, it won't default to empty list

There is a test case that has not been fixed yet, see below.


reference/tests/snapshots/snapshot__run_reference_examples@introduction-0.prql.snap
```diff
@@ -30,10 +30,17 @@ WITH table_0 AS (
...
+  AVG(salary),
+  AVG(salary + payroll_tax),
+  SUM(salary),
+  SUM(salary + payroll_tax),
+  AVG(salary + payroll_tax + benefits_cost),
+  sum_gross_cost,
+  table_0.ct,
   countries.*,
   country_code,
-  true AS always_true,
```
aljazerzen, line 36: We don't parse boolean literals yet, so this was parsed as ident. Which now got resolved :D

reference/tests/snapshots/snapshot__run_reference_examples@introduction-0.prql.snap
```diff
@@ -30,10 +30,17 @@ WITH table_0 AS (
...
 )
 SELECT
-  table_0.*,
+  title,
+  AVG(salary),
+  AVG(salary + payroll_tax),
+  SUM(salary),
+  SUM(salary + payroll_tax),
+  AVG(salary + payroll_tax + benefits_cost),
+  sum_gross_cost,
```
aljazerzen, line 39: This still has to be fixed, but can I do it a separate PR?

prql-compiler/src/prql.pest
```diff
@@ -61,17 +52,28 @@ ident = @{ !operator ~ (ASCII_ALPHA | "$") ~ (ASCII_ALPHANUMERIC | "." | "_" | "
...
 
-term = _{ ( s_string | f_string | func_call | term_simple ) }
-term_simple = _{ ( s_string | f_string | ident | parenthesized_expr | list | range | literal | inline_pipeline ) }
+term = _{ ( s_string | f_string | ident | nested_expr | nested_pipeline | list | range | literal ) }
 literal = _{ interval | number | string_literal }
-list = { "[" ~ NEWLINE* ~ named_expr ~ ("," ~ NEWLINE* ~ named_expr )* ~ ","? ~ NEWLINE* ~ "]" }
-parenthesized_expr = _{ "(" ~ expr ~ ")" }
+list = { "[" ~ (NEWLINE* ~ named_expr_call ~ ("," ~ NEWLINE* ~ named_expr_call )* ~ ","?)? ~ NEWLINE* ~ "]" }
+nested_expr = _{ "(" ~ expr_call ~ ")" }
+nested_pipeline = { "(" ~ expr_call? ~ pipe ~ pipeline ~ NEWLINE* ~ ")" }
```
aljazerzen, line 78: This is shorter but less descriptive. Feel free to revert.

max-sixty:
This looks extremely impressive @aljazerzen !!

I will look in more detail later today

prql-compiler/src/prql.pest
```diff
@@ -11,30 +11,21 @@ WHITESPACE = _{ " " | "\t" }
...
-// `named_arg`, then `x:y` will be parsed as `x` and then error on `:y`.
-// Possibly we can mandate whitespace to follow, but pest has annoying magic
-// which removes whitespace in non-atomic rules, so we'd need to explore that.
-func_def_params = { ( named_term_simple | ident )* }
-table = { "table" ~ ident ~ "=" ~ "(" ~ NEWLINE* ~ pipeline ~ NEWLINE* ~ ")" }
+func_def = { "func" ~ func_def_flags ~ ident ~ func_def_params ~ "=" ~ expr }
 
-pipe = _{ NEWLINE+ | "|" }
-pipeline = { transformation ~ (pipe ~ transformation)* }
+func_def_flags = @{ ("transform" | "aggregation" | "window" | "")? }
```
max-sixty, line 22: Should we have some sort of syntax for these "type" arguments? For the moment I'll suggest a TODO

```suggestion
// TODO: consider some syntax for these func "type" args; currently they're only 
// differentiated from params by their position
func_def_flags = @{ ("transform" | "aggregation" | "window" | "")? }
```

prql-compiler/src/prql.pest
```diff
@@ -11,30 +11,21 @@ WHITESPACE = _{ " " | "\t" }
...
-transformation = { WHITESPACE* ~ ident ~ named_expr_simple* }
+pipe = _{ NEWLINE | "|" }
+pipeline = { WHITESPACE* ~ func_curry ~ (pipe ~ func_curry)* }
 
-// We don't model this in the same way as a pipeline, since the first term isn't
-// a transformation. We could unify them at some point.
-inline_pipeline = { "(" ~ expr ~ (pipe ~ func_curry)+ ~ ")" }
-
-ident = @{ !operator ~ (ASCII_ALPHA | "$") ~ (ASCII_ALPHANUMERIC | "." | "_" | "*" )* }
+ident = @{ !operator ~ !keyword ~ (ASCII_ALPHA | "$") ~ (ASCII_ALPHANUMERIC | "." | "_" | "*" )* }
```
max-sixty, line 29:
```suggestion
// `$` is for SQL params
ident = @{ !operator ~ !keyword ~ (ASCII_ALPHA | "$") ~ (ASCII_ALPHANUMERIC | "." | "_" | "*" )* }
```

prql-compiler/src/prql.pest
```diff
@@ -101,19 +103,16 @@ range = ${ range_edge ~ ".." ~ range_edge }
...
 
-// These separate idents from each other — i.e. `foo bar and baz` will parse into
-// [`foo bar`, `and`, `baz`], where `foo bar` is presumably a function call.
-// TODO: review operators
-operator = ${ "and" | "or" | "not" | "==" | "!=" | "=" | ">" | "<" | ">=" | "<=" | "+" | "-" | "*" | "/" | "%" }
+operator = _{ operator_unary | operator_mul | operator_add | operator_compare | operator_logical }
+operator_unary = ${ "-" }
+operator_mul = ${ "*" | "/" | "%" }
+operator_add = ${ "+" | "-" }
+operator_compare = ${ "==" | "!=" | "=" | ">" | "<" | ">=" | "<=" }
```
max-sixty, line 112: I guess if we're doing these properly now, we could exclude `==`? (or close and we can assess separately if this isn't unambiguous)

```suggestion
operator_compare = ${ "!=" | "=" | ">" | "<" | ">=" | "<=" }
```

prql-compiler/src/prql.pest
max-sixty: ~Are there any examples of how these work? I see these as `FuncKind`s, but I couldn't see any practical examples. I'm presuming they're for giving the compile information about what the function is doing?~

I found the `transforms.prql`; I had been looking for `aggregation` examples

prql-compiler/src/semantic/resolver.rs
```diff
@@ -53,169 +73,359 @@ impl AstFold for Resolver {
...
+                column: (s.column.declared_at)
+                    .ok_or_else(|| anyhow!("Unresolved ident in sort?"))?,
+                direction: s.direction,
+            })
+        })
+        .try_collect()
+}
+
+/// Loads `internal.prql` which contains type definitions of transforms
+pub fn init_context() -> Context {
```
max-sixty, line 395: WDYT about making this the impl for `Default`? 

---

Zooming out: one thing I worry about is that as the code gets more powerful, it becomes more coupled — it's harder to trace through how a simple query gets parsed, resolved & translated.

I'll think more about this. To some extent, the complication is an inevitable cost of progress, but sometimes there are things we can do so that the simple case is more obvious — possible making default impls is one (I even went through a phase a month ago of reorganizing files so the important functions were at the top of files, as an example of another).

prql-compiler/src/sql/un_group.rs
```diff
@@ -0,0 +1,33 @@
...
+
+use crate::ast::ast_fold::*;
+use crate::ast::*;
+
+pub fn un_group(nodes: Vec<Node>) -> Result<Vec<Node>> {
+    UnGrouper {}.fold_nodes(nodes)
+}
+
+/// Traverses AST and replaces transforms with nested pipelines with the pipeline
+struct UnGrouper {}
```
max-sixty, line 11: I'm trying to think of a better name, so far unsuccessfully...

prql-compiler/tests/integration/compile.rs
```diff
@@ -1,164 +0,0 @@
-use insta::assert_snapshot;
```
max-sixty, line 1: How come we lost these test files? Is it that they're redundant given our unit tests?

reference/comparison-table.css
```diff
@@ -8,10 +8,14 @@ Would be good to fix this, or to make the page wider.
 
 */
 .comparison {
-    table-layout: fixed;
+    display: flex;
+    justify-content: center;
```
max-sixty, line 12: Do these make any practical difference? (I have no strong bond to the existing JS; I'm curious because I'm trying to learn how to make these better)

reference/tests/snapshots/snapshot__run_reference_examples@introduction-0.prql.snap
max-sixty: What's wrong with it now? That there could be a `countries.sum_gross_cost`?

prql-compiler/src/sql/translator.rs
```diff
@@ -99,40 +96,24 @@ pub fn translate_query(query: &Query) -> Result<sql_ast::Query> {
...
 }
 
-pub fn load_std_lib() -> Result<Vec<Node>> {
-    use crate::parse;
-    let std_lib = include_str!("./stdlib.prql");
-    Ok(parse(std_lib)?.nodes)
-}
-
-fn separate_pipeline(query: &Query) -> Result<(Vec<Table>, Vec<Node>, Vec<Pipeline>)> {
+fn into_tables(nodes: Vec<Node>) -> Result<Vec<Table>> {
```
max-sixty, line 105: The convention has been `tables_of_nodes`, though no strong view on the exact convention; it is nice to have one though

prql-compiler/src/sql/translator.rs
```diff
@@ -406,31 +398,26 @@ fn prepend_with_from(pipeline: &mut Pipeline, last_name: &Option<String>) {
...
+        .collect()
 }
 
 impl TryFrom<Item> for SelectItem {
     type Error = anyhow::Error;
     fn try_from(item: Item) -> Result<Self> {
         Ok(match item {
             Item::Expr(_) | Item::SString(_) | Item::FString(_) | Item::Ident(_) | Item::Raw(_) => {
-                SelectItem::UnnamedExpr(TryInto::<Expr>::try_into(item)?)
+                SelectItem::UnnamedExpr(Expr::try_from(item)?)
```
max-sixty, line 420: This does seem nicer...

max-sixty:
Overall — very impressive! Another great step forward.

I've reviewed:
- The high-level design
  - Unifying the funcs & transforms is really cool, very nice job
  - The `group` approach seems excellent
- The tests
- Some specific parts that I went deeper on as I tried to review how the new flow works
  - I get most of `semantic`; I need to spend more time with it. Running it as a single pass sounds great
  - Super re adding `Display`. We'll need to do things like include comments for it to be a full formatter.

I don't have particularly helpful code suggestions at the moment; I'll continue going through the code after we merge, and will engage more with it as I'm adding features.

As I wrote above, I will reflect on keeping the codebase as simple and modular as possible.

I didn't quite get why the test needed fixing, if it's as simple as qualifying a preceding CTE, then that can definitely go into another PR.

Good to go from my end, thank you @aljazerzen !

max-sixty (approved):

reference/tests/snapshots/snapshot__run_reference_examples@introduction-0.prql.snap
aljazerzen: That and the fact that these aggregation functions should not be here. Instead there should be columns referenced by name.

prql-compiler/src/semantic/resolver.rs
aljazerzen: I had that as a default trait, but default is usually used for a few assignes to fields and that's it.

This function actually parses a prql file with type definition for transforms, so it feelt too big for a drfault trait. Note that you dont have to call it yourself, it will get called if you pass `contex = None` to resolve function.

aljazerzen: I agree that modules should be decupled and should have one major function that you can call, and thats it. Thats the objective, that we currently slightly diviate from.

prql-compiler/src/sql/un_group.rs
aljazerzen: UnNester? DeNester? Flatten?

prql-compiler/tests/integration/compile.rs
aljazerzen: Oh yeah, I was fixing these tests and noticed that they are duplicates with unit tests and reference tests.

There is no reason to have duplicate tests if we have to maintain them over langugae changes

reference/comparison-table.css
aljazerzen: I added a container div, so the editor can be wider than the content column.

This is css for centering the table, even when it overflows.

reference/tests/snapshots/snapshot__run_reference_examples@introduction-0.prql.snap
max-sixty: Ah, sorry, I was myopically looking at just that line. Makes sense now

Issue #395, opened 2022-04-22
`cargo build` fails due to "linking with `cc` error"


max-sixty:
Currently the issue that we flagged has a wider impact than I thought — it causes `cargo build` at the workspace level to fail.

Weirdly, `cargo test` doesn't fail, which is how this got through CI. Possibly because this crate has no tests? Either we should add that as a step on the tests, or add some phantom test to this crate.

We should merge things quickly, and also revert things quickly — I'll exclude the `prql-python` crate from being built, and we can solve this issue at a more measured cadence. I'm happy to look into this next week.

```
   Compiling prql-python v0.0.2 (/Users/maximilian/workspace/prql/prql-python)
error: linking with `cc` failed: exit status: 1
  |
[...]

PR #399, opened 2022-04-22
Try adding back `prql-python`

max-sixty:
It sounds like at least the `cargo build` error might be due to my local setup, so adding a `cargo build` step to GHA and adding back `prql-python` crate

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -1,0 +1,15 @@
+[target.x86_64-apple-darwin]
+rustflags = [
+  "-C",
+  "link-arg=-undefined",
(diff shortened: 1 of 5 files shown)


max-sixty:
Adding the `py03` in a ` [target.'cfg(not(target_family="wasm"))'.dependencies]` block solves the wasm clippy issue. I will PR that separately.

Weirdly, the only platform this breaks on is Mac! So that's consistent with it breaking for me locally but not everyone.

max-sixty:
Closes https://github.com/prql/prql/issues/395

PR #400, opened 2022-04-22
Excluded py03 on wasm

max-sixty:


prql-python/Cargo.toml
```diff
@@ -7,6 +7,8 @@ edition = "2021"
 name = "prql_python"
 crate-type = ["cdylib"]
```
manning79, line 9: Didnt no that was possible, is this a standard rust / cargo feature ?!

max-sixty: Yes! We do it already in `prql-web`!

PR #401, opened 2022-04-23
Reenable prql-python workflow

max-sixty:


max-sixty:
@charlie-sanders  I'll merge but comment if I missed anything

PR #405, opened 2022-04-23
Adding to_json in prql-compiler/src/lib , and exposing to python

qharlie:
@max-sixty There are some python tests in prql-python/python/tests that would be nice to have run before the Release phase as a check before uploading, I could not get them to run correctly.  Pinging @rbpatt2019 also he might know how to integrate running the python tests before Release in .github/workflows/python.yaml


max-sixty:
PR looks good!

Why wouldn't those tests run correctly? What happens if we try running pytest on them?

max-sixty (approved):
Looks good!

It's also fine to put rust functions in the `prql-python` crate and then over time promote them into `prql-compiler`. `to_json` seems very reasonable to have in `prql-compiler` already though.

rbpatt2019:
Can definitely add a pytest step to this. ~~Does that need to come before or after the builds? My recommendation would be before, if possible - save ourselves the build time if the tests don't pass.~~ Answered my own question! we have to build, then install the module.

rbpatt2019:
Once I build locally, the tests fail complaining that there is no prql_python.compile method. Think we need to export this directly to pyo3 if you want to call it in python.

qharlie:
@rbpatt2019 after the local build are you calling 'maturin develop' ?  Thats what builds the python module, its happens with this step 
```
      - uses: messense/maturin-action@v1
        with:
          manylinux: auto
          command: build
          args: --release -o dist -m prql-python/Cargo.toml
```
, so we probably need a pytest after that for each platform.

@max-sixty the workflow would completely stop running when I was trying to develop the workflow , let me try again might have been a fluke.

rbpatt2019:
Yep. I build the wheel, install the wheel, `import prql_python as prql`, and then get `module prql_python has no attribute compile`.

Also, can't push commits, but I can definitely use a composite action to refactor the action into a single matrix covering all OS and python versions.

rbpatt2019:
Locally, I've got everything resolved (action for multiple os build, multiple python test) excepting the error. Just need to expose compile in some fashion...put I'll need push permission for the PR to share all this.

qharlie:
Oh in the python bindings they were renamed to `to_sql`, when I added the `to_json` method @rbpatt2019 , should we add a compile as well ?  Can you dir(prql_python) see what methods are defined on it ? 

Edit:  Oh I see the test is calling compile() my bad.  Ive added you as contributor to the charlie-sanders/prql fork if you want to push those changes :+1:

qharlie:
Closing this so not to dirty the Actions with my fumbling about

max-sixty:
> Closing this so not to dirty the Actions with my fumbling about

There's no dirtying! It's totally fine to do whatever is helpful on any non-main branch!

rbpatt2019:
Sorry I missed the bug here! What was causing the issues with the actions?

rbpatt2019:
I've fixed the python tests - specified the wrong directory for installing `prql_python`. They are pushed [here](https://github.com/charlie-sanders/prql).

.github/actions/setup_nox/action.yml
```diff
@@ -0,0 +1,33 @@
...
+        python-version: "3.10"
+    - name: "Install nox"
+      run: pipx install nox
+      shell: bash
+    - name: "Cache Nox"
+      uses: actions/cache@v2
+      with:
+        path: .nox
+        key:
+          nox-${{inputs.cache_number }}-${{ hashFiles('prql-python/noxfile.py') }}-${{ hashFiles('prql-python/requirements.txt') }}
```
max-sixty, line 33: FYI sometimes there are publicly available actions that embed this logic; e.g. the rust cache one we use does this. Not sure if there is a `nox` one though

.github/actions/build/action.yaml
```diff
@@ -0,0 +1,35 @@
+name: build
+description: "Use maturin to build python dists of prql-python. Conditional if checks ensure only the correct OX build is called."
```
max-sixty, line 2: `OS`!

.github/actions/setup_nox/action.yml
```diff
@@ -0,0 +1,33 @@
+name: Setup Nox
+description: 'Prepare and cache nox, install poetry.'
+
+inputs:
+  cache_number:
+    description: "A manually specified cache number. Useful for triggering new caches."
+    required: True
+    default: 0
+
+runs:
```
max-sixty, line 10: Interesting — maybe `test-cargo.yml` should be an action like this, rather than a workflow

.github/workflows/python.yml
```diff
@@ -23,74 +23,37 @@ concurrency:
...
+      - name: Run tests
+        run: nox -s tests -f prql-python/noxfile.py
   release:
     name: Release
     runs-on: ubuntu-latest
-    # TODO: Upload PYPI_API_TOKEN, confirm how we release all crates together.
-    # if: "startsWith(github.ref, 'refs/tags/')"
-    if: false
-    needs: [macos, windows, linux]
+    if: "startsWith(github.ref, 'refs/tags/')"
```
max-sixty, line 47: This should presumably be `false` until we upload the token?

prql-compiler/src/lib.rs
```diff
@@ -45,3 +50,19 @@ pub mod internals {
...
+    #[test]
+    fn test_to_json() -> Result<()> {
+        let json = to_json("from employees | take 10")?;
+        // Since the AST is so in flux right now just test that the brackets are present
+        assert_eq!(json.chars().next().unwrap(), '{');
+        assert_eq!(json.chars().nth(json.len() - 1).unwrap(), '}');
+
+        Ok(())
+    }
+}
```
max-sixty, line 68: Great rust @charlie-sanders !

prql-python/python/tests/test_all.py
```diff
@@ -0,0 +1,21 @@
...
+
+    # Since the AST is so in flux lets just take these dont throw exceptions
+    res = json.loads(prql.to_json(prql_query))
+    assert res is not None
+
+    res = prql.to_sql(prql_query)
+    assert res is not None
+
+
+if __name__ == '__main__':
```
max-sixty, line 20: I haven't used `nox` much, but traditionally in pytest, this isn't required — pytest will collect and run the tests

prql-python/requirements.txt
```diff
@@ -0,0 +1,3 @@
+pytest==7.0.0
```
max-sixty, line 1: FYI I've used poetry very successfully for defining requirements for the past few years (a bit heavier though, so this is maybe better at the moment)

prql-python/src/lib.rs
```diff
@@ -15,9 +15,21 @@ pub fn to_sql(query: &str) -> PyResult<String> {
     }
 }
 
+#[pyfunction]
+pub fn to_json(query: &str) -> PyResult<String> {
```
max-sixty, line 19: Really nice!

max-sixty (commented):
Really great, thanks @charlie-sanders & @rbpatt2019 

I left some nits for information, def no actions needed

.github/actions/setup_nox/action.yml
rbpatt2019: So the default setup nox action that is available doesn't do caching, and is currently unmaintained with a breaking error. So I've created one of our own to bypass this!

.github/actions/setup_nox/action.yml
rbpatt2019: Could be! I'm a big fan of composite actions - they make things so much more dry.

.github/workflows/python.yml
rbpatt2019: Yep I'm currently altering this. The problem is that `GitHub.ref` contains whatever triggered the build - so if we don't trigger on tags, then this will never evaluate to true.

prql-python/python/tests/test_all.py
rbpatt2019: Yep, the `if` statement is not required for nox. All it's doing is running purest manually in isolated virtualenvs of different versions.

prql-python/requirements.txt
rbpatt2019: I thought about poetry as well! It's my go to - but the reqs list was so small that I thought it was heavier. I'm also not sure how poetry would interact with Maturin - both require build sections, and I don't think pyproject.toml allows for more than one. I'll investigate!

.github/actions/setup_nox/action.yml
max-sixty: I just changed the cargo one to this design!

PR #429, opened 2022-04-27
Refactor semantic module & materializer

aljazerzen:


reference/tests/snapshots/snapshot__run_reference_examples@list-equivalence-2.prql.snap
```diff
@@ -6,6 +6,6 @@ input_file: reference/tests/examples/examples/list-equivalence-2.prql
 SELECT
   employees.*,
   salary + payroll_tax AS gross_salary,
-  salary + payroll_tax + benefits_cost AS gross_cost
+  gross_salary + benefits_cost AS gross_cost
```
max-sixty, line 9: Will this work?

max-sixty (commented):
Looks good! (though without having gone through every line...)

I left one comment — are we handling variables correctly there?

reference/tests/snapshots/snapshot__run_reference_examples@list-equivalence-2.prql.snap
aljazerzen: Ah. No it won't.

PR #431, opened 2022-04-27
Reference book tweaks & examples

aljazerzen:
Added examples to docs sidebar, tweaks to CSS

This was the problem:
image

Now it looks like this:
image


max-sixty:
I'm sure I've done things badly and there are ways of improving, but I'm not sure this looks great either:

image

and then if I scroll right:

image

max-sixty:
The existing mode has a scrolling issue with wide columns:

image

I had tried widening the max-width, which helped, but not completely...

One option is to adjust the examples by "inlining" the comments so the columns are much less wide.

---

One difference between out two views on the current impl — it looks like if one column has less text, the table background doesn't fill down — is that correct?

I tried it on Chrome (I generally use Firefox) and I see something similar — an ugly scroll bar across the cell. But the background does continue after.

Your screenshot:
image



My chrome screenshot:
image


[My firefox screenshot is above]

aljazerzen:
Yes, the reason I replaced `table` with `div display: flex` is non-full height of the code blocks.

I really don't like many small scrollbars in a page and prefer one for the whole page, but that is my preference. 

Also main content is easier to digest if the column is narrower.

With all these constraints there is only solution you suggested: make our code narrower by putting comments on their separate lines.

Or we can ditch side-by-side and make it wrap if the content is to big.

max-sixty:
Yeah, I agree re the small scrollbars being bad. 

I'll try and reduce the width of examples.

aljazerzen:
I made the code blocks wrap the lines, so they are one below the other on smaller screens. 

On larger screens, they are still side-by-side.

max-sixty:
@aljazerzen unfortunately this doesn't look great on non-wide screens — it's impossible to see the left side of the text, even with scrolling.

I'm happy to spend some time reducing the width of the examples and adding a test to keep them narrow. But we at least want people to be able to scroll when they're on narrower screens — WDYT?

I'm also happy to ask a friend who knows HTML to help...

image

aljazerzen:
I set a breakpoint for wrapping at screen size of 1400px. This is enough that HD screens have it side by side and that the code is always visible *if the sidebar is closed*:
image

It can be set to 1650px, so the code is visible regardless of the sidebar.

The problem here is that this breakpoint is dependent on width of the code blocks, so it may have to be adjusted when we change the book content.

It would be ideal if the breakpoint could be removed, so it would wrap when the blocks are too wide.

Another thing that may be wrong with this PR is this:
image

Some pages with too wide code blocks are not consistent with each other.

max-sixty:
OK great. Maybe I work on reducing the column width and then we take another pass at this.

PR #435, opened 2022-04-28
Window functions

aljazerzen:
```elm
from daily_orders
sort day
group month (
  sort num_orders
  derive rank
)
derive num_orders_last_week: (num_orders | lag 7)
```

now compiles into:
```SQL
SELECT
  daily_orders.*,
  RANK() OVER (
    PARTITION BY month
    ORDER BY
      num_orders
  ) AS rank,
  LAG(num_orders, 7) OVER () AS num_orders_last_week
FROM
  daily_orders
ORDER BY
  day
```

TODO:
- [ ] window frames via `window ()` context
- [ ] declare functions like `SUM` as both `aggregate` and `window`


max-sixty:
Really cool!! Congrats! (I only just got the notification on GH, weirdly)

Is the proposed design to have functions by default create a default window?

I think that probably makes sense for some functions where there isn't a window clause, like `lag`. But less so for functions like `sum` that conceptually require one, even if SQL defaults to `ROWS UNBOUNDED PRECEDING TO CURRENT ROW`[^1]

[^1]: we would represent this with something like `window ..current` or some syntax to represent `current`.

[BQ](https://cloud.google.com/bigquery/docs/reference/standard-sql/analytic-function-concepts?hl=en#navigation_function_concepts) defines three types:
- Navigation — e.g. lag, first_value
- Numbering — e.g. rank
- Aggregate — e.g. sum, avg

"window_frame_clause" is allowed for:
- all Aggregate functions
- some Numbering functions. 

It's maybe too manual, but I would propose that the end state has two distinct groups:
- Those that require a `window` transform — like `sum`. There's no concept of `sum` without a window. This is what I really like about our nested pipelines — the sum function will be orthogonal within a `window` transform as well as a `group` transform and on a whole column.
- Those that don't require, and probably don't allow for, a window transform — like `lag`. There's no concept of lag with a window; it's just coded that way because of SQL's inflexibility.

PR #442, opened 2022-05-01
Equals

### LINKED ISSUE #437: Operators for equality comparison, assigns and type annotations
Current syntax:

```elm
func add x to:1 = x + to

from employees
filter country = "USA"
derive [
  gross_salary: add salary payroll_tax,
  gross_cost: add gross_salary benefits_cost
]
join blah side:5 [blah_id]
group [title, country] (
    aggregate [
        sum_gross_cost: sum gross_cost,
        count: count,
    ]
)
```

On discord, we are having the following discussion:

aljazerzen:

> I've been working on window functions, the first step of which now works!
> Now I've been thinking how to move forward with function flags as scalar,  aggregation or window, but I noticed that this could be solved by type annotations on the parameters:
> - func transform filter conditions frame = ...
> + func filter conditions frame: Frame = ...
> 
> Here I had transform as a keyword (similar to public / static / final in java) to annotate that this is a transform of frames (tables). Instead, I we can just use type annotations on parameters.
> As @snth pointed out, : is often used for type annotations so it was my natural choice. Even if this will be used only in stdlib, I think it could be consistent with other modern languages.
> For that to work, I tried replacing current usage of : and = with = and == .  = is also used for function definitions which I replaced with ~
> This is the result:
> ```
> func add x to=1 ~ x + to
> 
> from employees
> filter country == "USA"                           # Each line transforms the previous result.
> derive [                                         # This adds columns / variables.
>   gross_salary = add salary payroll_tax,
>   gross_cost = add gross_salary benefits_cost    # Variables can use other variables.
> ]
> join blah side=5 [blah_id]
> group [title, country] (
>     aggregate [                                  # `by` are the columns to group by.
>         sum_gross_cost = sum gross_cost,
>         count = count,
>     ]
> )
> ```
> I know we have been over this multiple times, so I'll just throw it out there one last time:
> Current approach: `:` `=`      `func sum = ...`
> Equals approach: `=`  `==`  `func sum ~ ...`

snth:

> I'm definitely a fan of = for assignments in the derive clauses.
> The ~ for function definitions looks unfamiliar to me. How about using the ~ for the default values instead so you'd get something like: 
> ```elm
> func add x to~1 = x + to
> ```

> I was curious if you really need the == since in the filter clause you could probably get away without it but I guess if you want to do > boolean tests within assignments then you'll need it:
> 
> derive [ is_zero = (divisor == 0) ]

max-sixty:
> I am open to changes!
> 
> Personally, I don't find the `~` intuitive in `func add x to=1 ~ x + to` — I find splitting the function between the params and the body visually not easy. (This is also why I think Malloy's use of is isn't great — that split should be a specific piece of punctutation)
> 
> If we really have to, we could give up = as a comparison. Its advantage is that it's intuitive, and feels less programmy than ==. But it's not a religious issue.
> 
> If we want to add annotations, could we use something else than `:`? I don't think users will be using annotations much, so it's fine to make them ugly. We could use something like `func<frame>` though it's a similar-syntax-but-different-meaning to rust's use of `<foo>`.
> 
> To throw something else into the mix: the current syntax of using `=` in `table foo =` and `func foo x =` isn't completely coherent with the rest of the language; it would be more coherent to have `table foo:` and `func foo x`: — but that would conflict with named args for func.
> 
> So we really have 4 sorts of punctuation here:
> - `:` for named args; `interp lower:0 1600 sat_score`
> - `:` for assigns; `derive temp_c: (temp_f | celsius_of_fahrenheit)`
> - `=` for "fundamental assigns"; `func interp lower:0 higher x = (x - lower) / (higher - lower)`
> - annotations

aljazerzen:

> I agree that `func add x to=1 ~ x + to` does provide nice visual split between params and the body.

> For type annotations there are (a far as I know) only two standards: int a and a: int. So I would prefer to use one of those even if it's gonna be mainly used in our stdlib. When people lookup definition of, let's say, join transform, we can also use these annotations to convey the types.
> 
> I agree that table foo =  and func foo x =  are conceptually assigns, so to be coherent they should have same punctuation as assigns do. But I think that should be = instead of :. 
> 
> It's just that function definitions then have ambiguous parsing. 
> 

What about `func (add x to=1) = x + to` ? This is similar to how you (often must) call a function. 

Or  `func add x (to=1) = x + to`?

I also agree that function definitions have to be easy to parse visually, which means that `~` and `is` are not good separators.

aljazerzen:
Closes #437


reference/src/examples/variables.md
```diff
@@ -13,11 +13,29 @@ group [title, country] (                     # For each group use a nested pipel
...
 sort sum_gross_cost
 filter ct > 200
 take 20
 ```
+
+```prql
+from employees
+group [emp_no] (
+  aggregate [
+    emp_salary = average salary     # avg_salary should resolve to "AVG(salary)" (from stdlib)
```
max-sixty:
```suggestion
    emp_salary = average salary     # average salary resolves to "AVG(salary)" (from stdlib)
```

max-sixty (commented):
I didn't actually realize we wanted to change all the `:` to `=`! 

Rehashing the list from https://github.com/prql/prql/issues/437 (and adding on another item)

> So we really have 4 sorts of punctuation here:
> 
> * `:` for named args; `interp lower:0 1600 sat_score`
> * `:` for aliases; `from d:department_employees` [^1]
> * `:` for assigns; `derive temp_c: (temp_f | celsius_of_fahrenheit)`
> * `=` for "fundamental assigns"; `func interp lower:0 higher x = (x - lower) / (higher - lower)`
> * annotations — maybe `::`?
[^1]: added

I was thinking we would change the named args from `:` to `=`, but not for aliases & assigns. (And sorry if I was unclear. I hope asking this question is reasonable now given its importance. I realize @aljazerzen you've done most of the work for the past two weeks and I _really_ don't want to be getting in your way)

A framework for thinking through this:
1. Do we think `=` is better for each of named-args, aliases & assigns, and worth giving up using `=` for equality? I had preferred `:` for at least the assigns. If we do think `=` is always better, then we should merge this PR.
2. Do we think they should all be the same punctuation? 
    - I marginally thought so because, they were fairly similar. 
    - But there are semantic differences between a) named-args vs. b) aliases & assigns
      - named-args must correspond to a specific parameter — `lower:0` must be `lower` and not `l`!
      - aliases & assigns have any value in the lvalue `d:department_employees` or `dep:department_employees` are both legit
      - When I write aliases, I sometimes write `department_employees:d` because of this issue (but maybe that's just me)
    - If we do think they should all be the same punctuation and we can't have `:` because of type annotations, then we should merge this. Otherwise we should possibly _only_ change for named args.

The thing I _wouldn't_ weigh highly is having `:` as a type annotation — even if we use `=` extensively, I wouldn't vote to use a single `:` as a type annotation — I can't foresee them being used in user code enough to justify them taking such an important piece of punctuation. and the only SQL type annotations I could find are Snowflake's, which uses `::`. I do understand that `func add::number x::number to::number:1  = ...` is difficult to read because it mixes `::` and `:` together.

aljazerzen:
When writing PRQL I think of assigns, aliases and named args the same way: take r value and store it into l value.

The reason that you can arbitrarily change l value for assigns and not for named args, is that named args are expected and assigns are not.

It's so similar use case, that I think is should use the same punctuation. 

When I first saw PRQL, `:` was the least readable part of the language. Otherwise, one can start reading and understand how the underlying table is manipulated. Maybe that's my bias because I mostly know languages that already use `=` as assign operation, but after @snth expressed the same opinion, I think that `=` should be used for all named expressions.

For type annotations I agree, we can use `::` instead of `:`, since it won't be used that much. It could be only `:`, but let's leave that for any future needs.

To be honest, `func add::number x::number to::number:1 = ...` was my excuse to get started on this. But I do have justified reasons for changing `:` to `=`. Sorry for pushing my agenda by opening pull requests :D

max-sixty:
> Sorry for pushing my agenda by opening pull requests :D

Ha! This is as good a reason as any.

OK, I don't have a super-strong POV here.

Given the size of the change, could we ask for unfiltered thoughts from @charlie-sanders and @rbpatt2019 ? And @snth? (and anyone else who is interested)

snth:
First of, let me preface my comments with saying that I am writing as an expectant user of the language, so I don't know too much about what happens under the hood in terms of lvalues, rvalues, etc... . That said I will probably veer off into some philosophical comments as well.

I find the syntax in this PR much easier to read, i.e. I prefer the `=` over `:` in the cases addressed in this PR.

Since I first got involved in the discussion around this, I have also thought a bit about the semantics of it and I think that philosophically it is also correct. In something like
```elm
derive [ gross_salary = salary + payroll_tax ]
```
while you may think of it as an "assignment", it is an equality in that "gross_salary" is synonymous with "salary + payroll_tax" and you can use either in any expression and evaluate to the same thing.

Similarly in 
```elm
join de=dept_emp [emp_no]
```
"de" and "dept_emp" refer to the same table and can be used interchangeably (with the caveat that "de" is more specific than "dept_emp" if the "dept_emp" table is joined multiple times).

snth:
I think my previous comment could be extended to `filter` clauses and `join` conditions.

So for example
```elm
filter country_code = "USA"
```
to me would make sense because while above that line "country_code" is not synonymous to "USA", after that line it is and you can use the terms interchangeably.

Similarly for
```elm
join dm=dept_manager [dm.dept_no=de.dept_no]
```
, after that line you are guaranteed that `dm.dept_no` is the same as `de.dept_no`, i.e. they're equal. In terms of relational algebra, you know that you are restricting yourself to some diagonals of the cartesian product of the tables where the operands are equal (caveat again being that with outer joins one of the two sides may be null).

I know this isn't quite the syntax we have proposed but to me this would make sense.

snth:
The one big exception is equality testing in expressions.

So if I have
```elm
derive [ is_manager = job_title == "manager" ]
```
here it is not the case that `job_title` and "manager" are synonymous and the point of the expression is to test where they are and where they are not.

So here there is an active test being performed. I'm happy with the `==` notation here since that seems to be the commonly accepted one. Given how in my mind this is almost asking a question of the data I also thought about what about something like `?=` for this, e.g.
```elm
derive [ is_manager = job_title ?= "manager" ]
```
To me this almost makes it clearer that you are asking the data whether this is the case or not.

snth:
I'm not sure where I stand in terms of named args. If the argument is given then there is indeed an equality. If the argument is supplied then that's not the case. 🤷‍♂️ Having the `=` here is pretty well established in most languages so happy to go with it.

For type annotations a single `:` would be nicer in my opinion as that seems to be the common option in modern languages, Rust, Typescript, Python, are the ones that come to my mind.

The function syntax is a whole topic in itself and I want to comment on that elsewhere.

aljazerzen:
https://github.com/prql/prql/pull/442#issuecomment-1115887748

Oh, that is a bold move with `?=`. It's true that is better conveys what we are doing, but it's quite unconventional.

aljazerzen:
https://github.com/prql/prql/pull/442#issuecomment-1115873857

> join dm=dept_manager [dm.dept_no=de.dept_no]

I see the relation algebra argument here. But assign operation is not entirely equivalent to equality statement:
```
from employees
select first_name
select name = first_name
```
... from this point on, `name` will reference `employees.first_name`. But `first_name` will not be in scope anymore and may even get a new value assigned. 

Conceptually, assign is very similar to mathematical equality statement, but when writing code, there is a semantic difference. I know that many mathematicians struggle to accept that in programming, equality is expressed with `==`, but the truth is that assignment is much more common than equality comparison. That's why I believe that C-style notation got this right, and that R and Haskell are fighting an-uphill-battle.

qharlie:
> ```elm
> derive [ is_manager = job_title ?= "manager" ]
> ```
Really liking the look of that!  

>  [replace : with = ]

I was on the fence at first but after seeing these examples , the aliases in particular seem easier to read as an assignment.  
I like having named parameters and default values use `:` , especially if we are not using that for type annotations.

```elm
func my_func some_parm:21 = f''
derive bar my_func some_parm:32
```

max-sixty:
I appreciate @snth 's comments; to some extent that brings us to the conceptual discussion of equality vs identity.

> it is an equality in that "gross_salary" is synonymous with "salary + payroll_tax" and you can use either in any expression and evaluate to the same thing.

Yes, though let me adjust the terms slightly — when we say `gross_salary = salary + payroll_tax`, we're saying it's an _identity_ — i.e. it holds for every value of those variables — like `x^2 - y^2 = (x+y)(x-y)`. 
This sometimes uses the `≡` symbol in math (though I don't propose to go full APL and start adding unicode...)

When we say `job_title == "manager"` the `job_title == "manager"` is an _equality_ — we're saying for _some_ values of those the terms are the same, like `x^2 - y^2 = z^2`

---

> Oh, that is a bold move with `?=`. It's true that is better conveys what we are doing, but it's quite unconventional.

Similar sentiments — actually cool — but I'm not sure this is where we should spend our "novelty budget"

---

Another dimension I'm grappling with: whether we're defining a new object or not. Copying the list from above:

> * `:` for Named args; `interp lower:0 1600 sat_score`
> * `:` for Aliases; `from d:department_employees` [^1]
> * `:` for Assigns; `derive temp_c: (temp_f | celsius_of_fahrenheit)`
> * `=` for "Fundamental assigns"; `func interp lower:0 higher x = (x - lower) / (higher - lower)`
> * `::` for Annotations, maybe?

- Aliases & Assigns are creating something new: `gross_salary: salary + payroll_tax` — it's me choosing `gross_salary`.
- Named args are passing a value — `lower:0` — I'm not choosing the term `lower`.

Possibly I'm more sensitive to this than others, and so treating those differently would be overfitting to my own mind. The existing state was OK, but if we want to convert some of these to `=`, there is a coherent state of having Aliases & Assigns as different from Named args.

---

> For type annotations a single `:` would be nicer in my opinion as that seems to be the common option in modern languages, Rust, Typescript, Python, are the ones that come to my mind.

I quite strongly think that we should put very little weight on the presentation of type annotations. If there is any tradeoff of having type annotations better vs. something else in the language being better, we should weigh on the side of "something else". It should have a similar use to writing macros in Rust — something that most people never learn how to do, even though they reap the benefits every day. Types are really helpful and I've changed my mind on how important they'll be, but the compiler can do almost all the work, and normal user PRQL code can stay clean.

(I'm not saying others are saying anything else, and lmk if there are cases where the compiler wouldn't be able to work something out, given a table schema and code, and users would need to supply annotations)

---

I'm not sure where that leaves us — what do people think are viable options?

reference/tests/snapshots/snapshot__run_reference_examples@employees-0-0.prql.snap.new
```diff
@@ -0,0 +1,23 @@
...
+join t=titles [emp_no]
+join dept_emp side:left [emp_no]
+group [dept_emp.dept_no, t.title] (
+  aggregate [avg_salary = average emp_salary]
+)
+join departments [dept_no]
+select [dept_name, title, avg_salary]
+`;   --> 31:9
+   |
+31 | func lag<column> offset column ->  s"LAG({column}, {offset})"␊
```
max-sixty: FYI there are some unaccepted insta files with errors!

prql-compiler/src/sql/stdlib.prql
```diff
@@ -28,5 +28,5 @@ func in range value ->  s"{value} BETWEEN {range}"
 func as column type ->  s"CAST({column} AS {type})"
 
 # Window functions
-func window lag offset column ->  s"LAG({column}, {offset})"
-func window rank ->  s"RANK()"
+func lag<column> offset column ->  s"LAG({column}, {offset})"
```
max-sixty, line 31: I think these look great!

max-sixty (approved):
Overall this looks excellent, thank you v much as ever.

(I added one comment about some snapshot tests showing errors, lmk if you want me to look into those more)

max-sixty:
Great work @aljazerzen 

I think we're reached stability! :)

aljazerzen:
Jup, we've discussed all big parts of the language so changes from now on will all be lesser.

Issue #443, opened 2022-05-02
Code blocks in reference book


aljazerzen:
See #431

max-sixty:
I'm sending this out to people — I think I have to make some edits even if at the cost of the scroll bars! WDYT? I know you did some work on this so I don't want to be blindly reverting.

<img width="1404" alt="image" src="https://user-images.githubusercontent.com/5635139/166613089-931ac7dd-f80b-497d-bf43-d6f393230c6b.png">

Possibly we make the main page expand to fill the space, have the language blocks expand too, and have scroll bars in the code if it's too narrow — then they only show on thinner screens. 

And then work to make the code in the code blocks have shorter lines.

aljazerzen:
Oh no, that's terrible.

What about just having them wrap one below the other as soon as they are too wide?

I know that there is plenty space around, but narrow columns have a purpose of limiting the amount of information displayed and wide code blocks are violating that principle. That's why I had a hard time making the wider...

max-sixty:
> What about just having them wrap one below the other as soon as they are too wide?

That's an excellent idea, if it's possible! I probably need to do more on HTML, since they're currently structured as rows & columns, and presumably those won't easily wrap.

PR #446, opened 2022-05-04
Reference wrapping

aljazerzen:
#443


max-sixty:
Amazing — that looks very respectable indeed!

image

PR #453, opened 2022-05-05
Fix string escapes, refactor

### LINKED ISSUE #440: Escape single quote characters in strings
Following up to https://github.com/prql/prql/pull/439, Currently

```elm
derive x: "two households'"
```

compiles to:

```sql
SELECT
  ' two households ' '' AS x
```

Whereas I think it should compile to:

```sql
SELECT
  'two households\'' AS x
```

I'm actually not sure why it has spaces around it, and where `' ''` is coming from.

aljazerzen:
Closes #440

### DIFF
--- a/prql-compiler/src/sql/translator.rs
+++ b/prql-compiler/src/sql/translator.rs
@@ -12,11 +12,10 @@ use anyhow::{anyhow, bail, Result};
 use itertools::Itertools;
 use sqlformat::{format, FormatOptions, QueryParams};
 use sqlparser::ast::{
-    self as sql_ast, Expr, Function, FunctionArg, FunctionArgExpr, Join, JoinConstraint,
-    JoinOperator, ObjectName, OrderByExpr, Select, SelectItem, SetExpr, TableAlias, TableFactor,
-    TableWithJoins, Top, WindowSpec,
+    self as sql_ast, DateTimeField, Expr, Function, FunctionArg, FunctionArgExpr, Join,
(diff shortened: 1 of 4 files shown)


prql-compiler/src/sql/translator.rs
```diff
@@ -1438,4 +1437,40 @@ take 20
...
+            y = 'two households"',
+            z = f"a {x} b' {y} c",
+            v = f'a {x} b" {y} c',
+        ]
+        "###,
+        )?;
+
+        assert_display_snapshot!((resolve_and_translate(query)?), @r###"
+        SELECT
+          'two households''' AS x,
```
max-sixty, line 1456: FWIW I'm not sure this is quite right. At least it doesn't compile on BigQuery. (not saying it's on you, tbc) 
image

aljazerzen:
Indeed, the "two single quotes" escaping only works for [Oracle, SQL Server, MySQL and PostgreSQL](https://www.databasestar.com/sql-escape-single-quote/).

BQ uses [backslash escaping](https://cloud.google.com/bigquery/docs/reference/standard-sql/lexical) and also some other approaches.

Currently we cannot produce anything other the "two single quotes" escaping, because sqlparser does not support it (see `sqlparser::ast::value:183`). We will need a hack on out side or a PR on sqlparser.

max-sixty:
Ah great, TIL!

Yeah, it's a shame BQ isn't supported by sqlparser. It's nice to test because it's very easy to open a browser tab, but probably we need to use sql fiddle in these cases.

PR #461, opened 2022-05-07
Add Operator to AST

max-sixty:
Rather than just parsing as `Raw`. This will let us have decent code for
the `sort (-amount)` case.


aljazerzen:
Nice

PR #471, opened 2022-05-09
Allow passing a range to `take`

max-sixty:
This is a WIP. Ideally we can do

```
from employees
take 5..10
```

And it'll translate to a `LIMIT` & `OFFSET`

Passing `take 10` would be the same as passing `take 1..10`.


prql-compiler/src/sql/translator.rs
```diff
@@ -389,16 +404,52 @@ fn prepend_with_from(pipeline: &mut Pipeline, table: Option<TableRef>) {
     }
 }
 
+/// Aggregate several ordered ranges into one, computing the intersection.
+///
+/// Returns a tuple of `(start, end)`, where `end` is optional.
+fn range_of_ranges(ranges: Vec<Range>) -> Result<(i64, Option<i64>)> {
```
max-sixty: This seems like a very complicated implementation for what it's actually doing, but I'm not sure of an easier one...

prql-compiler/src/sql/translator.rs
```diff
@@ -667,6 +719,96 @@ mod test {
...
+
+        assert!(range_of_ranges(vec![range1.clone()])?.1.is_some());
+
+        assert_debug_snapshot!(range_of_ranges(vec![range1.clone()])?, @r###"
+        (
+            1,
+            Some(
+                10,
+            ),
+        )
```
max-sixty: Seems very verbose from `rustfmt`...

max-sixty:
Docs:
image

image

prql-compiler/src/sql/translator.rs
aljazerzen: would assert_yaml_snapshot work? I believe tuple and Option have Serializable trait...

prql-compiler/src/parser.rs
```diff
@@ -361,6 +361,39 @@ mod test {
...
+        ---
+        FuncCall:
+          name: take
+          args:
+            - Literal:
+                Integer: 10
+          named_args: {}
+        "###);
+
+        // Currently this parses but doesn't translate.
```
aljazerzen, line 379: Does it not translate?

prql-compiler/src/ast/item.rs
```diff
@@ -97,7 +97,7 @@ impl Range {
         }
     }
 
-    pub fn new_int(start: Option<i64>, end: Option<i64>) -> Self {
+    pub fn from_ints(start: Option<i64>, end: Option<i64>) -> Self {
```
aljazerzen, line 100: That's a better name, thanks!

prql-compiler/src/sql/translator.rs
```diff
@@ -389,16 +404,52 @@ fn prepend_with_from(pipeline: &mut Pipeline, table: Option<TableRef>) {
...
+        let current_length = if let Some(end_box) = range.end {
+            let current_end: i64 = dbg!(*end_box).item.into_literal()?.into_integer()?;
+            Some(current_end - current_start + 1)
+        } else {
+            None
+        };
+
+        // There has to be a more elegant approach than this? It's basically
+        // doing `np.nanmax(a, b)`.
+        length = length.map_or(current_length, |rl| {
```
aljazerzen: These should be equivalent, but none of them is very readable.
```
        length = current_length
            .zip(length).map(|(c, r)| c.min(r))
            .or(length)
            .or(current_length)
```

```
        length = current_length
            .or(length)
            .and_then(|cl| length.map(|rl| cl.min(rl)))
```

prql-compiler/src/sql/translator.rs
aljazerzen: Algorithmically, this is as concise as you can get (I think). But code-wise, this could be compressed, by:
- first mapping all ranges to Rust's `Range<i64>`,
- replacing `cs > length` with a check for negative length at the end,
- replace a the two `if let Some` with `unwrap_or` and `map`

aljazerzen:
Sorry, couldn't help it: I did a refactor of `range_of_ranges`. I just like to rotate code around until it looks nice.

But I found a mistake, best seen in:
```elm
take 101..110
```
```diff
- LIMIT 110 OFFSET 100
+ LIMIT 10 OFFSET 100
```
Limit directly tells how much rows should there be in the result - its applied after offset.

aljazerzen (approved):

aljazerzen:
Ahh. I've written these comments before and they are posted only after I finish the review... oops

prql-compiler/src/utils.rs
```diff
@@ -88,3 +88,24 @@ pub fn diff(a: &str, b: &str) -> String {
...
+pub trait OrMap<T> {
+    /// Merges two options into one using `f`.
+    /// If one of the options is None, results defaults to the other one.
+    fn or_map<F>(self, b: Self, f: F) -> Self
+    where
+        F: FnOnce(T, T) -> T;
+}
+
+impl<T> OrMap<T> for Option<T> {
+    fn or_map<F>(self, b: Self, f: F) -> Self
```
max-sixty, line 101: Great idea

prql-compiler/src/sql/translator.rs
```diff
@@ -329,9 +330,13 @@ fn atomic_pipelines_of_pipeline(pipeline: Pipeline) -> Result<Vec<AtomicTable>>
...
             }
-            // TODO: We'd be OK with two _consecutive_ takes (range_of_ranges),
-            // but they need to be consecutive.
-            "Filter" | "Sort" | "Take" => counts.get("Take").is_some(),
+            "Filter" | "Sort" => counts.get("Take").is_some(),
+
+            // There can be many takes, but they have to be consecutive
+            // For example `take 100 | sort a | take 10` can't be one CTE.
+            // But this is enforced by transform order anyway.
+            "Take" => false,
```
max-sixty, line 338: Thanks for adding this!

reference/tests/snapshots/snapshot__run_reference_prql@ranges-1.prql.snap
```diff
@@ -11,4 +11,4 @@ ORDER BY
   value DESC,
   date
 LIMIT
-  110 OFFSET 100
+  10 OFFSET 100
```
max-sixty, line 14: I guess this was wrong in my code...

prql-compiler/src/parser.rs
max-sixty: Sorry this was old

prql-compiler/src/sql/translator.rs
max-sixty: That is better!

max-sixty:
> Ahh. I've written these comments before and they are posted only after I finish the review... oops

Ha, yeah I do this a lot...

PR #477, opened 2022-05-11
Publish prql-js, update hljs

aljazerzen:
We should probably add a GH action for publishing to NPM...


max-sixty:
Thanks!

+1 re npm action

Issue #54, opened 2022-02-04
Table vs Value types


max-sixty:
@qorrect raises an interesting point on a [Discord thread](https://discord.com/channels/936728116712316989/939181039924944947).

Something I find frustrating about SQL is that a query result can have three types:
- a table
- a column
- a value — one column, one row

Some queries implicitly require one type. For example, in:

```sql
SELECT name
FROM facts
WHERE population = (SELECT MIN(population) FROM facts);
```

...the subquery `SELECT MIN(population) FROM facts` has to be a Value. But it's difficult to know that before running the query
> Here it's a full aggregation, so it is possible to statically check that, but it could be `select population where country = "USA"` which is ambiguous. And I'm not aware of live feedback on SQL tools to show that error even in the unambiguous case

Here's the PRQL of that query, separated into a `table` for effect:

```elm
table min_pop = (
  facts
  aggregate [min population]
)

from facts
select name
filter population = min_pop
```

We could make this be statically checked, by defining a `value` type:

```diff
-table min_pop = (
+value min_pop = (
   facts
   aggregate [min population]
 )
```

...and then having `aggregate [min population, max population]` would fail at "compile time".

Ambiguous subqueries such as `filter country = "USA"` could require a `take 1` to guarantee it's a single value.

qharlie:
Love the idea, and I if we can error out at compile time on things like if the select contains more than one field ( for value types ) that would be great.  Really liking the compile time checks.

max-sixty:
Reflecting on #478, there may be a 4th type — a "Row" type — with multiple columns but only one row. Modifying the example from above:

```sql
SELECT name
FROM facts
WHERE population = (SELECT population FROM min_population);
```

...here `min_population` needs to be a single row, but doesn't need to be a scalar.

PR #478, opened 2022-05-13
Implement window

### LINKED ISSUE #3: Window functions over groups
I've worked with SQL a lot and now already 5 years with Pandas, and sometimes both drive me absolutely mad. In some aspects Pandas in Python is nicer, but things like window functions are absolutely horrible.

Example: 

    orders
    ------
    cell   (a cell on a map)
    hour   (datetime but rounded to hours)
    amount

with some calculations we should produce this:

    orders
    ------
    cell
    hour_index (unix timestamp // 3600)
    hour
    day_of_year
    day_of_week
    week_of_year
    year

And then for each row, calculate

1. A = orders for the same day of week and same hour for the previous 10 weeks.
2. B = calculate mean orders per whole day for the same days of week for the previous 10 weeks.
3. C = calculate total orders yesterday.
4. A * C / B

I've done only part 1 in Pandas, and it's awful.

    dc = data.copy()
    data['mean_orders'] = 0
    for i in tqdm(range(1, 11)):
        dc.week += 1
        data = data.merge(dc[['cell', 'week', 'weekday', 'orders']], on=['cell', 'weekday', 'week'], suffixes=('', '_other'))
        data['mean_orders'] += data['orders_other'] / 10
        data.drop('orders_other', inplace=True, axis=1)

I can think of PostgreSQL equivalent (though window function names and params might not be exactly these):

    select cell, first(orders),
        sum(orders, 10) over (partition by cell, day_of_week, hour order by week) / 10 mean_same_hour_over_week,
        sum(orders, 168) over (partition by cell order by hour) / 7 mean_per_day_over_week,
        sum(b.orders)
    from orders a inner join orders y on y.hour_index between (a.hour_index - a.hour) and (a.hour_index - a.hour - 24)
    group by cell, orders;

...and then put it into a CTE and do the multiplication & division.

### LINKED ISSUE #420: Window functions
Following on from https://github.com/prql/prql/issues/300, we can now implement window functions. 

Much of the infrastructure has already been built by @aljazerzen , including using a similar construction to `group`. There will be some differences though, and it'll require some work in translation. It will be moderate-hard difficulty.

aljazerzen:
I've:
- added `window` transform, including docs,
- added a few window functions to stdlib,
- refactored literals into their own enum, to facilitate int and float literals, 
- removed raw literal,
- added `AnyOf` wrapper for types, so functions such as `sum` can return `column | scalar`
- made functions in `select` and `derive` be window functions and not aggregation

Closes #420
Closes #3


prql-compiler/src/ast/item.rs
```diff
@@ -93,6 +89,21 @@ pub struct Range {
...
+impl Range {
+    pub const fn unbounded() -> Self {
+        Range {
+            start: None,
+            end: None,
+        }
+    }
+
+    pub fn new(start: Option<i64>, end: Option<i64>) -> Self {
+        let start = start.map(|x| Box::new(Node::from(Item::Literal(Literal::Integer(x)))));
```
max-sixty, line 101: Re the `Integer` — I had done a PR to let ranges take any node; e.g. 

```elm
from employees
filter (dob | in @2019-01-01..@2020-01-01)
```

goes to 

```sql
SELECT
  employees.*
FROM
  employees
WHERE
  dob BETWEEN DATE '2019-01-01'
  AND DATE '2020-01-01'
```

If that doesn't work any longer then no great stress; I can modify in a follow up

prql-compiler/src/ast/literal.rs
```diff
@@ -0,0 +1,47 @@
...
+    Date(String),
+    Time(String),
+    Timestamp(String),
+}
+
+impl From<Literal> for anyhow::Error {
+    // https://github.com/bluejekyll/enum-as-inner/issues/84
+    #[allow(unreachable_code)]
+    fn from(item: Literal) -> Self {
+        // panic!("Failed to convert {item}")
```
max-sixty:
```suggestion
```

prql-compiler/src/ast/types.rs
```diff
@@ -0,0 +1,41 @@
...
+
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub enum Type {
+    Native(NativeType),
+    Parameterized(Box<Type>, Box<Node>),
+    AnyOf(Vec<Type>),
+}
+
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq, strum::EnumString, strum::Display)]
+pub enum NativeType {
```
max-sixty, line 15: Very nice. I added one comment [here](https://github.com/prql/prql/issues/54#issuecomment-1126337029) (for the future tbc)

prql-compiler/src/ast/types.rs
```diff
@@ -0,0 +1,41 @@
...
+        Type::Native(NativeType::Column)
+    }
+}
+
+impl PartialEq for Type {
+    fn eq(&self, other: &Self) -> bool {
+        match (self, other) {
+            (Self::Native(l0), Self::Native(r0)) => l0 == r0,
+            (Self::Parameterized(l0, l1), Self::Parameterized(r0, r1)) => l0 == r0 && l1 == r1,
+            (Self::AnyOf(many), one) | (one, Self::AnyOf(many)) => many.iter().any(|m| m == one),
```
max-sixty: I think this isn't transitive — i.e. `AnyOf[A, B] == A` and `AnyOf[A, B] == B`, but `A != B`. I think rust theoretically requires this but not sure if it matters in practice

prql-compiler/src/parser.rs
```diff
@@ -169,11 +171,23 @@ fn ast_of_parse_tree(pairs: Pairs<Rule>) -> Result<Vec<Node>> {
                     })
                 }
                 Rule::ident => Item::Ident(pair.as_str().to_string()),
+
+                Rule::number => {
+                    let str = pair.as_str();
+
+                    let lit = if let Ok(i) = str.parse::<i64>() {
```
max-sixty, line 178: Nice! I was thinking how to do this. This approach is nice and simple

prql-compiler/src/semantic/resolver.rs
```diff
@@ -650,7 +685,7 @@ mod tests {
         let result = parse(prql).and_then(resolve_and_translate).unwrap();
         assert_snapshot!(result, @r###"
         SELECT
-          first_name OVER (PARTITION BY last_name) AS first_name
+          first_name
```
max-sixty, line 688: Is this still a correct translation?

prql-compiler/src/semantic/transforms.rs
```diff
@@ -183,6 +183,83 @@ pub fn cast_transform(func_call: FuncCall, span: Option<Span>) -> Result<Transfo
 
             Transform::Group { by, pipeline }
         }
+        "window" => {
+            let ([pipeline], [rows, range, expanding, rolling]) =
+                unpack(func_call, ["rows", "range", "expanding", "rolling"])?;
```
max-sixty, line 188: Nice!

prql-compiler/src/sql/translator.rs
```diff
@@ -1311,7 +1352,9 @@ take 20
         )
         sort order_day # order:asc
```
max-sixty:
```suggestion
        sort order_day
```

prql-compiler/src/sql/translator.rs
```diff
@@ -1325,22 +1368,16 @@ take 20
           COUNT(ol.book_id) AS num_books,
           SUM(ol.price) AS total_price,
           SUM(COUNT(ol.book_id)) OVER (
-            PARTITION BY TO_CHAR(co.order_date, '%Y-%m')
-            ORDER BY
-              TO_CHAR(co.order_date, '%Y-%m-%d')
+            PARTITION BY TO_CHAR(co.order_date, '%Y-%m') RANGE BETWEEN UNBOUNDED PRECEDING
+            AND CURRENT ROW
```
max-sixty: V impressive!

max-sixty: Though do we need the `ORDER BY` here too?

reference/src/transforms/window.md
```diff
@@ -0,0 +1,56 @@
...
+For each row in result, its input segment is determined either by:
+
+- `rows`, which takes a range of rows relative to current row,
+- `range`, which takes a range of values relative to current row value.
+
+Start of the range is inclusive, end of range is exclusive. Index 0 references
+current row. If a bound is omitted, segment will extend until the end of the table (or group).
+
+For example:
+- `rows:0..3`   means current row plus two following,
```
max-sixty, line 18: I was thinking how to do `current` — this approach is perfect, can't believe I didn't think of it

reference/src/transforms/window.md
```diff
@@ -0,0 +1,56 @@
...
+- `rows:-3..0`  means two preceding rows plus current row,
+- `rows:-3..5`  means two preceding rows plus current row plus four following rows,
+- `rows:..1`    means all rows from the start of the table to and including current row,
+- `rows:0..`    means current row and all following rows until the end of the table,
+- `rows:..`     means all rows, which same as not having window at all.
```
max-sixty, line 23:
```suggestion
- `rows:..`     means all rows, which same as not having the `rows` argument at all.
```

reference/src/transforms/window.md
```diff
@@ -0,0 +1,56 @@
...
+
+> Note: currently, negative integer literals (`-3`) are not implemented.
+
+<!-- TODO: rows vs range example, with visualization -->
+
+For ease of use, there are two flags that override `rows` or `range`:
+
+- `expanding:true` is an alias for `range:..1`. Sum using this window is also known as "cumulative sum".
```
max-sixty, line 31: Wouldn't this be (but probably I'm misunderstanding):

```suggestion
- `expanding:true` is an alias for `rows:..1`. Sum using this window is also known as "cumulative sum".
```

?

reference/src/transforms/window.md
max-sixty: Though I'm still thinking through the answer to the question "what number is the current row?".

Did you think about making `0` the current row on both sides, and the range being inclusive?

It looks like that's how it's defined here:

```
`rows:-3..0`  means two preceding rows plus current row,
```

but possibly not here:

```
`rows:..1`    means all rows from the start of the table to and including current row,
```

Or am I missing something?

max-sixty:
Very impressive! I left a few comments re the `rows:..1` vs `rows:..0` syntax, and the `ORDER BY`.

I think this is the last big code feature in the compiler for 0.2! FYI I'll have some time this weekend to clean up things we need for 0.2 — docs / website / follow-ups etc.

prql-compiler/src/ast/item.rs
aljazerzen: This is still possible. `fn new` is just a helper function for constructing integer ranges. Should be named `new_int`.

prql-compiler/src/ast/types.rs
aljazerzen: That's true. Instead of `PartialEq`, this should be "isSubSetOf".

prql-compiler/src/semantic/resolver.rs
aljazerzen: In this case `first_name` is not a window function. It's a scalar function (identity actually), so it applies to a single row at a time, which does not depend on groups, sorting or window frames.

reference/src/transforms/window.md
aljazerzen: Oh, that's a mistake. It should be 
```
`rows:-2..1` means two preceding rows plus current row,
```

Such indexing was supported [by Dijkstra](https://www.cs.utexas.edu/users/EWD/transcriptions/EWD08xx/EWD831.html) it it seem most natural to me. I also wanted this to be similar to python and rust, while distancing from matlab...

max-sixty: Wonderful reference and explanation.

Do you think we should have a slightly different syntax for inclusive vs exclusive ranges, like rust has, then? My concern is that SQL ranges are by default inclusive; e.g. `BETWEEN 0 and 3`, or `in 0..3` includes both 0 & 3, as does `range:0..3`

There are a few areas like this where it's difficult to completely upgrade SQL, because we can't see through to the data. So we can either separate off the parts we can — e.g. different approach to `rows` than other ranges — or compromises to their standards — e.g. just use inclusive for everything.

WDYT?

reference/src/transforms/window.md
max-sixty: @aljazerzen just wanted to check you saw this. IIUC this only matters when there are duplicate values; otherwise they're the same (or possibly my SQL knowledge is rusty)

PR #479, opened 2022-05-13
Filter does not take lists

### LINKED ISSUE #469: Filter with a list returns an incorrect result
```elm
from employees
filter [age > 25, age < 40]
```

```elm
SELECT
  employees.*
FROM
  employees
WHERE
  age > 25 age < 40
```

(in this case we can use `in` & `..` but the principle holds)

---

I don't see much downside in disallowing lists and joining them with `and`, even though it's very easy to do two `filter`s:

```elm
from employees
filter age > 25
filter age < 40
```

aljazerzen:
Closes #469


max-sixty:
Nice, thank you!

PR #481, opened 2022-05-14
Add coalesce operator

### LINKED ISSUE #475: NULL-COALESCING Operator
Handling `NULL` values is a very common operation in SQL queries. Different SQL dialects have different functions for handling these, see for example https://www.w3schools.com/sql/sql_isnull.asp. The `COALESCE` function appears to be the most widely supported but is rather verbose to type and takes up a lot of space when reading queries. IMHO infix operator expressions are also easier to read but opinions on this may vary.

According to Wikipedia `??` seems to be the most common choice for this operator: https://en.wikipedia.org/wiki/Null_coalescing_operator

For example this is supported in [C#](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/operators/null-coalescing-operator), [Dart](https://dart.dev/codelabs/dart-cheatsheet#null-aware-operators) and [Javascript](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing_operator).

This would allow one to write expressions like 
```sql
ABS(COALESCE(a.amount, 0) - COALESCE(b.amount, 0)) AS difference
```
as
```elm
derive difference = (a.amount ?? 0) - (b.amount ?? 0) | abs
```
in PRQL.

I therefore propose implementing this operator in PRQL and have it transpile to the `COALESCE` function.

Related [discussion on Discord](https://discord.com/channels/936728116712316989/945832228011798528/973306057701093496).

max-sixty:
Closes #475


max-sixty:
The outrageously large diff is because we have to parse `expr_coalesce` differently to other `expr`, and actually extract the value on its left & right.

(Maybe we should reduce the size of the parse trees in snapshots given how big they are now)

aljazerzen (approved):

PR #486, opened 2022-05-15
Tighten up datetime parsing with an end_expr item

max-sixty:
My guess is that we might have lots of these sorts of things, where the parsing is a bit too lax, and becomes confusing on small mistakes...


aljazerzen (approved):

aljazerzen:
I've noticed this a few times, but brushed it of as "it a user mistake". Such tweaks remove sharp corners of the language, so great!

max-sixty:
Yes, agree. 

And if you can think of a better approach, lmk — e.g. if we add struct syntax, then we have to remember to add `}`. I thought about having "end" be anything not alpha numeric but maybe that causes other issues...

PR #489, opened 2022-05-16
Add dependabot

max-sixty:


max-sixty:
Not sure this is the correct config etc, but let's see

PR #505, opened 2022-05-17
Add docs for dialect

max-sixty:
Also compress the terms down to one each, unless anyone has any objection


max-sixty:
> Also compress the terms down to one each

@aljazerzen lmk any objections!

aljazerzen:
Its better to have one way of doing things.

PR #506, opened 2022-05-17
Website and webapp

aljazerzen:
- removed prql-web.
- added website
- added webapp
- migrated README content into website. Probably I moved too much - @max-sixty can you take a look on how to organize the content?

TODO:
- [x] GH actions for deployment (I left old ones in they can be removed)
- [x] figure out domains
- [ ] tasks for installing hugo


README.md
```diff
@@ -23,361 +23,7 @@ News](https://news.ycombinator.com/item?id=30060784#30062329) and
 [Lobsters](https://lobste.rs/s/oavgcx/prql_simpler_more_powerful_sql) earlier
 this year when it was just a proposal.
 
-## Overview
```
max-sixty, line 26: If we leave this in, I'm happy to do this separately and ensure there's a "Links" section at the top.

(possibly we also want to keep a small example here, I personally prefer being able to see an example when I go to repos, and I know other similar projects got flak for being too abstract in their Readmes...)

Cargo.toml
```diff
@@ -4,7 +4,6 @@ members = [
   "prql-python",
   "prql-compiler",
   "prql-js",
-  "prql-web",
```
max-sixty, line 7: RIP, this was a journey. Ashes to ashes...

max-sixty (approved):
Thanks a lot for doing the website!

- Do we need two gh-pages apps for this repo? I guess the proposed `.github` workflows won't work yet, which is fine, but when they do, will they be compatible with the language reference, which also uses gh-pages? I can have a look at this more.
- Possible we should add a license file around the logo saying it can only be used by the PRQL team (i.e. including you), rather than just anyone. I don't think we need to worry about this at stage though.

max-sixty:
> Do we need two gh-pages apps for this repo? I guess the proposed `.github` workflows won't work yet, which is fine, but when they do, will they be compatible with the language reference, which also uses gh-pages? I can have a look at this more.

One option is to host all the source files in a single repo, and then use https://github.com/cpina/github-action-push-to-another-repository to deploy them to another repo, which hosts the content. That seems like a decent model; happy to set it up.

README.md
aljazerzen: Yeah sure, but I think is should be shorter.

Cargo.toml
aljazerzen: Sorry... it will remain in history and our hearts :D

aljazerzen:
It just hit me: we don't need multiple domains. We can build each of the sites and copy them into one filesystem, deployed to GH pages.

```
website/
- index.html
- ...
- reference/
  - index.html
  - ...
- compiler/
  - ...
```

This will probably be easier than copying between repos.

Also, have you considered buying "prql-lang.org"?

Cargo.toml
max-sixty: 😂

max-sixty:
Yes that definitely works! We'll have a site that has very different themes between each mini-site, but I think that's completely fine. The main website will have links to the other two (and GH etc).

> Also, have you considered buying "prql-lang.org"?

Yes, this probably is more conventional than `prql.builders` :). I got prql-lang.org & prql-lang.com. Ofc open to any other ideas. (And if we continue our momentum I'll donate them to some entity so it's not me personally holding them.)

max-sixty:
FYI I removed `test-web` from required tests. It doesn't seem to have updated, I'll try merging main to see if that refreshes it

webapp/src/workbench/prql-syntax.js
```diff
@@ -0,0 +1,60 @@
...
+      [/"/,        { token: 'string.quote', bracket: '@close', next: '@pop' } ]
+    ],
+
+    whitespace: [
+      [/[ \t\r\n]+/, 'white'],
+      [/\/\*/,       'comment', '@comment' ],
+      [/\/\/.*$/,    'comment'],
+    ],
+  },
+};
```
max-sixty, line 59: Awesome, thanks for doing this!

website/content/pages/motivation.md
```diff
@@ -0,0 +1,188 @@
+---
+title: "Motivation"
+url: motivation
+---
```
max-sixty, line 4: These are great. I'll do a run through of them when we merge.

website/themes/prql-theme/LICENSE
```diff
@@ -0,0 +1,20 @@
+The MIT License (MIT)
+
+Copyright (c) 2022 PRQL core developers
```
max-sixty: Not important for me, but:
- Did you choose MIT above Apache? We have Apache for the rest of the repo.
- I had thought that every author retained their copyright, but licensed it under the license.

(I am fine giving copyright of my work, I don't have a preference here, I know some people might in the future — fine to push this off)

max-sixty (approved):
Very nice indeed! Thank you @aljazerzen 

When we merge I'll (finally) ask the web designed to start from here.

max-sixty:
If you merge this, I can look at the GHA for Hugo etc.

website/themes/prql-theme/LICENSE
aljazerzen: Whoops, that was auto-generated by hugo, will remove, so it is covered by the main LICENCE file.

Do you know, how would we add separate licences for the logo?

max-sixty: > Do you know, how would we add separate licences for the logo?

I'm not sure actually; I can hopefully have a look at some point. Possibly we leave it until it matters and hope it's not too late...

PR #526, opened 2022-05-19
Remove markdown table hack

max-sixty:
ref https://github.com/Byron/pulldown-cmark-to-cmark/issues/44


aljazerzen:
Ah great, I was wondering why the table was not rendering...

max-sixty:
Yeah this one was a journey!

PR #543, opened 2022-05-20
Rename webapp / compiler to Playground

max-sixty:


max-sixty:
As discussed @aljazerzen !

PR #548, opened 2022-05-20
Remove News from website

max-sixty:
I agree this would be good to have. But my concern is that it won't get updated, and that it's much worse to have something stale than nothing at all.

What do others think?


aljazerzen:
I ve added this mosty to showcase how to add pages with hugo.

Actually i think that updates should be on github, where development happens. Website is targeted at end users.

PR #549, opened 2022-05-20
Update frontpage examples

aljazerzen:


README.md
```diff
@@ -24,25 +22,25 @@ News](https://news.ycombinator.com/item?id=30060784#30062329) and
...
 
 ```elm
-from employees
-filter country = "USA"                        # Each line transforms the previous result.
+from employees                                # Each line transforms the previous result.
+filter start_date > @2021-01-01               # Clear date syntax.
 derive [                                      # `derive` adds columns / variables.
-  gross_salary: salary + payroll_tax,
-  gross_cost:   gross_salary + benefits_cost  # Variables can use other variables.
+  gross_salary = salary + payroll_tax,
```
max-sixty, line 28: D'oh!

reference/src/queries/tables.md
```diff
@@ -5,15 +5,16 @@ We can create a table — similar to a CTE in SQL — with `table`:
...
 )
 
-from a            # Starts a new pipeline
+from another_table      # Starts a new pipeline
 ```
 
-In PRQL `table`s are far less common that CTEs are in SQL, since a single
-pipeline of logic can always be represented without simple pipelines alone.
+In PRQL `table`s are far less common than CTEs are in SQL, since a linear
+series of CTEs can be represented with a single pipeline.
```
max-sixty, line 17: Much better

README.md
```diff
@@ -3,9 +3,7 @@
 <!-- User badges on first line (language docs & chat) -->
 [![Language Docs](https://img.shields.io/badge/DOCS-LANGUAGE-blue?style=for-the-badge)](https://prql-lang.org)
 [![Discord](https://img.shields.io/discord/936728116712316989?label=discord%20chat&style=for-the-badge)](https://discord.gg/eQcfaCmsNc)
-[![VSCode](https://img.shields.io/visual-studio-marketplace/v/prql.prql?label=vscode&style=for-the-badge)](https://marketplace.visualstudio.com/items?itemName=prql.prql)
```
max-sixty, line 6: Is it that you think we're over-badged here? I think the VSCode plugin is good and it's useful to show that we're catering to developers too. Ironically we have better VSCode support than our friends at dbt!

max-sixty (approved):

README.md
max-sixty: I could add _all_ `.md` files to the testing infra which run tests on the docs

README.md
aljazerzen: It's a bit crowded I think. We can have vscode plugin under integrations.

max-sixty: OK

Issue #517, opened 2022-05-19
Negative bounds in windows


max-sixty:
As per the docs, this doesn't quite work yet:

```
from employees
group employee_id (
  sort month
  window rows:-5..0 (
    derive semi_annual_comp = sum paycheck
  )
)
```

(labeling as "bug" even though I know we knew this, hope that's OK!)

max-sixty:
I had a brief attempt at this. It seems harder than I had thought, so writing up my findings

The constraint is based on:
- we want `-5..` to bind the `-` to the `5`, and parse `-5` & `..` (not `-` & `5..`)
- but also in `a - b` we don't want the `-` to bind to  `b`, and be like a function call of `a` and `-b`

We currently allow leading `-` in sort directions like `sort [-value]`. The `operator_add` which allows `-`  is in the `expr_call` definition:

https://github.com/prql/prql/blob/ef2800163c6fa9e464f4a3a52fb9b8e589564c7e/prql-compiler/src/prql.pest#L68

Having it in `expr_call` prevents `a - b` from being parsed as `a` & `-b`, because `a - b` is an expression but `b` isn't — it's a term: https://github.com/prql/prql/blob/ef2800163c6fa9e464f4a3a52fb9b8e589564c7e/prql-compiler/src/prql.pest#L76

We can't allow leading `-` in terms, since that will allow `-b` to be parsed as a term, because then `a - b` will be a function call of `a` and `-b`.

But we also can't allow an expr in a `range_edge`; shown here for reference:

https://github.com/prql/prql/blob/ef2800163c6fa9e464f4a3a52fb9b8e589564c7e/prql-compiler/src/prql.pest#L111-L115

Because then `..` will recursively parse into an expr, which is a range, which is an expr.

max-sixty:
Closed by #552

PR #552, opened 2022-05-21
Enable negative bounds in ranges

max-sixty:
Overcomes the issues in #517 by changing how the `sort` terms are
parsed.


aljazerzen:
Great, this was not trivial!

PR #553, opened 2022-05-21
Allow descending sort without parentheses

max-sixty:


max-sixty:
Enabled by #552 parsing changes

Issue #530, opened 2022-05-19
Handling of `true` literal


max-sixty:
Currently, in the example in the web compiler, we have:

```elm
...
derive [
  always_true = true,
  db_version = s"version()",    # An S-string, which transpiles directly into SQL
]
```

which complies to:

```sql
SELECT
  table_0.*,
  countries.*,
  country_code,
  countries.true AS always_true,
  version() AS db_version
FROM
  table_0
  LEFT JOIN countries USING(country_code)
```

Is `countries.true` correct? Probably we want to prioritize that as a literal, even though it may mean disallowing it as a column name.

max-sixty:
Weirdly I'm not getting this in `prql-compiler`, maybe the versions are different

max-sixty:
Closed by https://github.com/prql/prql/pull/555

PR #554, opened 2022-05-21
Add tests for interpreting `true` literal

max-sixty:


max-sixty:
Confirming https://github.com/prql/prql/issues/530 is fixed

PR #555, opened 2022-05-21
Use latest version of `prql-js` in `playground`

max-sixty:
And did a pass on the readmes etc


playground/package.json
```diff
@@ -1,43 +1,47 @@
...
-      "last 1 firefox version",
-      "last 1 safari version"
-    ]
+  "homepage": "https://prql-lang.org/playground/",
+  "name": "prql-playground",
+  "private": true,
+  "scripts": {
+    "build": "react-scripts build",
+    "eject": "react-scripts eject",
+    "preinstall": "cd ../prql-js && wasm-pack build",
```
max-sixty, line 42: This file is mostly autoformatting changes annoyingly, but this is the important line

PR #564, opened 2022-05-25
Group and take a range

### LINKED ISSUE #9: Group and select rows which have min/max value
I wonder if PRQL can solve this very practical problem that many people suffer from every day: https://stackoverflow.com/q/3800551

### LINKED ISSUE #421: `group` not working as intended with `take`
Possibly we need to implement #420 given our plan of how to implement this query? Though in the meantime, should this raise an error rather than ignoring `group`?

```elm
from employees
group role (
  sort join_date  # taken from above
  take 1
)
```

```sql
SELECT
  employees.*
FROM
  employees
ORDER BY
  join_date
LIMIT
  1
```

aljazerzen:
Closes #421


max-sixty:
Very impressive @aljazerzen !

PR #565, opened 2022-05-26
build: Add insta test to prql-python (currently failing)

max-sixty:
To confirm whether there's an error on https://github.com/mitsuhiko/insta/issues/227

(though given build issues we may not be able to resolve this)


max-sixty:
This was coincidentally added in https://github.com/PRQL/prql/pull/1499, and works! So we can close this and the upstream issue.

PR #587, opened 2022-06-07
Preliminary work for type analysis

### LINKED ISSUE #414: Variables at the start of pipelines
Currently, this works:

```elm
func iden x = x

from foo
derive bar: (temp_c | iden)
```

```sql
SELECT
  foo.*,
  temp_c AS bar
FROM
  foo
```

But this doesn't:

```elm
func iden x = x

from foo
derive bar: (
  temp_c
  iden
)
```

```
Error { span: Some(Span { start: 42, end: 48 }), reason: NotFound { name: "temp_c", namespace: "function" }, help: None }
```

I think this should probably work:
- I had thought pipelines should be equivalent between `|` and linebreaks
- It means we can't have pipelines in functions that span more than one line, IIUC. I found this trying to change the final compiled example at https://lang.prql.builders/functions.html, which requires lots of scrolling; but unsuccessfully.

aljazerzen:
Language changes:
- Solved the `( |` workaround for inline pipes. `from employees | group dept (take 1)` is now a valid query.
- Functions, variables and tables now all have the same namespace. This means that you cannot reference column named `rank` without table name prefix. Reason is internal compiler architecture, but I also think is a good change, because it makes the language simpler: there is only one namespace.

AST changes:
- pipelines don't have `value` field anymore. Value is stored as first node of the pipeline.
- `Expr` and `Operator` were replaced with `Binary` and `Unary`, which represent binary and unary expressions. Operators are now stored in as enums instead of strings.

Closes #414


prql-compiler/src/ast/item.rs
```diff
@@ -20,17 +19,65 @@ pub enum Item {
...
+    Type(Ty),
     Table(Table),
     SString(Vec<InterpolateItem>),
     FString(Vec<InterpolateItem>),
     Interval(Interval),
     Windowed(Windowed),
 }
 
+#[derive(Debug, PartialEq, Clone, Serialize, Deserialize, strum::Display, strum::EnumString)]
+pub enum BinOp {
```
max-sixty, line 42: Nice, thanks for doing this

book/src/transforms/sort.md
```diff
@@ -22,7 +23,7 @@ sort age
 
 ```prql
 from employees
-sort -age
+sort [-age]
```
max-sixty, line 26: Very reasonable

prql-compiler/src/ast/types.rs
```diff
@@ -5,39 +5,70 @@ use serde::{Deserialize, Serialize};
...
-    AnyOf(Vec<Type>),
+pub enum Ty {
+    Literal(TyLit),
+    Named(String),
+    Parameterized(Box<Ty>, Box<Node>),
+    AnyOf(Vec<Ty>),
+
+    /// Means that we have no information about the type of the variable and
+    /// that it should be inferred from other usages.
+    Infer,
```
max-sixty, line 16: 👍

max-sixty (commented):
I haven't looked at it all but the result is very nice indeed! Thank you v much @aljazerzen .

Not a good reason for us to have delayed 0.2 but now it's in for 0.2, that's good!

PR #594, opened 2022-06-10
Merging release.yml 

roG0d:


max-sixty:
I'm having some trouble merging `main` into this branch. But GH seems to be fine merging the branch into main.

Some of these diffs seem like they might be a bit old — e.g. https://github.com/prql/prql/pull/594/files#diff-e81e13daadd1745de51d8b8d85fce1fcd9be355732b64d60a6b56e18c28388caR14 is already in main.

Is it just `release.yml`? If so, could you maybe start from main, add that file, and then force push the result so that's the only diff? 

Thank you!

roG0d:
Yeh ofc! The easiest path will be to force push my prql repo with the actual data of prql/prql with the .yml included, that way will be equal.

The problem might come because I deleted the other .yml in order to avoid errors, sorry about that 😅

roG0d:
Max, I've just copy everything from prql main to rog0d/prql main and I've inlcuded the release.yml file... Now it detects even more files... I've even checked both files from both repos and I still find no sense about it

roG0d:
Sorry about the [54b112e](https://github.com/prql/prql/pull/594/commits/54b112edc52dcedfa9c6cb90fbe65f277e68c27a) commit, I slipped a little 😅, now it looks good enough  I think https://github.com/prql/prql/pull/594/files#diff-e81e13daadd1745de51d8b8d85fce1fcd9be355732b64d60a6b56e18c28388caR14

max-sixty:
Great!

Do we also need to merge https://github.com/prql/homebrew-prql/tree/actions in on the homebrew-prql side?

Anything else?

Thanks @roG0d

max-sixty:
BTW tests are very nice! https://github.com/prql/homebrew-prql/blob/actions/.github/workflows/tests.yml

max-sixty:
I created a new user, as it doesn't seem possible to limit access to a single repo, and put that token in the PRQL repo as a secret: `HOMEBREW_PRQL_TOKEN`.

image

Let me know if I'm missing anything!

roG0d:
Yeah! We would need to merge `actions` into `main` on `prql/homebrew-prql`, with special care regarding the actual version parameter of the `prql.rb`, as soon as I can, I will review it and change it if necessary!

Perfectly decided as for the new user, I've checked the scopes available and it's a pity that you can't actually select some specific repo... If you want I can change  the workflow with the new secret 🙌

roG0d:
HOMEBREW_PRQL_TOKEN is now changed: [81f5711](https://github.com/prql/prql/pull/594/commits/81f571170b46e0afbc6bf2fc44ef6415b12adfea)

Correct version of prql/homebrew-prql is changed: https://github.com/prql/homebrew-prql/pull/3/commits/f36328bb949854c5a0930edaffc23e31db455fbd

And a pull request on prql/homebrew-prql has been created: https://github.com/prql/homebrew-prql/pull/3

PR #595, opened 2022-06-13
website rework

aljazerzen:
Merged from https://github.com/sualsoft/prql, but with heavily stripped dependencies and fonts. I did not use the designs for examples and roadmap pages, because I found them too "promotional" and not enough "technical".

I have also added:
- a showcase element with basic examples,
- a better looking footer,
- Avenir Next font,
- some stripes in the background :D


aljazerzen:
PS: i've put the theme in it's own directory so i may be easily reused.

PR #604, opened 2022-06-16
Try adding OpenGraph to website

max-sixty:


max-sixty:
Not really sure how to test this, so let's give it a go

PR #606, opened 2022-06-17
Make a bunch of refinements to homepage

max-sixty:


max-sixty:
Merging but lmk any feedback / objections — feel free to revert if any of these are disliked

PR #624, opened 2022-06-21
Use Hugo navbar

max-sixty:


website/config.yml
```diff
@@ -4,29 +4,51 @@ title: PRQL
...
       url: /roadmap/
       pageref: /roadmap/
       weight: 3
+    - identifier: Examples
+      name: Examples
+      title: Examples
+      url: /examples/
+      pageref: /examples/
+      weight: 4
+    - identifier: FAQ
```
aljazerzen, line 36: This could be cleaned-up: `weight`, `pageref` and `title` are unused. 

I don't know why the author of the theme even included them here...

aljazerzen (approved):

website/config.yml
max-sixty: Seems like `weight` is actually required to know the order; it doesn't respect the ordering of the yaml.

Other things are removed!

Issue #627, opened 2022-06-21
Alternating background color for website sections?


max-sixty:
The background colors for the sections on the website might be a bit confused. Currently we have the colored background on the first section, and the "Showcase"

With the usual caveat that I'm no UI person and very happy to defer to someone who knows what they're doing; possibly we should have:
- One background for the first section
- Alternating backgrounds for the others, to break them up. This may or may not be the same background as the first section

roG0d:
I tried to follow your guidelines @max-sixty , But in the end I tried a simpler approach:

- The main color of the webpage now correspond to the blue one of the logo.
- The webpage keeps a flow of light blue that keeps everything together. Giving a common sensation. 
- It's enclosed by the same `repeating-linear-gradient` to keep it consistent.
- The showcase section is an especial one cause have a little bit more of interactivity, that's why I suppose It was highlighted before. I tried to keep that but I didn't find an elegant solution. I'll keep thinking and trying what could it be better

It isn't a significant change to the design, but I tried a not-to-mess-thing-up approach as priority. I didn't came with a nice an elegant idea of break apart the different blocks neither. I think that is cause every webpage now follow this markdown-style of cohesion 😅

I created a pull request since I didn't find a better way for you to check it out, feel free to cancel it if you don't like the result 😁

https://github.com/prql/prql/pull/630

max-sixty:
I really like this!

PR #630, opened 2022-06-21
MODIFIED: style frontpage

roG0d:


max-sixty:
I think this looks great!

As discussed in #627, we don't have the alternating background by section — but I think that's fine. I like the phase out at the end a lot!

Thank you v much @roG0d .

I'll merge; if anyone else has feedback then ofc feel free to comment.

PR #631, opened 2022-06-21
Fixes the fahrenheit_from_celsius function in the website showcase

snth:
Hi,

A small PR to fix the fahrenheit_from_celsius function in the Showcase section of the website.

I think the Showcase section is really nice and it's one of the first things I was drawn to when reviewing the new website just now. Having an incorrect formula in there just kinda irks me and I also thought naming the function `fahrenheit_from_celsius` rather than `fahrenheit_of_celsius` is a little bit nicer.

### DIFF
--- a/website/content/_index.md
+++ b/website/content/_index.md
@@ -144,13 +144,13 @@ showcase_section:
     - id: functions
       label: Functions
       prql: |
-        func fahrenheit_of_celsius temp -> temp * 3 + 32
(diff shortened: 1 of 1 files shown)


max-sixty:
@snth many thanks indeed! And wonderful to have you here as well as on Discord — you deserve to be on the list of contributors! (I won't pretend to keeping the incorrect formula here as a strategy, but it would have been clever :) )

PR #633, opened 2022-06-21
Fixes the cte-0.prql example in the Playground

snth:


max-sixty:
Thanks a lot @snth !

max-sixty:
If you want to add a bug for the original issue then that would be great! (I can otherwise)

PR #635, opened 2022-06-21
Quotes

roG0d:


max-sixty:
Thanks a lot for adding this!

Do you think it's possible for the cards to be a similar size to the tweets? Hopefully we can add more.

And currently on a smaller screen, they don't shrink that well. Though I'm not sure whether that's a realistic goal or not — no stress if not:
image

roG0d:
For sure! something like this?

Captura de pantalla_2022-06-22_13-02-35


I tried that it looks good on a smaller screen! (more zoom on the browser in my case)

Captura de pantalla_2022-06-22_13-03-25

max-sixty:
@roG0d excellent! I'll merge this.

I added some TODOs to the code — if you want to continue on this, then addressing some of those would be valuable.

PR #642, opened 2022-06-23
Quotesfix

roG0d:
Referring to:
> TODO:

          - Move the content of the testimonial into `_index.md` so we can have a
          few of these, and templated. (maybe renaming this page from tweets to
          something more general?)
          - Format the testimonial more like a "card" in the other sections, including a light
          border. Currently it has an idiosyncratic blue border only on the left. Do
          we need to unify some of `card` format in style.css across the sections?
          - Currently the items overlap at some monitor sizes — can we prevent
          that? Does the browser seem to think the items are smaller than they are?

- Now the content of testimonial has been moved to `_index.md` looking for a more template, ordered code. Every "tweet" mention on the code has been renamed to "comments" (for a more general view).
- Testimonial cards now have a similar style than the others on the website.
- As for the overlap in tweets. I think the main problem is that the columns and rows divs of boostrap don't synchronize with the embedded tweet provided by twitter API. (I've added some width limits to the iframe tag just to secure it). I've managed to keep it consistent but at expense the column divs for the tweets now are bigger, so the tweets display can adapt to the progressive size of the columns. 
Some preview:

Captura de pantalla_2022-06-23_13-38-36


max-sixty:
The code is much nicer, thank you @roG0d ! And great preview.

As you mention, they're now bigger. Would it be possible to make them all the same width as the cards? I think that would look more ordered. IIRC it's possible to ask the Twitter API for a certain width, if that's helpful. Lmk if you want me to find a link for that.

(I also probably preferred them smaller, but no strong view, and do whatever you think looks best / is implementable).

roG0d:
I've tried my best max, but I can't find a total solution to this: 

Here, divs containing the iframe (twitter-widget) seems to respect the width and height followed by the `cards`  element of bootstrap:
Captura de pantalla_2022-06-23_23-14-54

The main problem is that the iframe itself doesn't respect the resolution of the parent element, so it overflows and produce the overlapping:
Captura de pantalla_2022-06-23_23-14-14

Although I lack of bootstrap and CSS knowledge, I've managed to impose the parent resolution resulting in this: 

Captura de pantalla_2022-06-23_23-28-53

this width limit starts appearing at x200 zoom on the browser, it corrects when the cards oversize corresponding to the zoom browser (x300 more less)

When normal zoom and until x150 it seems to works nice:

Captura de pantalla_2022-06-24_00-22-26

Sorry I didn't find a better solution to this...! 😓
I'll push this new fix, if you don't find it good enough feel free to cancel the PR with no problems whatsoever!

max-sixty:
That's looking pretty good! Thank you v much @roG0d . We don't need to get to perfection to make progress.

I'll see if I can do anything later but this is very decent.

PR #646, opened 2022-06-24
Tweak tweets display on website

max-sixty:
@roG0d WDYT?

FWIW this stuff takes way more time than I would have expected, thanks for taking it so far.


max-sixty:
image

roG0d:
That's great! You've achieved! Now tweets display on a different row before it can be cutted down

Captura de pantalla_2022-06-24_09-14-29

PR #647, opened 2022-06-24
Refine layout on home page

max-sixty:


max-sixty:
It's looking pretty slick now, and the code has also got simpler.

I had no idea I'd learn so much about web doing this...

PR #724, opened 2022-06-28
Test against SQLite & postgres

### LINKED ISSUE #291: Create a way to test PRQL-generated SQL queries against real databases

aljazerzen:
I created a crate that:
- compiles all queries in a directory,
- executes SQL in a local SQLite database,
- executes SQL over a postgres connection,
- serializes both results to CSV and tests them as a snapshot.

Now the second part, how to run that while having a running Postgres instance, loaded with data?

I created two Dockerfiles that construct two containers:
- one for Postgres (with data already loaded in) 
- and one for the crate above (+sqlite database file), compiled with --tests

There is also docker-compose.yml file that builds and starts both services in correct order.

This solution is not working completely yet, but it could work - the question is if we *want* to do it this way?

It would be great if there would be no need to also run the test crate in a container, because it complicates the development. But we could have two configs for the crate: one would execute queries against the localhost where developers would be running postgres (probably the same docker-compose config) and another which could be run in a container so it could be run during automated tests in GH actions.

Closes #291


max-sixty:
Amazing @aljazerzen !

I think containers are the right approach — otherwise we get a lot of complexity around managing dependencies. I'm happy to be wrong there — maybe just `brew install postgres` would be sufficient — but ensuring that works across versions & OSs might be quite difficult. I'd also hope that we don't need to run these tests in a tight loop; they're something that does a check in CI or when we find a specific problem with a DB.

I know there are some efforts to bind SQLite or DuckDB — if that worked then great — but that's another issue.

qharlie:
This is killer!

snth:
This is great!

Regarding the postgres piece, how about using DuckDB instead? I'll explain below.

- DuckDB can be run locally like sqlite so should be easier to test without containers. 
- We already rely on DuckDB heavily in the PyPRQL Jupyter magic so we'll want to ensure compatibility for that.
- DuckDB aims to be the sqlite for analytical workloads so is probably the best fit for the kind of workloads PRQL aims to be great at.

Most importantly regarding the postgres compatibility though, [DuckDB uses the PostgreSQL query parser under the hood](https://duckdb.org/why_duckdb)  (see **SQL parser** under the **Standing on the shoulders of giants** section). So by testing DuckDB compatibility we should be ensuring postgres compatibility as well, at least at the SQL dialect level.

You could then split the postgres container testing part out into a separate PR and keep this one simple without that complexity.

aljazerzen:
Oh, I didn't know that about DuckDB, that's great news. I was planning to add it anyway, along with some other standalone databases.

My last idea was that I run SQLite (and DuckDB) always , but run postgres only when connection URL is supplied. Then we can test either:
- as normal cargo test, without postgres URL, using only SQLite and DuckDB
- in container with postgres URL, against many standalone databases. I'm not sure how to run this during GH actions. Good news is that now, it is not necessary, since two major DBMS parsers are already covered with the local tests.

So i'll leave all the Dockerfile-s and docker-compose config in the PR, but will have them on disabled.

snth:
Actually on second thoughts I wonder if what I said is still the case. It says so on the DuckDB website but they recently added a whole lot of features like `SELECT * EXCLUDE` and `GROUP BY ALL` which aren't supported by postgres and I don't know if they support all of postgres' JSON functions so it would seem like there are some differences between the two? 🤷

aljazerzen:
TODO:
- [x] also query DuckDB
- [x] don't run these tests on wasm
- [x] actually add some PRQL queries :D

max-sixty:
Awesome @aljazerzen !

PR #732, opened 2022-06-29
Create one NPM package for node/browsers/webpack

qharlie:
This uses wasm-pack to build all 3 targets into a 'fat binary' that is useable for any javascript purpose.

Normally package.json is generated but for our purposes we use our own.  Before this is merged the NPM_TOKEN secret will need to be added with access to https://www.npmjs.com/package/prql-js


.github/workflows/publish-npm.yaml
```diff
@@ -0,0 +1,60 @@
...
+        run: npm publish
+        working-directory: prql-js/
+        env:
+          NODE_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }}
+
+      - name: Bump npm version
+        run: npm version patch
+        working-directory: prql-js/
+
+      - name: Commit version bump
```
max-sixty: How do you see the relationship between the version in the file and the tag of the repo?

Do we tag the repo, and then it bumps the version of the npm package? Is that OK that the tagged repo has the previous version?

(at the moment, this workflow runs on `main` so I guess this would run in a loop, but I'm guessing we remove it from running on `main`?)

FWIW I think this isn't easy given we have a monorepo (though good for other reasons); if we have to do the versions manually for a bit to work it out, then that's OK.

prql-js/README.md
```diff
@@ -1,61 +1,70 @@
...
+## Notes
+
+
+This uses
+[`wasm-pack`](https://rustwasm.github.io/docs/wasm-pack/tutorials/npm-browser-packages/index.html)
+to generate bindings[^1].
+
+[^1]:
+though we would be very open to other approaches, and used `trunk`
+successfully in a rust-driven approach to this, RIP `prql-web`.
```
max-sixty:
```suggestion
[^1]: though we would be very open to other approaches, and used `trunk`
successfully in a rust-driven approach to this, RIP `prql-web`.
```

max-sixty (approved):
Thanks a lot @charlie-sanders !

aljazerzen:
I've added NPM_TOKEN

.github/workflows/publish-npm.yaml
manning79: Unfortunately NPM's versions ( unlike pythons ) cannot be overwritten, once you publish a 0.2.0 you can never update it, so it's not going to match the PRQL version except for the initial release, we can keep the major and minor the same probably and just bump the patch number.

When committing back using the GITHUB_TOKEN it won't trigger the actions again, so this will publish the package, then increment the number, then commit it back, will that work with the other release scripts we have ?

max-sixty: Thanks

> (at the moment, this workflow runs on main so I guess this would run in a loop, but I'm guessing we remove it from running on main?)

Even if it won't retrigger, it will currently create another commit for every commit on main; is this intended?

ref https://github.com/prql/prql/issues/736

manning79: Well the intent is to increase the patch number on successful publishing of the NPM package in package.json,  the side effect is the extra commit - I'm not sure how to do it otherwise.  This will publish on  every merge re #736

max-sixty: I see; I think let's split this part out and decide on the release approach separately, if you agree. 

The NPM content is great here, so I don't want to slow that down.

manning79: Sounds good, Ive disabled the bump-version-and-commit piece.  @aljazerzen is the NPM_TOKEN the Automation kind ? That will bypass the 2-factor auth thats normally required.

aljazerzen: Yes, the token is Automation so there is no 2FA required.

max-sixty:
@charlie-sanders lmk if you want a hand finishing this — it looks like the only failing test is a pre-commit one; which should be quite easy to fix...

qharlie:
Ok yes please, I thought I had fixed it but looks like it's still failing.  I'll have computer access tommorow.

.github/workflows/publish-npm.yaml
```diff
@@ -0,0 +1,57 @@
...
+
+jobs:
+  publish-to-npm:
+    name: Release
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+        with:
+          token: ${{ secrets.GITHUB_TOKEN }}
```
max-sixty: This shouldn't be needed, right?

prql-js/README.md
```diff
@@ -1,61 +1,66 @@
...
-import compile from "prql-js";
+Currently the `compile(prql_string)` function is the only function available.
 
-const { sql, error } = compile(`from employees | select first_name`);
+### From NodeJS
+
+```javascript
+const prql = require("prql-js/dist/node/prql_js.js");
+
+const sql = prql.compile("from employees | select first_name").sql;
```
max-sixty: I had a merge conflict here — is this correct? `compile` returns an `Option` in rust

max-sixty:
> Ok yes please, I thought I had fixed it but looks like it's still failing. I'll have computer access tommorow.

OK, lmk if OK to hit the buttan

.github/workflows/publish-npm.yaml
manning79: The checkout?  Not if we add it to the existing publish

prql-js/README.md
manning79: I would use their example for compile the destructuring of SQL and error looks better

max-sixty:
OK — am merging — @charlie-sanders are you OK to handle any follow-ups? Thank you!

qharlie:
> OK — am merging — @charlie-sanders are you OK to handle any follow-ups? Thank you!

Yep! ill be back in town tomorrow night.

Issue #721, opened 2022-06-28
Raw s-string


max-sixty:
As discussed on [discord](https://discord.com/channels/936728116712316989/991384476313714749).

Crystal uses variables like `{?foo}`; which would be good to pass through.

Two options I can see:
- Work this into the parser, like we did for dbt. `{?foo}` is passed through to SQL just as `{{ foo }}` is.
- Allow for a "raw s-string", like `sr"{?foo}"`. This doesn't special-case the parser so much, though is a bit less visually appealing for Crystal users.
  - The issue with `s"{?foo}" is that s-strings interpret code between `{ }` as a PRQL expression.

Edit: As per @mattdharmon's suggestion, let's go with the raw s-string

mattdharmon:
I think raw s-string might be a good idea since there are things like SISS packages that need to be called from time to time that a vendor has made.

max-sixty:
@mattdharmon are there any docs on these `{?foo}` variables? I'd like to use them as a test case and couldn't immediately find a description of them online.

max-sixty:
Proposal here: https://github.com/prql/prql/pull/752#issuecomment-1187625092

@mattdharmon — would this cover your use case? It would mean the crystal variable would be `"{?foo}"` in postgres, or surrounded by backticks in MySql. So if a Crystal layer is replacing those with variables, it needs to do so inside the quotes.

PR #752, opened 2022-07-02
Add double-bracket strings

max-sixty:


max-sixty:
Would close https://github.com/prql/prql/issues/721. I need to add to the docs.

I had previously implemented an `rs"foo"` string. But actually I think the double-bracket approach is fairly standard, and more closely aligns with the current problem. Definitely open to feedback though.

prql-compiler/src/prql.pest
```diff
@@ -136,9 +136,16 @@ operator_compare = ${ "==" | "!=" | ">" | "<" | ">=" | "<=" }
 operator_logical = ${ ("and" | "or") ~ &WHITESPACE }
 operator_coalesce = ${ "??" }
 
-s_string = ${ "s" ~ opening_quote ~ ( interpolate_string_inner | ( "{" ~ expr_call ~ "}" ))* ~ POP }
-f_string = ${ "f" ~ opening_quote ~ ( interpolate_string_inner | ( "{" ~ expr_call ~ "}" ))* ~ POP }
-interpolate_string_inner = { ( !( PEEK | "{" ) ~ ANY )+ }
+// If we have lots more string prefixes then wwe could just have a type
```
max-sixty:
```suggestion
// If we have lots more string prefixes then we could just have a type
```

max-sixty:
Currently the SQL formatter is changing `{?foo}` to `( ? foo }`. I need to think about the best way of dealing with this — is there some generalized way (i.e. all valid across dialects) of surrounding the expression so that the SQL formatter doesn't mangle it, but it's still valid?

Otherwise we'll have to replace it with something and then swap it back after the formatting, which would be a bit inelegant.

max-sixty:
Possibly the way to do this is to surround double-bracket strings with dialect-specific "quotes" — e.g. for MySQL & BigQuery, we use backticks; for others we use double-quotes.

I can't see an escape hatch like we do for jinja expressions (https://github.com/prql/prql/blob/0.2.0/prql-compiler/src/sql/translator.rs#L46).

mklopets:
> Currently the SQL formatter is changing `{?foo}` to `( ? foo }`

fyi, just as a related data point, our PRQL pipeline already does this kind of replacing post-compilation as we use postgres's `@?` inside of an S-string, which gets mangled to `@ ?` (i.e. invalid sql)

max-sixty:
Thanks for that case @mklopets 

I increasingly think we should parse to some representation of "raw ident", e.g. backticks, and then based on the dialect convert to either backticks or double-quotes.

max-sixty:
> I increasingly think we should parse to some representation of "raw ident", e.g. backticks, and then based on the dialect convert to either backticks or double-quotes.

I'm planning to pursue this, combined with the approach in #809.

So:

```elm
from foo
select `{{bar}}`
```

becomes, for dialects that use double-quotes to surround "raw identifiers":

```sql
select "{bar}" from foo
```

or for those that use backticks:

```sql
select `{bar}` from foo
```

The only change here would be to compile `{{bar}}` into `{bar}` — i.e. double brackets become single brackets with no evaluation of the contained expression. Currently is an error since it attempts to evaluate `{bar}` as a PRQL expression.

---

Edit: for any future travelers, I didn't end up pursuing this exact approach — the s-string approach seems to cover the cases, with fewer additional language features.

max-sixty:
Not sure what's going on with the commit history but it'll be squashed...

Added some docs too; lmk if anything is unclear!

PR #755, opened 2022-07-02
Update playground UI to not wrap tabs

### LINKED ISSUE #750: Didn't understand Playground until window was > 1160px wide
I wasn't sure what to do with the [Playground](https://prql-lang.org/playground/) when I first went there. It was only when I made the window wider (the threshold seems to be 1160 px) that I saw the SQL pane on the right. Sure enough, changes on the left updated the right pane. Slick!

My suggestion would be to let the two panes share the space, even if they're as narrow as 100px... Thanks!

Jaspooky:
Resolves https://github.com/prql/prql/issues/750

### DIFF
--- a/playground/src/workbench/Workbench.css
+++ b/playground/src/workbench/Workbench.css
@@ -7,18 +7,18 @@
 }
 .tabs {
   display: flex;
-  flex-wrap: wrap;
   flex-grow: 1;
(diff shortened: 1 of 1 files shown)


max-sixty:
Thanks a lot @Jaspooky !

PR #758, opened 2022-07-03
prql-compiler java api

### LINKED ISSUE #728: Provide antlr4 grammar file?
If there is an antlr4 grammar file, use spark to implement, support distributed compute

doki23:
Closes #728 

Implement a java wrapper with java native interface.
But I am not familiar with cross compilation, so it now supports 
1. osx x86_64, aarch64
2. linux x86_64
3. windows x86_64


max-sixty:
This looks excellent, thanks @doki23 !

A couple of comments:
- Looks like the `.so` files are committed into the repo, which we should remove before merging.
- How confident are you that this works in the current state? Should we set up a test as a workflow? If this is easy, we could add it here. Otherwise fine to leave to another PR (great if you want to take it, otherwise please could you add an issue?)
- We can add it to https://github.com/prql/prql/blob/main/website/content/_index.md?plain=1#L336, but fine for another PR
- I don't know Java well; are there any considerations that users need to be aware of?
- Is there a package manager for Java that we should publish this to? (again, for another PR)

Thanks again!

doki23:
@max-sixty 
- Those `.so` files need to be published together with the code. So I place them into the java package manually to make sure it can work now. It's difficult to me to provide a script doing the cross compilation automatically now, but I'd like to make another issue to keep tracking with it.
- I'd like to set up a java unit test as a workflow. But I have no idea about how to create a workflow now. After learning about this, I hope to add the workflow in this pr. So I'll change it to a draft pr.
- After supporting all platforms, users don't need to be aware of anything, just use it as common third-party libs.
- I'm afraid I don't have the right to publish it to maven(the java package manager) which should be done by the maintainer. [This doc](https://docs.github.com/en/packages/working-with-a-github-packages-registry/working-with-the-apache-maven-registry#publishing-a-package) may be helpful

max-sixty:
> Those `.so` files need to be published together with the code. So I place them into the java package manually to make sure it can work now. It's difficult to me to provide a script doing the cross compilation automatically now, but I'd like to make another issue to keep tracking with it.

Do you know what libraries generally do? Generally artifacts are not stored in the repository, since repos accumulate in size, and these artifacts are much larger than everything else in the repo combined. Often artifacts are published as part of releases in "Assets" (e.g. https://github.com/prql/prql/releases/tag/0.2.0).

If there's some way of merging incremental progress, that's fine; we don't need to have everything perfect — just a decent design and some part of the implementation.

> I'd like to set up a java unit test as a workflow. But I have no idea about how to create a workflow now. After learning about this, I hope to add the workflow in this pr. So I'll change it to a draft pr.

As you wish; it's also fine to do this later. Check out https://github.com/prql/prql/blob/main/.github/workflows/release.yml#L26-L42 which calls https://github.com/prql/prql/blob/main/.github/workflows/python.yaml as an example. I imagine there are lots of java examples around the web — feel free to copy one in and we can iterate from there.

> After supporting all platforms, users don't need to be aware of anything, just use it as common third-party libs.

Awesome!



> I'm afraid I don't have the right to publish it to maven(the java package manager) which should be done by the maintainer. [This doc](https://docs.github.com/en/packages/working-with-a-github-packages-registry/working-with-the-apache-maven-registry#publishing-a-package) may be helpful

Great, sounds like something we can do later.

---

Thanks @doki23, looking forward to adding this! Please comment back here for any help, lots of us are willing to pitch in.

Baoqi:
Hi, doki23, I just have a small concern, in "NativeLibraryLoader.java" and "Environment.java", there are following notice:

// Copyright (c) Facebook, Inc. and its affiliates. All Rights Reserved.

This may be conflict with prql's Apache v2 license (I'm not sure about this). Maybe we can simplify those (we may need to work with PowerPC, AIX, etc. plus "MuslLibc".

doki23:
@Baoqi Thanks for the remind. I copied the source code from rocksdb and it is dual-licensed under both the GPLv2 and Apache 2.0 License, so I think it's ok. But I'm not sure whether I need add some comments on the top of it.

max-sixty:
Yes please, we can just add a link to the original license. Thanks!

doki23:
Hi @max-sixty :
I've resolved these issues:
1. Add the link to original license on the top of "NativeLibraryLoader.java" and "Environment.java".
2. Add java unit test workflow running on ubuntu and macos. 
3. Remove `.so`, `.dylib` and `.dll` files from the codebase.

But it still needs a release workflow to cross compile libs for every platforms we want to support and then publish the java package to maven public repository.
It's difficult to me now, I'm going to make another issue to keep track with it.

max-sixty:
@doki23 this is awesome! Thank you very much.

I'll merge now. If you have the patience, would you mind adding a short README with 2-3 bullets explaining how it works? Or are these files from a standard location that we could link to? At least for me, there is a lot of code here that I'm not sure whether it's PRQL-specific or generic, or how it works.

doki23:
It's ok, I'll add a README then.

aljazerzen:
I tested this out and it's great! I didn't delve too much into the build process, since it built without any errors.

I've posted the demo project here <https://github.com/prql/prql-java-demo>

First I tried to install [this one](https://search.maven.org/artifact/io.github.doki23/prql-java/0.0.1/jar), but it didn't have the binary for linux64. After building it locally, it was without problems.

One suggestion: should the artifact group be named `org.prqllang` instead of `org.prql`? Since we have <http://prql-lang.org>

doki23:
Yes of course. I've created another issue to provide a release workflow and I will change the groupid in that pr.

Issue #736, opened 2022-06-30
Release process


max-sixty:
Now that some people are actually using PRQL, we should decide how to do releases.

#### Principles

- Low release latency. When someone makes a change, that should flow to users quickly — it's good for users and encouraging for contributors. There's very little weighing on the side of releasing slower — we have extensive tests, we can always roll-back if we make a mistake, no one is relying on it for production.
- Automated. Can we make this as automated as possible?

#### Tagging

We have the advantage of a monorepo (except for `dbt-prql` & `pyprql`, for their own reasons), but it means that it might too broad to just version everything based on a single unified repo tag; at least for patch versions — what do others think? 

The upside is that it's very simple. The downside is that if we make a change to only `prql-python` and want to do a release, we get a new version of `prql-compiler` with zero changes. That would weigh more if we had something like `dbt-prql` here, because it's more than just a binding for `prql-compiler`.

#### Automation

- @rbpatt2019 & @charlie-sanders have set up some quite impressive automation in `pyprql`. It a bit heavier duty — each commit is labeled with whether it's a breaking change or not, and then it works out what version to use — but very manageable. (any thoughts?)
- Bumping the version in the source definition seems quite reasonable, particularly if we have different versions by library.
- We could attempt to publish to [crates.io](http://crates.io/) / PyPI / npm on every single merge, and allow them to fail if they're already created (though we'd need to parse the error message to ensure it didn't fail for another reason). At least for cargo, there doesn't seem to be a flag for "pass if it's already there".
  - Once pushed, a process could make the tags automatically (it needs to be that way around — we can't change the files based on pushing tags, since the tag would be attached to the old version)

#### Cadence

- We could do something like "nightly, assuming there's been a merge". Is there anything downstream that would be upset by that many releases? Maybe something with dependabot or a low latency, constantly filing PRs?
- The extreme would be "every merge"...

Any thoughts?

mklopets:
a couple of thoughts:
- I quite like using [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/) for commit messages in any case
- there's something called [Release Please](https://github.com/googleapis/release-please) for automation on top of conventional commits
- I like the idea of versions being in sync across the different packages, so that prql-js 1.2.3 would match compiler 1.2.3, but obviously, changing the JS package shouldn't trigger a change for e.g. prql-python users
- what if the minor version matched between all packages, but patches wouldn't need to?
[...]

PR #760, opened 2022-07-03
Bump package-lock.json in playground

max-sixty:
I don't know what this means, but it keeps being bumped when I run `task build-all`, so I'm guessing it's updating the version. (Though I don't think we actually use this; we instead include the current version?).

CC @charlie-sanders, if you know / have a view :)

### DIFF
--- a/playground/package-lock.json
+++ b/playground/package-lock.json
@@ -27,7 +27,7 @@
     },
     "../prql-js/pkg": {
       "name": "prql-js",
-      "version": "0.1.2",
+      "version": "0.1.3",
       "license": "Apache-2.0"
     },
     "node_modules/@ampproject/remapping": {


mklopets:
How about changing the playground's dependency to use the build from npm instead of `file:..`?


```diff
-    "prql-js": "file:../prql-js/pkg",
+    "prql-js": "prql-js@^0.13.0", 
```

The downside is that the version needs to be bumped in `playground` every time a new `prql-js` version is released, but the upside is that everyone can be certain the playground is using a published version, not something random that happened to be in `main`.

For local dev, you would need to use [`npm link`](https://docs.npmjs.com/cli/v8/commands/npm-link)

max-sixty:
That's what we used to have. I appreciate there's a tradeoff between having a dependable version vs. the latest code on the playground. 

One synthesis of this is to do faster releases — https://github.com/prql/prql/issues/736; possibly I'll prioritize that.

I'm open-minded on the tradeoff — I would lean towards always having the latest, but I can understand both sides, and I'm likely biased by doing lots of development.

qharlie:
Just FYI if we use the (new) NPM published package we will need to change the import/require path.

I would suggest using the local build ( like we have now ) until we get to stable.

max-sixty:
~OK — @charlie-sanders until that's done is it OK to merge this? (Or maybe it's almost done and we should close this — your call!)~

Edit: actually does https://github.com/prql/prql/pull/792 negate this, just merged a minute ago?

qharlie:
I thought package-lock.json was generated from the build process, I'm not sure what this change does ?

max-sixty:
Right, running `task build-all` makes this change — should it be committed into main?

max-sixty:
This doesn't seem to happen any longer, so closing!

PR #763, opened 2022-07-06
Clarify documentation of `derive` and `select`.

roldugin:


max-sixty:
Thanks @roldugin ! Those are better :)

PR #765, opened 2022-07-07
Fix prql-js usage example in readme

mklopets:
`compile` returns a `CompileResult` that has `sql` and `error` keys

just console.logging the output of `compile` would result in something like
```
CompileResult { ptr: 1179656 }
```


max-sixty:
Thanks @mklopets 

IIRC we've discussed which version is better on discord recently.

Ofc regardless of which we choose, the example should be correct — so I'll merge — others please feel free to comment / adjust.

PR #777, opened 2022-07-08
Revert "Create one NPM package for node/browsers/webpack (#732)"

max-sixty:
This reverts commit e14b85a297448dd572744d4263d5b98c0ebe1569, since the Playground is broken.

I imagine it's a very small change, but reverting quickly as per https://github.com/prql/prql/blob/main/CONTRIBUTING.md.

@charlie-sanders do you know what we need to change? Is there any quick test we can have to ensure the playground works? As you know, JS is not my area...


qharlie:
Ugh this was a sloppy mistake on my part, this should not have changed any rust code, let me try again.

max-sixty:
No prob @charlie-sanders ! "Better to have merged and reverted than never have merged at all"

PR #779, opened 2022-07-08
Fix windows build

max-sixty:
Follow-up to https://github.com/prql/prql/pull/778


max-sixty:
ref https://github.com/wangfenjin/duckdb-rs/issues/62

Issue #776, opened 2022-07-08
Add from_json and to_json to prql-js


florent-martineau:
pyPRQL has a to_json method, but prql-js doesn't. Moreover, if #775 is merged, we can add support to from_json.

FYI, code is ready

qharlie:
Closed in #782

Issue #775, opened 2022-07-08
Add from_json support to compiler


florent-martineau:
We can currently serialize PRQL into JSON, but we cannot deserialize JSON into PRQL.

FYI, code is ready

max-sixty:
Thanks @florent-martineau ! So I understand, WDYM "code is ready"?

florent-martineau:
I meant the code is ready and works locally, I just haven't made the PR yet. Same thing for #776.

I was saying code is ready to avoid having people waste their time on it, since I have already made a first version.

max-sixty:
Oh amazing! Thanks @florent-martineau

PR #782, opened 2022-07-09
Add from_json to compiler, add from_json and to_json to prql-js

florent-martineau:
The goal of this PR is:
- To add a from_json capability to the compiler (#775): using to_json we can serialize PRQL to JSON, using from_json we can deserialize JSON to PRQL.
- To add from_json and to_json capabilities to prql-js (#776): in PyPRQL we can use to_json, in prql-js there is no such capability. The goal is to add both.

Please note that it's my first PR on an open source repository, plus I had never seen a line of Rust until yesterday, so please be indulgent :sweat_smile:


prql-compiler/src/ast/item.rs
```diff
@@ -320,7 +330,7 @@ impl Display for Item {
                 f.write_char('>')?;
             }
             Item::Literal(literal) => {
-                write!(f, "{:?}", literal)?;
+                write!(f, "{}", literal)?;
```
florent-martineau, line 333: When generating PRQL, it would write Integer(50) instead of 50, hence this change.

Example:
It would have written take Integer(50) instead of take 50

prql-compiler/src/ast/item.rs
```diff
@@ -293,6 +293,11 @@ impl Display for Item {
                 UnOp::Not => write!(f, "not {}", expr.item)?,
             },
             Item::FuncCall(func_call) => {
+                match func_call.name.as_str() {
```
florent-martineau, line 296: When generating a group, we would have had something like: 
```
group [foo, bar]
aggregate [baz = sum foobar]
```

There were missing parentheses around aggregate, which led to invalid PRQL. With this code, we write an opening parenthesis before aggregate.

Another block of code writes the closing parenthesis, so that we get:
```
group [foo, bar] (
  aggregate [baz = sum foobar]
)
```

prql-compiler/src/ast/item.rs
```diff
@@ -301,6 +306,11 @@ impl Display for Item {
                 for arg in &func_call.args {
                     write!(f, " {}", arg.item)?;
                 }
+
+                match func_call.name.as_str() {
```
florent-martineau, line 310: This is the closing parenthesis around aggregates, see more detailed comment above.

prql-compiler/src/lib.rs
```diff
@@ -7,13 +7,16 @@ mod semantic;
 mod sql;
 mod utils;
 
+use crate::{ast::Item};
```
florent-martineau, line 10: I don't know if you have a conventions (eg. should you put imports in alphabetical order, separate blocks of imports by line breaks, etc.)

Plus, I didn't research the difference between `use` and `pub use`.

prql-compiler/src/semantic/declarations.rs
```diff
@@ -80,7 +80,7 @@ impl Debug for Declarations {
 impl Display for Declaration {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
         match self {
-            Declaration::Expression(node) => write!(f, "{}", node.item),
+            Declaration::Expression(node) => write!(f, "({})", node.item),
```
florent-martineau, line 83: For instance, when generating PRQL for a `table`, the parentheses were missing: 
```
table top_employees
from employees
sort [rating]
take 20
```

Thanks to this code, we now get: 
```
table top_employees (
  from employees
  sort [rating]
  take 20
)
```

prql-compiler/src/ast/item.rs
florent-martineau: Actually this is wrong, as groups do not only have aggregates, eg.:
```
group employee_id (
  sort month
  window rolling:12 (
    derive [trail_12_m_comp = sum paycheck]
  )
)
```

We should really put parentheses around the group's "content" and not around aggregate.

max-sixty:
@florent-martineau this is really good. Congratulations on your first PR!

As discussed on Discord, I'm happy to help finish this off — your call on how much you want to do yourself vs ask for help.

To merge, we need a couple of additions:
- Run `cargo fmt`. 
  - You also get this from running `task test-all` (https://github.com/prql/prql/blob/main/CONTRIBUTING.md#tasks), after installing `Task` & running `task setup-dev` (as described on that page too), if you have the patience.
- Add a couple of tests. Would it be helpful for me to add an initial test to this module to show how we generally do tests?

prql-compiler/src/lib.rs
max-sixty: `cargo fmt` will organize these!

prql-compiler/src/ast/item.rs
max-sixty: Great!

And it's fine to leave a comment in the code for something that's not quite there yet!

max-sixty:
@florent-martineau FYI I merged the latest main into your branch, so run `git pull` to get the latest

max-sixty:
@florent-martineau thanks for kicking this off! We merged #829 so I'll close this. But congrats on your first open-source PR, and we'd be happy to work together on another.

PR #785, opened 2022-07-09
Allow leading underscores for idents

mklopets:
As briefly discussed on [discord](https://discord.com/channels/936728116712316989/945832046121607208/994914239028400148), this adds support for leading underscores in front of any idents.

In my case, this will allow us to get rid of a lot of s-strings!

This is my first rodeo both with prql's code base as well as Rust & pest – a through review is probably wise.


max-sixty:
Great, thanks @mklopets !

PR #790, opened 2022-07-10
Collapse book crates together

max-sixty:
Particularly since we don't have the playground in the book any longer, there's not much point having a separate crate.

If you agree @aljazerzen, I might do the same for test-databases, as an integration test crate for prql-compiler?


aljazerzen:
Great idea, i support that and also moving test-databases

PR #793, opened 2022-07-10
Fix >= (gte) and <= (lte) operators

mklopets:
Because of how pest works, `<=` was always first matching `<` and `>=` was matching `>`, and the `=` was then left just hanging there, erroring:

```
Error: during parsing
   ╭─[:3:13]
   │
 3 │ filter age >= 100
   ·             ┬  
   ·             ╰── expected expr_mul
───╯
```

For reference, here are the [relevant pest docs](https://pest.rs/book/grammars/peg.html#ordered-choice).

### DIFF
--- a/prql-compiler/src/parser.rs
+++ b/prql-compiler/src/parser.rs
@@ -1906,4 +1906,71 @@ join `my-proj`.`dataset`.`table`
                     named_args: {}
         "###)
     }
+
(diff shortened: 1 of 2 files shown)


max-sixty:
Thank you v much @mklopets !

I'll give you merge access so no need to wait for anyone for things like this

PR #796, opened 2022-07-10
Use correct (bundler) build of prql-js for playground

mklopets:
Following https://github.com/prql/prql/pull/792, this makes the playground use the correct build of `prql-js`.


max-sixty (approved):
This looks great! Thanks @mklopets

PR #809, opened 2022-07-12
Dialects for ident escaping

aljazerzen:
Make PRQL is case sensitive and produce correct ident escaping, depending on the dialect.

Probably conflicts with #752

I had to refactor much of `translator`, mainly converting `From<Item> for sql_ast::Expr` into functions that also take `Dialect`.

I've removed support for `` `project-bar`.`dataset`.`table` ``. It can be added back, but it would require better ident parsing with PEST and I don't think it's needed. You can think that PRQL has no nested namespaces such as tables and schemas, but has a single namespace that can contain `schema1.my_table.column` and also `schema1..column` all of which are valid names.

### DIFF
--- a/prql-compiler/src/parser.rs
+++ b/prql-compiler/src/parser.rs
@@ -222,11 +222,10 @@ fn ast_of_parse_pair(pair: Pair<Rule>) -> Result<Option<Node>> {
         }
         Rule::ident | Rule::jinja => {
             let inner = pair.as_str();
-            let stripped = if inner.starts_with('`') && inner.ends_with('`') {
-                &inner[1..inner.len() - 1]
-            } else {
-                inner
-            };
+            let stripped = inner
+                .strip_prefix('`')
+                .and_then(|s| s.strip_suffix('`'))
+                .unwrap_or(inner);
             Item::Ident(stripped.to_string())
         }
 
--- a/prql-compiler/src/parser.rs
+++ b/prql-compiler/src/parser.rs
@@ -220,7 +220,15 @@ fn ast_of_parse_pair(pair: Pair<Rule>) -> Result<Option<Node>> {
                 pipeline: Box::new(pipeline),
             })
         }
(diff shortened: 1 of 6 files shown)


prql-compiler/src/parser.rs
```diff
@@ -220,7 +220,15 @@ fn ast_of_parse_pair(pair: Pair<Rule>) -> Result<Option<Node>> {
                 pipeline: Box::new(pipeline),
             })
         }
-        Rule::ident | Rule::jinja => Item::Ident(pair.as_str().to_string()),
+        Rule::ident | Rule::jinja => {
+            let inner = pair.as_str();
+            let stripped = if inner.starts_with('`') && inner.ends_with('`') {
+                &inner[1..inner.len() - 1]
```
max-sixty: String slicing like this can fail on unicode; we should instead use `.chars`

(Or add a Todo and I'm happy to take this)

max-sixty (commented):
Yes I think this is a very good design. The multiple part Ident is OK to give up for the moment I think.

We should add BigQuery as similar to MySql, since that's the main use of backticks atm. I can add that on later IYL.

prql-compiler/src/parser.rs
aljazerzen: Oh, yeah that is a must fix before merging, thanks.

max-sixty:
And re #752 — let's do this and then I'll adjust that to use this, that's a great synthesis of both issues.

prql-compiler/src/sql/translator.rs
```diff
@@ -1445,17 +1440,80 @@ take 20
...
+        )?;
+
+        assert_display_snapshot!((resolve_and_translate(query)?), @r###"
+        SELECT
+          `anim"ls`.*,
+          `BeeName` AS `čebela`,
+          `bear's_name` AS medved
+        FROM
+          `anim"ls`
+        "###);
```
max-sixty, line 1515: Great tests!

Issue #818, opened 2022-07-16
Link to bindings is broken


MFAshby:
Hi! 

I was hoping to try out prql in a toy app, and I was excited to see in the FAQ that there are bindings for Rust. 

However the link is broken. Page: https://prql-lang.org/faq/ link for Bindings points to: https://prql-lang.org/content/_index.md#Bindings, which is a 404.

max-sixty:
Thanks @MFAshby ! Closed by #821

PR #821, opened 2022-07-18
Remove incorrect link

max-sixty:
Removing this because it generates a 404, as per #818.

Is there a way to link to a section on the homepage? Linking to prql-lang.org/index.html#Bindings doesn't seem to work either.


mklopets:
> Is there a way to link to a section on the homepage?

opened https://github.com/prql/prql/pull/823 with the simplest way I could think of

PR #823, opened 2022-07-18
Add back link to list of bindings from FAQ

mklopets:
I haven't used Hugo in many years, and there might be a smarter way, but I added something to make home page sections link-able

Addresses already-closed issue https://github.com/prql/prql/issues/818


max-sixty:
Thanks a lot @mklopets !

PR #825, opened 2022-07-18
Add "copy to clipboard" button to playground for output SQL

mklopets:
I spent **way** too much time manually copying long output SQL when debugging – I assume this tiny addition will be helpful to others as well.

Should work on non-IE browsers, tested Chrome and Safari.

### DIFF
--- a/playground/src/workbench/Workbench.js
+++ b/playground/src/workbench/Workbench.js
@@ -23,6 +23,7 @@ class Workbench extends React.Component {
     filename: "input.prql",
     sql: "",
     prql: "",
+    justCopied: false,
   };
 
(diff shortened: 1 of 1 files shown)


max-sixty:
Brilliant! I'm going to find this very helpful myself too!

max-sixty:
@mklopets this doesn't seem to work for me, I get this error in Dev Tools. Does it work for you on https://prql-lang.org/playground/? I'm using Firefox on Mac, is that a special case?

image

mklopets:
@max-sixty turns out Firefox is the only major browser that doesn't support this API without a feature flag

opened a PR: https://github.com/prql/prql/pull/880

image

PR #829, opened 2022-07-21
Add from_json to compiler

### LINKED ISSUE #775: Add from_json support to compiler

arrizalamin:
Sorry for a duplicate PR since #782 hasn't been updated for 10 days and our team needs this to build a query builder. From what I understand with https://github.com/prql/prql/pull/782#discussion_r917243880 and https://github.com/prql/prql/pull/782#discussion_r917291829 is that the parser can't parse a function call on an argument, so we have to wrap it in a pipeline which then will be unwrapped from the AST if it contains only 1 node. But now the AST doesn't have enough information to rebuild the original query, so in the `Display` trait I wrap all function call on an argument with a parenthesis. This should close #775 and #776.

### DIFF
--- a/prql-compiler/src/lib.rs
+++ b/prql-compiler/src/lib.rs
@@ -40,6 +40,11 @@ pub fn to_json(prql: &str) -> Result<String> {
     Ok(serde_json::to_string(&parse(prql)?)?)
 }
 
+// Convert JSON AST back to PRQL string
+pub fn from_json(json: &str) -> Result<String> {
+    Ok(display(serde_json::from_str(json)?))
+}
+
 #[cfg(test)]
 mod test {
     use super::*;
--- a/prql-compiler/src/lib.rs
+++ b/prql-compiler/src/lib.rs
@@ -40,7 +40,7 @@ pub fn to_json(prql: &str) -> Result<String> {
     Ok(serde_json::to_string(&parse(prql)?)?)
 }
 
-// Convert JSON AST back to PRQL string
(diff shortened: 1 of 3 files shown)


prql-compiler/src/lib.rs
```diff
@@ -40,6 +40,11 @@ pub fn to_json(prql: &str) -> Result<String> {
     Ok(serde_json::to_string(&parse(prql)?)?)
 }
 
+// Convert JSON AST back to PRQL string
```
max-sixty:
```suggestion
/// Convert JSON AST back to PRQL string
```

prql-compiler/src/lib.rs
```diff
@@ -40,6 +40,11 @@ pub fn to_json(prql: &str) -> Result<String> {
...
+pub fn from_json(json: &str) -> Result<String> {
```
max-sixty, line 44: Possibly these should be returning a `Result<Query>`, similar for `to_json`. But this is fine for now

max-sixty (approved):
Thanks @arrizalamin !

Merging now because you need this. It would be great to add a test around this; I recognize we don't have tests for `Display` yet though, so this would be a first.

Do you need a release? I'm planning to do a release this weekend after the quotes issue is sorted.

max-sixty:
Actually I don't seem to be able to edit the PR; could you approve my suggestion re the docstring? You have merge privileges so can hit the button without me

PR #831, opened 2022-07-22
Add test for from_json

qharlie:


max-sixty:
Very nice @charlie-sanders ! Thanks!

arrizalamin:
@charlie-sanders I think we can use compiler test cases from `book/tests/prql/` since most of the language features are not covered by this test.

qharlie:
@arrizalamin this is just to test that from_json generates the same SQL as compile() does , although the prql in this test is very minimal, Im not sure I follow re /book/tests/prql ( though I am going to update this test with better prql from those examples :) )

arrizalamin:
@charlie-sanders I mean from these test cases https://github.com/prql/prql/tree/main/book/tests/prql, although this is also good.

qharlie:
Ahh gotcha, yeah I took it from those tests ( employees0 )

PR #833, opened 2022-07-23
Fixes #748, tests for prql-js 

### LINKED ISSUE #748: JS unittests for generated javascript

qharlie:


.github/workflows/release.yaml
```diff
@@ -78,6 +78,10 @@ jobs:
           node-version: "18.x"
           registry-url: "https://registry.npmjs.org"
 
+      - name: Run prql tests for node
```
max-sixty, line 81: Cool! 

We should probably move this into `test-all.yaml`; at the moment this will only run on a release, which would be a bit late, and mean the tagged version would never be released.

There are lots of setup steps though — we could split it up into a separate action like `test-python/action.yaml`

(not high priority though)

max-sixty (commented):
Great tests! Thanks!

.github/workflows/release.yaml
manning79: Yes good point ( though I do think it should be a part of the release also ... if it fails tests then it shouldn't release ).  

The setup is pretty time consuming though as it has to compile the project and `wasm-pack` for all targets, always seems to be the last GHA to finish.

max-sixty: Yeah, the `test-all.yaml` only runs on each merge, so doesn't slow down PRs. For python there's also a workflow that runs on changes to `prql-python`, if we want to also get fancy with `js`.

> ( though I do think it should be a part of the release also ... if it fails tests then it shouldn't release ).

Yes, fair!

Issue #609, opened 2022-06-18
Precedence issue


max-sixty:
Currently:

```elm
from cities
select temp_c = (temp - 32) * 3
```

compiles to:

```sql
SELECT
  temp - 32 * 3 AS temp_c
FROM
  cities
```

...which has the wrong precedence — we should have `(temp - 32) * 3`

aljazerzen:
This comment has been here from before I started contributing:

https://github.com/prql/prql/blob/be13c12f91bcf3cd147f9bd30098293da3ea5191/prql-compiler/src/sql/translator.rs#L487-L489

and answer is yes, but not in all cases :D

max-sixty:
As ever, s-strings are a way of mitigating this until it's fixed:

```
from n
derive x = s"(y or z) and (a or b)"
```

is correctly parsed as:

```
SELECT
  n.*,
  (
    y
    or z
  )
  and (
    a
    or b
  ) AS x
FROM
  n
```

mmm444:
I have run into a variant of this bug:
```
from x
derive [
  n = a + b,
  r = a/n,
]
```
translates as
```
SELECT
  x.*,
  a + b AS n,
  a / a + b AS r
FROM
  x
```

max-sixty:
Thanks @mmm444 and sorry for the trouble . 

For the next day or two™, you can get around it with an s-string: 

```elm
from x
derive [
  n = s"(a + b)",
  r = a/n,
]
```

```sql
SELECT
  x.*,
  (a + b) AS n,
  a / (a + b) AS r
FROM
  x
```

We'll aim to fix this soon though.

max-sixty:
I'm not sure of the best way to solve this — we can put parentheses around `left` and `right` here: https://github.com/prql/prql/blob/dbbd8ddc9c2a6acb767c6481d88eeb69a431ef55/prql-compiler/src/sql/translator.rs#L493-L512

but it'll generate a _lot_ of parentheses.

While I can imagine writing the code to handle this properly (e.g. "is the current operator higher precedence than any of the operators within `right", assuming operators are left-associative") — there's probably some well-understood technique to this. If anyone is familiar, please comment! 

Otherwise I'll ask someone who is more experienced than me to ensure we're not reinventing stuff.

max-sixty:
This is the "put parentheses around everything" approach;

```diff
[...]

PR #836, opened 2022-07-23
fix some display trait bugs

arrizalamin:
I tested this to all queries in https://github.com/prql/prql/tree/main/book/tests/prql/, the only error is missing parenthesis on a binary operator, related to #609. Also, I wrapped all identity with a backtick to prevent them from being parsed as other items, eg: `from project-foo.dataset.table `


max-sixty:
Awesome, thanks @arrizalamin !

It would be great some tests for this — we could do some fairly simple ones with insta, like we have in lots of places (e.g. `parser.rs` has dozens). Otherwise I worry someone changes something and all the checks you've done to improve this don't necessarily persist. If you want me to do a couple of example ones lmk!

PR #838, opened 2022-07-24
Fix quoting issues in idents

max-sixty:


max-sixty:
@mklopets @huw @blackary — I think this fixes the quoting issue. Let me know if you have any other test cases to add in

max-sixty:
If there are no objections, I'll probably do a release tomorrow

PR #839, opened 2022-07-24
Fix precedence issue

max-sixty:
Not yet complete, but solves the immediate issue, and a nice framework to build off, thanks to @aljazerzen


max-sixty:
Merging, but from https://github.com/prql/prql/issues/609:

> Does anyone have other cases we should test against? This currently only works for a few operators which I think should be sufficient for most cases — what are the ones we're currently missing?

qharlie:
Nice!  Looking forward to going through it

PR #856, opened 2022-07-25
Add some workflows to track our compilation time

max-sixty:
This adds some workflows to track our compilation time:
- On every merge, with caches (which is most about what we care about; almost all PRs run with caches)
- Every day with & without caches

It then stores the files as artifacts. (Even better would be a store we could track / query, but that's for another day)

This is because compilation of the tests is getting quite long now — the CI PR workflow take 90-150 seconds, almost all of which is compilation, even with a cache. Most of that is from https://github.com/wangfenjin/duckdb-rs/issues/70.


max-sixty:
@matklad thought you might be interested in this; inspired by your blog posts!

.github/actions/time-compilation/action.yaml
```diff
@@ -0,0 +1,43 @@
...
+      uses: Swatinem/rust-cache@v1
+      # 'true' seems to require quotes (and using `inputs.use_cache` doesn't
+      # work); I'm really not sure why. There are some issues on the interwebs
+      # around this, but I couldn't find one that explained it.
+      if: inputs.use_cache == 'true'
+    - name: 🏭 Compile
+      uses: actions-rs/cargo@v1
+      with:
+        command: build
+        args: --timings=html -Z unstable-options
```
matklad, line 37: timings is now stable, no need for nightly or -Z

.github/actions/time-compilation/action.yaml
```diff
@@ -0,0 +1,43 @@
...
+    required: false
+  use_cache:
+    required: true
+    type: boolean
+
+runs:
+  using: composite
+  steps:
+    - name: 🔗 Setup toolchain
+      uses: actions-rs/toolchain@v1
```
matklad, line 19: actions-rs was deprecated recently. Folks migrate to https://github.com/dtolnay/rust-toolchain or (as rust-analyzer) just not using any action at all https://github.com/rust-lang/rust-analyzer/pull/12764/files

.github/actions/time-compilation/action.yaml
max-sixty: Thanks. IIUC `--timings=html` is not stable though... (arguably we shouldn't be using that for CI — it does make things easier to look at though!)

.github/actions/time-compilation/action.yaml
max-sixty: Thanks! https://github.com/prql/prql/issues/862

Issue #852, opened 2022-07-25
Fix Quoting issue for BigQuery


blackary:
Though #838 may have resolved issues with other dialects, it hasn't fixed the BigQuery case

```
prql dialect:bigquery
from db.schema.table
```

Now yields the result with no quotes at all

```sql
SELECT
  db.schema.table.*
FROM
  db.schema.table
```

But, this is invalid. In BigQuery, it actually needs to have `s around the full name in the select and from.

```sql
SELECT
  `db.schema.table`.*
FROM
  `db.schema.table`
```

max-sixty:
As long as the identifiers are valid, BQ didn't seem to require backticks.

Do you have a specific case where it raises and error?

Thank you

blackary:
Here's a fully-functioning example:

```
prql dialect:bigquery
from `bigquery-public-data.pypi.file_downloads`
filter timestamp > '2024-08-01'
```

-> 

```sql
SELECT
  `bigquery-public-data`.pypi.file_downloads.*
FROM
  `bigquery-public-data`.pypi.file_downloads
WHERE
  timestamp > '2024-08-01'
```

Note that it puts backticks around `biquery-public-data` because of the hyphens, and I get the error 

```
Unrecognized name: `bigquery-public-data` at [2:3]
```

I get the same error if the db name doesn't have hyphens in it.

If I change it so that there are backticks around the whole db.schema.table, it works (though, given the date, returns no results)

```sql
SELECT
  `bigquery-public-data.pypi.file_downloads`.*
FROM
  `bigquery-public-data.pypi.file_downloads`
WHERE
  timestamp > '2024-08-01'
```

max-sixty:
Ah, sorry about that @blackary , mea culpa.

IIUC, the initial example _would_ be OK — i.e., without the hyphen, it would be fine without the backticks. But if any part of the table expression has an invalid character, the whole identifier needs backticks. Lmk if that's inconsistent with how you're thinking about it.

Ironically I originally tested this with BQ, so the very first approach worked like this. But it's incompatible with Postgres (#822), which requires each part to be quoted separately.

So we'll have to split the implementation up further between dialects, I'll prioritize this and get to it later this week.

blackary:
Sorry, @max-sixty, that wasn't the best example, due to the hyphenated project name -- I used that one because it's using a public dataset. Here's something closer to the real example I'm working with:

```
prql dialect:bigquery
from company.apps.users
```

```sql
SELECT
[...]

PR #858, opened 2022-07-26
Fix BigQuery quoting

max-sixty:
Not sure this will be the end of it, but progress


max-sixty:
(I need to do some checks before merging, e.g. should the * be outside the backticks)

max-sixty:
(Continuing on from https://github.com/prql/prql/issues/852)

It's actually a bit more complicated than this!

Because it's required to have the column name (or `*`) outside the quote:

```sql
SELECT
  `bigquery-public-data.pypi.file_downloads`.country_code
-- or  
--    `bigquery-public-data.pypi.file_downloads`.*
FROM
  `bigquery-public-data.pypi.file_downloads`
WHERE
  timestamp = '2021-08-01'
```

This is not valid:

```diff
SELECT
-  `bigquery-public-data.pypi.file_downloads`.country_code
+  `bigquery-public-data.pypi.file_downloads.country_code`
\--or
---    `bigquery-public-data.pypi.file_downloads`.*
+--    `bigquery-public-data.pypi.file_downloads.*`
```

But the rule of "leave the final ident outside of the quotes" doesn't work, because the whole table must be quoted in the `FROM` clause.

So we probably need to branch higher in the translator and operate differently for column vs. table names. 

Another option would be to have more semantic understanding of the identifier — e.g. whether something is a table vs a column name — in the translator. Probably longer term this has more general utility, but I'm not sure how far it is now; I'm getting more up to speed with `semantic`. (CC @aljazerzen !)

max-sixty:
I don't think I made `translator.rs`'s code quality go up, but I _think_ it works now. Please reply to any bugs here! 

I'm also going to kick off a couple of issues — for BQ integration tests, and for our `translator.rs` architecture.

blackary:
@max-sixty Thanks! It looks good from the playground. I'll let you know if I run into any more issues.

PR #859, opened 2022-07-26
Precedence for unary operators and IS NULL

aljazerzen:
I've moved `binding_strength` from `item.rs` into `translator.rs` because it returns strength of **SQL** expressions.

There is also a few more tests for unary expressions and corrected precedence order which I copied from Postgres and Microsoft SQL references.


aljazerzen:
#609
#839

max-sixty:
Great, thanks @aljazerzen !

PR #863, opened 2022-07-26
cli revamp

aljazerzen:
Replaces the single `compile` command and format argument with different commands:

    $ prql-compiler compile input.prql output.prql
    $ prql-compiler parse input.prql output.prql
    $ prql-compiler debug input.prql output.prql
    $ prql-compiler annotate input.prql output.prql

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,6 +2,8 @@
 
 dist
 .vscode/settings.json
+.vscode/launch.json
 .idea
-
+/*.prql
+*/*.prql
 target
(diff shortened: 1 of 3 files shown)


.gitignore
```diff
@@ -2,6 +2,8 @@
 
 dist
 .vscode/settings.json
+.vscode/launch.json
 .idea
-
+/*.prql
+*/*.prql
```
max-sixty, line 8: Sorry if I messed with these before!

max-sixty (commented):
Wonderful!

I'm also hoping to add some more steps — e.g. get a parse tree vs an AST — which can be quite helpful for debugging — and let people see each modular step. We can also design tests around those steps for a given query.

So this is a good framework to start on that.

aljazerzen:
Yea, I started working on this because I wanted to add "resolve" step that outputs IR.

PR #869, opened 2022-07-27
docs: Architecture

aljazerzen:


max-sixty:
Awesome start!

We can start adding to it as we go; e.g. we can link to files, flesh out some parts. I think this would be a good exercise for me, as I become more accustomed to your code. 

My vote would be to keep it quite short, not describe every function (we have the code for that...), instead offer it as an entry point.

We could link to it from `CONTRIBUTING`?

aljazerzen:
I agree with all you said: link from contributing.md, keep it short, omit docs of each function.

I see architecture as described here https://matklad.github.io/2021/02/06/ARCHITECTURE.md.html 

It should be easy to read, should only cover the big picture, without details that are prone to change.

max-sixty:
Great, very much agree, shall we merge?

PR #874, opened 2022-07-28
Final fix of BQ quotes (hopefully)

max-sixty:


max-sixty:
Closes #852, hopefully. Please say where I'm wrong!

blackary:
@max-sixty I tried running from the playground, and checking out the repo and installing prql-python from the repo, but I'm getting an error on the case of hyphenated-db-name

```
In [7]: prql.to_sql("""
   ...: prql dialect:bigquery
   ...: from bigquery-public-data.pypi.file_downloads
   ...: """)
Traceback (most recent call last):

  File ~/.pyenv/versions/3.9.11/lib/python3.9/site-packages/IPython/core/interactiveshell.py:3369 in run_code
    exec(code_obj, self.user_global_ns, self.user_ns)

  Input In [7] in <cell line: 1>
    prql.to_sql("""

  File <string>
SyntaxError: Error { span: Some(Span { start: 28, end: 68 }), reason: Expected { who: None, expected: "ident", found: "`bigquery` - `public` - `data.pypi.file_downloads`" }, help: Some("`from` does not support inline expressions. You can only pass a table name.") }
```

It works fine with a non-hyphenated db name. Will this update automatically be applied to the local prql-python API, or will that come later?

max-sixty:
Thanks a lot for trying it out!

The playground is on main so is indeed the best place to test recent changes.

For hyphens, the PRQL would require backticks. Does that work? Apologies if not!

max-sixty:
@blackary I'm at a computer now, and the backticks work, I think:

```elm
prql dialect:bigquery
from `bigquery-public-data.pypi.file_downloads`
```

```sql
SELECT
  `bigquery-public-data.pypi.file_downloads`.*
FROM
  `bigquery-public-data.pypi.file_downloads`
```

Requiring that we surround operators like minus in backticks is lets us have a more context-free parser — it can parse the subtraction `a-b` in `select a-b` without knowing it's in a `select` rather than a `from`, and `a-b` isn't a table. I hope that's a reasonable tradeoff!

blackary:
@max-sixty Adding backticks works for me. That seems like a reasonable tradeoff, and that should work fine for my use-case. As soon as there's a new release of prql-python, I'll be trying this out in some real-world cases, and will let you know if I run into any other issues.

max-sixty:
Great, I just pushed that, so it should be out in the next few minutes

PR #875, opened 2022-07-28
Switch implicit `compile` tests to actual `compile` tests

max-sixty:


max-sixty:
I'm going to try and develop something that 
a) aggregates the tests into fewer cases
b) tests each stage of the pipeline for each case

But merging this for the moment, so we have a better place to put the simple tests

PR #880, opened 2022-07-29
Fix copying to clipboard on firefox

mklopets:
Turns out there was a way simpler way all along, and one that's supported by Firefox.

Bug report by @max-sixty under https://github.com/prql/prql/pull/825#issuecomment-1197219408


max-sixty:
Thanks a lot for fixing @mklopets !

PR #894, opened 2022-07-30
feat: Add union operator

max-sixty:
This is still a WIP -- I'm getting a error in semantic which I need to spend some more time on.

Any thoughts on the language design? Currently this has a required positional param to specify the type of the union -- should we do that vs. an optional named param?

### DIFF
--- a/prql-compiler/src/ast/pl/expr.rs
+++ b/prql-compiler/src/ast/pl/expr.rs
@@ -277,6 +277,7 @@ pub enum TransformKind {
         range: Range,
         pipeline: Box<Expr>,
     },
+    Concat(Box<Expr>),
 }
 
 #[derive(Debug, PartialEq, Eq, Clone, Serialize, Deserialize)]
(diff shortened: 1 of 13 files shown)


aljazerzen:
I like the language design: in most cases you'd write `union all`, which is quite self-explanatory. We could use a default value here, but we'd have to convert to a named argument:

```elm
from table1
union kind:distinct table2
```

----

Regarding the comment about resolving: to be honest I don't remember how to enable table resolving at `main` branch because I redid that at `semantic` branch 2 times :D Can this PR wait a week for `semnatic` to be merged?

max-sixty:
> Can this PR wait a week for `semantic` to be merged?

For sure!

aljazerzen:
I've tried to convert this to a "good-first-issue", but it included work on many parts of the codebase, so I just did it myself.

It needs docs, though.

This is how it works:

```elm
from employees
concat managers
union other_employees
```

max-sixty:
Awesome!! I am happy to do the docs

PR #902, opened 2022-08-02
fix(parser): fix parse error when new lines are present in `table`

### LINKED ISSUE #901: Empty lines in `table` not parsing
```elm
table x = (
  from x_table

  select only_in_x = foo
)

from x
```

...raises an error because of the empty line.

For anyone familiar with an analytical parser, this should be a fairly simple PR, and even without experience, this should be quite tractable and a good intro.

sebastiantoh:
Resolves https://github.com/prql/prql/issues/901

### DIFF
--- a/prql-compiler/src/prql.pest
+++ b/prql-compiler/src/prql.pest
@@ -25,7 +25,7 @@ type_term = { ident ~ type_def? }
 
 table = { "table" ~ ident ~ "=" ~ nested_pipeline }
 
-pipe = _{ NEWLINE | "|" }
+pipe = _{ NEWLINE+ | "|" }
 pipeline = { WHITESPACE* ~ expr_call ~ (pipe ~ expr_call)* }
 
 // See parser.rs for why this is so complicated to parse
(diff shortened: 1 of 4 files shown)


prql-compiler/src/prql.pest
```diff
@@ -25,7 +25,7 @@ type_term = { ident ~ type_def? }
 
 table = { "table" ~ ident ~ "=" ~ nested_pipeline }
 
-pipe = _{ NEWLINE | "|" }
+pipe = _{ "|" | NEWLINE* }
```
max-sixty: I _think_ this should be a `+` rather than a `*` — i.e. it should have at least one newline. I'm kinda surprised the whole thing works with a `*` — meaning nothing at all is a valid `pipe`!

```suggestion
pipe = _{ "|" | NEWLINE+ }
```

Is that right?

max-sixty (commented):
@sebastiantoh thank you!

I added one comment.

Great to have you as a contributor. Let me know any feedback on getting set up on the repo / running the tests etc.

prql-compiler/src/prql.pest
sebastiantoh: oops, you're right. I've updated my PR!

max-sixty:
Thanks a lot @sebastiantoh ! Great to have you as a contributor.

PR #904, opened 2022-08-03
refactor: Add `fold_query` to ASTFold

### LINKED ISSUE #820: Same column name in multiple tables causes buggy behaviour
When multiple PRQL `table`s share an identical column name (e.g. `id`): joining them together causes the compiler to reuse the meaning of one `id` in multiple places (even where it's incorrect).

This reduced example should illustrate it best:

```
table x = (
  from x_orig # table with columns: id
  derive [
    x_id = id
  ]
  select [x_id]
)

table y = (
  from y_orig # table with columns: id, x_id
  derive [
    y_id = id
  ]
  select [y_id, x_id]
)

from x
join y [x_id]
```

compiled SQL:
```
WITH x AS (
  SELECT
    id AS x_id
  FROM
    x_orig
),
y AS (
  SELECT
    x_id AS y_id, --------- this should instead be `id AS y_id`
    x_id
  FROM
    y_orig
)
SELECT
  x.*,
  y.*,
  x_id
FROM
  x
  JOIN y USING(x_id)
```

Note that if the PRQL derive statement of `y_id = id` is changed to `y_id = y_orig.id`, then the buggy output SQL line correctly becomes `id AS y_id`. But, we can't expect users to always add these (unless this becomes an explicit requirement)

max-sixty:
I think to fix #820 we're going to need to know "where we are" in the
folding — what tables we've seen before, what is currently visible;
which I think means we want to control how each table is parsed from a
higher method.

So this:
- Adds a `fold_query` method
- Moves some functions from taking nodes to taking a `Query` object. I'm
  not that confident this is better everywhere, so we can revert parts.

### DIFF
--- a/prql-compiler/src/semantic/frame.rs
+++ b/prql-compiler/src/semantic/frame.rs
@@ -19,9 +19,9 @@ pub struct Frame {
 /// Columns we know about in a Frame. The usize value represents the table id.
 #[derive(Debug, Clone, PartialEq, Serialize, Deserialize)]
 pub enum FrameColumn {
-    /// Equivalent of `foo_table.*`
+    /// Used for `foo_table.*`
     All(usize),
-    // TODO: when is this used?
+    /// Used for `derive a + b` (new column has no name)
     Unnamed(usize),
     Named(String, usize),
 }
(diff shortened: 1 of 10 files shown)


aljazerzen:
Good change, it should be like this all along

prql-compiler/src/semantic/frame.rs
```diff
@@ -19,9 +19,9 @@ pub struct Frame {
 /// Columns we know about in a Frame. The usize value represents the table id.
 #[derive(Debug, Clone, PartialEq, Serialize, Deserialize)]
 pub enum FrameColumn {
-    /// Equivalent of `foo_table.*`
+    /// Used for `foo_table.*`
     All(usize),
-    // TODO: when is this used?
+    /// Used for `derive a + b` (new column has no name)
```
max-sixty, line 24: Thanks!

PR #907, opened 2022-08-05
fix(compiler): ident escape during fmt

aljazerzen:
- only emit ident backticks during Item::display if they are required
- test for fmt

### DIFF
--- a/prql-compiler/src/ast/item.rs
+++ b/prql-compiler/src/ast/item.rs
@@ -227,7 +227,7 @@ impl Display for Item {
                 match ne.expr.item {
                     Item::FuncCall(_) => {
                         // this is just a workaround for inserting parenthesis
-                        // full approach would include "binding strength" 
+                        // full approach would include "binding strength"
                         // checking, just as we do for SQL
                         write!(f, "{} = ({})", Item::Ident(ne.name.clone()), ne.expr.item)?;
                     }
--- a/prql-compiler/src/ast/item.rs
+++ b/prql-compiler/src/ast/item.rs
@@ -206,21 +206,39 @@ impl Display for Item {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
         match self {
             Item::Ident(s) => {
-                write!(f, "`{s}`")?;
+                fn forbidden_start(c: char) -> bool {
+                    !(('a'..='z').contains(&c) || matches!(c, '_' | '$'))
+                }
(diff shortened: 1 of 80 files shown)


prql-compiler/src/ast/item.rs
```diff
@@ -206,21 +206,39 @@ impl Display for Item {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
         match self {
             Item::Ident(s) => {
-                write!(f, "`{s}`")?;
+                fn forbidden_start(c: char) -> bool {
```
max-sixty, line 209: Should we add a Todo for unifying with the Translator, or is this semantically different?

prql-compiler/src/cli.rs
```diff
@@ -197,4 +197,37 @@ sort full
...
+ derive      `želva_means_turtle`   =    (`column with spaces` + 1) * 3
+group a_column (take 10 | sort b_column | derive [the_number = rank, last = lag 1 c_column] )
+        "#,
+        )
+        .unwrap();
+
+        // this test is here just to document behavior - the result is far from being correct:
+        // - indentation does not stack
+        // - operator precedence is not considered (parenthesis are not inserted for numerical
+        //   operations but are always inserted for function calls)
```
max-sixty, line 216: 👍

prql-compiler/src/cli.rs
```diff
@@ -197,4 +197,37 @@ sort full
         sort full                                              # [?, full, gender]
         "###);
     }
+
+    #[test]
+    fn format_test() {
+        let output = Cli::execute(
```
max-sixty, line 203: FYI there is an insta library for CLI now! (I added a link to the issue on CLI tests, can find it when back on a laptop)

max-sixty (approved):
Great!

prql-compiler/src/ast/item.rs
aljazerzen: This is different - functionally it's the same, but it translates into different languages, which means it could be different.

prql-compiler/src/cli.rs
aljazerzen: :+1:

max-sixty:
As I was preparing the changelog, I noticed that the commit message for this was the original PR title, not the one when we merged. That's despite having this option selected:
image

That renders the optional check for the PR title not that helpful. Though this isn't the thing that's decide whether PRQL is successful or not :)

image

PR #923, opened 2022-08-11
feat: add prql library generation to be used in other platforms (e.g. Golang)

sigxcpu76:
I've added a small shim that exports `to_sql` and `to_json` and generates static and dynamic libraries that can be used through FFI in other platforms. I am using it in Go.


max-sixty:
Thanks @sigxcpu76 ! Great to have you as a contributor.

Is the intention this would be released to crates.io? If not, could we add `publish = false` (e.g. https://github.com/prql/prql/blob/main/prql-js/Cargo.toml#L6) to the cargo.toml?

I don't have much experience for building these, so I defer to you / others. As long as you're OK with changes in the future, it's fine to merge this as-is.

Cheers!

sigxcpu76:
I have no opinion on this. I am a Rust newbie, so I don't even know too much about publishing. If this can be integrated "as is" so be it. I am OK with changes in the future as long as they keep it simple (which I assume it is the default). Using external libraries from Go is a PITA anyway, so I tried to make it as simple as I could.

max-sixty:
OK cool. How do you use it from go? 

My guess is that this shouldn't be published to crates.io, (or if it is, it should possibly be integrated into `prql-compiler`?) Would you be OK to add that line so it doesn't automatically get published there?

max-sixty:
> My guess is that this shouldn't be published to crates.io, (or if it is, it should possibly be integrated into `prql-compiler`?) Would you be OK to add that line so it doesn't automatically get published there?

Ah, sorry, you already did! I missed it.

sigxcpu76:
I see that it already has "publish = false".
```[package]
edition = "2021"
name = "prql-lib"
publish = false
version = "0.2.6"```

**Comment by max-sixty** [comment]
One more thing — if you have the patience, it would be great to have a very short `README` on how you've used it from `go`. I'm OK merging as-is though.

**Comment by sigxcpu76** [comment]
I was about to add a README. Give me 5 minutes.

**Comment by max-sixty** [comment]
(merge main to get the rust 1.63 lints fix!)

File: prql-lib/src/lib.rs [hunk 1]
```diff
@@ -0,0 +1,57 @@
+extern crate libc;
+
+use libc::{c_char, c_int};
```
max-sixty, line 5: I think we want this line: https://github.com/prql/prql/blob/main/prql-python/src/lib.rs#L0-L1, to not attempt to build this with wasm

prql-lib/README.md
```diff
@@ -0,0 +1,81 @@
...
+
+Copy the `.a` and `.so` files in a convenient place and add the following compile flags to Go (cgo):
+
+`CGO_LDFLAGS="-L/path/to/libprql_lib.a -lprql_lib -pthread -ldl" go build`
+
+## Code
+
+Below is an example from an actual application that is using PRQL in Go.
+
+```
```
max-sixty:
```suggestion
```go
```

prql-lib/src/lib.rs
sigxcpu76: I don't know how to prevent that. It is above my Rust knowledge.
Can you tell me what directives/pragmas are needed for that?

prql-lib/README.md
sigxcpu76: I've pushed a commit with this change. Thanks

prql-lib/src/lib.rs
sigxcpu76: Found it in python library. I'll add it.

max-sixty: Sorry, I just meant — could we put the linked line at the top of this file too?

prql-lib/README.md
```diff
@@ -0,0 +1,81 @@
+# PRQL library target
```
max-sixty, line 1: Great readme! Thanks @sigxcpu76

max-sixty:
If you can enable me to push to the PR (https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork) I can help fix the lints. 

Or here's at least some of the diff, from running `pre-commit run -a`:

```diff
commit 2859efd166f8c4ba90d711f4feefb28df17071cf
Author: Maximilian Roos <m@maxroos.com>
Date:   Thu Aug 11 14:07:14 2022 -0700

    Lint

diff --git a/prql-lib/Cargo.toml b/prql-lib/Cargo.toml
index 60c0392..d83cbe8 100644
--- a/prql-lib/Cargo.toml
+++ b/prql-lib/Cargo.toml
@@ -15,4 +15,4 @@ serde_json = "1.0"
 [package.metadata.release]
 shared-version = true
 tag-name = "{{version}}"
-tag-prefix = ""
\ No newline at end of file
+tag-prefix = ""
diff --git a/prql-lib/README.md b/prql-lib/README.md
index 0191613..8a9042d 100644
--- a/prql-lib/README.md
+++ b/prql-lib/README.md
@@ -1,6 +1,7 @@
 # PRQL library target
 
 ## Description
+
 This module compiles PRQL as a library (both `.a` and `.so` are generated). This allows embedding in languages that support FFI - looking at Golang.
 
 ## Usage
@@ -13,7 +14,7 @@ ## Code
 
 Below is an example from an actual application that is using PRQL in Go.
 
-```go
+````go
 package prql
 
 /*
@@ -39,7 +40,7 @@ ## Code
     // from the PRQL compiler with a very short query
 	cStringBufferLength := 1024
 
-    // allocate a buffer 3 times the length of the PRQL query to store the 
+    // allocate a buffer 3 times the length of the PRQL query to store the
     // generated SQL query
 	if len(prql)*3 > cStringBufferLength {
 		cStringBufferLength = len(prql) * 3
@@ -78,4 +79,4 @@ ## Code
 
 	return "", errors.New(C.GoString(cstr))
 }```
-
+````
diff --git a/prql-lib/src/lib.rs b/prql-lib/src/lib.rs
index c77b455..526b548 100644
--- a/prql-lib/src/lib.rs
+++ b/prql-lib/src/lib.rs
@@ -56,4 +56,4 @@ pub extern "C" fn to_json(query: *const c_char, out: *mut c_char) -> c_int {
     } else {
         return 0;
     };
-}
\ No newline at end of file
+}
```

I'm also getting some lint errors from clippy though.

sigxcpu76:
I guess that's all good now. Thanks for your patience.

max-sixty:
I'm still seeing errors from cargo clippy. Some of those require `git merge main`. Some are very easy (e.g. remove `return` statements), one I need to look at around `unsafe`.

You can see them with `cargo clippy`, or if you can enable me to push to the PR (https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork) I can help fix them.

sigxcpu76:
It looks like I cannot give you push access to this PR. I don't have that checkbox "Allow edits from maintainers"

sigxcpu76:
@max-sixty all good now. Happy Clippy!

max-sixty:
Awesome! Thanks for your patience resolving the lints. Great to have you as a contributor @sigxcpu76

gedw99:
hey @sigxcpu76 

I am a golang dev, and total Rust newbie.

But this shows a nice way to do packaging for golang: https://github.com/rajveermalviya/go-webgpu

The rust binary output is committed to github. Yeah its not great but it makes it a "go run" or "go install" for us dummies.

max-sixty:
Thanks for commenting @gedw99. Generally we should avoid adding binaries to the repo, given they'll grow the repo size on every change. But GitHub has a feature where releases can add "Assets" — e.g. [here](https://github.com/prql/prql/releases/tag/0.2.7) — we'd be very open to having them there. They're also downloadable with a simple HTTP call. Let us know whether that's something you'd be interested to explore further. Thanks!

max-sixty:
FYI one project that I just saw which has lots of built binaries in their release is Task: https://github.com/go-task/task/releases/tag/v3.17.0, if we want to use that as a template (though it's in `go` rather than `rust`)

PR #957, opened 2022-09-10
fix: Fixes #929 , removing line and row numbers from javascript exception message

### LINKED ISSUE #929: Fix line and row numbers in exceptions for prql-js
Line numbers and columns are both off by -1
2022-08-18 21_17_14-localhost_3000

qharlie:
These are already being reported in from the rust prql error.


max-sixty (approved):
Nice! Favorite type of PR :)

At some point we could have tests for this, but another day.

PR #975, opened 2022-09-16
ci: Build wasm on explicit rust version

max-sixty:


max-sixty:
This ended up being quite the yak shave

PR #1004, opened 2022-10-04
build: Upgrade clap to from version 3 to 4

aj-bagwell:


aj-bagwell:
Clap 4 adds a bunch of anoying restrictions to the lifetimes of things, so I had to make some changes to the book generator, an alternative would be to turn on the [string feature](https://epage.github.io/blog/2022/09/clap4/#removing-lifetimes) of clap.

max-sixty:
Thanks a lot @aj-bagwell !

> an alternative would be to turn on the [string feature](https://epage.github.io/blog/2022/09/clap4/#removing-lifetimes) of clap.

Whatever you think is simplest, I'm def fine with another feature.

I'll merge for now!

PR #1005, opened 2022-10-04
feat: expose plain error line

aljazerzen:
Replaces output of `format_error` which was `(String, Option<SourceLocation>)` with `FormattedError`.

Maybe we should also rename `message` into `output` (or some better name), so `FormattedError` would be:

```rust
struct FormattedError {
   line: String,
   output: String,
   location: Option<SourceLocation>,
}
```


max-sixty (approved):
Great, thanks a lot!

(I thought about whether it made sense to have this info on the `Error` itself, but it would require lots of context, this seems very reasonable indeed)

PR #1016, opened 2022-10-10
docs: Add changelog for next version

max-sixty:
I've bumped the minor version here, to 0.3.0. This might be a bit
conservative, because I see the stdlib issue as more of a bug than a
change.

But maybe people like @mklopets are relying on that?

(I guess 0.1.0 & 0.2.0 were both big releases, and this is very small.
But they don't _have_ to be coupled)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,11 +1,11 @@
 # PRQL Changelog
 
-## 0.3.0 — 2022-10-17
+## 0.2.8 — 2022-10-17
 
-0.3.0 is another modest release with some fixes, doc improvements, bindings
-improvements, and lots of internal changes. We're bumping the minor part of the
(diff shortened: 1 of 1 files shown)


mklopets:
Not sure if you meant us relying on semantic versioning or these 2 stdlib functions but we haven't been relying on either one!  IMO conservative versioning (at least when choosing between minor/patch) is good.

Due to ~being uninformed~ reasons we'd currently been using S-strings for both of these 😅

mklopets:
On second thought, for "PR & marketing purposes" pre-1.0, reserving minor releases for things that are relatively major _might_ be a good move. 🤷

max-sixty:
OK, doing 0.2.8! @mklopets's second thought persuaded me...

PR #1022, opened 2022-10-12
feat: Add Dockerfile to simplify developer tool installation

richb-hanover:
I am a [fan of using Docker](https://randomneuronsfiring.com/more-fun-with-docker-tooling/) to create reproducible build environments, especially if they're complicated to set up. An added advantage is that once a Docker container is working, it's not subject to the vagaries of version changes when using the "latest" tools.

The Dockerfile in this PR replicates the `cargo test` environment, and lets people set it up without even needing to install rust or cargo. (They must, of course, install Docker.)

Some thoughts about the state of this PR:

* It's a FIRST CUT. 
* My tests and examples display the Playground, Language Book, and Website. 
* I am unsure how to test the other components - please consider updating the USING_DOCKER.md document (or tell me how to do it).
* I have not/do not know how to pin various version numbers to all the dependencies. This would be a good second step so that the build environment is entirely reproducible. 
* As an example, when working on the `book`, there's a warning about version conflicts using mdbook-admonish (0.4.18 vs 0.4.21) that results in rebuilding something when I expect the versions should be the same.

I am happy to work with reviewers to make the Dockerfile and its documentation better.


richb-hanover:
Hmmm... Interesting commit error messages. Let me take a day to work on this before you spend time reviewing it.

.gitignore
```diff
@@ -7,3 +7,20 @@ dist
 /*.prql
 */*.prql
 target
+playground/node_modules
```
max-sixty, line 10: `node_modules` is in https://github.com/prql/prql/blob/main/playground/.gitignore#L4 — is there a reason we need it here too?

.gitignore
```diff
@@ -7,3 +7,20 @@ dist
...
+book/integrations
+book/language-features
+book/queries
+book/transforms
+book/*.js
+book/*.html
+book/*.css
+book/.nojekyll
+book/*.png
+book/*.svg
```
max-sixty, line 26: All these artifacts should go into `book/book`, and that's gitignored in `book`. Did it work differently on your system?

README.md
```diff
@@ -72,8 +72,11 @@ To stay in touch with PRQL:
 - Join us on [Discord](https://discord.gg/eQcfaCmsNc)
 - Star this repo
 - [Contribute][contributing] — join us in building PRQL, through writing
-  code or inspiring others to use it. It's easy to get started — the project can
-  be built in a couple of commands, and we're a really friendly community!
+  code or inspiring others to use it.
+- See the [development](DEVELOPMENT.md) documentation for PRQL.
```
max-sixty, line 76: 👍

Taskfile.yml
```diff
@@ -82,26 +117,24 @@ tasks:
...
-      - which hugo
-      - which npm
-    preconditions:
-      - sh: which brew
-        msg: |
-
+  install-hugo:
+    cmds:
+      # - /home/linuxbrew/.linuxbrew/bin/brew install hugo
+      - curl -L https://github.com/gohugoio/hugo/releases/download/v0.91.2/hugo_0.91.2_Linux-64bit.deb -o hugo.deb
```
max-sixty, line 130: One advantage of using `brew` is that it'll work on both Mac & Linux. 

My guess is that a majority of people are on MacOS? At the very least a decent proportion. So I don't think we should only provide Linux scripts. (Maybe stereotyping but Mac folks are also likely to benefit more from the scripts!)

Is there a reason we can't just `brew install hugo npm`?

Taskfile.yml
```diff
@@ -35,17 +57,25 @@ tasks:
...
         wasm-bindgen-cli
         wasm-pack
+
+  install-brew:
+    # brew might be installed, so check first
+    status:
+      - which brew
+    cmds:
+      # https://stackoverflow.com/questions/58292862/how-to-install-homebrew-on-ubuntu-inside-docker-container
+      - useradd -m -s /bin/zsh linuxbrew
```
max-sixty, line 67: Adding users is quite invasive, so it's fine to do in a Docker container, but not on someone's system. So I would suggest we put these in the Dockerfile, and revert to the "Please install brew" here. (and it needs to be compatible with MacOS here as below)

Taskfile.yml
```diff
@@ -58,20 +88,25 @@ tasks:
...
+      # If vscode isn't installed, or there are no missing extensions,
+      # return 0 and exit early.
       - '[ ! -x "$(which code)" ] || {{ not (contains "❌" .missing_extensions) }}'
     silent: true
     cmds:
       - |
         echo "
 
-        🟡 It looks like VSCode is installed but doesn't have all recommended extensions installed:
+        🟡 It looks like VSCode is installed but doesn't have all
```
max-sixty, line 99: This rewrap changes the semantics, it shouldn't rewrap automatically here!

USING_DOCKER.md
```diff
@@ -0,0 +1,93 @@
+# Using the Dockerfile
```
max-sixty, line 1: I think we should put this in `DEVELOPING.md`, it fits well and will be easy to find there. WDYT?

USING_DOCKER.md
```diff
@@ -0,0 +1,93 @@
...
+
+**Website** Use the command above, then enter:
+
+```
+cd website; hugo server --bind 0.0.0.0 -p 3000
+```
+
+_[I don't know how to run/test the following components]_
+
+**prql-compiler** Use the command above, then enter:
```
max-sixty, line 59: For all of these, people can run from the root, and cargo takes care of it (e.g. passing `-p prql-compiler` to cargo will only compile that crate

USING_DOCKER.md
```diff
@@ -0,0 +1,93 @@
...
+**Playground** Use the command above, then enter:
+
+```
+cd playground; npm start
+```
+
+**Language Book** Use the command above, then enter:
+
+```
+cd book; mdbook serve -n 0.0.0.0 -p 3000
```
max-sixty, line 48: Nice. A potential follow-up would be to set the default port in `mdbook` confighs so we don't even need to supply it here (and same for hugo)

Taskfile.yml
```diff
@@ -21,10 +21,32 @@ vars:
...
+      - task: install-hugo
+      - task: install-nodejs
+      - task: install-npm-dependencies
+      # We only suggest, rather than install;
+      #    we don't want to intrude (maybe we're being too conservative?).
+      - task: suggest-vscode-extensions
+      - task: install-precommit-install-hooks
+      - cmd: |
+          echo "
+          🟢 Set up complete! ✅🚀"
```
max-sixty, line 41:
```suggestion
          🟢 Setup complete! ✅🚀"
```
(my fault!)

max-sixty (commented):
Thanks a lot for putting the time in here @richb-hanover !

I think we should ensure it's compatible with Mac — it doesn't mean we have to script everything for Mac, just not run anything that can't be run, and instead direct people (e.g. "please install brew"!). And ensure anything that could run on people's system is safe. We definitely shouldn't have anything that involves `sudo`.

(FWIW I find `cargo` provides excellent isolation, so I don't use Docker for rust development. Even with python, I use `poetry` now. But I appreciate that it's not always the case and some people like Docker. So I'm fine to have this and appreciate your work)

max-sixty:
> As an example, when working on the `book`, there's a warning about version conflicts using mdbook-admonish (0.4.18 vs 0.4.21) that results in rebuilding something when I expect the versions should be the same.

Yeah, I don't think this is important, but false positives aren't great. My guess is that regenerating the `mdbook-admonish` artifacts will make the warning go away. But low priority

TaxFairness:
A quick response to your observations...

- The macOS or Linux installation procedure can, and should, be different from Docker - they will serve different audiences. 
- I realize I screwed up the `cargo test` (non-Docker) installation procedure in the Taskfile.yml. I'll fix that by using brew to install `hugo` and `npm` in the `setup-dev` task. (I removed `brew` because it takes 5+ minutes to install in Docker. It wasn't worth the effort when those programs take take ~10 seconds to install individually...) 
- I realize that poetry for Python (or pyenv), and nvm for Node projects, and other "environment management systems" are popular with people who are heavily invested in a particular language. But for a software dilettante like me (who hoovers up projects in loads of different language on a regular basis), my stomach begins to hurt when someone says, "Just run xxx in [ poetry | pyenv | nvm | ... ]" to set up an environment. It's never "just do anything" - it requires you to buy into that system wholesale... Case in point: I originally had an ancient version of rust on my macOS machine that was initially throwing all sorts of errors. I had to Google "how to uninstall rust" and then re-install a current version.
- The things that attract me to the Docker process include:
  - It's identical on macOS, Linux, or Windows. No special instructions necessary
  - It's "can't fail" - a well-constructed Dockerfile will always replicate the same environment
  - Most importantly, it will never "pollute" my other development environments. 
- I had originally combined **Using the Dockerfile** in the DEVELOPMENT.md file, but realized that there was only incidental commonality between them. Although both list the components and give hints for making them run, the Docker hints are somewhat different from those of the "standard install". 

I'll work further on this in the next day or so. Thanks for your review!

max-sixty:
> (I removed `brew` because it takes 5+ minutes to install in Docker. It wasn't worth the effort when those programs take take ~10 seconds to install individually...)

Very reasonable. I think if it's _just_ for Docker, then I would vote to put them in the Dockerfile, and have the Taskfile for more general usage

> "Just run xxx in [ poetry | pyenv | nvm | ... ]" to set up an environment. It's never "just do anything" - it requires you to buy into that system wholesale... Case in point: I originally had an ancient version of rust on my macOS machine that was initially throwing all sorts of errors. I had to Google "how to uninstall rust" and then re-install a current version.

Yes I very much appreciate this. e.g. if someone wanted to make an edit to the Book and didn't have a rust toolchain, it's quite useful.

Thank you!

.github/workflows/super-linter.yml
```diff
@@ -0,0 +1,18 @@
+name: Super-Linter
```
max-sixty, line 1: (not sure if you want feedback while you're iterating — let me know either way! — if you do — this is reasonable but let's add to `pull-request.yaml`; it doesn't need its own workflow. And fine to PR things in separately if that lets you move faster)

richb-hanover:
Oh dear. I had forgotten that I had the PR outstanding, and worse, that it bothers you while I stumble around in my ignorance. I am going to close this and work on it privately until I have something to show. 

(I didn't have Github Actions turned on in my fork, so I had never seen the hailstorm of grumpy notes that were emitted when I made a PR.)

PS If I have strength, I may work on some of the warnings about dependencies. The one that I don't know how to correct is that all the ~~workflows~~ actions have `using: composite`, and this elicits warnings about Node12. Where is `composite` defined, and how would I change it to use Node16? Thanks.

max-sixty:
Sorry @richb-hanover — I didn't mean to interfere! It's totally fine — I get lots of GH notifications, and can easily ignore the ones if you'd rather wait until the end to review.

Totally as you wish re working on an open PR vs your own branch. You can also use "Draft" if you want to make that legible.

I appreciate your contributions so far and hope you can continue to be part of PRQL!

richb-hanover:
Your comments don't cause a problem. I enjoy learning about how all these bits work together (it's a series of progressive discoveries... Oh - look - that's what happens if I wiggle this lever... :-)

Any thoughts on the `using: composite` question? Thanks again

max-sixty:
> Any thoughts on the `using: composite` question? Thanks again

I _think_ `composite` is the default (some details, but no great summary, at https://docs.github.com/en/actions/creating-actions/creating-a-composite-action & https://wallis.dev/blog/composite-github-actions).

I think probably the way around that is adding something like https://github.com/actions/setup-node to the workflow steps which use node, rather than relying on the old default version. Lmk any issues!

max-sixty:
I see what you mean about the node 12 warnings — they're everywhere, e.g. https://github.com/prql/prql/actions/runs/3253076848

I think many of those will need to be updated upstream, e.g. in the warning above: `Please update the following actions to use Node.js 16: amannn/action-semantic-pull-request`

richb-hanover:
> I think many of those will need to be updated upstream, e.g. in the warning above: Please update the following actions to use Node.js 16: amannn/action-semantic-pull-request

https://github.com/amannn/action-semantic-pull-request/blob/main/CHANGELOG.md#501-2022-10-14

You can see the future... :-)

richb-hanover:
Completely OT, and since I'm being a noodge... :-) The README currently says:

> **Current status**
>
> After several months of building, PRQL is ready to use! Check out the [0.2 Release Notes](https://github.com/prql/prql/releases/tag/0.2.0)!

Is there any news now (or on the near horizon) that could update the README? It feels as if you have accomplished most of what's implied here. What is next on your roadmap? What are the next "big issues"? (Integration with various tools seems interesting. What are your thoughts about how to accomplish this?) Thanks, as always.

max-sixty:
> Is there any news now (or on the near horizon) that could update the README? It feels as if you have accomplished most of what's implied here. What is next on your roadmap? What are the next "big issues"? (Integration with various tools seems interesting. What are your thoughts about how to accomplish this?) Thanks, as always.

Definitely, I think it's definitely due for an update, and makes it look a bit stale at the moment. Thanks a lot for suggesting. I'll give it an update, will open an issue to remind myself

PR #1028, opened 2022-10-14
docs: book chapter on semantics

aljazerzen:
@max-sixty, I've adapted your draft of "High level design" of name semantics we discussed on Discord.
I also added a page on functions.

I've written this as it already has consensus, but that's just so it can be part of the book.
It is open to discussion. It is also a summary of how name currently works, so I's a nice intro to the compiler, I think.

I'm planning to also add how currying should work and how functions are resolved.


book/src/semantics/functions.md
```diff
@@ -0,0 +1,68 @@
...
+a | foo 3 | bar 'hello' 'world' | baz
+```
+
+... is equivalent to ...
+
+```prql_no_test
+baz (bar 'hello' 'world' (foo 3 a))
+```
+
+You may not have noticed, but transforms are regular functions too!
```
max-sixty:
```suggestion
As you may have noticed, transforms are regular functions too!
```

nice to be optimistic about the reader! :)

book/src/semantics/functions.md
```diff
@@ -0,0 +1,68 @@
...
+
+```elm
```
max-sixty: (fyi, for the book, we can write `prql` or `prql_no_test`; it's only GH which requires `elm`. But no need to change)

book/src/semantics/functions.md
```diff
@@ -0,0 +1,68 @@
...
+filter age > 50 (from employees) | sort name
+```
+
+... which is the same as:
+
+```elm
+sort name (filter age > 50 (from (employees))
+```
+
```
max-sixty:
```suggestion
<!-- TODO @max-sixty to add something on idiomatic usage? i.e. don't write that last one, this is just to explain how it works -->
```

book/src/semantics/name-resolving.md
```diff
@@ -0,0 +1,76 @@
+# Name resolving
+
+Because PRQL primarily handles relational data, it has specialized scoping rules for referencing columns.
+
+## Scopes
+
+For pruposes of this book, a scope is the collection of all names one can reference from a specific point in the program.
```
max-sixty:
```suggestion
In PRQL's compiler, a scope is the collection of all names one can reference from a specific point in the program.
```
(nice definition btw!)

book/src/semantics/name-resolving.md
```diff
@@ -0,0 +1,76 @@
...
+
+In PRQL, names in the scope are composed from namespace and variable name which are spearated by a dot, similar to SQL.
```
max-sixty:
```suggestion
In PRQL, names in the scope are composed from namespace and variable name which are separated by a dot, similar to SQL.
```

book/src/semantics/name-resolving.md
```diff
@@ -0,0 +1,76 @@
+# Name resolving
```
max-sixty: I wonder whether this should be in an "Internals" section. Maybe it's also helpful for users doing debugging?

(no need to resolve now tbc)

book/src/semantics/name-resolving.md
```diff
@@ -0,0 +1,76 @@
...
+For pruposes of this book, a scope is the collection of all names one can reference from a specific point in the program.
+
+In PRQL, names in the scope are composed from namespace and variable name which are spearated by a dot, similar to SQL.
+Namespaces can contain many dots, but variable names cannot.
+
+> **Example**
+>
+> Name `my_table.some_column` is a variable `some_column` from namespace `my_table`.
+>
+> Name `foo.bar.baz` is a variable `baz` from namespace `foo.bar`.
```
max-sixty:
````suggestion
```admonish example
Name `my_table.some_column` is a variable `some_column` from namespace `my_table`.

Name `foo.bar.baz` is a variable `baz` from namespace `foo.bar`.
```
````

File: book/src/semantics/name-resolving.md [hunk 8]
```diff
@@ -0,0 +1,76 @@
...
+
+## Resolving
+
+For each ident we want to resolve, we search the scope's items in order. One of three things can happen:
+
+- Scope contains an exact match, e.g. a name that matches in namespace and the variable name.
+
+- Scope does not contain an exact match,
+  but the ident did not specify a namespace, so we can match a namespace that contains a `*` wildcard.
+  The matched namespace is also updated to contain this new variable name.
```
max-sixty:
```suggestion
  If there's a single namespace, the matched namespace is also updated to contain this new variable name.
```

book/src/semantics/name-resolving.md
```diff
@@ -0,0 +1,76 @@
...
+
+  In the case that there are multiple namespaces with a wildcard,
+  we don't match with neither of the namespaces, but match as `*.*` name.
+  This is a special feature that allows a column name which may reside in two different tables
+  not to be associated with any of them.
+  Instead, it is translated into the column name only, so database can determine which table it belongs to.
+  Note that this may lead to PRQL passing on ambigous queries to the database, instead of resulting error early.
+
```
max-sixty:
```suggestion
<!-- TODO @max-sixty to add a reference to roadmap about db cohesion? -->
```

max-sixty (approved):
Awesome! Thanks a lot!

I'm happy to take a pass at this, or other parts, to the extent it helps me get into the newer compiler code, and frees you up. Or even better if we go back & forth (much of this is much better than I could have done fwiw!)

book/src/semantics/name-resolving.md
aljazerzen: Yea, this whole semantics chapter is meant for advanced users or maybe even for compiler team only.

book/src/semantics/name-resolving.md
aljazerzen: Good idea.

PR #1056, opened 2022-10-20
chore(semantic): Ignore failing tests

max-sixty:


max-sixty:
While not all tests pass, `cargo test -p prql-compiler --lib` passes, so I'm using that locally 

I won't comment out the GHAs for expediency, but will come back to that if it causes issues.

PR #1058, opened 2022-10-20
feat: detect version mismatch

### LINKED ISSUE #1037: Reject queries that have newer version than the compiler itself
When the query specifies a version, but the compiler version is lower, reject the query with message:
```
This query uses version of PRQL that is not supported by your prql-compiler. You may want to upgrade it.
```

MarinPostma:
close #1037

### DIFF
--- a/book/src/queries/dialect_and_version.md
+++ b/book/src/queries/dialect_and_version.md
@@ -45,7 +45,7 @@ take 10
 PRQL allows specifying a version of the language in the PRQL header, like:
 
 ```prql
-prql version:1
+prql version:"0.2"
 
 from employees
 ```
--- a/prql-compiler/src/ast/item.rs
+++ b/prql-compiler/src/ast/item.rs
@@ -4,7 +4,7 @@ use std::fmt::{Display, Write};
 use anyhow::anyhow;
 use enum_as_inner::EnumAsInner;
 use itertools::Itertools;
-use semver::{Comparator, Op, Prerelease, VersionReq};
+use semver::VersionReq;
 use serde::{Deserialize, Serialize};
 
 use super::*;
@@ -43,19 +43,6 @@ impl Item {
     pub fn parse_version(self) -> std::result::Result<VersionReq, Self> {
         match self {
             Self::Literal(Literal::String(ref s)) => VersionReq::parse(s).map_err(|_| self),
-            Self::Literal(Literal::Integer(major)) => {
-                let comp = Comparator {
-                    op: Op::Exact,
-                    major: major.try_into().map_err(|_| self)?,
-                    minor: None,
-                    patch: None,
-                    pre: Prerelease::EMPTY,
-                };
-
-                Ok(VersionReq {
-                    comparators: vec![comp],
-                })
-            }
             _ => Err(self),
         }
     }
--- a/prql-compiler/src/ast/item.rs
+++ b/prql-compiler/src/ast/item.rs
@@ -4,6 +4,7 @@ use std::fmt::{Display, Write};
 use anyhow::anyhow;
 use enum_as_inner::EnumAsInner;
 use itertools::Itertools;
+use semver::{Comparator, Op, Prerelease, VersionReq};
 use serde::{Deserialize, Serialize};
 
 use super::*;
@@ -36,6 +37,28 @@ pub enum Item {
     FString(Vec<InterpolateItem>),
     Interval(Interval),
     Windowed(Windowed),
+}
+
+impl Item {
+    pub fn parse_version(self) -> std::result::Result<VersionReq, Self> {
+        match self {
+            Self::Literal(Literal::String(ref s)) => VersionReq::parse(s).map_err(|_| self),
+            Self::Literal(Literal::Integer(major)) => {
+                let comp = Comparator {
+                    op: Op::Exact,
+                    major: major.try_into().map_err(|_| self)?,
+                    minor: None,
+                    patch: None,
+                    pre: Prerelease::EMPTY,
+                };
+
+                Ok(VersionReq {
+                    comparators: vec![comp],
+                })
+            }
+            _ => Err(self),
+        }
+    }
 }
 
 #[derive(
@@ -241,7 +264,7 @@ impl Display for Item {
             }
             Item::Query(query) => {
                 write!(f, "prql dialect:{}", query.dialect)?;
-                if let Some(version) = query.version {
+                if let Some(ref version) = query.version {
                     write!(f, " version:{}", version)?
                 };
                 write!(f, "\n\n")?;
@@ -383,6 +406,7 @@ impl Display for Item {
                 write!(f, "{}", literal)?;
             }
         }
+
         Ok(())
     }
 }
--- a/prql-compiler/src/lib.rs
+++ b/prql-compiler/src/lib.rs
@@ -11,8 +11,13 @@ pub use anyhow::Result;
 #[cfg(feature = "cli")]
 pub use cli::Cli;
 pub use error::{format_error, FormattedError, SourceLocation};
+use once_cell::sync::Lazy;
 pub use parser::parse;
+use semver::Version;
 pub use sql::translate;
+
+pub(crate) static PRQL_VERSION: Lazy<Version> =
+    Lazy::new(|| Version::parse(env!("CARGO_PKG_VERSION")).expect("Invalid PRQL version number"));
 
 /// Compile a PRQL string into a SQL string.
 ///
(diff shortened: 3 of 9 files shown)


prql-compiler/src/parser.rs
```diff
@@ -2041,4 +2055,46 @@ join `my-proj`.`dataset`.`table`
                     named_args: {}
         "###)
     }
+
+    #[test]
+    fn check_valid_version() {
```
max-sixty, line 2068: Very nice tests!

prql-compiler/src/parser.rs
```diff
@@ -2041,4 +2055,46 @@ join `my-proj`.`dataset`.`table`
...
+            r#"
+        prql version:"{}"
+        "#,
+            env!("CARGO_PKG_VERSION_MAJOR")
+        );
+        assert!(parse(&stmt).is_ok());
+
+        let stmt = format!(
+            r#"
+            prql version:"{}.{}"
```
max-sixty, line 2079: Do we want this to be a string (i.e. `version:"1.0"`), or just bare `version:1.0`. 

The existing approach is bare.

But if we're doing full semver, potentially `1.0.0` will not be a valid value

prql-compiler/src/parser.rs
```diff
@@ -28,7 +30,21 @@ pub(crate) type PestRule = Rule;
...
+    if let Some(ref version) = query.version {
+        check_query_version(version, &PRQL_VERSION)?;
+    }
+
+    Ok(query)
+}
+
+fn check_query_version(query_version: &VersionReq, prql_version: &Version) -> Result<()> {
+    if !query_version.matches(prql_version) {
+        bail!("This query uses version of PRQL that is not supported by your prql-compiler. You may want to upgrade it.");
```
max-sixty:
```suggestion
        bail!("This query uses version of PRQL that is not supported by your prql-compiler. You may want to upgrade the compiler.");
```

prql-compiler/src/ast/item.rs
```diff
@@ -38,6 +39,28 @@ pub enum Item {
     Windowed(Windowed),
 }
 
+impl Item {
+    pub fn parse_version(self) -> std::result::Result<VersionReq, Self> {
+        match self {
+            Self::Literal(Literal::String(ref s)) => VersionReq::parse(s).map_err(|_| self),
+            Self::Literal(Literal::Integer(major)) => {
```
max-sixty: Does this currently accept a bare int like `version:0`? (sorry if I'm being slow at reading the code)

I'm not sure whether we should have the version as a string or bare, but I would vote to only accept one.

max-sixty (commented):
Very nice code @MarinPostma ! Thank you

I have one question on how this should work at the language level. Let me know if you have thoughts, and CC @prql/prql-team

max-sixty:
We will also need to change the book to the thing that now works; i.e. using `0` rather than `1`. 

If you are up for adding a whatsnew & adding a brief explanation to the book, that would be great, but it's also fine to just do the rust, and we'll do the docs changes later.

prql-compiler/src/ast/item.rs
MarinPostma: yes, we basically accept either a string representing a semver version requirement or an integer representing the major version number.

prql-compiler/src/parser.rs
MarinPostma: there are a few way we can go about that I think:
- Make the version a syntax element, I don't like that, because it causes the syntax to become ambiguous
- Current state: accept integer or string that parses into version, when an int is passed, interpret it as a major
- Current version + support for floats, so we also support `major.minor` versions as plain numbers

MarinPostma:
I'll take a look at the book too :) once we settle on the syntax

MarinPostma:
I have failing tests to, I need to take care of that

prql-compiler/src/parser.rs
aljazerzen: I also don't like having a new syntax just for semver, but also don't like it to be parsed as float, because it's not a float (`1.10 != 1.1`).

So I'd argue against having bare `version:1.0` or `version:1.0.0`. I'm partial about shortcut `version:1`.

MarinPostma: Do you suggest that we just go full string? This is probably the sanest solution, but it break backward compat

aljazerzen: It would be the sanest option, true. I doubt that anyone is using this yet so don't worry about compat.

We can add int shortcut/special version syntax element later, if there is pressure to make it more elegant.

prql-compiler/src/parser.rs
```diff
@@ -28,7 +30,21 @@ pub(crate) type PestRule = Rule;
...
+    if let Some(ref version) = query.version {
+        check_query_version(version, &PRQL_VERSION)?;
+    }
+
+    Ok(query)
+}
+
+fn check_query_version(query_version: &VersionReq, prql_version: &Version) -> Result<()> {
+    if !query_version.matches(prql_version) {
+        bail!("This query uses version of PRQL that is not supported by your prql-compiler. You may want to upgrade the compiler.");
```
snth: Minor typo. I think it should read "This query uses **a** version of PRQL ...".

prql-compiler/src/lib.rs
```diff
@@ -11,9 +11,14 @@ pub use anyhow::Result;
 #[cfg(feature = "cli")]
 pub use cli::Cli;
 pub use error::{format_error, FormattedError, SourceLocation};
+use once_cell::sync::Lazy;
```
snth, line 14: If it's not too much trouble, for my education, could you explain what this does and why it's necessary?

prql-compiler/src/parser.rs
snth: Yeah, I also think backward compat is much of a concern at this stage.

I wish it could work without needing to quote the strings but I don't know enough about the parser grammar so happy to go with whatever you guys decide.

MarinPostma: We could add it to the grammar, but it would be ambiguous with floats, so we'd have to treat the whole version:xxx differently, to disambiguate

snth: Sorry about the typo, what I meant to say is I think backward compatibility is NOT much of a concern at this stage.

---
Right, your full string proposal sounds like the best then.

prql-compiler/src/parser.rs
max-sixty: @snth good spot (feel free to use the suggestion feature for something like this; i.e. 

```suggestion
        bail!("This query uses a version of PRQL that is not supported by your prql-compiler. You may want to upgrade the compiler.");
```

prql-compiler/src/parser.rs
max-sixty: > Do you suggest that we just go full string? This is probably the sanest solution, but it break backward compat

Yeah, I think at this stage we can be flexible, and make a guess if it's going to break anyone, and I would guess the same as others, that's it's fine. (And we already break things, because we're going to `0`!)



> We can add int shortcut/special version syntax element later, if there is pressure to make it more elegant.

👍

prql-compiler/src/lib.rs
MarinPostma: This creates a static variable at runtime that is initialized exactly once. This is because we parse the version from an env string, and we don't want to do that for every request. Maybe you are familiar with `lazy_static`, this does basically the same thing.

Also, once_cell is already a transitive dependency to most crates, including this, so adding it is free :)

prql-compiler/src/parser.rs
```diff
@@ -1866,6 +1880,14 @@ join `my-proj`.`dataset`.`table`
         )
         .is_err());
 
+        assert!(parse(
+            r###"
+        prql dialect:bigquery version:"2"
```
max-sixty:
```suggestion
        prql dialect:bigquery version:"25"
```

So we don't get a surprising failure when we release a new version, for a long time :)

book/src/queries/dialect_and_version.md
```diff
@@ -45,7 +45,7 @@ take 10
 PRQL allows specifying a version of the language in the PRQL header, like:
 
 ```prql
-prql version:1
+prql version:"0"
```
max-sixty:
```suggestion
prql version:"0.2"
```

(so it does something)

max-sixty:
Excellent @MarinPostma ! Thank you, this is high quality code, the tests especially.

I made a couple of tiiiny comments — we can do them another day if you want. Same for the book explanation and an entry in the changelog (I'm happy to do these at the next release)

I have given you merge permissions so hit the button whenever you're ready.

Issue #1041, opened 2022-10-18
Refactor semantic::resolve not to return semantic::Context


aljazerzen:
`semantic::Context` should be used only within `semantic` module. Currently it was exposed to the outside, because of the ambition to resolve many queries one after another, using the same context. We can put this off for now and make the API for resolving simpler.

So `semantic::resolve` has to be updated, and all of its usages fixed.

PR #1059, opened 2022-10-21
refactor: Remove `context` from resolving

max-sixty:
Closes https://github.com/prql/prql/issues/1041


aljazerzen (approved):
Looks great

max-sixty:
While these are small, I actually think they're reasonable issues, particularly for people who are new to the whole codebase. So if we can leave some for folks, and they get traction, great (and if not then we can either leave them longer or do them ourselves...)

PR #1061, opened 2022-10-22
build: Bump rust toolchain to 1.64.0

max-sixty:
While we generally rely on a cron workflow to bump the toolchain https://github.com/prql/prql/blob/f62908a9ce14a1e3fc59bfb93b2e9e3ee62a2a51/.github/workflows/cron.yaml#L48-L57, we need to bump this now — currently `task setup-dev` fails, because `cargo-release` has a MSRV of `1.64.0` (from https://github.com/crate-ci/cargo-release/pull/559), and cargo doesn't seem to have a way of finding the latest compatible version, or using a different toolchain for `cargo install` than it's using for the current project, so it just fails.

@epage I don't have some religious view on how fast that should be bumped, but @-ing you in case that's helpful context for your decision (thanks as ever for `cargo release`!)

### DIFF
--- a/rust-toolchain.toml
+++ b/rust-toolchain.toml
@@ -1,5 +1,5 @@
 [toolchain]
-channel = "1.63.0"
+channel = "1.64.0"
 components = ["rustfmt", "clippy"]
 # We want two targets: wasm32, and the default target for the platform, which we
 # don't list here. (i.e. we use each platform to test each platform)


epage:
> @epage I don't have some religious view on how fast that should be bumped, but @-ing you in case that's helpful context for your decision (thanks as ever for cargo release!)

Normally I keep a couple versions back.  This was more of a one off thing due to the importance of workspace inheritance for the scope of this release.

However, this change was big enough that you should probably double check your config and expected workflow.  It might be good in general to do `cargo-install@0.21` (or `cargo-install@0.22`) so you don't get versions across breaking releases.

max-sixty:
> However, this change was big enough that you should probably double check your config and expected workflow. It might be good in general to do `cargo-install@0.21` (or `cargo-install@0.22`) so you don't get versions across breaking releases.

Thanks — I'll do a separate PR to update. Thanks the subcommands, that fits with our workflow better!

PR #1064, opened 2022-10-22
docs: Update docs for releasing

max-sixty:
for cargo-release 0.22. Replaces #1062 for the moment

(But it's not exactly working how I expected, will post an issue)


max-sixty:
I'll merge, because it's not actually coupled to that issue; but without a resolution, we'll need to figure out the `cargo-release` version next time we do a release

PR #1065, opened 2022-10-22
build: Upgrade maturin version, use workspace versioning

max-sixty:


prql-python/pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [build-system]
```
messense, line 1: `cargo-extra-args = "--features pyo3/extension-module"` at the end of this file should be changed to `features = ["pyo3/extension-module"]`

See also https://www.maturin.rs/migration.html

max-sixty: Thanks so much @messense !

PR #1073, opened 2022-10-23
fix: Don't quote uppercase strings

### LINKED ISSUE #820: Same column name in multiple tables causes buggy behaviour

max-sixty:
This is a temporary workaround for @mklopets & team, while we resolve #820.

@aljazerzen -- not sure if you think this is an acceptable tradeoff? I'll wait for your review.

### DIFF
--- a/prql-compiler/src/lib.rs
+++ b/prql-compiler/src/lib.rs
@@ -235,18 +235,18 @@ table UPPER = (
 from UPPER
 join some_schema.tablename [id]
         "###)?), @r###"
-        WITH "UPPER" AS (
+        WITH UPPER AS (
           SELECT
             lower.*
           FROM
             lower
         )
         SELECT
-          "UPPER".*,
+          UPPER.*,
           some_schema.tablename.*,
           id
         FROM
-          "UPPER"
+          UPPER
           JOIN some_schema.tablename USING(id)
         "###);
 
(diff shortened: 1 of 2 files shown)


max-sixty:
Sounds like @mklopets already has a workaround, so maybe this isn't worth merging.

But @mklopets if it's 5x easier to change something in PRQL than your lib, then fine to change it in PRQL, even if it's not in the right direction for a narrow definition of correctness.

aljazerzen:
As I said on discord, I would prefer if PRQL would remain case sensitive.

max-sixty:
> As I said on discord, I would prefer if PRQL would remain case sensitive.

OK, given @mklopets has a workaround to the s-string issue, I'm closing

Issue #1031, opened 2022-10-15
Handling escaped idents which contain periods


max-sixty:
@brunorpinho suggested we use backticks for the ClickHouse dialect (https://discord.com/channels/936728116712316989/1030229933299077180). "That'll be super easy" I said. But actually the example presents a more general issue, regardless of the the specific dialect issue.

In SQL, `"event.type"` is treated as a column name. But it's not currently possible to represent an ident containing a period in PRQL. For example, in:

```elm
from github_json
derive [event_type_dotted = `event.type`]
```

``` `event.type` ``` will be parsed into an `event.type` ident in the PRQL AST — i.e. a single term with the period. And then later in the compiler, that term will be resolved into a namespace `event` and a column `type`. But we want a column name `event.type`!

---

I can see two options:

1. Change `Ident` into a proper struct — replace https://github.com/prql/prql/blob/10811eb25c0614fe99ff62b55d2a8b389d3bbfe5/prql-compiler/src/ast/mod.rs#L23 with a Struct that can hold a namespace and column name. 

    Then:
    - `event.type` is parsed into `{namespace: "event", column: "type"}` 
    - ``` `event.type` ``` is parsed into `{namespace: None, column: "event.type"}`

    ...but we use `Ident` for function names / variables / etc at the moment, and the parser deliberately doesn't have the context to discriminate. So we'll be carrying around a `{namespace: None, column: "sum"}` for the `sum` function[^1] 

2. Carry around the backticks where they're supplied in the PRQL query in the `Ident` struct until we get to the resolver, rather than parse them. So ``` `event.type` ``` stays as ``` `event.type` ``` in the AST, and then when we get to resolving the name, we check for backticks, and produce the column ``` `event.type` ```.

3. (I don't think this is viable) Have another type `EscapedIdent`, which would contain `event.type`, but knowing that it shouldn't be resolved. But how would ``` `table with space`.column ``` be parsed?

If `Ident` were always a table name, I would suggest the Struct option, even though it would be quite the mindless refactor. But because it's also used for functions, I'm tempted to go with carrying around the backticks, even though it's some violation of the abstraction (now the name resolver is parsing backticks?!).

Any thoughts?

[^1]: though we could rename the `column` field to `name` or `term` or something

aljazerzen:
As mush as this is a pain in the ass, but it shows that we are not parsing idents correctly.

[...]

PR #1074, opened 2022-10-24
refactor: Prepare for including a NS with Ident

max-sixty:
This adds an additional Struct while we move over; so we don't need to do one huge PR with the merge conflicts that entails

ref #1031

I'm adding this to the semantic branch to avoid future merge conflicts, even though I'm not a great fan of abandoning main (I also wanted to do _something_ this weekend and wasn't making enough progress with the "onboarding" issues, so chose this)

While they both exist, the only tax is some `.into()` calls


prql-compiler/src/ast/expr.rs
```diff
@@ -60,6 +60,46 @@ pub enum ExprKind {
 /// A name. Generally columns, tables, functions, variables.
 pub type Ident = String;
 
+// We are moving `Ident` from being a string to containing the structure of a
+// namespace. Eventually we can remove this and just use `Ident` everywhere.
+// https://github.com/prql/prql/issues/1031
+#[derive(Debug, PartialEq, Eq, Clone, Serialize, Deserialize)]
+pub struct IdentWithNamespace {
+    pub namespace: Option<String>,
+    pub ident: Ident,
```
aljazerzen, line 69: I think we need a better name for this. Ident is the whole node, an identifier, while this field is the specific name in a namespace. I used "variable name" or "var_name", but I'm open to suggestions.

prql-compiler/src/ast/expr.rs
```diff
@@ -60,6 +60,46 @@ pub enum ExprKind {
 /// A name. Generally columns, tables, functions, variables.
 pub type Ident = String;
```
aljazerzen, line 61: I believe that this can be fully replaced by IdentWithNamespace. Because all idents can have namespace, right?

aljazerzen (approved):
Nice integration with current ident handling and well prepared for improved ident parsing.

prql-compiler/src/ast/expr.rs
max-sixty: Yes, my plan was to move it incrementally; then probably rename `IdentWithNamespace` back to `Ident`. I wanted to do the closest thing first, so we didn't get conflicts etc. (But maybe it's not that hard and we could do it in one fell swoop...)

prql-compiler/src/ast/expr.rs
max-sixty: Yes agree. `var_name` is fine; how about just `var`?

max-sixty:
Merging to ensure we don't get more conflicts — will continue iterating on it!

prql-compiler/src/ast/expr.rs
aljazerzen: `var` it is

prql-compiler/src/ast/expr.rs
aljazerzen: That's my philosophy on semantic branch - just do it in one fell swoop :D

max-sixty: Ha! We are so aligned on so many dimensions; "average PR size" is the outlier!

PR #1075, opened 2022-10-24
test: Ignore rather than panic

max-sixty:
Not to get into a revert war :), but when these are `should_panic`, it's not possible to update other snapshots without going through the ones we know need to fail, which is quite laborious. 

Is it OK to have these as ignore?


aljazerzen:
Sure :D

We've spent quite an amount of words over this and the funny thing is that I don't have much of a preference...

max-sixty:
Thanks :)

PR #1078, opened 2022-10-24
build: Merge Cargo files from main so GHA uses cache

max-sixty:
Currently it busts the cache every time, which means automated tests take ages


prql-compiler/Cargo.toml
```diff
@@ -65,8 +67,8 @@ rusqlite = {version = "0.28.0", features = ["bundled", "csvtab"]}
 
 # Re-enable on windows when duckdb supports it
 # https://github.com/wangfenjin/duckdb-rs/issues/62
-# [target.'cfg(not(any(target_family = "windows", target_family="wasm")))'.dev-dependencies]
-# duckdb = {version = "0.5.0", features = ["bundled", "chrono"]}
+[target.'cfg(not(any(target_family = "windows", target_family="wasm")))'.dev-dependencies]
```
max-sixty: Ah, I forgot about this. We can ignore this PR if this still causes an issue for you @aljazerzen

aljazerzen: Whoops, this shouldn't be on semantic branch, I reverted the change with ed036e8.

snth:
> Currently it busts the cache every time, which means automated tests take ages

This is also a problem I have with the `prql-query` releases/builds (>30 min sometimes).

What are the key changes here that allow caching of rust / cargo crates on GHA?

aljazerzen:
I don't know much about how GH cache work.

But this commit pulls changes from main that break cargo build, because code needs to be adapted to the new sqlparser.

If you want to have this @max-sixty, we can maybe overwrite `semantic` branch with `semantic-rebased`, which I'm keeping up-to-date with commits from `semantic` while also rebasing on top of `main`.

max-sixty:
> If you want to have this @max-sixty, we can maybe overwrite `semantic` branch with `semantic-rebased`, which I'm keeping up-to-date with commits from `semantic` while also rebasing on top of `main`.

Yes let's do this, and then cherry-pick any changes on `main` on this from here on.

@MarinPostma are you OK to switch over to a rebased branch for any work you're doing? This will be the only time we do this, sorry for the noise.

max-sixty:
> What are the key changes here that allow caching of rust / cargo crates on GHA?

Mostly just having a matching `Cargo.lock`; the rust-cache readme has fuller details though

max-sixty:
This is done through the rebase that @aljazerzen just pushed. 

@MarinPostma (and anyone else working on that branch) you'll need to `git reset --hard origin/semantic` or similar. This will be the only time we do this, sorry for the noise.

PR #1079, opened 2022-10-25
feat: Allow idents to contain `.` with backticks 

max-sixty:
RIP `Ident`, long live `Ident`!

I suggested `namespace` & `name` as the two fields; is that reasonable?

(We should merge this soon to prevent conflicts)

### DIFF
--- a/prql-compiler/src/ast/expr.rs
+++ b/prql-compiler/src/ast/expr.rs
@@ -66,23 +66,6 @@ pub struct Ident {
     pub namespace: Option<String>,
     pub name: String,
 }
-
-impl From<Ident> for String {
-    fn from(ident: Ident) -> Self {
-        match ident.namespace {
-            Some(namespace) => format!("{}.{}", namespace, ident.name),
-            None => ident.name,
-        }
-    }
-}
-impl From<String> for Ident {
-    fn from(ident: String) -> Self {
-        Ident {
-            namespace: None,
-            name: ident,
-        }
-    }
-}
 impl ToString for Ident {
     fn to_string(&self) -> String {
         match &self.namespace {
@@ -324,7 +307,11 @@ pub enum JoinSide {
 
 impl Expr {
     pub fn new_ident<S: ToString>(name: S, declared_at: usize) -> Expr {
-        let mut node: Expr = ExprKind::Ident(name.to_string().into()).into();
+        let mut node: Expr = ExprKind::Ident(Ident {
+            namespace: None,
+            name: name.to_string(),
+        })
+        .into();
         node.declared_at = Some(declared_at);
         node
     }
--- a/prql-compiler/src/ast/expr.rs
+++ b/prql-compiler/src/ast/expr.rs
@@ -57,9 +57,6 @@ pub enum ExprKind {
     FString(Vec<InterpolateItem>),
 }
 
-// We are moving `Ident` from being a string to containing the structure of a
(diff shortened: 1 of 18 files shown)


prql-compiler/src/ast/expr.rs
```diff
@@ -64,11 +64,11 @@ pub struct Ident {
     pub name: String,
 }
 
-impl From<String> for Ident {
-    fn from(name: String) -> Self {
+impl Ident {
+    pub fn new_name<S: ToString>(name: S) -> Self {
```
max-sixty, line 68: OK nice, thanks. Maybe I call this `from_name`?

prql-compiler/src/sql/codegen.rs
```diff
@@ -404,62 +403,36 @@ pub(super) fn translate_join(t: &Transform, context: &mut Context) -> Result<Joi
...
-        Dialect::BigQuery => {
-            if ident.split('.').count() > 2 {
-                vec![sql_ast::Ident::with_quote(
-                    context.dialect.ident_quote(),
-                    ident,
-                )]
-            } else {
-                vec![sql_ast::Ident::new(ident)]
-            }
+pub(super) fn translate_ident(
```
max-sixty, line 411: @aljazerzen thanks for adding things — did I miss something, or is this just resolving merges? Hit me with criticism!

prql-compiler/src/lib.rs
```diff
@@ -54,7 +54,6 @@ mod test {
     use insta::{assert_display_snapshot, assert_snapshot};
 
     #[test]
-    #[ignore]
```
max-sixty, line 57: Nice! We can run with `--ignored` to get all the tests (though I'm not sure it's like python which will flag `xfail` tests that pass)

prql-compiler/src/sql/anchor.rs
```diff
@@ -88,20 +88,19 @@ impl AnchorContext {
...
             let table = self.table_defs.get(tid).unwrap();
 
-            format!("{}.{}", table.name, col_name)
-        } else {
-            col_name
-        }
+            table.name.clone()
+        });
+        (table, col_name)
     }
```
aljazerzen, line 104: Here, materialize_name was concatinating table name and var name. We separated them for the whole pipeline, but merged them together as a last step before split them back up in `translate_ident`.

prql-compiler/src/sql/codegen.rs
aljazerzen: translate_ident was operating with `String`, which has concatenated namespace and varname just before. So I allowed namespace and var name into translate_ident separately.

prql-compiler/src/sql/anchor.rs
max-sixty: Much nicer!

(also fine to point me towards things like this, also fine to DIY like this tho)

aljazerzen: Yeah, sorry I took over your PR :D I was excited to see it mostly done, because I knew it composes well with other parts of translator.

max-sixty: No prob at all — thanks for finishing it!

prql-compiler/src/ast/expr.rs
aljazerzen: Now that I look at `new_name` I don't know what I was thinking when choosing it! `from_name` is much better.

PR #1094, opened 2022-10-29
feat: Omit table name when only one ident in SELECT

### LINKED ISSUE #1081: Omit table names in idents when there is only one table in SELECT
> this issue is part of onboarding, see more info [here](https://github.com/prql/prql/pull/593#issuecomment-1282120001)

----

When we compile to very basic `SELECT` statements, table name prefixes in indents are not needed:
```sql
SELECT title FROM movies;
```
As soon as there are multiple tables, prefixes are needed:

```sql
SELECT movies.title FROM movies JOIN genres USING(genre_id);
```

This is because table `genres` could have column with name `title` which would cause conflict.

On branch `semantic`, we always include prefix, just to be sure, but it would be better to omit when possible. To make simple queries simpler.

(this is not critical to get done, but it's easy. @max-sixty I think this is too easy for you, leave it to @MarinPostma or @priithaamer or someone else)

---- 

Best way to achieve this is to add a flag `omit_ident_prefix: bool` on [`translator::Context`](https://github.com/prql/prql/blob/22c79b42155621df84ef5fd0429f594ea103abf8/prql-compiler/src/sql/translator.rs#L23-L26). It should start as `false`.

Then, before starting to build a SELECT query, we must count all tables in `From` and `Join` transforms. This can be done by a new struct that implements an `IrFold`, similar to `CidRedirector` (https://github.com/prql/prql/blob/22c79b42155621df84ef5fd0429f594ea103abf8/prql-compiler/src/sql/anchor.rs#L263-L272.

Something like:
```rust
struct TableCounter {
    count: u16,
}
impl IrFold for TabelCounter {
   fn fold_transform(...) { ... }
}
```

Folding of the SELECT (represented by `Vec<Transform>`) can be invoked here:

https://github.com/prql/prql/blob/22c79b42155621df84ef5fd0429f594ea103abf8/prql-compiler/src/sql/translator.rs#L120-L126

and set the flag:
```rust
context.omit_ident_prefix = tableCounter.count == 1;
```

To actually omit prefixes, `translate_ident` with have to be updated: https://github.com/prql/prql/blob/22c79b42155621df84ef5fd0429f594ea103abf8/prql-compiler/src/sql/codegen.rs#L406-L431

When this is all done, many of currently failing tests should start working. You can check that by running:
```
cargo test --lib -- --ignored
```
You can now remove `#[ignored]` from tests that pass.

MarinPostma:
fixes #1081

### DIFF
--- a/prql-compiler/src/ir/mod.rs
+++ b/prql-compiler/src/ir/mod.rs
@@ -1,10 +1,12 @@
 mod expr;
 mod id_gen;
 mod ir_fold;
+mod table_counter;
 
 pub use expr::{Expr, ExprKind, UnOp};
 pub use id_gen::IdGenerator;
 pub use ir_fold::*;
+pub use table_counter::TableCounter;
 
 use serde::{Deserialize, Serialize};
 
--- a/prql-compiler/src/ir/table_counter.rs
+++ b/prql-compiler/src/ir/table_counter.rs
@@ -1,0 +1,27 @@
+use std::collections::HashSet;
+
+use super::{IrFold, TId, Transform};
+
+/// Folder that counts the number of table referenced in a PRQL query.
+#[derive(Debug, Default)]
+pub struct TableCounter {
+    tables: HashSet<TId>,
+}
+
+impl TableCounter {
+    pub fn count(&self) -> usize {
+        self.tables.len()
+    }
+}
+
+impl IrFold for TableCounter {
+    fn fold_transforms(&mut self, transforms: Vec<Transform>) -> anyhow::Result<Vec<Transform>> {
+        for transform in &transforms {
+            if let Transform::Join { with: tid, .. } | Transform::From(tid) = transform {
+                self.tables.insert(*tid);
+            }
+        }
+
+        Ok(transforms)
+    }
+}
--- a/prql-compiler/src/lib.rs
+++ b/prql-compiler/src/lib.rs
@@ -225,7 +225,6 @@ mod test {
     }
 
     #[test]
-    #[ignore]
     fn test_quoting() {
         // GH-#822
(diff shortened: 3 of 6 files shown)


max-sixty:
This is good; I think providing more concise SQL is important and this contributes for small queries.

I was trying to think if it would handle queries with multiple stages, like:

```elm
from employees
take 5
group [name] (
  aggregate [sum salary]
)
```

On `main` this compiles to:

```sql
WITH table_0 AS (
  SELECT
    employees.*   # would this be `*`?
  FROM
    employees
  LIMIT
    5
)
SELECT
  name,
  SUM(salary)
FROM
  table_0
GROUP BY
  name
```

Unfortunately `semantic` can't compile that yet, so I couldn't try it out. But I _think_ it will, since this is on the `sql_of_atomic_table`, and so counts the tables in the CTE rather than the whole query.

If it doesn't yet, no stress, though maybe we add a comment in the code.

I'll let @aljazerzen take a glance too.

Thanks @MarinPostma !

max-sixty (approved):

MarinPostma:
@max-sixty I think it does. This:
```
from employee
take 1..10
sort name
take 1..5
```

compiles to:
```
WITH table_0 AS (
  SELECT
    * AS _expr_0
  FROM
    employee
  LIMIT
    10
)
SELECT
  _expr_0
FROM
  table_0
ORDER BY
  name
LIMIT
  5
```

#1096  fixes the aliasing issue

max-sixty:
Excellent!

Feel free to merge if you're confident overall (easy to revert if @aljazerzen spots something)

aljazerzen:
This is great! I dont even have any suggestions!

aljazerzen (approved):

PR #1096, opened 2022-10-29
feat: prevent wildcard aliasing

MarinPostma:
This pr prevents wildcard aliasing. This is done in the `ensure_column_name` method, by checking if the considered column is a wildcard, in which case, we set its name to `"*"`.

This is pretty hacky, by also works with ctes.

Maybe a better approach would be to change `ColumnDef` to something like:

```rust
enum ColumnDef {
    Wildcard,
    Def {
        id: CId,
        name: Option<String>,
        expr: Expr,
    }
}
```

By my knowledge of the project is not yet good enough to judge all the implications.

I haven't re-enabled any tests yet, it makes more sense to do that once #1094 is merged.

### DIFF
--- a/prql-compiler/src/sql/anchor.rs
+++ b/prql-compiler/src/sql/anchor.rs
@@ -66,10 +66,17 @@ impl AnchorContext {
         let def = self.columns_defs.get_mut(cid).unwrap();
 
         if def.name.is_none() {
-            let id = self.next_col_name_id;
-            self.next_col_name_id += 1;
+            match &def.expr.kind {
+                ExprKind::ExternRef { variable, .. } if variable == "*" => {
+                    def.name = Some("*".into());
+                }
+                _ => {
+                    let id = self.next_col_name_id;
+                    self.next_col_name_id += 1;
 
-            def.name = Some(format!("_expr_{id}"));
+                    def.name = Some(format!("_expr_{id}"));
+                }
+            }
         }
 
         def.name.clone().unwrap()


prql-compiler/src/semantic/resolver.rs
```diff
@@ -226,7 +226,6 @@ impl Resolver {
                 match res {
                     Ok(id) => id,
                     Err(e) => {
-                        dbg!(&self.context);
```
MarinPostma, line 229: this _should_ not be there, I can remove that from this pr if you want

max-sixty:
(I've seen this but I think @aljazerzen by far the better reviewer — IIUC he is traveling but is back quite soon)

max-sixty:
> Maybe a better approach would be to change `ColumnDef` to something like:

I agree, this is also similar to the proposal for now name resolution would work; now outlined in the `internals` in the book. But yes let's see what @aljazerzen thinks

prql-compiler/src/sql/anchor.rs
```diff
@@ -66,10 +66,17 @@ impl AnchorContext {
         let def = self.columns_defs.get_mut(cid).unwrap();
 
         if def.name.is_none() {
-            let id = self.next_col_name_id;
-            self.next_col_name_id += 1;
+            match &def.expr.kind {
+                ExprKind::ExternRef { variable, .. } if variable == "*" => {
+                    def.name = Some("*".into());
```
aljazerzen, line 71: This is clever but a bit hacky. Passing a `"*"` does indeed prevent alias to be emitted, because it uses rule that SQL assumes alias from ident names only, so we omit explicit alias.

If we don't want to rely on this rule, I have to do a little refactor.

aljazerzen (commented):
I'll do a refactor and add more info on how to do this.

aljazerzen:
Hey, my little refactor got out of hand :D I ended up doing exactly what you suggested @MarinPostma . I wanted to give more directions on how to do it, but it ended up being very connected to many different parts of the compiler, which is not ideal for onboarding.

This is the commit that partially does this:
https://github.com/prql/prql/commit/4aae055e8453971d2e503d27f22721996ba8360c

And this is the other part:
https://github.com/prql/prql/commit/fd5b5c400c4edbf2b0f8cc7e90a3f471ff9b32be

max-sixty:
@MarinPostma thanks for the effort — we're v happy to have you as a regular contributor, and there's much more to do!

PR #1098, opened 2022-10-31
docs: Add a 'Why PRQL' section to the website

max-sixty:
Probably needs some wordsmithing; very open to any suggestions on content too.

The formatting is OK, could be improved -- any ideas for the layout?

### DIFF
--- a/website/content/_index.md
+++ b/website/content/_index.md
@@ -32,6 +32,34 @@ hero_section:
     sort [sum_gross_cost, -country]
     take 1..20
 
+####################### Why PRQL section #########################
+why_prql_section:
+  enable: true
+  title: "Why PRQL?"
+  items:
+    - title: For data engineers
+      content:
+        - PRQL is database agnostic, compiling to many dialects of SQL
+        - PRQL is concise; with abstractions like variables & functions
+        - PRQL doesn't cap you — it can embed SQL when necessary
+    - title: For analysts
+      content:
+        - PRQL is ergonomic for data exploration — for example, commenting out
+          transforms or columns maintains a valid query
+        - PRQL is simple to learn, with a small number of powerful concepts
+    - title: For tools
+      content:
+        - PRQL is database agnostic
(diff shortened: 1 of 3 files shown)


max-sixty:
Rendered:
image

snth:
I love the "For HN enthusiasts" section 😂

---
This is somewhat off-topic and not a priority right now but given the amount of blowback Gitea is getting at the moment, in order to satisfy the HN FOSS stalwarts, longer term you'll probably have to found a foundation and grant the PRQL IP to that in order to make the "PRQL will never be a commercial product" have enough gravitas.

max-sixty:
> longer term you'll probably have to found a foundation and grant the PRQL IP to that in order to make the "PRQL will never be a commercial product" have enough gravitas.

I'm up for doing this! Probably needs some admin but when the time comes...

max-sixty:
> I love the "For HN enthusiasts" section 😂

I'm up for more comedy in docs — preferably better comedy than mine!

I was thinking of a "For CIOs" with random buzzwords like "PRQL drives value throughout the organization" but couldn't make it funny enough. Open to contributions!

aljazerzen:
This reads nicely. Well written!

Another thing I noticed now, a few months after we published the website, is that typography doesn't look all that well. I cannot pinpoint what's wrong, but I feel it's something with combination of title & body fonts. But that is a challenge for someone how knows this stuff better than me. Will we find someone if we open an issue?

max-sixty:
Interesting! A part of me agrees, and possibly I've become accustomed to it already.

We can open an issue and see if there are any takers. My guess is that this is the sort of thing we need to hire a contractor for, or figure out ourselves (happy to hire someone for design stuff, though just doing the fonts would be a v small project, so maybe we bundle with other things?)

Maybe I try using one font everywhere and see whether that helps?

aljazerzen:
Yeah go ahead if you have time, otherwise I don't think this is a priority.

Let's open an issue if someone feels a designers urge :D

PR #1115, opened 2022-11-06
refactor: aggregate

### LINKED ISSUE #1045: Add windowing params to ir::Transform and ir::ColumnDef
Edit: I've updated to be in sync with latest changes and split this task into two parts.

`ir::Transform` should be more convenient for compiling into SQL.

## Part 1

Aggregate can be just a "command" to aggregate, without specifying columns to select. Because of that a Transform::Select must be [injected before Transform::Aggregate](https://github.com/prql/prql/blob/e4c767964972c65a3d114b520502171b673e203f/prql-compiler/src/semantic/lowering.rs#L195-L200) to actually select the columns.
This will have the effect of simplifying [`determine_select_columns`](https://github.com/prql/prql/blob/e4c767964972c65a3d114b520502171b673e203f/prql-compiler/src/sql/anchor.rs#L168-L188) function.
```rust
pub enum Transform {
    ...
    Aggregate { by: Vec<CId> },
    ...
}
```

## Part 2

I have to think about it.

MarinPostma:
refactor `Transform::Aggregate` as per #1045

### DIFF
--- a/prql-compiler/src/semantic/lowering.rs
+++ b/prql-compiler/src/semantic/lowering.rs
@@ -195,7 +195,7 @@ impl Lowerer {
             ast::TransformKind::Aggregate { assigns, tbl } => {
                 let select = self.declare_as_columns(assigns, &mut transforms)?;
 
-                transforms.push(Transform::Aggregate(select));
+                transforms.push(Transform::Aggregate { by: select });
                 Some(tbl)
             }
             ast::TransformKind::Sort { by, tbl } => {
--- a/prql-compiler/src/sql/anchor.rs
+++ b/prql-compiler/src/sql/anchor.rs
@@ -175,7 +175,7 @@ impl AnchorContext {
                     columns = table_def.columns.iter().map(|c| c.id).collect();
                 }
                 Transform::Select(cols) => columns = cols.clone(),
-                Transform::Aggregate(cols) => columns = cols.clone(),
+                Transform::Aggregate { by } => columns = by.clone(),
                 Transform::Join { with, .. } => {
                     let table_def = &self.table_defs.get(with).unwrap();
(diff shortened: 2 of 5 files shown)


prql-compiler/src/semantic/lowering.rs
```diff
@@ -195,7 +195,7 @@ impl Lowerer {
             ast::TransformKind::Aggregate { assigns, tbl } => {
                 let select = self.declare_as_columns(assigns, &mut transforms)?;
 
-                transforms.push(Transform::Aggregate(select));
+                transforms.push(Transform::Aggregate { by: select });
                 Some(tbl)
```
aljazerzen, line 199: `by` should now contain `transform_call.partition` instead of assings (select columns).

Assigns should go into a `Transfrom::Select` that is pushed into transforms *before* `Aggregate`.

I should have written more about this in the issue.

prql-compiler/src/sql/anchor.rs
```diff
@@ -175,7 +175,7 @@ impl AnchorContext {
                     columns = table_def.columns.iter().map(|c| c.id).collect();
                 }
                 Transform::Select(cols) => columns = cols.clone(),
-                Transform::Aggregate(cols) => columns = cols.clone(),
+                Transform::Aggregate { by } => columns = by.clone(),
                 Transform::Join { with, .. } => {
```
aljazerzen, line 179: Aggregate should not be here anymore, because it doesn't have any effects on columns anymore.

aljazerzen (changes requested):
Hey, sorry that I didn't see your commit before pushing new changes to semantic, so there will be some merge conflicts.

Otherwise, this looks to be on the right track.

aljazerzen:
I've merged the changes via commit a2207105d7b81cc28f91abfbe22805dbe0a88361

PR #1136, opened 2022-11-16
fix: Solve precedence issue in materializer

max-sixty:
This is not great; hopefully there is a better way  -- it creates lots of parentheses where they're not needed. One challenge is that we don't store parentheses in our AST; we use the parser for this and encode as `BinOp` in the correct order. But that facility is no longer available for us during the materialization phase.

Maybe one inelegant solution is to check whether we need parentheses and only add them if so.

### DIFF
--- a/prql-compiler/src/lib.rs
+++ b/prql-compiler/src/lib.rs
@@ -129,7 +129,7 @@ mod test {
         select [result = c * sum_1 + sum_2]
         "###)?), @r###"
         SELECT
-          c * (a + b) + (a + b) AS result
+          c * (a + b) + a + b AS result
         FROM
           numbers
         "###);
@@ -194,11 +194,11 @@ mod test {
         SELECT
           numbers.*,
           c - (a + b),
-          c + (a - b),
-          c + (a - b),
-          c + (a + b),
-          (c + a) - b,
-          (c - d) - (a - b)
+          c + a - b,
+          c + a - b,
+          c + a + b,
+          c + a - b,
+          c - d - (a - b)
         FROM
           numbers
         "###
--- a/prql-compiler/src/lib.rs
+++ b/prql-compiler/src/lib.rs
@@ -182,26 +182,27 @@ mod test {
           r###"
           from numbers
           derive x = (y - z)
-          derive [
+          select [
             c - (a + b),
             c + (a - b),
             c + a - b,
             c + a + b,
             (c + a) - b,
             ((c - d) - (a - b)),
+            ((c + d) + (a - b)),
             +x,
             -x,
           ]
           "###
           )?, @r###"
         SELECT
-          numbers.*,
           c - (a + b),
           c + a - b,
           c + a - b,
           c + a + b,
           c + a - b,
           c - d - (a - b),
+          c + d + a - b,
           y - z AS x,
           -(y - z)
         FROM
--- a/prql-compiler/src/lib.rs
+++ b/prql-compiler/src/lib.rs
@@ -129,7 +129,7 @@ mod test {
         select [result = c * sum_1 + sum_2]
         "###)?), @r###"
         SELECT
-          c * (a + b) + a + b AS result
+          c * (a + b) + (a + b) AS result
         FROM
           numbers
         "###);
@@ -173,6 +173,32 @@ mod test {
         )?, @r###"
         SELECT
           a + b IS NULL
+        FROM
+          numbers
+        "###
+        );
+
+        assert_display_snapshot!(compile(
+          r###"
+          from numbers
+          derive [
+            c - (a + b),
+            c + (a - b),
+            c + a - b,
+            c + a + b,
+            (c + a) - b,
+            ((c - d) - (a - b)),
+          ]
+          "###
+          )?, @r###"
+        SELECT
+          numbers.*,
+          c - (a + b),
+          c + (a - b),
+          c + (a - b),
+          c + (a + b),
+          (c + a) - b,
+          (c - d) - (a - b)
         FROM
           numbers
         "###
--- a/prql-compiler/src/lib.rs
+++ b/prql-compiler/src/lib.rs
@@ -181,6 +181,7 @@ mod test {
         assert_display_snapshot!(compile(
           r###"
           from numbers
+          derive x = (y - z)
           derive [
             c - (a + b),
             c + (a - b),
@@ -188,6 +189,8 @@ mod test {
             c + a + b,
             (c + a) - b,
             ((c - d) - (a - b)),
+            +x,
+            -x,
           ]
           "###
           )?, @r###"
@@ -198,7 +201,9 @@ mod test {
           c + a - b,
           c + a + b,
           c + a - b,
-          c - d - (a - b)
+          c - d - (a - b),
+          y - z AS x,
+          -(y - z)
         FROM
           numbers
         "###
--- a/prql-compiler/src/snapshots/prql_compiler__test__prql_to_sql_2.snap
+++ b/prql-compiler/src/snapshots/prql_compiler__test__prql_to_sql_2.snap
@@ -9,14 +9,14 @@ SELECT
   SUM(salary),
   AVG(salary + payroll_tax),
   SUM(salary + payroll_tax),
-  AVG(salary + payroll_tax + benefits_cost),
-  SUM(salary + payroll_tax + benefits_cost) AS sum_gross_cost,
+  AVG((salary + payroll_tax) + benefits_cost),
+  SUM((salary + payroll_tax) + benefits_cost) AS sum_gross_cost,
   COUNT(*) AS ct
 FROM
   employees
 WHERE
   country = 'USA'
-  AND salary + payroll_tax + benefits_cost > 0
+  AND (salary + payroll_tax) + benefits_cost > 0
 GROUP BY
   title,
   country
--- a/prql-compiler/src/snapshots/prql_compiler__test__prql_to_sql_2.snap
+++ b/prql-compiler/src/snapshots/prql_compiler__test__prql_to_sql_2.snap
@@ -9,14 +9,14 @@ SELECT
   SUM(salary),
   AVG(salary + payroll_tax),
   SUM(salary + payroll_tax),
-  AVG((salary + payroll_tax) + benefits_cost),
-  SUM((salary + payroll_tax) + benefits_cost) AS sum_gross_cost,
+  AVG(salary + payroll_tax + benefits_cost),
+  SUM(salary + payroll_tax + benefits_cost) AS sum_gross_cost,
   COUNT(*) AS ct
 FROM
   employees
 WHERE
   country = 'USA'
-  AND (salary + payroll_tax) + benefits_cost > 0
+  AND salary + payroll_tax + benefits_cost > 0
 GROUP BY
   title,
   country
--- a/prql-compiler/src/sql/translator.rs
+++ b/prql-compiler/src/sql/translator.rs
@@ -854,16 +854,38 @@ fn translate_operand(
     })
 }
 
+pub enum Associativity {
+    Left,
+    Right,
+    Both,
+}
+
 trait SQLExpression {
     /// Returns binding strength of an SQL expression
     /// https://www.postgresql.org/docs/14/sql-syntax-lexical.html#id-1.5.3.5.13.2
     /// https://docs.microsoft.com/en-us/sql/t-sql/language-elements/operator-precedence-transact-sql?view=sql-server-ver16
     fn binding_strength(&self) -> i32;
+    fn associativity(&self) -> Associativity;
 }
+
 impl SQLExpression for Expr {
     fn binding_strength(&self) -> i32 {
         // Strength of an expression depends only on the top-level operator, because all
         // other nested expressions can only have lower strength
+        match self {
+            Expr::BinaryOp { op, .. } => op.binding_strength(),
+
+            Expr::UnaryOp { op, .. } => op.binding_strength(),
+
+            Expr::Like { .. } | Expr::ILike { .. } => 7,
+
+            Expr::IsNull(_) | Expr::IsNotNull(_) => 5,
+
+            // all other items types bind stronger (function calls, literals, ...)
+            _ => 20,
+        }
+    }
+    fn associativity(&self) -> Associativity {
         match self {
             Expr::BinaryOp { op, .. } => op.binding_strength(),
 
--- a/prql-compiler/src/sql/translator.rs
+++ b/prql-compiler/src/sql/translator.rs
@@ -516,7 +516,8 @@ fn translate_item(item: Item, dialect: &dyn DialectHandler) -> Result<Expr> {
                     left: translate_operand(
                         left.item,
                         op.binding_strength(),
-                        // No right-associativity in SQL
+                        // No right-associativity in SQL,so we never need to fix
+                        // the left associativity.
                         false,
                         dialect,
                     )?,
@@ -536,7 +537,9 @@ fn translate_item(item: Item, dialect: &dyn DialectHandler) -> Result<Expr> {
                 UnOp::Neg => UnaryOperator::Minus,
                 UnOp::Not => UnaryOperator::Not,
             };
-            let expr = translate_operand(expr.item, op.binding_strength(), true, dialect)?;
+            let fix_associativity = matches!(op.associativity(), Associativity::Left);
+            let expr =
+                translate_operand(expr.item, op.binding_strength(), fix_associativity, dialect)?;
             Expr::UnaryOp { op, expr }
         }
 
@@ -877,7 +880,7 @@ fn translate_operand(
 }
 
 /// Associativity of an expression's operator.
-/// There is now exponent symbol in SQL, so we don't seem to have a `Right` variant.
+/// Note that there's no exponent symbol in SQL, so we don't seem to require a `Right` variant.
 /// https://en.wikipedia.org/wiki/Operator_associativity
 pub enum Associativity {
     Left,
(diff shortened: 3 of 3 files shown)


max-sixty:
I didn't update snapshots as it's a huge diff, but here's a sample of what it does now:

```elm
from foo
  select [
    s = (a+b),
    diff = (c-s)
]
```

```sql
SELECT
  +(a + b) AS s,
  +(c - +(a + b)) AS diff
FROM
  foo
```

At the very least we'd have to run over the result and strip away the unneeded parentheses; but hopefully there's an easier way

CC @aljazerzen if you have ideas

aljazerzen:
While this does work, we have a system for determining when parenthesis are needed based on "binding strength". The problem is probably with this system and how it handles left-to-right and right-to-left precedence.

I'll look into it and write up what's to be done.

aljazerzen:
Ok, so here is the deal with precedence in SQL (and other languages including standard math notation):

## Detailed explanation how our translator works

(can you add this to the source code?)

If there are no parenthesis, first we determine operator precedence. This is called "binding strength" in our code in `translator.rs`. Operators with higher binding strength are applied first:

```
f == a + b * c or d
(f == (a + (b * c))) or d
```

Second rule is needed when there are multiple operators of the same strength (i.e. `a + b - c`). The rule is called associativity in programming languages, but it has some naming conflicts with math so let's call it operator application "direction". It can be left or right:

```
input: a + b + c
left dir:  (a + b) + c
right dir:  a + (b + c)
```

In our case, we have an AST which has to be converted to SQL and have parenthesis inserted where needed. We do this in `translate_operand` function in `translator.rs`.

There are basically two cases that can happen:
```
case 1:    case 2:
  +          *      
 / \        / \     
c   *      c   +    
   / \        / \   
  a   b      a   b  
```
In case 1, outer operator has lower strength than inner operator, so parenthesis are not needed.
In case 2, outer operator has higher strength than inner operator, so without parenthesis it would overrule the inner operator, so parenthesis are needed.

This is implemented by passing max_binding_strength to `translate_operand`. If operand (in this case inner node) has greater binding strength than requested, it gets nested.

Now cases when operators has the same binding strength:
```
 case 3       case 4
    -            -      
  /   \        /   \    
 a     -      -     c   
      / \    / \        
     c   d  a   b
```
(I used minus for all ops, but it's the same for any combination of operators with same binding strength).

Case 3: 
- If the outer operator has **left** direction (this is the case for minus),
  it would override the inner operator, so we need parenthesis.
- If the outer operator has **right** direction,
  it would **not** override inner operator, so we don't need parenthesis.


Case 4: 
- If the outer operator has **left** direction (this is the case for minus), 
  it would **not** override the inner operator, so we don't need parenthesis.
- If the outer operator has **right** direction
  it would override inner operator, so we need parenthesis.

General rule is that parenthesis are needed when `outer operator direction != operand position`.

Note that some operators (such as plus) could have both application directions and still have the same meaning:

```
plus has left and right direction:
(a - b) + c = a - b + c (ok)
a + (b - c) = a + b - c (ok)

minus has only left direction:
(a + b) - c = a + b - c (ok)
a - (b + c) = a - b + c (nope)
```

Formally, this is called associativity: https://en.wikipedia.org/wiki/Operator_associativity

## TODO

1. add enum `Associativity` with values `Left, Right, Both` with functions `is_left` and `is_right`.
2. add function `associativity` to `SQLExpression` trait
3. implement it for `sql::BinaryOperator` (most ops have Both)
4. add `correct_associativity` param to `translate_operand` and nest expressions if binding strength is equal and associativity is not correct.
5. when translating `UnOp`, `current_associativity` can be always true.
6. when translating `BinOp`, left operand:
   ```
   current_associativity = op.associativity().is_left()
   ```
9. when translating `BinOp`, right operand:
   ```
   current_associativity = op.associativity().is_right()
   ```

richb-hanover:
Thanks for this description of the process. I would definitely support adding this explanation to the code. A couple questions:

1. I couldn't find a definition for `operand position` as used in the general rule below:

   > General rule is that parenthesis are needed when outer `operator direction != operand position.`

2. Do you have test cases for testing out all kinds of aberrant expressions? Do you need some?

3. A naive question: Is it worth creating an intermediate step that simply places parentheses around each variable substitution? Sure - it generates lots of paren's (but I don't think we would need a leading "+" as in some of the examples above.) That would create testable (and hopefully correct) version that we could use as a benchmark for any optimizations like the general rule you propose.

Thanks again

aljazerzen:
> 1. operand position

Oh, I just meant position of the operand within a binary operation. That is: Left or Right.

> 2\. Do you have test cases

I have these, but more is always better.

```rust 
        assert_display_snapshot!(compile(
          r###"
          from numbers
          select c - (a + b)
          "###
          )?, @r###"
          SELECT
            c - (a + b)
          FROM
            numbers
          "###
        );

        assert_display_snapshot!(compile(
          r###"
          from numbers
          select c + (a - b)
          "###
          )?, @r###"
          SELECT
            c + a - b
          FROM
            numbers
          "###
        );

        assert_display_snapshot!(compile(
          r###"
          from numbers
          select (c + a) - b
          "###
          )?, @r###"
          SELECT
            c + a - b
          FROM
            numbers
          "###
        );

        assert_display_snapshot!(compile(
          r###"
          from numbers
          select ((c - d) - (a - b))
          "###
          )?, @r###"
          SELECT
            c - d - (a - b)
          FROM
            numbers
          "###
        );
```

> 3\. Is it worth creating

That's a great question. I've spent much time optimizing generated SQL, so the queries are as readable as possible. It would be easier and less error prone to just always emit parenthesis, but I do think that readability is a big plus.

People are currently copy-pasting from prql compiler to SQL engines, and learning PRQL from the generated code, so it does hold some value.

As for your suggestion to always emit and compare results: 
- always emitting is easy (best to be done in `translate_operand`),
- comparing is hard, because we need SQL evaluation engine. It could be done in integration tests that run on SQLite & DuckDB. Compile and execute only with necessary parenthesis, then compile and execute with all parenthesis and compare the results. For this we'd need some kind of compiler options that you pass along the PRQL query...

richb-hanover:
**Update:** On reflection, I want to modify my response.

1. Thanks again for the detailed answer. I agree with your goals - readable SQL helps people learn PRQL, a highly-desired outcome. But right now, I can't use PRQL for "real work" because of this precedence error. I would be OK with a "parenthesis-laden" version in the interim (unless you think this can be completed soon.)

2. I have been thinking of a parallel "end to end" test. It would push a bunch of expressions through Python (for example - it's a language that does "ordinary arithmetic") and check its output for numerical accuracy to the values created by the PRQL compiler & SQL engine. I started with the following: It's not so easy to automate, but lets us create arbitrary test cases quickly. (And could be integrated with SQLite & DuckDB, as you indicate)

```python
      a = 11.0
    b = 13.0
    c = 17.0
    d = 19.0
    e = 23.0
    f = 29.0
    g = 31.0
    h = 37.0
    sum = a + b
    sum2 = sum + sum
    diff = c - d
    diff2 = diff - diff
    prod = e * f
    prod2 = prod * prod
    quot = g / h
    quot2 = quot / quot
   
    print("a=", a)
    print("b=", b)
    print("c=", c)
    print("d=", d)
    print("e=", e)
    print("f=", f)
    print("g=", g)
    print("h=", h)
    print("sum=", sum)
    print("sum2=", sum2)
    print("diff=", diff)
    print("diff2=", diff2)
    print("prod=", prod)
    print("prod2=", prod2)
    print("quot=", quot)
    print("quot2=", quot2)

    print("exp1:", sum + diff)
    print("exp2:", sum - diff)
    print("exp3:", sum * diff)
    print("exp4:", sum / diff)
    print("exp5:", sum2 + diff2)
    print("exp6:", sum2 - diff2)
    print("exp7:", sum2 * diff2)
    # print("exp8: %s", sum2 / diff2)
 ```
produces:
```
a= 11.0
b= 13.0
c= 17.0
d= 19.0
e= 23.0
f= 29.0
g= 31.0
h= 37.0
sum= 24.0
sum2= 48.0
diff= -2.0
diff2= 0.0
prod= 667.0
prod2= 444889.0
quot= 0.8378378378378378
quot2= 1.0
exp1: 22.0
exp2: 26.0
exp3: -48.0
exp4: -12.0
exp5: 48.0
exp6: 48.0
exp7: 0.0
```

**Comment by aljazerzen** [comment]
Good idea, this could be fuzzed, to make sure that simple expressions are surly working right. I wrote a script to generate random expressions:

```python
import random

values = [11.0, 13.0, 17.0, 19.0, 23.0, 29.0, 31.0, 37.0]

def gen(max_depth, ops, terminal, rng: random.Random) -> str:
    if max_depth == 1 or rng.random() < 0.25:
        return terminal(rng)

    op = rng.choice(ops)

    l = gen(max_depth - 1, ops, terminal, rng)
    r = gen(max_depth - 1, ops, terminal, rng)
    return f'({l} {op} {r})'


def gen_number(rng: random.Random) -> str:
    return str(rng.choice(values))


def gen_arthmetic(rng: random.Random) -> str:
    return gen(4, ["+", "-", "*", "/", "%"], gen_number, rng)


def gen_boolean(rng: random.Random) -> str:
    return gen(1, ["<", ">", "==", "<=", ">=", "!="], gen_arthmetic, rng)


def gen_logic(rng: random.Random) -> str:
    return gen(3, ["and", "or"], gen_boolean, rng)


rng = random.Random(0)

for i in range(10):
    print(gen_logic(rng))
```
```
((((19.0 * 17.0) * (23.0 + 17.0)) / ((37.0 * 13.0) + (19.0 * 37.0))) or (11.0 or 31.0))
((29.0 or (19.0 + 17.0)) and (29.0 or (23.0 / ((19.0 % 23.0) % (31.0 % 29.0)))))
((((17.0 + 17.0) * (31.0 + 23.0)) + ((31.0 % 23.0) - (13.0 * 29.0))) or (((23.0 - 19.0) * ((13.0 + 17.0) - (11.0 - 13.0))) or 19.0))
(((11.0 + 17.0) or (11.0 / ((13.0 % 23.0) + 13.0))) and ((((19.0 / 23.0) + (17.0 / 19.0)) + ((17.0 - 29.0) + (37.0 + 17.0))) or 31.0))
(((11.0 * ((23.0 + 17.0) + 37.0)) or (((17.0 % 23.0) % (13.0 / 11.0)) * (19.0 * (31.0 / 11.0)))) or ((13.0 + ((11.0 / 11.0) - (37.0 * 11.0))) or (13.0 / ((31.0 / 29.0) - 11.0))))
((19.0 and ((31.0 - (23.0 + 37.0)) * ((29.0 - 13.0) + (11.0 * 11.0)))) and (((37.0 + 37.0) * (17.0 * 19.0)) % (17.0 * (29.0 * 13.0))))
(23.0 and ((((13.0 % 23.0) + (23.0 * 31.0)) / 37.0) and (13.0 * (29.0 / (17.0 - 31.0)))))
(19.0 and 31.0)
((((23.0 * (31.0 - 19.0)) - ((37.0 + 19.0) + 31.0)) or (((19.0 + 37.0) - (17.0 % 13.0)) + (31.0 / (37.0 / 29.0)))) or (((29.0 - (17.0 + 23.0)) - ((23.0 % 37.0) - 13.0)) or (17.0 + 11.0)))
(((((11.0 % 13.0) % (13.0 % 31.0)) / ((31.0 % 37.0) - (19.0 / 11.0))) and (((29.0 * 13.0) * (23.0 * 31.0)) + (17.0 / 29.0))) or 31.0)
```

richb-hanover:
Cool! That fuzzing program is clever. I like how it adds other operators I hadn't imagined. But I am not sure that it addresses my concern. 

I'm thinking about test cases that we can feed to PRQL, and then evaluate the resulting SQL to get values we can compare to known good values (say, values produced by a Python program's "ordinary arithmetic".) 

I tweaked the fuzzing program to use strings ("a", "b", "c"...) instead of prime numbers. It happily generated some bodacious expressions. I also added "derive"s for the eight numbers, and pasted them into the Playground. (see below)

I then pasted the resulting SQL (see below) into SQLite, and got the results: It's not as interesting as I hoped - note how most of the expressions are 1 or 0.
```
a	b	c	d	e	f	g	h	exp_0	exp_1	exp_2	exp_3	exp_4	exp_5	exp_6	exp_7	exp_8	exp_9
11	13	17	19	23	29	31	37	1	1	1	1	1	0	1	1	1	1
````
Here's the generated PRQL:

```
from albums
select [ 

	a = 11.0,
    b = 13.0,
    c = 17.0,
    d = 19.0,
    e = 23.0,
    f = 29.0,
    g = 31.0,
    h = 37.0,
    
exp_0 = ((((d * c) * (e + c)) / ((h * b) + (d * h))) or (a or g)),
exp_1 = ((f or (d + c)) and (f or (e / ((d % e) % (g % f))))),
exp_2 = ((((c + c) * (g + e)) + ((g % e) - (b * f))) or (((e - d) * ((b + c) - (a - b))) or d)),
exp_3 = (((a + c) or (a / ((b % e) + b))) and ((((d / e) + (c / d)) + ((c - f) + (h + c))) or g)),
exp_4 = (((a * ((e + c) + h)) or (((c % e) % (b / a)) * (d * (g / a)))) or ((b + ((a / a) - (h * a))) or (b / ((g / f) - a)))),
exp_5 = ((d and ((g - (e + h)) * ((f - b) + (a * a)))) and (((h + h) * (c * d)) % (c * (f * b)))),
exp_6 = (e and ((((b % e) + (e * g)) / h) and (b * (f / (c - g))))),
exp_7 = (d and g),
exp_8 = ((((e * (g - d)) - ((h + d) + g)) or (((d + h) - (c % b)) + (g / (h / f)))) or (((f - (c + e)) - ((e % h) - b)) or (c + a))),
exp_9 = (((((a % b) % (b % g)) / ((g % h) - (d / a))) and (((f * b) * (e * g)) + (c / f))) or g),
]
```
And the generated SQL:

```
SELECT
  11 AS a,
  13 AS b,
  17 AS c,
  19 AS d,
  23 AS e,
  29 AS f,
  31 AS g,
  37 AS h,
  19 * 17 * (23 + 17) / (37 * 13 + 19 * 37)
  OR 11
  OR 31 AS exp_0,
  (
    29
    OR 19 + 17
  )
  AND (
    29
    OR 23 / 19 % 23 % 31 % 29
  ) AS exp_1,
  (17 + 17) * (31 + 23) + 31 % 23 - 13 * 29
  OR (23 - 19) * (13 + 17 - 11 - 13)
  OR 19 AS exp_2,
  (
    11 + 17
    OR 11 / (13 % 23 + 13)
  )
  AND (
    19 / 23 + 17 / 19 + 17 - 29 + 37 + 17
    OR 31
  ) AS exp_3,
  11 * (23 + 17 + 37)
  OR 17 % 23 % 13 / 11 * 19 * 31 / 11
  OR 13 + 11 / 11 - 37 * 11
  OR 13 / (31 / 29 - 11) AS exp_4,
  19
  AND (31 - 23 + 37) * (29 - 13 + 11 * 11)
  AND (37 + 37) * 17 * 19 % 17 * 29 * 13 AS exp_5,
  23
  AND (13 % 23 + 23 * 31) / 37
  AND 13 * 29 / (17 - 31) AS exp_6,
  19
  AND 31 AS exp_7,
  23 * (31 - 19) - 37 + 19 + 31
  OR 19 + 37 - 17 % 13 + 31 / 37 / 29
  OR 29 - 17 + 23 - 23 % 37 - 13
  OR 17 + 11 AS exp_8,
  11 % 13 % 13 % 31 / (31 % 37 - 19 / 11)
  AND 29 * 13 * 23 * 31 + 17 / 29
  OR 31 AS exp_9
FROM
  albums
limit 1
```

**Comment by max-sixty** [comment]
I started doing this on my flight last night, I have a couple of very basic questions for @aljazerzen to ensure I'm not missing anything:

> * add `correct_associativity` param to `translate_operand` and nest expressions if binding strength is equal and associativity is not correct.
> * when translating `UnOp`, `current_associativity` can be always true.

To confirm, `current_associativity` == `correct_associativity`, that's not intended as a difference, right?

> add enum `Associativity` with values `Left, Right, Both` with functions `is_left` and `is_right`.

These functions are as simple as returning true for the respective value & `Both`, and false otherwise — I'm not missing anything there?

---

Re fuzzing, fine. There is https://github.com/bheisler/criterion.rs. Though to some extent if we get the design right, the fuzzing becomes less important (this was not a subtle bug!)

**Comment by max-sixty** [comment]
I've done & pushed a less-hacky-but-still-hacky version which just adds parentheses when the binding strengths are equal. I'm going to try and finish the one above — the only thing I'm tripping up on is at what level to do this — in the parent or child node.

**Comment by max-sixty** [comment]
OK this is done correctly now! I need to have another glance but then I think we can merge.

Hopefully the merge into `semantic` won't be too difficult...

**Comment by richb-hanover** [comment]
I've generated a test case to compare Python's calculations (using "ordinary arithmetic") to the output of the SQL engine.

```
from albums
select [
 a = 11.0,
    b = 13.0,
    c = 17.0,
    d = 19.0,
    e = 23.0,
    f = 29.0,
    g = 31.0,
    h = 37.0,
    sum = a + b,
    sum2 = sum + sum,
    diff = c - d,
    diff2 = diff - diff,
    prod = e * f,
    prod2 = prod * prod,
    quot = g / h,
    quot2 = quot / quot,
    
    exp1 = sum + diff,
    exp2 = sum - diff,
    exp3 = sum * diff,
    exp4 = sum / diff,
    exp5 = sum2 + diff2,
    exp6 = sum2 - diff2,
    exp7 = sum2 * diff2,
]
take 1
```
Python says the results should be:

```
a= 11.0
b= 13.0
c= 17.0
d= 19.0
e= 23.0
f= 29.0
g= 31.0
h= 37.0
sum= 24.0
sum2= 48.0
diff= -2.0
diff2= 0.0
prod= 667.0
prod2= 444889.0
quot= 0.8378378378378378
quot2= 1.0
exp1: 22.0
exp2: 26.0
exp3: -48.0
exp4: -12.0
exp5: 48.0
exp6: 48.0
exp7: 0.0
```
Once your code stabilizes, we can check the output of SQL to see if it matches. Thanks.

max-sixty:
OK merging! Happy to address any feedback after

max-sixty:
OK merging! Happy to address any feedback after as ever

richb-hanover:
Good news! The Python & PRQL output produce the same results. 

And the SQL output is quite readable. Thanks!

PR #1141, opened 2022-11-17
fix: Retain floats through translation

### LINKED ISSUE #1139: Need to preserve "zero" decimal places in variables
It's a common SQL technique to multiply a value by 1.0 (or by 100.0 if computing percentages) to force a cast to REAL. But it looks as if the 0.2.9 compiler drops zeroes after the decimal point:
```
from foo
derive [
x = 1.0,
y = 1.00000000001,
z = 1.1,
q = 100.0,
]
```
gives

```
SELECT
  foo.*,
  1 AS x,
  1.00000000001 AS y,
  1.1 AS z,
  100 AS q
FROM
  foo
```

max-sixty:
Closes https://github.com/prql/prql/issues/1139

### DIFF
--- a/prql-compiler/src/lib.rs
+++ b/prql-compiler/src/lib.rs
@@ -307,6 +307,23 @@ select `first name`
     }
 
     #[test]
+    fn test_numbers() -> Result<()> {
+        let query = r###"
+        derive x = 5
+        derive y = 5.0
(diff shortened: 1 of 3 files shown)


aljazerzen (approved):
Nice

PR #1156, opened 2022-11-22
refactor: associativity

aljazerzen:


prql-compiler/src/sql/translator.rs
```diff
@@ -951,13 +955,6 @@ impl SQLExpression for UnaryOperator {
             _ => 9,
         }
     }
-
-    fn associativity(&self) -> Associativity {
-        match self {
-            UnaryOperator::Minus => Associativity::Left,
-            _ => Associativity::Both,
-        }
-    }
```
aljazerzen, line 960: Associativity does not make sense for unary operators.

prql-compiler/src/sql/translator.rs
```diff
@@ -673,9 +658,8 @@ fn try_into_is_null(
             return Ok(None);
         };
 
-        let min_strength = Expr::IsNull(Box::new(Expr::Value(Value::Null))).binding_strength();
-        // TODO: should we always be fixing the associativity here?
-        let expr = translate_operand(expr, min_strength, true, dialect)?;
+        let strength = Expr::IsNull(Box::new(Expr::Value(Value::Null))).binding_strength();
+        let expr = translate_operand(expr, strength, true, dialect)?;
```
aljazerzen: We should never be fixing assoc here, because this is unary op.

prql-compiler/src/sql/translator.rs
```diff
@@ -510,25 +510,12 @@ fn translate_item(item: Item, dialect: &dyn DialectHandler) -> Result<Expr> {
...
                     BinOp::Coalesce => unreachable!(),
                 };
-                let fix_right_associativity = matches!(op.associativity(), Associativity::Left);
-
-                Expr::BinaryOp {
-                    left: translate_operand(
-                        left.item,
-                        op.binding_strength(),
-                        // No right-associativity in SQL, so we never need to fix
-                        // the left associativity.
```
aljazerzen, line 520: I've added Right, for sake of completeness and the fact that there are some right associ operators in SQL:

https://dev.mysql.com/doc/refman/5.7/en/operator-precedence.html

max-sixty:
Awesome, thanks @aljazerzen !

PR #1175, opened 2022-11-28
feat: update anchoring algorithm

aljazerzen:


max-sixty:
Nice! The SQL is indeed better now

PR #1176, opened 2022-11-28
feat: change self equaity op from ~ to ==

aljazerzen:


max-sixty:
Thanks; releasing tomorrow!

PR #1181, opened 2022-11-29
test: An initial error message test

max-sixty:
Hopefully we can expand on these; maybe they could be an interesting way for folks to get started.

I left a comment / question in the code about the `FormattedError` struct


max-sixty:
Hmmm, we're going to have a fight between ariadne and pre-commit here: https://github.com/prql/prql/pull/1181/commits/80ae4319e1558e81025f6f72fe5f87f851e6fcda

We may need to put these in separate snapshot files rather than inline

max-sixty:
OK merging, it's not complete, but it's a start, and #1182 will make it a bit better

Issue #1167, opened 2022-11-27
Public API of prql-compiler and binding libraries


aljazerzen:
This is structure of our current constructs and all supported conversions at the moment:
```
            PRQL string
                │ ▲
                │ │
          parse │ │ format
                │ │
       ──────►  ▼ │
PL JSON        PL AST
       ◄──────   │
                 │
                 │resolve & lower
                 │
       ──────►   ▼
RQ JSON        RQ AST
       ◄──────   │
                 │
                 │translate
                 │
                 ▼
            SQL string
```

The question is what functions should we provide as public API of prql-compiler and bindings for other languages?

We had:
```rust
fn parse(string: &str) -> Result<Vec<Stmt>>;
fn semantic::resolve(statements: Vec<Stmt>) -> Result<Query>;
fn sql::translate(query: Query) -> Result<String>;

fn compile(prql: &str) -> Result<String> {
    parse(prql).and_then(semantic::resolve).and_then(sql::translate)
}

pub fn format(prql: &str) -> Result<String>;
pub fn to_json(prql: &str) -> Result<String>;
pub fn from_json(json: &str) -> Result<String>;
```

And I replaced the 3 last functions with:
```rust
fn pl_to_prql(pl: Vec<Stmt>) -> Result<String>;
fn pl_to_json(pl: Vec<Stmt>) -> Result<String>;
fn json_to_pl(json: &str) -> Result<Vec<Stmt>>;
fn json_to_rq(json: &str) -> Result<Query>;
```

Bindings current have `compile` function along with `to_json` and `from_json` (and some of them `to_sql`). I think we should unify the API to be the same of all of the bindings.

What should be our naming schema? 

Should we rename `parse`, `resolve` and `translate` to `x_to_y()`?

eitsupi:
+1 for renaming to `x_to_y`.
Considering the possibility of future compilations to other targets such as substrait, etc., I think it makes sense to clearly state what is being converted from what.

aljazerzen:
Yeah, I'll probably change the API to what I proposed. This is updated schema:
```
                   PRQL

                   │ ▲
        prql_to_pl │ │ pl_to_prql
                   │ │
    json_to_pl     ▼ │
        ────────►
PL JSON           PL AST
        ◄────────
    pl_to_json     │
                   │
                   │
          pl_to_rq │
                   │
                   │
    json_to_rq     ▼    rq_to_substrait
        ────────►        ─────────►
RQ JSON           RQ AST            Substrait
[...]

PR #1182, opened 2022-11-29
refactor: FormattedError

max-sixty:
Not the most essential code change — and open to feedback on whether it's useful on net! — but trying to make the interface simpler as I was adding a test for errors:
- Moved related functions into a Trait
- Renamed the Trait containing the error message from `FormattedError` to `ErrorMessage`, since it's not an `Error` type (maybe there are better alternatives though?)

Still not confident about the overall interface. Should `compile` return a `Result` whose `Error` type has the full message on? At that point we have `source`, albeit not the `source_id`. Or should we have one `Error` type and one "Error Output" (i.e. `FormattedError` / `ErrorMessage`)


max-sixty:
(this builds on https://github.com/prql/prql/pull/1181)

aljazerzen:
I like this trait approach. Renaming is also better now.

Looking at this I'd say that compile could format the error by default. It's a convenience wrapper function anyway -- if you don't want to run formatting you can still call underlying processing functions.

Related to #1167

max-sixty:
OK great. 

- If `compile` returns `Result<String, ErrorMessage>`, should `ErrorMessage` (nee `FormattedError`) implement the `StdError` trait? Currently it doesn't, it's just a normal Struct.
- The alternative would be a single `Error` type which has a `source` & `source_id` of `Optional`, and can produce an  error message where those are `Some`.
  - We may need to implement this on `anyhow::Error`, given we return that from almost all functions (and it sometimes contains our own `Error`, not always though)
- If this is a lot of decisions, happy to merge this and then decide more as we gain more experience — not the most critical part of our code

aljazerzen:
I don't any reasons to change current setup:
- std::error::Error trait is meant for errors in sense of exceptions that will provide info like causes and descriptions and backtrace to developers. We don't need that - we need structured data that is passed to the frontend (or the CLI)
- having source in the Error may be too costly memory-wise. It could be a reference, borrowing the input, but I don't want to deal with that.

Current setup allows you to call any amount of functions and at the end, optionally format the error. This is modular and has ability not to format errors in if you want for example check validity of the source at every keystroke.

max-sixty:
OK cool @aljazerzen , that sounds v reasonable, I'll finish this off at the weekend

max-sixty:
> Looking at this I'd say that compile could format the error by default. It's a convenience wrapper function anyway -- if you don't want to run formatting you can still call underlying processing functions.

I'm actually not sure this quite works — then it's not possible to call `.compile(...)?`, because the `?` requires `ErrorMessage` to implement `StdError`.

If you agree, I'll do the "close" thing now, and then we can iterate on it later?

PR #1186, opened 2022-11-29
fix: same column being selected multiple times

### LINKED ISSUE #1185: Same column gets selected multiple times
Using the latest / semantic version:
```elm
from account
filter account.name != null
select [name = account.name]
```
compiles to
```sql
WITH table_0 AS (
  SELECT
    name,
    name <---------------- ooops
  FROM
    account
)
SELECT
  name
FROM
  table_0
WHERE
  name IS NOT NULL
```

Notice `name` appearing twice in `table_0`'s `SELECT`. This happens whenever you explicitly select a field that you're also filtering on.

This causes the database (in my case postgres) to error on line 9 of the SQL:
```
column reference "name" is ambiguous
```

If the last line was changed like shown below, the output would be perfectly fine. But, I'd argue the first query wasn't incorrect, and there are cases where this kind of explicit selecting is required.
```diff
-select [name = account.name]
+select [name]
```
```sql
SELECT
  name
FROM
  account
WHERE
  name IS NOT NULL
```

mklopets:
Fixes #1185 following @aljazerzen's tips

As you'll see, it's not quite perfect, as the `table_0` CTE still selects both `name` and `renamed_name` in the second test, but that's just slightly suboptimal, not actively broken.

As an unintended positive side effect, this also removed one case of double-selecting in an unrelated test.

### DIFF
--- a/prql-compiler/src/lib.rs
+++ b/prql-compiler/src/lib.rs
@@ -1398,8 +1398,7 @@ take 20
         WITH table_0 AS (
           SELECT
             employees.emp_no,
-            d.name,
-            employees.emp_no
+            d.name
           FROM
             employees
             JOIN department AS d ON employees.dept_no = d.dept_no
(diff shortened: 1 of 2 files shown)


max-sixty:
> As you'll see, it's not quite perfect, as the `table_0` CTE still selects both `name` and `renamed_name` in the second test, but that's just slightly suboptimal, not actively broken.

Yes, very marginal! Fine to add a TODO for these sorts of things

max-sixty:
Thank you v much @mklopets !

max-sixty:
And thanks @aljazerzen for the tips — happy to see that gradually others (including me!) are coming up to speed with the compiler

aljazerzen (commented):
Nice

PR #1192, opened 2022-11-30
fix: Fix prql-js tests

max-sixty:


prql-js/tests/test_all.js
```diff
@@ -19,7 +19,7 @@ group [dm.emp_no, gender] (
     salary_sd = stddev emp_salary
   ]
 )
-derive mng_no = dm.emp_no
+derive mng_no = emp_no
```
max-sixty, line 22: This was the change. And I think the current way it works is very reasonable.

PR #1195, opened 2022-11-30
feat: inline table expressions

aljazerzen:


max-sixty:
Amazing! This is great to have!

snth:
This is very cool and the test case example
```elm
from employees
select [emp_id, name, surname, type, amount]
join s = (from salaries | select [emp_id, salary]) [==emp_id]
```
also makes it quite compelling. Also, SQL has this so people coming from SQL probably also expect this.

All that said, I am still against it. I fully expect to lose this argument but I just want to put it here for the record because until now it's only been captured in some discussions on Discord and maybe the odd Github issue.

My reasons for opposing this are:
* In large SQL queries, I often find quite large inline table expressions like this which break the flow of the code.
* In the PRQL case even more so I expect this to break the flow of the pipeline for the reader.
* Not allowing this and forcing it to be written as a table expression at the top will lead to more self-documenting code by forcing the author to name the pipeline.
* Philosophically, I believe that shoehorning people into only certain allowed behaviours will lead to simpler more readable code in the long run. I look at Python here as an example of this which IMHO worked. See the Python discussions around lambdas for example.
* Another example is Apple and their curtailing of user choices as opposed to IBM in he 80s and 90s and Android in the last decade.

---
As a compromise, as I do see the usefulness of it for quick modifications of some source table as in the example above, what about only allowing single line pipelines using the `|` form in inline expressions and disallowing multiline ones with `\n`? I understand that this may not be possible in the current implementation but as something to perhaps aim towards?

What I'm trying to guard against is sprawling pipelines within pipelines within pipelines ... so if we could accept small inline expressions but disallow large ones then I think I would be happy with that. I think that's kinda the same compromise that Python came to with their lambdas.

aljazerzen:
I fully agree @snth - this is a bad practice and allows you to construct real ugly and unintuitive queries that start resembling SQL. 

But it's a logical consequence of how we define our expressions and type system, and forbidding it entirely seems too harsh to me.

Instead, I'd add a warning for this. Just as rust allows you have unused variables or functions, but emits a warning. This allows you to run the code even though it's unfinished but also makes basically all published crates fix the warnings otherwise the users downstream complain about the warnings.

PR #1201, opened 2022-12-01
ci: Updated actions/workflow to minimize deprecation warnings

richb-hanover:
This PR updates the README to mention the `0.3` release in the "Current Status" section. It also begins to accumulate changes in the 0.3.1 Changelog

It also updates the Dockerfile to use rust 1.64.

Finally, it updates the `.github` actions and workflows to remove most of the deprecation warnings. 

- Most of the deprecation warnings were caused by running node12 in Github Actions, which will start requiring node16 soon.

- Most of the upstream actions have been updated, so these actions/workfiles only needed to refer to a new version number

- baptiste0298/cargo-install@next seems to work, although it may not be a stable fix. (I'm not confident that that the author is paying attention to that branch.)

- In `actions/test-rust/action.yml` the `actions-rs/cargo@v1` hasn't been updated, even after I created an issue requesting it. I cloned it to `richb-hanover/cargo@v1.1.0` to see if it would work. It did, but PRQL should not rely on my faulty memory, and should seek an alternate repo.

- Similarly, `actions-rs/tarpaulin@v0.1` has not been fixed/updated to node16. (Its `action.yml` file is still `using: 'node12'). Unless the upstream fixes the Node version, it would need to be cloned in the same manner as richb-hanover/cargo.

- Both the Java and Python workflows give warnings about using macos-12 soon. The official date for cutover is 1 Dec 2022 (tomorrow), so I just left their versions as "macos-latest" which should switch on the appointed day.


.github/actions/setup-nox/action.yaml
```diff
@@ -4,23 +4,23 @@ description: Prepare and cache nox
 runs:
   using: composite
   steps:
-    - uses: actions/setup-python@v3
+    - uses: actions/setup-python@v4
```
max-sixty, line 7: Interesting — do you know whether this is new or not?

If it's not, we should look at why it's not being upgraded automatically by dependabot; it should be doing these itself — e.g. https://github.com/prql/prql/pulls?q=is%3Apr+is%3Aclosed+label%3Agithub_actions

.github/actions/test-rust/action.yaml
```diff
@@ -15,29 +15,29 @@ runs:
         # Without inputs.target, it seems to select linux x64 in the wasm32
         # tests, though in the docs it says it uses the target as the key?
         key: ${{ inputs.target_option }}
-    - uses: jetli/wasm-bindgen-action@v0.1.0
+    - uses: jetli/wasm-bindgen-action@v0.2.0
       if: inputs.target_option == '--target=wasm32-unknown-unknown'
     - name: 📎 Clippy
-      uses: actions-rs/cargo@v1
+      uses: richb-hanover/cargo@v1.1.0
```
max-sixty, line 21: Nice! It's probably time for someone to make a fork. This could be the canonical fork! Or if you're not planning to maintain it given the comments above, we can keep it there for the moment, but could we add a comment that we should try and find the alternative in time?

.github/workflows/publish-web.yaml
```diff
@@ -26,13 +26,13 @@ jobs:
 
       # Website setup
       - name: Setup hugo
-        uses: peaceiris/actions-hugo@v2.6.0
+        uses: peaceiris/actions-hugo@v2
```
max-sixty: The nice thing about having this explicit is that we know when it changes. And dependabot should reduce the maintenance burden to almost zero.

CHANGELOG.md
```diff
@@ -1,5 +1,10 @@
 # PRQL Changelog
 
+## 0.3.1 - Unreleased
+
+- Update Github Actions and Workflows to current version numbers
+  (and avoid using Node12 anymore)
```
max-sixty, line 6: Thanks! Leading by example :)

Would you be up for adding your GH handle? No pressure

README.md
```diff
@@ -61,20 +61,23 @@ For more on the language, more examples & comparisons with SQL, visit
...
-PRQL is being actively developed by a growing community. It's ready to use for
+PRQL is being actively developed by a growing community. It's ready to use by
 the intrepid, either as part of one of our supported extensions, or within your
 own tools, using one of our supported language bindings.
 
+With the announcement of the `0.3` release,
+PRQL now has a semantic understanding of expressions and
+fixes a number of lingering bugs.
+It also unlocks development of longer-term priorities.
+The [CHANGELOG.md](./CHANGELOG.md) gives more information.
```
max-sixty, line 74: Very nice! Thanks for this

max-sixty (approved):
Thank you very much @richb-hanover !

.github/actions/setup-nox/action.yaml
richb-hanover: I don't really how Github actions / dependabot work, so I can't offer an opinion about why dependabot didn't notice...

.github/actions/test-rust/action.yaml
richb-hanover: Yes, a fork for this and tarpaulin, both in github.com/actions-rs, might be in order. Is this something to consider under the PRQL umbrella? Maybe `prql/actions-rs-cargo `and `prql/actions-rs-tarpaulin`? (And maybe the `baptiste0289` repo?)

I say this because you're way more tuned in to the rust ecosystem and tools. Plus ~~you're firmly committed to keeping this software running.~~ PRQL is a better home for "rust development stuff" than the grab-bag that's over at richb-hanover...

.github/workflows/publish-web.yaml
richb-hanover: Ah yes. Dependabot. I think I changed to `v2` because I got bitten by another package that I did overspecify. I changed it back to v2.6.0 (which it used in 0.2.11)

CHANGELOG.md
richb-hanover: Thanks, but I'm OK standing on the sidelines.

.github/actions/setup-nox/action.yaml
max-sixty: Hmmm, it's been out for a while, let me have a look at why it's not updating it. Thanks for finding this!

.github/workflows/publish-web.yaml
max-sixty: Thanks!

.github/actions/test-rust/action.yaml
max-sixty: Yeah, I think keeping the prql org focused on the PRQL lang keeps the scope limited, and having a proper home — either some org like `actions-rs` was intending to be, or someone dedicated to maintaining it, will give it the focus it needs.

So if you have the patience, add a comment, but no great stress; I'll try and remember

.github/actions/setup-nox/action.yaml
max-sixty: It seems that the `action.yaml` files require an explicit directory in the `.dependabot.yml` config: https://github.com/dependabot/dependabot-core/issues/4178#issuecomment-1118492006; e.g. https://github.com/ergebnis/.github/blob/main/.github/dependabot.yml

If you want to add that, feel free, otherwise I'm happy to do it

.github/workflows/docker-build-test.yml
```diff
@@ -0,0 +1,19 @@
...
+jobs:
+  build:
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v3
+      - name: Build Docker image
+        run: docker build -t prql .
+      # - name: Run tests inside the container
+      #   run: docker run samplepy poetry run pytest
```
max-sixty: Nice, this is good to have.

Let's add it as a job to `test-all.yaml`? That's where these wider tests go; we don't need a separate workflow for each

.github/actions/setup-nox/action.yaml
richb-hanover: You should do it. I'm at the very edge of my understanding of how all this stuff interlocks... Thanks.

.github/actions/test-rust/action.yaml
richb-hanover: I just wrote a note to the author of that repo, asking if there's someone maintaining them... I'll let you know what I find.

.github/workflows/docker-build-test.yml
richb-hanover: Again, I am happy to have you do this. Thanks.

.github/actions/test-rust/action.yaml
richb-hanover: PS I'm happy to leave cargo@v1.1.0 there for the time being.

.github/workflows/docker-build-test.yml
max-sixty: Ok, though then let's remove it from this PR?

richb-hanover: Moved into the test-all.yaml file

.github/workflows/test-all.yaml
```diff
@@ -87,3 +87,15 @@ jobs:
...
+  # from https://stackoverflow.com/questions/67134410/how-to-test-my-dockerfile-for-my-python-project-using-github-actions
+  build-docker:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - name: Build Docker image
+        run: docker build -t prql .
+      # In the future, do this if needed
+      # - name: Run tests inside the container
+      #   run: docker run samplepy poetry run pytest
```
```suggestion
      # TODO: run tests as well as building
```

README.md
```diff
@@ -61,20 +61,23 @@ For more on the language, more examples & comparisons with SQL, visit
 To experiment with PRQL in the browser, check out
 [PRQL Playground][prql playground].
 
-## Current status — October 2022
+## Current status — November 2022
```
max-sixty:
```suggestion
## Current status — December 2022
```

max-sixty:
Super! Thank you v much @richb-hanover , really appreciate the addition

PR #1203, opened 2022-12-01
fix: report direct table references

### LINKED ISSUE #1199: Resolving table name inside S-string
```elm
from x

filter field > 0 # this causes an intermediary table_1 to be created

join y [something]

select [
  # this compiles to `x.field` while it should be `table_1.field`
  field = s"{x}.field"
]
```
The resulting query will fail on the DB level:
```sql
WITH table_1 AS (
  SELECT
  FROM
    x
  WHERE
    field > 0
)
SELECT
  x.field AS field <---------- there's no `x` available here – instead, this should reference table_1.field
FROM
  table_1
  JOIN y ON something
```

Curly brackets inside of S-strings nicely resolve fields (`s"{x.field}"`) but resolving only the table part doesn't (`s"{x}.field"`).

aljazerzen:
Closes #1199

### DIFF
--- a/prql-compiler/src/semantic/resolver.rs
+++ b/prql-compiler/src/semantic/resolver.rs
@@ -172,6 +172,14 @@ impl AstFold for Resolver {
 
                     DeclKind::Expr(expr) => expr.as_ref().clone(),
 
+                    DeclKind::InstanceOf(_) => {
+                        bail!(Error::new(Reason::Simple(
+                            "table instance cannot be referenced directly".to_string(),
+                        ))
+                        .with_span(span)
+                        .with_help("did you forget to specify the column name?"));
+                    }
+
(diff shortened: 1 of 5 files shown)


prql-compiler/src/semantic/resolver.rs
```diff
@@ -172,6 +172,14 @@ impl AstFold for Resolver {
 
                     DeclKind::Expr(expr) => expr.as_ref().clone(),
 
+                    DeclKind::InstanceOf(_) => {
+                        bail!(Error::new(Reason::Simple(
+                            "table instance cannot be referenced directly".to_string(),
+                        ))
+                        .with_span(span)
+                        .with_help("did you forget to specify the column name?"));
```
max-sixty, line 180: 👍 🙏

PR #1204, opened 2022-12-01
fix: anchoring sorts

### LINKED ISSUE #1187: Renaming via select breaks preceding sorting
When a field that's used for sorting gets renamed using select, things break:
```elm
from x
derive somefield = "something"
sort [somefield]
select [renamed = somefield]
```
```sql
SELECT
  'something' AS renamed
FROM
  x
ORDER BY
  somefield
```

This will error in the database – the `ORDER BY` should now be using `'something'` instead (or however that derived field is defined – it's probably also referencing other columns):
```
Query 1 ERROR: ERROR:  column "somefield" does not exist
LINE 6:   somefield
          ^
```

### LINKED ISSUE #1198: Missing select with join
```elm
from account

filter name == "PRQL" # this creates the intermediary table_1 which triggers the problem

join user_in_account [==account_id]

select name
```
Compiles to
```sql
WITH table_1 AS (
  SELECT <------------ this should also select `name`
    account_id
  FROM
    account
  WHERE
    name = 'PRQL'
)
SELECT
  name <-------------  this does not exist in table_1
FROM
  table_1
  JOIN user_in_account ON table_1.account_id = user_in_account.account_id
```
```
Query 1 ERROR: ERROR:  column "name" does not exist
LINE 10:   name
           ^
```

If the original PRQL is modified:
```diff
-select name
+select account.name
```
then the output SQL is fine.

We could either:
1. require the PRQL `select` at the end to explicitly specify the table
2. automatically select `account.name` as we already know there's a `name` there for a fact – this is clearly somewhat less safe, but also less verbose
3. do a combo of these – compile like option 2, but output a warning (not error) about the table being inferred

aljazerzen:
Closes #1187

### DIFF
--- a/prql-compiler/src/ast/rq/fold.rs
+++ b/prql-compiler/src/ast/rq/fold.rs
@@ -61,9 +61,9 @@ fn fold_column_decl<F: ?Sized + IrFold>(
     Ok(ColumnDecl {
         id: cd.id,
         kind: match cd.kind {
-            ColumnDefKind::Wildcard => ColumnDefKind::Wildcard,
-            ColumnDefKind::ExternRef(name) => ColumnDefKind::ExternRef(name),
-            ColumnDefKind::Expr { name, expr } => ColumnDefKind::Expr {
+            ColumnDeclKind::Wildcard => ColumnDeclKind::Wildcard,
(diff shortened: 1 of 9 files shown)


max-sixty:
Very nice! I get most of the code, I need to do some more issues to get more up-to-speed.

Also lol
image

PR #1205, opened 2022-12-01
feat: prohibit ambigious wildcards

### LINKED ISSUE #1193: Ambigious wildcard references
I'm opening this issue as a followup on a recent call and @snth's [comment here](https://github.com/prql/prql/issues/1168#issuecomment-1329162324).

First I suggest to read current [name resolution](https://prql-lang.org/book/internals/name-resolving.html) algorithm.

This issue is about this paragraph:

> In the case that there are multiple namespaces with a wildcard, we don’t match with neither of the namespaces, but match as *.* name. This is a special feature that allows a column name which may reside in two different tables not to be associated with any of them. Instead, it is translated into the column name only, so database can determine which table it belongs to. Note that this may lead to PRQL passing on ambiguous queries to the database, instead of resulting error early.

In practice this means that such queries:

```
from employees
join salaries []
select first_name
```
... are translated to:
```sql
SELECT 
  first_name  -- no table name, even though there are multiple tables in the query
FROM employees JOIN salaries ON ...;
```

-----

Note that this happens only when both tables names still contain wildcards.

In this case, `first_name` is know to be in employees and possibility that it may be in salaries is not considered:
```elm
from employees
select first_name
join salaries []
select first_name
```

In this case, `first_name` is known to be in both, so ambiguity error is raised:

```elm
table sal = (from salaries | select first_name)

from employees
select first_name
join sal []
select first_name # error here
```

-----

As @snth put it, there are two options:

> 1. Name references resolving to multiple wildcards should throw an error straight away at compile time, forcing the user to specify which parent table the column references.
> 2. Ambiguity is ok and we simply produce the SQL and we let the underlying database/RDBMs resolve the ambiguity at runtime.
>
> It was mentioned that the burden of the strict name resolution option could be lessened through tooling, which at authoring time could suggest the ambiguous parent object names as possible autocompletions.

With semantic branch, this option 2 became a little worse, because ambiguous idents are converted to s-strings earlier which means that anchoring does not know about them and may cause problems.

aljazerzen:
Closes #1193 and #1197 

I won't merge this yet, waiting for resolution of the discussion.

### DIFF
--- a/prql-compiler/src/semantic/context.rs
+++ b/prql-compiler/src/semantic/context.rs
@@ -4,7 +4,7 @@ use itertools::Itertools;
 use serde::{Deserialize, Serialize};
 use std::{collections::HashMap, fmt::Debug};
 
-use super::module::{Module, NS_DEFAULT_DB, NS_NO_RESOLVE, NS_SELF, NS_STD};
+use super::module::{Module, NS_DEFAULT_DB, NS_SELF, NS_STD};
 use crate::ast::pl::*;
 use crate::error::Span;
 
@@ -163,20 +163,10 @@ impl Context {
                 Ok(module_ident + Ident::from_name(ident.name.clone()))
             }
 
-            // don't report ambiguous variable, database may be able to resolve them
+            // ambiguous
             _ => {
-                // insert default
-                let ident = NS_NO_RESOLVE.to_string();
-                self.root_mod
-                    .names
-                    .insert(ident, Decl::from(DeclKind::NoResolve));
-
-                log::debug!(
-                    "... could either of {:?}",
-                    decls.iter().map(|x| x.to_string()).collect_vec()
-                );
-
-                Ok(Ident::from_name(NS_NO_RESOLVE))
+                let decls = decls.into_iter().map(|d| d.to_string()).join(", ");
+                return Err(format!("Ambiguous reference. Could be from any of {decls}"));
             }
         }
     }
--- a/prql-compiler/src/semantic/context.rs
+++ b/prql-compiler/src/semantic/context.rs
@@ -122,7 +122,7 @@ impl Context {
                 // ambiguous
                 _ => {
                     let decls = decls.into_iter().map(|d| d.to_string()).join(", ");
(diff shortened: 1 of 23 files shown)


prql-compiler/src/semantic/context.rs
```diff
@@ -163,20 +163,10 @@ impl Context {
...
-                    .insert(ident, Decl::from(DeclKind::NoResolve));
-
-                log::debug!(
-                    "... could either of {:?}",
-                    decls.iter().map(|x| x.to_string()).collect_vec()
-                );
-
-                Ok(Ident::from_name(NS_NO_RESOLVE))
+                let decls = decls.into_iter().map(|d| d.to_string()).join(", ");
+                return Err(format!("Ambiguous reference. Could be from any of {decls}"));
```
snth: Do you have a test case for this error? I'd quite like to see an example of the message that it produces.

snth:
Looking through the changes it appears to be mostly examples and tests.

I like how the new examples look because it makes it much easier for the reader to see what source table a column comes from.

The litmus test will be to see whether it will be annoying when writing queries. However, similarly to the experience with Rust, I suspect that after an initial adjustment period one will actually come to appreciate the concreteness this brings.

We have quite a few changes with this 0.3.x line of releases and it looks like we'll plan to bring quite a few new user facing syntax features in so I think this is a good time to try this.

It's a +1 from me.

prql-compiler/src/semantic/context.rs
aljazerzen: No I don't, maybe I should I one.

```
from x
join y []
select a
Error: 
   ╭─[_a.prql:3:8]
   │
 3 │ select a
   ·        ┬  
   ·        ╰── Ambiguous name. Could be from any of _frame.y.`*`, _frame.x.`*`
───╯
```

max-sixty: FYI I started error message tests; need to fix the whitespace issue and then it should be easy to build off: https://github.com/prql/prql/pull/1181

max-sixty:
Thanks @aljazerzen !

PR #1211, opened 2022-12-02
fix: anchor filter & computes

### LINKED ISSUE #1207: Multiple columns with same name created
This issue arises whenever you define a column with the same name as something that's already been filtered on
```elm
from x

filter name != "Bob"

select name = name ?? "Default"
```
```sql
WITH table_1 AS (
  SELECT
    COALESCE(name, 'Default') AS name,
    name
  FROM
    x
)
SELECT
  name
FROM
  table_1
WHERE
  name <> 'Bob' <> 'Bob' <---------- this will fail as table_1 has two columns named `name`
```

The last line of PRQL could be anything that creates a column called `name`, even something that dismisses the original value:
`select name = "Always Bob"`

The correct query should look something like this (notice that the coalescing is done after & independently of filtering):
```sql
WITH table_1 as (
  SELECT
    name
  FROM
    x
  WHERE
    name <> 'Bob'
)
SELECT
  COALESCE(name, 'Default') AS name
FROM
  table_1
```

aljazerzen:
Closes #1207

### DIFF
--- a/book/tests/snapshots/snapshot__run_reference_prql@introduction-0.prql.snap
+++ b/book/tests/snapshots/snapshot__run_reference_prql@introduction-0.prql.snap
@@ -8,39 +8,28 @@ WITH table_1 AS (
     title,
     country,
     salary + COALESCE(tax, 0) + benefits_cost AS gross_cost,
-    salary + COALESCE(tax, 0) AS gross_salary,
-    start_date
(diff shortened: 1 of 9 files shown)


max-sixty:
Quite the pace of fixes this week! I will spend more time on the code this weekend. 

It's important that I & others get up to speed with it too, and it's not all falling on you again.

PR #1213, opened 2022-12-03
test: check that prql.to_json returns a pipeline

### LINKED ISSUE #1210: Fix JS test
I commented a JS test out at https://github.com/prql/prql/pull/1194/files#diff-30240ba768ec3bee6477f84682d25945c00c4a5bee601ff55ed7d40a3659571aR59-R60, so we could at least start running the JS tests in `test-all`. 

I _think_ the tests passed locally a short while ago (as part of https://github.com/prql/prql/pull/1192/files), but i can't be sure.

If anyone is familiar with why `js` is null on that line, please have a go — JS is probably my worst language...

mklopets:
Closes #1210

There was a change in output for `to_json` due to the intermediary format changing, so there was no longer a `.nodes` inside this JSON to access.

### DIFF
--- a/prql-js/tests/test_all.js
+++ b/prql-js/tests/test_all.js
@@ -56,8 +56,7 @@ describe("prql-js", () => {
   describe("to_json", () => {
     it("should return valid json from valid prql", () => {
       const js = JSON.parse(prql.to_json(employee_prql));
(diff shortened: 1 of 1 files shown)


max-sixty:
Thanks a lot! Great run @mklopets !

Issue #670, opened 2022-06-26
Nicer looking FAQ


max-sixty:
Does anyone have ideas for making the FAQ look better? Can we add a template for a clickable FAQ to our current theme, or would it conflict? 

Currently it's not that elegant: https://prql-lang.org/faq/ — both not clickable, and very wide...

spenserblack:
It's a bit simple, but I think a ToC to jump to and from items could help. E.g.
```markdown
## ToC

- [**Cool story Hansel, but what can I actually do with PRQL now?**](#cool-story-hansel-but-what-can-i-actually-do-with-prql-nowtoc)
- [**Something here reminds me of another project, did you take the idea from them?**](#something-here-reminds-me-of-another-project-did-you-take-the-idea-from-themtoc)

## [Cool story Hansel, but what can I actually do with PRQL now?](#toc)

...

## [Something here reminds me of another project, did you take the idea from them?](#toc)
```
Only issue is that this will put the linked sections at the *top* of the page, which currently means underneath the fixed-position header.

max-sixty:
That would be great! Can we create that in Hugo automatically?

It's possible we need to write our own template, but maybe we can leverage those already out there...

spenserblack:
I'm not familiar with Hugo, but I definitely agree that an auto-generated table of contents would be ideal. Just based on familiarity with similar tools, maybe each FAQ could be represented with a data format that points to FAQ contents, somewhat like this?
```yaml
faqs:
  - title: "Cool story..."
    contents: partials/faqs/cool-story.md
```
```markdown
<!-- partials/faqa/cool-story.md -->
We're still early...
```
Again, I'm not familiar with Hugo at the moment, so I may be *way* off the mark with this :slightly_smiling_face:

max-sixty:
Yes! That's exactly how hugo works actually — here's our front page: https://github.com/prql/prql/blob/main/website/content/_index.md?plain=1

max-sixty:
This was done by #1217, thanks @mklopets !

Issue #1107, opened 2022-11-03
Touchup website design


aljazerzen:
We do have a website that is not bad looking, but it's not great.

There are a few specific issues like "[FAQ is too wide](https://github.com/prql/prql/issues/670)" and there is also my recent observation that typography doesn't look all that well. I cannot pinpoint what's wrong, but I feel it's something with combination of title & body fonts.

I'm opening this issue as an invitation for anyone who wants to help with the design.

Some info to get started:
- the website is in [`website/`](https://github.com/prql/prql/tree/main/website) directory,
- it uses [Hugo static-site-generator](https://gohugo.io/),
- you can build & serve it locally by running `hugo server`.

PR #1217, opened 2022-12-03
docs: website style tweaks

mklopets:
Mostly affects the [FAQ page](https://prql-lang.org/faq/) (closes issue #670); also slightly decreased the body font-size site-wide.

Related: #1107

### DIFF
--- a/website/content/faq.md
+++ b/website/content/faq.md
@@ -2,7 +2,7 @@
 title: "FAQ"
 ---
 
-## Cool story Hansel, but what can I actually do with PRQL now?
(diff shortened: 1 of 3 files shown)


aljazerzen (approved):
Looking better. I like smaller font, but compared now the difference between p and to h2/h3 is too big. Maybe also smaller headers - just on subpages and not landing page?.

#670 is also mentioning making body narrower, but I don't know how that would look.

Great job on making the questions collapsible!

max-sixty:
Very nice! This is much better.

I'll merge and then we can iterate more if people want?

PR #1219, opened 2022-12-03
docs: integrate book and playground into main website

### LINKED ISSUE #632: Move playground within the website
As discussed on Discord:

It would be less of a jump, and so fewer people giving up exploring the language, if `Playground` could share a header and formatting with the rest of the site.

I don't think it can't be build purely with Hugo given it requires wasm & JS, but it could be copied into a path as part of a build step (as we do in the [build](https://github.com/prql/prql/blob/main/.github/workflows/publish.yaml#L50-L58)).

It would require adjusting the Playground's CSS to be more similar to the site's.

This was originally inspired by @alonso's [Rulex](https://rulex-rs.github.io/). Though it looks like they've recently regressed a bit and it's not exactly the same header.

mklopets:
Closes #632 

Two notes:
- new ugly URLs – this keeps the book & playground's full-screen versions at /book and /playground, and creates new routes (see GIF) that the main menu now links to. Suggestions welcome!
- the styles of the overall website, the book & the playground are fairly different – not ideal, but I think this change still moves the site in the right direction (we could open a followup issue to theme the playground similarly to the book & code samples)

### DIFF
--- a/website/config.yml
+++ b/website/config.yml
@@ -12,11 +12,11 @@ menu:
     # TODO: Can we use the info from pages directly? https://gohugo.io/templates/menu-templates/#menu-entries-from-the-pages-front-matter
     - identifier: Book
       name: Book
-      url: "https://prql-lang.org/book/"
+      url: /book_integrated/
       weight: 1
     - identifier: Playground
       name: Playground
-      url: "https://prql-lang.org/playground/"
+      url: /playground_integrated/
       weight: 2
     - identifier: Roadmap
       name: Roadmap
--- a/website/content/book_integrated.html
+++ b/website/content/book_integrated.html
@@ -1,0 +1,6 @@
+---
+title: "Book"
+layout: big_iframe
+---
+
+<iframe src="https://prql-lang.org/book/"></iframe>
--- a/website/content/playground_integrated.html
+++ b/website/content/playground_integrated.html
@@ -1,0 +1,6 @@
+---
+title: "Playground"
+layout: big_iframe
+---
+
+<iframe src="https://prql-lang.org/playground/"></iframe>
--- a/website/themes/prql-theme/layouts/_default/baseof.html
+++ b/website/themes/prql-theme/layouts/_default/baseof.html
@@ -10,7 +10,9 @@
     <div class="d-flex flex-column flex-grow-1 overflow-scroll">
       {{- block "main" . }}{{- end }}
 
-      {{ partial "footer.html" . }}
+      {{- block "footer" . }}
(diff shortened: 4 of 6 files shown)


aljazerzen (approved):
Nice. I'm +0 on having these things integrated, but if you are at least +1, merge!

max-sixty:
Thanks a lot @mklopets !

I'm +0.8, on some standard deviation scale (is that what we're using? I've seen @aljazerzen say +2 !). The thing I really like about it being integrated is the lack of any friction to trying the lang — reducing even very small frictions keeps people in the funnel... 

I also think this would look even nicer if we made the formatting consistent in the playground, but that's something we might need to outsource

FWIW, interestingly, the one I originally linked from https://github.com/prql/prql/issues/632, https://pomsky-lang.org/ doesn't have their playground integrated any longer

(one could imagine that opening page example being a live playground, but we wouldn't want to wait to download the playground to show that...)

mklopets:
I remember being annoyed with this when starting out – the header nav made sense, but the links in the left sidebar of the playground were something I only noticed this week (and I've been using the playground **a lot.**).

I often found myself going to the playground to try something, realising I didn't know the syntax, and looking for the book. Or, going to the book, and then wanting to test it out in the playground. Having the unified nav makes this way more natural.

So, I'm a +1.5(?)

> if we made the formatting consistent in the playground

Do you essentially mean just the colour scheme?

max-sixty:
Yes good points. 

I think we can merge given those votes, we'd need some big negatives to not do this...

Re the formatting — yes, the color scheme, and the fonts. Makes sense to have different margins though.

aljazerzen:
Voting is from python world. It has started as [-1, -0, +0, +1](https://peps.python.org/pep-0010/#voting-scores), but is sometimes extended to -2, +2.

mklopets:
I'll take a quick stab at unifying the styles a bit more before merging.

max-sixty:
> I'll take a quick stab at unifying the styles a bit more before merging.

OK great! TBC, I didn't mean we had to bundle them. Great if you're up for it but fine to merge any change, however marginal :)

max-sixty:
@mklopets shall we hit the button?

mklopets:
Apologies for the delay; I updated the branch to make it work with the latest changes to the website. 

But, I didn't get around to doing anything to unify the colours between the different parts of the site. @max-sixty I'll leave this with you – please merge it if you'd like, or close it if you don't care for it. My +/- points are listed above

max-sixty:
Great, thank you @mklopets !

PR #1224, opened 2022-12-03
fix: make website footer sticky

### LINKED ISSUE #1220: Website footer doesn't stick to the bottom
When the page isn't tall enough to fill the viewport, the footer doesn't stick to the bottom, but rather floats awkwardly.

image

mklopets:
Fixes #1220
image


max-sixty:
Super, thanks!

PR #1237, opened 2022-12-06
build: Add `package.lock`s?

max-sixty:
These get added when we run the code in [https://github.com/prql/prql/pull/1236](https://github.com/prql/prql/pull/1236/files#diff-cd2d359855d0301ce190f1ec3b4c572ea690c83747f6df61c9340720e3d2425eR186-R187) — should they be added to the repo?

I'm much less familiar with the node world, so appreciate any direction here


aljazerzen:
Yes, lock files should be committed.

PR #1238, opened 2022-12-07
feat: Quote keywords when translating

### LINKED ISSUE #896: Reserved column names like "from" and "to"
I have trouble using column names that are reserved keywords, for example `from` and `to`. There are two sides to this issue:
1. PRQL has syntax errors when parsing these names.
2. The generated SQL code does not put the names into backquotes.

I managed to use S-Strings as a partial workaround, but they don't work at the left side of `=`.
```
s"`from`"
```

max-sixty:
Closes https://github.com/prql/prql/issues/896

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2033,6 +2033,7 @@ dependencies = [
  "globset",
  "insta",
  "itertools",
+ "lazy_static",
  "log",
  "once_cell",
  "pest",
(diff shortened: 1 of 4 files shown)


max-sixty:
(Merging but feel free to comment and I'll clean anything up)

aljazerzen:
Nice, this is as efficient as it gets.

PR #1239, opened 2022-12-07
refactor: Consolidate ident quote check in a regex

max-sixty:


prql-compiler/src/sql/codegen.rs
```diff
@@ -549,28 +550,22 @@ fn is_keyword(ident: &str) -> bool {
...
-    fn starting_forbidden(c: char) -> bool {
-        !(('a'..='z').contains(&c) || matches!(c, '_' | '$'))
-    }
-    fn subsequent_forbidden(c: char) -> bool {
-        !(('a'..='z').contains(&c) || ('0'..='9').contains(&c) || matches!(c, '_' | '$'))
+    lazy_static! {
+        // One of:
+        // - `*`
+        // - An ident starting with `a-z_\$` and containing other characters `a-z0-9_\$`
+        static ref VALID_BARE_IDENT: Regex = Regex::new(r"^((\*)|(^[a-z_\$][a-z0-9_\$]*))$").unwrap();
```
max-sixty, line 560: I'm actually not sure this is simpler! But it is consolidated. And it's much faster, though probably we shouldn't weigh that too highly.

aljazerzen: Actually it's not too bad, complexity wise. And now that I think about performance - before it was much worse.

If you want to make it more readable, we could replace the regex string with call to [pomsky_macro](https://docs.rs/pomsky-macro/0.7.0/pomsky_macro/) ([docs](https://pomsky-lang.org/docs/get-started/quick-start/)).

Equivalent pomsky expression would be (I hope at least):
```
^ ('*' | [ascii_lower '_$'] [ascii_lower ascii_digit '_$']* ) $
```

prql-compiler/src/sql/codegen.rs
```diff
@@ -549,28 +550,22 @@ fn is_keyword(ident: &str) -> bool {
...
     }
 
-    let is_asterisk = ident == "*";
+    let is_match = dbg!(VALID_BARE_IDENT.is_match(&ident));
```
aljazerzen:
```suggestion
    let is_bare = dbg!(VALID_BARE_IDENT.is_match(&ident));
```

aljazerzen (approved):

prql-compiler/src/sql/codegen.rs
max-sixty: Oh yes I'm up for that! I've been following the project albeit from a distance. 

I won't do it quite yet — requires adding a dependency — but v open to doing it in the future

prql-compiler/src/sql/codegen.rs
```diff
@@ -549,28 +550,22 @@ fn is_keyword(ident: &str) -> bool {
...
-    if !is_asterisk
-        && !is_jinja
-        && (ident.is_empty()
-            || ident.starts_with(starting_forbidden)
-            || (ident.chars().count() > 1 && ident.contains(subsequent_forbidden)))
-        || is_keyword(&ident)
-    {
-        sql_ast::Ident::with_quote(ctx.dialect.ident_quote(), ident)
-    } else {
+    if is_jinja || is_match && !is_keyword(&ident) {
```
max-sixty:
```suggestion
    if is_jinja || is_bare && !is_keyword(&ident) {
```

PR #1260, opened 2022-12-12
build: Fix docker test

max-sixty:
CC @richb-hanover


Taskfile.yml
```diff
@@ -19,8 +19,6 @@ vars:
     mitsuhiko.insta
     prql-lang.prql-vscode
     rust-lang.rust-analyzer
-  default_target:
```
max-sixty, line 22: The changes in this file are to fix a separate error that I added last week

PR #1264, opened 2022-12-13
docs: Fix string highlighting in book

max-sixty:


max-sixty:
## Before

image


## After

image

PR #1269, opened 2022-12-15
docs: Add parentheses in syntax

max-sixty:
Because of @richb-hanover's fix, the correct formula for celsius no longer had parentheses!

I also tweaked a few things; giving more room for the other columns by reducing the length of the "Syntax" column

@richb-hanover your changes are great, both per-se, and them inspiring more changes!


max-sixty:
image

image

(feedback welcome as ever)

PR #1270, opened 2022-12-15
docs: Don't include extra string example

max-sixty:
As discussed in
https://github.com/prql/prql/commit/acfe4b3c14a930a4ff3652be3c50d1d230a2ae51

(also some gentle reformatting)


richb-hanover:
I agree the triple double-quote example is too clever by half. It might be worth mentioning on the Strings page.

Good links to Assigns & Aliases

I see you have re-enabled Prettier. My monitor's not wide enough to see the full table - that's why I experimented with turning it off.

But... LGTM. Ship it! Thanks.

PR #1271, opened 2022-12-15
feat: hide intermadiate named from RQ

aljazerzen:
This removes names from column definitions in RQ. Reason behind this is that we don't want to depend on names, because they can cause collisions, but instead used `CId`s.

Instead of names directly on column defs, we do define names of *columns on the table*.

The end result is that:
- CId is used to reference column within a table (in a pipeline)
- RelationColumn (basically a column name) is used to reference a column from outside the table. For example when instantiating the table in some pipeline downstream.

This way, RQ:
- contains column names for each exposed table,
- does not contain column names for intermediate columns.

Practical effect is this:

```diff
  from x
  # declare a column
  derive [intermediate = a + 1]
  
  # trigger a pipeline split
  take 10
  filter a > 0

  # don't expose the intermediate
  select [final = intermediate +2]

  WITH table_1 AS (
    SELECT
-     a + 1 AS intermediate,
+     a + 1 AS _expr_0,
      a
    FROM
      x
    LIMIT
      10
  )
  SELECT
-   intermediate + 2 AS final
+   _expr_0 + 2 AS final
  FROM
    table_1
  WHERE
    a > 0% 
```

----

This commit also fixes some ambiguities on name shadowing in scenarios like this:
```
from x
select [a, a = b, a = b + 1]
```
The behavior that I implemented resolves this table to have this frame: 
```
[<unnamed>, <unnamed>, a]
```

### DIFF
--- a/book/tests/snapshots/snapshot__run_reference_prql@examples__variables-0.prql.snap
+++ b/book/tests/snapshots/snapshot__run_reference_prql@examples__variables-0.prql.snap
@@ -7,8 +7,8 @@ WITH table_1 AS (
   SELECT
     title,
     country,
-    salary + payroll_tax + benefits_cost AS gross_cost,
-    salary + payroll_tax AS gross_salary,
+    salary + payroll_tax + benefits_cost AS _expr_0,
+    salary + payroll_tax AS _expr_1,
     salary
   FROM
     employees
@@ -19,16 +19,16 @@ SELECT
   title,
   country,
   AVG(salary),
-  AVG(gross_salary),
+  AVG(_expr_1),
   SUM(salary),
-  SUM(gross_salary),
-  AVG(gross_cost),
-  SUM(gross_cost) AS sum_gross_cost,
+  SUM(_expr_1),
+  AVG(_expr_0),
+  SUM(_expr_0) AS sum_gross_cost,
   COUNT(*) AS ct
 FROM
   table_1
 WHERE
-  gross_cost > 0
+  _expr_0 > 0
 GROUP BY
   title,
   country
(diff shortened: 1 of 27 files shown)


book/tests/snapshots/snapshot__run_reference_prql@examples__variables-0.prql.snap
```diff
@@ -19,16 +19,16 @@ SELECT
   title,
   country,
   AVG(salary),
-  AVG(gross_salary),
+  AVG(_expr_1),
   SUM(salary),
-  SUM(gross_salary),
-  AVG(gross_cost),
-  SUM(gross_cost) AS sum_gross_cost,
+  SUM(_expr_1),
```
max-sixty, line 24: One downside of this is that DBs will often give a reasonable name for unnamed aggregations.

So we might have got `sum_gross_salary` prior, but now we get `sum__expr_1`.

(as ever, not some verdict on the overall change, and from the peanut gallery)

prql-compiler/src/sql/translator.rs
```diff
@@ -2,20 +2,15 @@
 //! then to a String. We use sqlparser because it's trivial to create the string
 //! once it's in their AST (it's just `.to_string()`). It also lets us support a
 //! few dialects of SQL immediately.
-// The average code quality here is low — we're basically plugging in test
```
max-sixty, line 5: 👍 !

book/tests/snapshots/snapshot__run_reference_prql@examples__variables-0.prql.snap
aljazerzen: Yes that is true - we basically have to replicate their naming rules and if the target database does infer some name, override it with what the RQ dictates.

This would be done here: https://github.com/prql/prql/blob/7a23912e998886c6df38870d19c9119f1dc72998/prql-compiler/src/sql/codegen.rs#L346-L352

prql-compiler/src/sql/translator.rs
aljazerzen: Yee, it's not low anymore :D

Issue #1240, opened 2022-12-07
PRQL after s-string in `from`


max-sixty:
I was adding some docs for the new s-string functionality (which is great!), but hitting an issue re any PRQL which comes after the `from`:

```elm
from s"""SELECT * FROM employees"""
filter country == "USA"
```

Note the `* = 'USA'`

```sql
WITH table_1 AS (
  SELECT
    *
  FROM
    employees
)
SELECT
  *
FROM
  table_1 AS table_0
WHERE
  * = 'USA'
```

If we try and be more explicit, we can't resolve the table name:

```elm
from e=s"""SELECT * FROM employees"""
filter e.country == "USA"
```

```
Error: 
   ╭─[:2:8]
   │
 2 │ filter e.country == "USA"
   ·        ─────┬────  
   ·             ╰────── Unknown name e.country
───╯
```

Possibly s-strings goes against our push for explicitness. I did really like this s-string feature — it's a great escape hatch for anything that isn't currently possible in PRQL. Happy to work on it if it's a tractable problem.

PR #1273, opened 2022-12-15
feat: ensure unique SQL names

### LINKED ISSUE #1249: Generated SQL has duplicate column
PRQL input:
```prql
from report
derive r = s"RANK() OVER (PARTITION BY timeperiod, campaignid ORDER BY _sdc_report_datetime DESC)"
filter r == 1
```

Output:
```sql
WITH table_1 AS (
  SELECT
    *,
    RANK() OVER (
      PARTITION BY timeperiod,
      campaignid
      ORDER BY
        _sdc_report_datetime DESC
    ) AS r
  FROM
    report
)
SELECT
  *,
  r
FROM
  table_1
WHERE
  r = 1
```

This output is not good, because the `SELECT *, r` generates two columns with the same name `r`. One gets renamed to `r_1` by BigQuery. In some situation it can even cause an error "duplicate column name".

aljazerzen:
Closes #1249 
Fixes the first problem with #1240

### DIFF
--- a/prql-compiler/src/ast/pl/frame.rs
+++ b/prql-compiler/src/ast/pl/frame.rs
@@ -61,7 +61,23 @@ impl Default for SortDirection {
 }
 
 impl Frame {
-    pub fn push_column(&mut self, col_name: Option<String>, id: usize) {
+    pub fn apply_assign(&mut self, expr: &Expr) {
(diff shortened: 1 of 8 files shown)


max-sixty:
Very nice, thank you!

PR #1274, opened 2022-12-15
fix: table literal aliases

### LINKED ISSUE #1240: PRQL after s-string in `from`

aljazerzen:
Closes #1240

### DIFF
--- a/prql-compiler/src/lib.rs
+++ b/prql-compiler/src/lib.rs
@@ -1950,6 +1950,26 @@ join y [foo == only_in_x]
           country = 'USA'
         "###
         );
+
+        assert_display_snapshot!(compile(r###"
+        from e=s"""SELECT * FROM employees"""
+        filter e.country == "USA"
+        "###).unwrap(),
+            @r###"
+        WITH table_1 AS (
+          SELECT
+            *
+          FROM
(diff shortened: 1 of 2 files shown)


max-sixty:
Awesome, thanks @aljazerzen !

PR #1276, opened 2022-12-15
docs: Add FAQ re capitalization

max-sixty:
The important questions!

Do folks this is a good idea? Or we could do lowercase for everything?


aljazerzen:
It's an abbreviation - so it should be uppercase, right?

max-sixty:
Yes, fair point. But then should we rename everything to the uppercase? e.g. this repo? Or we say repos are generally lowercase but we use uppercase in all prose?

aljazerzen:
Yes, exactly. We just use kebab-casing for naming repos and crates, but it's uppercase actually.

max-sixty:
OK, I'm fine with that — I'll make the change unless anyone else has a view

max-sixty:
And are we OK to rename the organization to `PRQL` so this repo is `PRQL/prql`?

snth:
I'm happy with the proposal to rename the organisation to `PRQL`.

aljazerzen:
Go ahead with rename and merge of this.

max-sixty:
I changed the org name, I'll update the PR and do some more renames (unless anyone gets there first, but not expecting anyone to)

Issue #504, opened 2022-05-16
`if` or `case` expressions


max-sixty:
I'm thinking from the language POV what might be the biggest gaps, since we're so close to 0.2.

Something we don't yet have in PRQL is `if` or `case when then else end` expressions.

This might be the biggest gap in the language atm. Should we do anything on this for 0.2 or leave them? (adding a 0.2 tag for the sake of the question).

A few options:
- A tertiary expression, like `x ? y : z` (probably not with a colon though)
- An `if` function, like `x | if y z` — though written this way it would need to have a signature `f y z x`, maybe not conventional
- A case pipeline like (without having thought about it sufficiently):
```
case (
  x -> y
  _ -> z
)
```
- A match pipeline like (without having thought about it sufficiently):
```
match a (
  b -> c
  d -> e
  _ -> f
)
```

I'll close https://github.com/prql/prql/issues/50 in favor of this.

max-sixty:
Malloy uses `pick` for these:

```
size ?
  pick 'small' when < 10
  pick 'medium' when < 20
  else 'large'
```

Tangentially, something nice in Malloy is "[partial comparison](https://looker-open-source.github.io/malloy/documentation/language/expressions.html#partial-comparison)"; the `pick` becomes an expression rather than a statement; or another way of saying this in that it curries `size ?`. So the above is equivalent to:

```
  pick 'small' when size < 10
  pick 'medium' when size < 20
  else 'large'
```

ref #71

Ryman:
@max-sixty Is there any workaround for this currently? Have you had any further thoughts on what direction you'd like to go? 

It seems like an `if` expression might be nice: 

```prql
select [ 
    property = if x { if y > 200 { 'foo' } else { 'bar' } } else { 'baz' } 
]
```

Seems like you can use S-strings to at least get *something* working and have the benefit of reusing definitions:
```psql
# in the employees sample
derive [
    tenure_or_zero = tenure ?? 0,
    tenure_type = s"""IF {tenure_or_zero} > 4 THEN 'senior' ELSE IF {tenure_or_zero} > 10 'legend' ELSE 'normal' END"""
  ]
  ```
generates:
```sql
IF COALESCE(newest_employees.tenure, 0) > 4 THEN 'senior'
  ELSE IF COALESCE(newest_employees.tenure, 0) > 10 'legend'
  ELSE 'normal'
END AS tenure_type
```

max-sixty:
[...]

PR #1278, opened 2022-12-15
feat: switch

aljazerzen:
First stab at #504

```elm
from employees
derive display_name = switch [
    nickname != null -> nickname,
    true -> f'{first_name} {last_name}'
]
```

Note that:
- default case can be set by using `true` condition.
- if there is no `true` condition, default is NULL (unlike in SQL where an exception is the default).

### DIFF
--- a/prql-compiler/src/ast/pl/expr.rs
+++ b/prql-compiler/src/ast/pl/expr.rs
@@ -59,6 +59,7 @@ pub enum ExprKind {
     TransformCall(TransformCall),
     SString(Vec<InterpolateItem>),
     FString(Vec<InterpolateItem>),
+    Switch(Vec<SwitchCase>),
 }
 
 impl ExprKind {
@@ -69,8 +70,6 @@ impl ExprKind {
         }
     }
 }
-
-mod modname {}
 
 #[derive(
     Debug, PartialEq, Eq, Clone, Serialize, Deserialize, strum::Display, strum::EnumString,
@@ -221,6 +220,12 @@ impl Range {
             false
         }
     }
+}
+
+#[derive(Debug, PartialEq, Eq, Clone, Serialize, Deserialize)]
+pub struct SwitchCase<T = Expr> {
+    pub condition: T,
+    pub value: T,
 }
 
 /// FuncCall with better typing. Returns the modified table.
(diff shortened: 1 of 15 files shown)


max-sixty:
Amazing — huge win! (and such a small PR!)

vanillajonathan:
Isn't it rather misleading to use `true` for this?
Then you would assume it would use that branch when it was actually literally `true`?
It makes no sense that `true` is used as the default case.

I propose instead to do it the way Rust does and use `_` to mean for anything else.

max-sixty:
> Then you would assume it would use that branch when it was actually literally `true`?

This isn't matching against a value, it's switching based on which condition is `true`; does that make sense?

vanillajonathan:
> > Then you would assume it would use that branch when it was actually literally `true`?
> 
> This isn't matching against a value, it's switching based on which condition is `true`; does that make sense?

Oh okay, I think I get it now, but it is a bit unconventional and different from the C-style `switch` and the Rust-style `match` that I am used to. Seeing the name `switch` might lead someone to believe it would work something like a C-style switch statement. Despite sharing the same name it works entirely different.

An alternative could be:
```elm
from employees
derive display_name = condition [
    if nickname != null -> nickname,
    else -> f'{first_name} {last_name}'
]
```

This would avoid confusion by not using the name name "switch" , and instead using some other name such as "condition". Using the keywords `if` and `else` could also further be used to make it easier to grasp.

max-sixty:
I actually think you're right about the switch statement. I'm not sure why I missed that before. I think we were going between `switch` and `match` and `case`, and deciding whether there was a control value, and we ended up here.
Note that SQL uses `case` like we use `switch`, though other languages use `case` as part of a traditional `switch` statement.

We did spend a lot of time discussing this, and we've spent a lot of time discussing the syntax (!), so maybe folks' patience for more discussions is waning. But I would propose we either:
- Rename `switch` to something like `case`
- or have a control value, which could be `true` to get the current behavior, or could be `foo` to evaluate against `foo`
- or some other synthesis of the tradeoffs

I'm less of a fan of the `if` / `else` here fwiw, though there are reasonable alternative constructions of `if` & `else`)

PR #1279, opened 2022-12-15
refactor: Use `Iterator` for `IdGenerator`

max-sixty:
Another minor PR to let me become more acquainted with the new compiler, while doing something of at least positive value.

Though as ever @aljazerzen lmk if this isn't positive value.


aljazerzen:
I don't like this. Unwrap gives an impression that it can fail.

I would be up for merging this, if we used IdGenerator as iterator somewhere. But because we don't, I'm -1.

max-sixty:
Yes fair. I felt like we were doing our own `Iterator`, but you're right that `Iterator`s return type adds noise without much benefit

PR #1280, opened 2022-12-15
refactor: public API

aljazerzen:
First stab at #1167

Ref #1182

I haven't done any of the bindings yet, waiting for your opinion @max-sixty


prql-compiler/src/sql/mod.rs
```diff
@@ -3,7 +3,38 @@
...
 
-pub use translator::translate;
+pub use dialect::Dialect;
+pub use translator::compile;
+
+/// Compilation options for SQL backend of the compiler.
+pub struct Options {
+    /// True for passing generated SQL string trough a formatter that splits
+    /// into multiple lines and prettifies indentation and spacing.
+    pub format: bool,
```
max-sixty, line 19: Awesome, great idea

prql-compiler/src/lib.rs
```diff
@@ -5,13 +5,16 @@ mod error;
...
 mod utils;
 
 #[cfg(feature = "cli")]
 pub use cli::Cli;
 pub use error::{ErrorMessage, IntoErrorMessage, Result, SourceLocation};
-pub use parser::parse;
-pub use sql::translate;
+pub use parser::parse as prql_to_pl;
+pub use semantic::resolve as pl_to_rq;
+pub use sql::compile as rq_to_sql;
```
max-sixty: I have to admit I was trying to find `rq_to_sql` for a while! I haven't seen this renaming much in rust, but if it's relatively standard then great; there is a nice elegance to it

prql-js/src/lib.rs
```diff
@@ -2,7 +2,7 @@
 #![allow(clippy::drop_non_drop)]
 mod utils;
 
-use prql_compiler::{pl_to_json, pl_to_prql, translate, ErrorMessage, IntoErrorMessage};
+use prql_compiler::{pl_to_json, pl_to_prql, compile, ErrorMessage, IntoErrorMessage};
```
max-sixty: How should we deal with `options` on the language bindings? Would we need an `Options.default()` for each language?

Would there be any logic in having `Option<Options>`, and then we could easily pass a `None` value? It possibly adds a layer of indirection within rust though.

max-sixty (commented):
I thought the plan was great — ideal architecture! Though I haven't thought about any big tradeoffs, if there are any.

I left a couple of small comments, and I'm still looking through.

Thank you v much! Quite the recent run over past few days... :)

prql-compiler/src/lib.rs
aljazerzen: Yeah, it's not the most conventional way, but I also don't want to rename parse and resolve base functions.

prql-js/src/lib.rs
aljazerzen: Good idea, will do that.

aljazerzen:
This can be merged, BUT it's a breaking change and if we merge, shouldn't release any more 0.3.x but 0.4 as the next release.

max-sixty:
> This can be merged, BUT it's a breaking change and if we merge, shouldn't release any more 0.3.x but 0.4 as the next release.

I think that given your recent run of feature additions, it might be time for a 0.4 anyway... So I think no reason to wait from that logic

prql-compiler/src/error.rs
```diff
@@ -21,12 +21,14 @@ pub struct Error {
     pub help: Option<String>,
 }
 
-#[derive(Debug, Clone)]
+/// Location within the source file.
+/// Tuples contain:
+/// - line number (1-based),
+/// - column number within that line (1-based),
```
max-sixty, line 28: 👍

prql-compiler/src/error.rs
```diff
@@ -97,115 +111,92 @@ impl Display for Error {
...
-        let config = Config::default().with_color(color);
-
-        if let Some(error) = self.downcast_ref::<Error>() {
-            let message = error.reason.message();
+}
 
-            if let Some(span) = error.span {
-                let span = Range::from(span);
+impl ErrorMessage {
+    pub fn as_json(&self) -> String {
```
max-sixty: FWIW technically I think this should be `to_json` based on https://rust-lang.github.io/api-guidelines/naming.html#ad-hoc-conversions-follow-as_-to_-into_-conventions-c-conv, but no huge deal

prql-compiler/src/lib.rs
```diff
@@ -22,2218 +57,56 @@ static PRQL_VERSION: Lazy<Version> =
...
+    parser::parse(prql)
+        .map_err(error::downcast)
+        .map_err(|e| e.composed("", prql, false))
 }
 
-/// Serialize PL AST into JSON
-pub fn pl_to_json(pl: Vec<Stmt>) -> Result<String> {
-    Ok(serde_json::to_string(&pl)?)
+/// Perform semantic analysis and convert PL to RQ.
+pub fn pl_to_rq(pl: Vec<Stmt>) -> Result<Query, ErrorMessage> {
```
max-sixty: These are both conceptually elegant, like the previous iteration, and easy to find 👏

book/src/main.rs
```diff
@@ -112,7 +112,8 @@ fn replace_examples(text: &str) -> Result<String> {
             Event::Start(Tag::CodeBlock(CodeBlockKind::Fenced(lang))) if lang == "prql".into() => {
                 if let Some(Event::Text(text)) = parser.next() {
                     let prql = text.to_string();
-                    let html = table_of_comparison(text.as_str().unwrap(), &compile(&prql)?);
+                    let html =
+                        table_of_comparison(text.as_str().unwrap(), &compile(&prql).unwrap());
```
max-sixty, line 116: We have lots of functions like `{output}_of_{input}`, like the one showing in this diff. 

I had marginally preferred that to `{input}_to_{output}`, since the output intuitively seems like higher-bit information, and makes for better autocomplete. (and in rust we get the `to_{ouput}` in methods on the `Input` already). I think I picked that up from my occasional Ocaml work where it's more of a standard.

Totally fine to move to `{input}_to_{output}` — which the proposed code does. If we want to move to that standard, we can add that to some "Conventions" section in `Development.md` and we can gradually move our functions to that standard?

prql-compiler/src/error.rs
aljazerzen: I didn't know there is actually a guide on that! I'll change it.

book/src/main.rs
aljazerzen: Actually, I do prefer `{output}_of_{input}`, it just that I didn't think of it when refactoring.

Since this API won't change for a long time, hopefully, I think it's worth correcting it now.

aljazerzen:
Another thing that should be done now, to prevent future changes to the API, is to return `Vec<ErrorMessage>` instead of `ErrorMessage` (wrapped in Result::Err in both cases).

book/src/main.rs
snth: Sorry to chime in here when I haven't really been involved and my comments are also purely at the English level, not the semantic or implementation level. Overall I am in favour of having a consistent Convention that makes the most sense to people.

Is `{something}_of_{something}` a common practice in some languages? I'm particularly talking about the `_of_` bit here? I'm familiar with things like `to_string()` in Rust and for me the opposite to that is always a `from_{something}` like `from_str()`. So for me `{something}_from_{something}` would be the natural choice. For example I used to find the `fahrenheit_of_celsius` example quite confusing because to me `of` means something slightly different and `fahrenheit_from_celsius` would have made more sense to me.

`from` is slightly longer to type though so not quite as convenient.

---
Regarding `output_(of|from)_input` vs `input_to_output`, I don't think that I actually have particularly strong feelings on this point but I'm just going to set the cat amongst the pigeons here by suggesting that isn't this a bit like the original SQL argument for having the SELECT come first, i.e. to focus on what the outputs are which is what a user might care most about rather than where the data comes from?

So if you compare how a full conversion might be written

```rust
sql = prql.prql_to_pl().pl_to_rq().rq_to_sql()
```
vs
```rust
sql = prql.pl_from_prql().rq_from_pl().sql_from_rq()
```
I would argue that the first one is easier to mentally verify that you have strung the conversions together in the correct manner whereas the second one is quite hard to read IMHO.

Actually, after having written this up I find myself do having a rather strong preference for the `input_to_output()` form because it makes it easy to mentally check that what's going in on the left is of the correct form/type and it let's you know what comes out on the RHS so at that point you can decide what you want to carry on doing with it.

aljazerzen: You make some good points for `input_to_output` and now I'm split between the two conventions.

For now this easy to change, but after we release 0.4, people may depend on the API. Which means this we should decide now. The change takes me 10min at most.

@max-sixty WDYT?

max-sixty: I'm very open-minded!

A couple of quick points, can come back to this later:

max-sixty: Overall I'm v open-minded, really, and place a lot of weight on whatever people will expect — so if that's `X_to_Y` then that's fine. 

A couple of quick points:

---

> but I'm just going to set the cat amongst the pigeons here by suggesting that isn't this a bit like the original SQL argument for having the SELECT come first, i.e. to focus on what the outputs are which is what a user might care most about rather than where the data comes from?

lol, very good! I haven't quite reconciled this in my mind yet, the dissonance is too strong

---

The method chaining is somewhat different, since the methods know the type of the input, and so can just be:

```diff
-sql = prql.prql_to_pl().pl_to_rq().rq_to_sql()
+sql = prql.to_pl().to_rq().to_sql()
```

This ofc doesn't work for string objects containing json / rq / sql / prql. For actual types, having `to_X` is great though.

---

I first heard about `of` in Ocaml. It's actually quite hard to Google for "of" :) to find references, but e.g. in the ReasonML book on P31 they cover it: http://reasonmlhub.com/exploring-reasonml/downloads/exploring-reasonml-preview-book.pdf

aljazerzen: I've searched crates.io for any kind of converters and took a look of naming they use.

They are not major crates with multi-million downloads, but they do have a few thousand and should reflect Rust naming conventions.

`get_future_record_from_struct` https://crates.io/crates/schema_registry_converter
`convert_etrs89_to_osgb36_threaded` https://crates.io/crates/lonlat_bng
`ms_into_time` https://crates.io/crates/ms-converter
`dos2unix` https://docs.rs/newline-converter/0.2.2/newline_converter/

The most common pattern I found was a trait with function `into_{output}`, but this doesn't fit the way I want people to think about these functions.

So in conclusion, there is no convention, namings are all over the place, but I did not find a mention of `{output}_of_{input}`.

snth: On the [Rust API Guidelines:Naming](https://rust-lang.github.io/api-guidelines/naming.html) page, apart from the [Ad-hoc conversions follow as_, to_, into_ conventions (C-CONV)](https://rust-lang.github.io/api-guidelines/naming.html#ad-hoc-conversions-follow-as_-to_-into_-conventions-c-conv) @max-sixty already linked to above, the only other bit is [Casing conforms to RFC 430 (C-CASE)](https://rust-lang.github.io/api-guidelines/naming.html#casing-conforms-to-rfc-430-c-case) which includes:
image

How strongly typed are the input objects, would `into_{output}` not be sufficient?

> The most common pattern I found was a trait with function into_{output}, but this doesn't fit the way I want people to think about these functions.

How is it that you want people to think about these functions? Why do you want the input type to be included?

---
Edit: I guess a `to_sql()` could be a compound function, i.e. it could either be `rq_to_sql()` or `prql_to_pl().pl_to_rq().rq_to_sql()` or something like that, so it's not unique.

Issue #1253, opened 2022-12-11
How to use Table Aliases?


max-sixty:
[Moving from discussion, looks like a surprising bug, thank you @richb-hanover ]
### Discussed in https://github.com/prql/prql/discussions/1252

<div type='discussions-op-text'>

<sup>Originally posted by **richb-hanover** December 11, 2022</sup>
I would like to use (I think they're called) Table Aliases to simplify code. That is, I would like to use:

```
from l=TableA
join r=TableB [==Col1]
select [
  l.Col1,
  r.Col1,
]
filter [l.Col1 != r.Col1]
```

* If I use this form, the Playground gives an error: `Unknown name l.Col1` in the filter expression.

* If I move the filter between the `join` and `select`, I get a long `Cannot lower to IR expr:...` message.

What am I missing? Thanks.</div>

max-sixty:
This looks like a bug; surprisingly quite a big one. Thanks for finding it @richb-hanover .

Adding the `filter` here gives a long error:

```elm
from a
join b [a.id == b.id]
filter [b.id == 5]   # works without this
```

```
"Cannot lower to IR expr: `Expr { id: None, kind: List([Expr { id: Some(15), kind: Binary { left: Expr { id: 
Some(16), kind: Ident(Ident { path: [\"_frame\", \"b\"], name: \"id\" }), span: Some(span-chars-38-43), 
target_id: Some(11), ty: Some(infer), needs_window: false, alias: None }, op: Eq, right: Expr { id: Some(17), 
kind: Literal(Integer(5)), span: Some(span-chars-46-47), target_id: None, ty: Some(integer), needs_window: 
false, alias: None } }, span: Some(span-chars-38-47), target_id: None, ty: Some(bool), needs_window: false, 
alias: None }]), span: Some(span-chars-37-48), target_id: None, ty: None, needs_window: false, alias: None 
}`"
```

(line breaks added)

richb-hanover:
> This looks like a bug; surprisingly quite a big one...

Sometimes I feel like Inigo Montoya wandering around with my eyes closed, coming up with interesting stuff...

aljazerzen:
This is indeed a bug, but the original question has an interesting answer:

> If I use this form, the Playground gives an error: Unknown name l.Col1 in the filter expression.

If I add frame annotations, this will make it clear:

```
from l=TableA           # [l.*]
join r=TableB [==Col1]  # [l.*, l.Col1, r.*, r.Col1]
[...]

PR #1285, opened 2022-12-17
docs: Add docs clarifying how aliases flow

max-sixty:
Based on https://github.com/prql/prql/issues/1253

If I'm thinking about this correctly, our implementation is different to SQL, which carries the table / namespace through the query.

I don't think it's necessarily bad, and we can have a nice error message saying "Did you mean `first_name`?" if we see a `e.first_name` column

### DIFF
--- a/book/src/transforms/select.md
+++ b/book/src/transforms/select.md
@@ -41,8 +41,7 @@ e.first_name`, such that the table / namespace is lost. So this would not work:
 ```prql_no_test
 from e=employees
 select e.first_name
-# Can't find `e.first_name`
-derive fn = e.first_name
+select fn = e.first_name # Error: cannot find `e.first_name`
 ```
 
 ...and would instead need to be:
--- a/book/src/transforms/select.md
+++ b/book/src/transforms/select.md
@@ -41,7 +41,7 @@ e.first_name`, such that the table / namespace is lost. So this would not work:
 ```prql_no_test
 from e=employees
 select e.first_name
-select fn = e.first_name # Error: cannot find `e.first_name`
+select fn = e.first_name  # Error: cannot find `e.first_name`
 ```
 
 ...and would instead need to be:
--- a/book/src/transforms/select.md
+++ b/book/src/transforms/select.md
@@ -47,7 +47,7 @@ derive fn = e.first_name
 
 ...and would instead need to be:
 
-```prql_no_test
+```prql
 from e=employees
 select e.first_name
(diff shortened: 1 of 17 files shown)


book/src/transforms/select.md
```diff
@@ -3,30 +3,52 @@
...
+```
+
+Note that currently `select e.first_name` is an alias for `select first_name =
+e.first_name`, such that the table / namespace is lost. So this would not work:
+
+```prql_no_test
+from e=employees
+select e.first_name
+# Can't find `e.first_name`
+derive fn = e.first_name
```
aljazerzen:
```suggestion
select e.first_name
select fn = e.first_name # Error: cannot find `e.first_name`
```

aljazerzen (approved):

book/src/transforms/select.md
max-sixty: Possibly we should show errors in the docs, WDYT?

book/src/transforms/select.md
```diff
@@ -3,30 +3,51 @@
...
+select [e.first_name, e.last_name]
+```
+
+Note that currently `select e.first_name` is an alias for `select first_name =
+e.first_name`, such that the table / namespace is lost. So this would not work:
+
+```prql_no_test
+from e=employees
+select e.first_name
+select fn = e.first_name # Error: cannot find `e.first_name`
```
max-sixty:
```suggestion
select fn = e.first_name  # Error: cannot find `e.first_name`
```

book/src/transforms/select.md
```diff
@@ -3,30 +3,51 @@
...
-select [
-  name = f"{first_name} {last_name}",
-  age_eoy = dob - @2022-12-31,
-]
+from e=employees
+select [e.first_name, e.last_name]
+```
+
+Note that currently `select e.first_name` is an alias for `select first_name =
+e.first_name`, such that the table / namespace is lost. So this would not work:
```
max-sixty:
```suggestion
e.first_name`, such that the table / namespace is not retained. So this would not work:
```

book/src/transforms/select.md
aljazerzen: Good idea, yes

aljazerzen: If it's easy, otherwise don't bother.

max-sixty: It needs some adjustment so it's something we can do soon if not immediately. Will be good to show off some of our error messages!

Issue #1102, opened 2022-11-01
Lezer / CodeMirror grammar


max-sixty:
As discussed with @antonycourtney, a [Lezer]( https://lezer.codemirror.net/docs/guide/) grammar would be useful, to the extent he explores using PRQL in some of his tools. 

IIUC this will give us a CodeMirror grammar for free. It's a similar design to treesitter.

For context we have a startling array of grammars for syntax highlighting:
- [hljs](https://github.com/prql/prql/blob/8c402e06ef7393211fc947c6f1684f8865c5ab02/book/highlight-prql.js), for the website & book
- [textmate](https://github.com/prql/prql-code/blob/d9946a6897688f5364e3c710e61a8fd77c60c006/syntaxes/prql.tmLanguage.json), for VSCode
- [monaco](https://github.com/prql/prql/blob/8c402e06ef7393211fc947c6f1684f8865c5ab02/playground/src/workbench/prql-syntax.js), for the playground

aljazerzen:
Similar to #285, related to #131

max-sixty:
Closing as completed. I'm yet to work this into the Jupyter extension though.

Issue #285, opened 2022-03-30
Write a language specification (EBNF syntax)


dlescos:
Write a language specification document in Extended Backus-Naur Form (see the [Go document for an example](https://go.dev/ref/spec#Notation)) may help us:

* Ensure everyone share the same understanding of particular aspects of the syntax
* Help newcomers to dig in the details
* Provide a safe layout for external projects which would like to interface with us (pgadmin, phpmyadmin...)

max-sixty:
I'm definitely open to this. 

As we continue iterating on the language, we wouldn't want this to block changes. But if there were a willing contributor, we could create one at a point-in-time, and refresh it asynchronously.

Once the language is more stable, then possibly we could support it and provide a guarantee that it's current.

Does that make sense @dlescos ?

dlescos:
Sure! And I agree the document should follow the code and not make things to rigid.

qharlie:
We use Lark for PyPrql which is like eBNF but with a few extras , https://github.com/qorrect/PyPrql/blob/main/pyprql/lang/prql.lark .  

The extras are `[]` means `()?` , and a `.N` after the rule name ( where N is a number ) assigns priorities to the rules if there is a conflict.

chris-pikul:
In working on a Go implementation I actually started taking notes in the form of a rough RFC-ish markdown document. I would like to convert the syntax to Extended Backus-Naur Form and actually will start doing that now.

My point in bringing this up, is that it may serve as a starting point for more in-depth official specifications for PRQL. I found the "book" to be a little vague on the finer points of how the compiler/parser actually works, or syntax behaviors expected.

Anyways, let me know if this is a good direction and something that is wanted officially. If it is, I can work it out as a more robust version for the language in general.

[github.com/chris-pikul/go-prql/SYNTAX-NOTES.md](https://github.com/chris-pikul/go-prql/blob/main/SYNTAX-NOTES.md)

max-sixty:
@chris-pikul that looks very pressive, thanks for writing it up.

I'm open to adding this to our docs. I think for the moment the canonical source should be the actual implentation, but over time the canonical source could be this document. In particular, proposing changes (e.g. how inline pipelines should work), could be in this doc, rather than a random issue / a prototype implementation.

It all looks quite accurate. Couple of tiny things on first glance:
- `prql` is not _required_ at the start of queries
- ranges don't require an `in` function in the lang; though atm they practically do, given that's the only way of using them
- we could add dates and times
[...]

PR #1289, opened 2022-12-18
feat: Initial Lezer grammar

max-sixty:
Based on https://github.com/prql/prql/issues/1102

I've been using https://lezer-playground.vercel.app/, which is great!  And I struggled to set up a JS project that was good for iterating; though we'll need to do that at some point.

This parses our standard query (in examples.md) though it gives a couple of warnings around linebreaks — I still haven't really got my head around the way `@precedence` works.

It doesn't yet parse things like `func` & `table` though those aren't difficult. It heavily relied upon our pest file, though requires some different approaches in some areas.

---

Where should this go? In this repo? How should we package it? https://github.com/Snowflake-Labs/lezer-snowsql is an example of a full repo & package for a Lezer grammar


aljazerzen:
I like the syntax of Lezer grammars, because it's quite similar to EBNF #285. 

But I don't see the point of having another grammar definition, along with 4 others, each of which is not totally correct... We need a way to unify them somehow.

max-sixty:
> But I don't see the point of having another grammar definition, along with 4 others, each of which is not totally correct... We need a way to unify them somehow.

In a way I couldn't agree more — I've been quite surprised at how fragmented the space is.

But what's the alternative? They're too different to automatically translate — at least no one has built a tool to do that. And we need CodeMirror / Lezer for PRQL to a) be included in DuckDB's new tool (though we have no guarantee it would be included even if we do make this), and b) have syntax highlighting in Jupyter (though again, it's still non-trivial to get that to work).

But totally mood-affiliated and very open to ideas!

(And maybe someone will come and ask for a treesitter grammar — that would be a 5th!)

aljazerzen:
Ok, that all I needed to hear: there is a need for them. 

The grammar should not change much so there is not a lot of maintenance anyway. And even it they are broken, the worst effect is bad syntax highlighting.

max-sixty:
This is in a pretty good state now! 

I now need to work out how to get it to work with Jupyter etc; it may go in its own repo. I may use https://github.com/Snowflake-Labs/lezer-snowsql as a template

max-sixty:
Done!

This contributes towards (but doesn't close) https://github.com/PRQL/pyprql/issues/45

PR #1290, opened 2022-12-18
docs: Update `select` discussion of table / namespace

richb-hanover:
I wasn't quite sure I understood the explanation. Here's my attempt to explain it. Thanks.

### DIFF
--- a/book/src/transforms/derive.md
+++ b/book/src/transforms/derive.md
@@ -3,7 +3,11 @@
 Computes one or more new columns.
 
 ```prql_no_test
-derive [{new_name} = {expression}]
+derive [
+  {new_name} = {expression},
+  # or
+  {expression}
+]
 ```
 
 ## Examples
--- a/book/src/transforms/select.md
+++ b/book/src/transforms/select.md
@@ -35,20 +35,21 @@ from e=employees
 select [e.first_name, e.last_name]
 ```
 
-Note that currently `select e.first_name` is an alias for `select first_name =
-e.first_name`, such that the table / namespace is lost. So this would not work:
+_Note:_ In the example above, the `e` representing the table / namespace
+is no longer available after the `select` statement.
+So the `filter` statement below would give a "Can't find" error:
 
 ```prql_no_test
 from e=employees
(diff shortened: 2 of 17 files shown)


max-sixty:
Thanks @richb-hanover — unfortunately it seems like we might have done the same thing! Our explanations are fairly similar!

I did make an effort to tag the original issue in the PR, sorry if that wasn't visible enough.

Feel free to edit my explanation with anything you think can be improved!

richb-hanover:
Hmmm... I was responding to the description that (I think) is still [in the Book](https://prql-lang.org/book/transforms/select.html#examples)

When I first read it, I didn't quite understand what was being said, and the examples didn't help me. I am experimenting with the language to make it more explicit about the thing that isn't available. (I will be glad for corrections if I've got it wrong.)

richb-hanover:
I also realize I don't entirely understand how to connect the example PRQL code and the resulting SQL...

max-sixty:
Did you see the PR at https://github.com/prql/prql/pull/1285? It was not merged when you wrote this PR (but was opened 6 hours earlier...). 

It's now merged, so that means we have some conflicts with your branch:
 
image

richb-hanover:
@max-sixty  ~~No - I hadn't seen that. Let me read it over... Thanks.~~

Actually, I had seen it - here's my reasoning for the change...

- I wanted to be clear that the `e` represents the `employees` table
- I wanted to note that it's not so much an "alias issue", but that the `e` is just _gone_ after the `select`
- I wanted to use a `filter` statement in the example since it's kind of orthogonal to the use of `select`
- I switched to using `fname` as the variable name, because my brain initially saw `fn` as "function"
- I wanted to remind the user that the `fname` variable referred to the column
- I removed the apparent time-sensitive word "currently" - it makes it seem as if a (major?) change might be in the wings.

max-sixty:
Very open to changes — do you want to merge the current version into yours, and then we can see the diff?

richb-hanover:
Abandoned - merging (new) main branch into mine...

PR #1303, opened 2022-12-20
fix: prql-js public API

aljazerzen:


max-sixty:
Thanks — lmk if I broke something

PR #1305, opened 2022-12-20
feat: playground duckdb

aljazerzen:


max-sixty:
As discussed on discord, this is huge!

snth:
Totally agree, this is amazing!

snth:
Minor niggle, the dates in employees.prql and invoices.prql are difficult to interpret as they are currently displayed:

image

If I had to guess, I would say you probably need some conversion around Output.js:63 depending on the column type.

playground/src/output/Output.js
```diff
@@ -0,0 +1,99 @@
...
+      const arrow = this.props.content.arrow;
+
+      const header = arrow.schema.fields.map((f, index) => {
+        return <th key={index}>{f.name}</th>;
+      });
+
+      const data = arrow.toArray().map((x) => [...x]);
+      const rows = data.map((x, index) => {
+        const cells = x.map(([_name, value], index) => (
+          <td key={index}>{"" + value}</td>
```
aljazerzen, line 63: @snth you exactly right. Here I'd have to do a `new Date(value)` when type is timestamp/date.

That could be determined from looking into `arrow.schema.fields[index].type`.

I probably won't get around to doing this, but I'm leaving a comment here for anyone who might want to.

PR #1307, opened 2022-12-20
build: Upgrade dependencies

max-sixty:
Will clear a non-significant security warning in GH


max-sixty:
This ended up being a giant yak-shave, since upgrading meant we brought in a dependency that failed to compile under wasm.

So we now exclude much more from compiling under wasm.

(I find rust's approach to target compatibility not that friendly — there's no way for crates to signal whether they're compatible with a certain target. Instead, the compilation just fails. `cargo tree` was helpful at tracking things down.)

PR #1300, opened 2022-12-20
docs: Tweak "select" wording

richb-hanover:
This tracks the conversation in https://github.com/prql/prql/pull/1290 to update the language re: `select` and loss of the table/namespace alias

PR #1310, opened 2022-12-21
ci: Deploy GH Pages unless on a fork

max-sixty:


max-sixty:
Thanks to @richb-hanover for spotting it hadn't deployed in https://github.com/prql/prql/pull/1300

PR #1312, opened 2022-12-21
refactor: public API

aljazerzen:
Re #1280

Given
- rust naming conventions, 
- research described in #1280, 
- the fact that we often use the following chaining pattern,
  ```
  Ok(prql).and_then(prql_to_pl).and_then(pl_to_rq)
  ```
- and my personal preference

... I'm changing public API to use `{input}_to_{output}` naming convention.

Additionally, I future proofed serialization and moved it into `json` module.

### DIFF
--- a/prql-compiler/src/lib.rs
+++ b/prql-compiler/src/lib.rs
@@ -10,21 +10,21 @@
 //!    (parse) │ ▲
 //! prql_to_pl │ │ pl_to_prql
 //!            │ │
-//!            ▼ │      json::to_pl
+//!            ▼ │      json::from_pl
 //!                   ────────►
 //!           PL AST            PL JSON
 //!                   ◄────────
-//!            │        json::from_pl
+//!            │        json::to_pl
 //!            │
 //!  (resolve) │
 //!   pl_to_rq │
 //!            │
 //!            │
-//!            ▼        json::to_rq
+//!            ▼        json::from_rq
 //!                   ────────►
 //!           RQ AST            RQ JSON
 //!                   ◄────────
-//!            │        json::from_rq
+//!            │        json::to_rq
 //!            │
 //!  rq_to_sql │
 //!            ▼
--- a/prql-compiler/src/lib.rs
+++ b/prql-compiler/src/lib.rs
@@ -104,7 +104,7 @@ pub fn pl_to_prql(pl: Vec<ast::pl::Stmt>) -> Result<String, ErrorMessages> {
 pub mod json {
     use super::*;
 
-    /// JSON serialization    
+    /// JSON serialization
     pub fn from_pl(pl: Vec<ast::pl::Stmt>) -> Result<String, ErrorMessages> {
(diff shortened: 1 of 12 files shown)


prql-compiler/src/lib.rs
```diff
@@ -8,25 +8,25 @@
 //!            PRQL
 //!
 //!    (parse) │ ▲
-//! pl_of_prql │ │ prql_of_pl
+//! prql_to_pl │ │ pl_to_prql
 //!            │ │
-//!            ▼ │      pl_of_json
+//!            ▼ │      json::to_pl
```
snth: Based on the tests below, I think the labels in lines 13 and 17 should be the other way around.

prql-compiler/src/lib.rs
```diff
@@ -8,25 +8,25 @@
...
-//!            │        json_of_pl
+//!            │        json::from_pl
 //!            │
 //!  (resolve) │
-//!   rq_of_pl │
+//!   pl_to_rq │
 //!            │
 //!            │
-//!            ▼        rq_of_json
+//!            ▼        json::to_rq
```
snth: Based on the tests below, I think the labels in lines 23 and 27 should be the other way around.

prql-compiler/src/test.rs
```diff
@@ -36,8 +36,8 @@ fn test_stdlib() {
 
 #[test]
 fn json_of_test() {
```
snth, line 38: Suggestion: rename to `json_from_prql_test` (or `prql_to_json_test`).

prql-java/src/lib.rs
```diff
@@ -33,7 +33,7 @@ pub extern "system" fn Java_org_prql_prql4j_PrqlCompiler_toJson(
         .expect("Couldn't get java string!")
         .into();
     let rs_json_str: String =
-        { pl_of_prql(&prql_query).and_then(json_of_pl) }.expect("Couldn't get prql json of query!");
+        { prql_to_pl(&prql_query).and_then(json::from_pl) }.expect("Couldn't get prql json of query!");
```
snth: Suggested change:

    .expect("Couldn't get json from prql query!")

snth (changes requested):
Just the diagram labels that I think need to be changed.

I also made some suggestions for name changes but those are up to you and not essential.

FWIW, having read through some of the tests with the `.and_then(...)` chains, I find them much easier to read with the new naming convention. Thanks

prql-compiler/src/lib.rs
aljazerzen: Oh shit, you're right

snth (approved):
LGTM now!

prql-compiler/src/lib.rs
snth: No worries. That's why we have code reviews to catch typos and such!

max-sixty:
Super, thank you! And re `of` vs `to` — never am I happier than when we do something better than I came up with

PR #1313, opened 2022-12-21
fix: playground

### LINKED ISSUE #1309: Playground "copy to clipboard" no longer works
I'm just getting `undefined` hitting the clipboard. I tried on Firefox & Chrome

aljazerzen:
Closes #1309

### DIFF
--- a/prql-compiler/src/error.rs
+++ b/prql-compiler/src/error.rs
@@ -24,8 +24,8 @@ pub struct Error {
 
 /// Location within the source file.
 /// Tuples contain:
-/// - line number (1-based),
-/// - column number within that line (1-based),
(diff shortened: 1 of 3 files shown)


prql-compiler/src/error.rs
```diff
@@ -24,8 +24,8 @@ pub struct Error {
 
 /// Location within the source file.
 /// Tuples contain:
-/// - line number (1-based),
-/// - column number within that line (1-based),
+/// - line number (0-based),
```
max-sixty, line 27: Should we change this to 1-based? It seems quite non-standard, no?

(as ever — not saying you have to do it)

max-sixty:
Thanks a lot

PR #1314, opened 2022-12-21
feat: compiler signature

### LINKED ISSUE #1154: Add PRQL Version comment to the generated SQL
I was wondering whether it might be useful to have a comment like
```sql
-- Generated by PRQL version 0.2.11 (www.prql-lang.org)
```
at the top (or bottom) of the generated SQL?

Maybe in the Playground as well as in `prqlc` with the option to disable / enable it whatever the default ends up being?

That way when you copy and paste SQL to Github issues or dbt models or wherever, you have a reference point where it came from.

If you have Github issues that remain open for a long time, you may not be able to reproduce the output as it was posted in the issue and this would allow you to identify when the behaviour was as described in the report.

---
I also think it could serve as a form of marketing: if someone comes across such SQL code in the wild and they don't know PRQL then they may be prompted to investigate what PRQL is.

aljazerzen:
Closes #1154


prql-compiler/src/sql/mod.rs
```diff
@@ -30,13 +75,40 @@ pub struct Options {
...
         Self {
             format: true,
             dialect: None,
+            signature_comment: true,
         }
     }
 }
+
+impl Options {
+    pub fn no_format(mut self) -> Self {
```
max-sixty, line 96: I had a look at how bools were generally handled in the builder pattern; one factory crate just uses the field name for them all: https://docs.rs/derive_builder/latest/derive_builder/

It at least makes sense to me to handle bools similarly to others

(this is nitty though, no need to change)

prql-compiler/src/test.rs
```diff
@@ -1,8 +1,16 @@
 //! Simple tests for "this PRQL creates this SQL" go here.
-use super::*;
-use crate::parser::parse;
+// use super::*;
+use crate::{parser::parse, sql};
 use insta::{assert_display_snapshot, assert_snapshot};
 
+pub fn compile(prql: &str) -> Result<String, crate::ErrorMessages> {
```
max-sixty, line 6: FWIW I do find myself asking if the functions are useful for seeing test outputs, surely they're useful as top-level methods in the crate?

Maybe there are defaults for tests that are specific, but the use case of "convert prql to sql with default options" seems fairly general!

max-sixty:
This is breaking java tests. That seems a bit annoying to revert for tests that I don't think have much impact, but let's still revert. I'm happy to have a look and try and fix.

prql-compiler/src/sql/mod.rs
aljazerzen: This is a small digression from the pattern and I would argue it makes more sense here.

prql-compiler/src/test.rs
aljazerzen: You're right. `compile` should also take `Option<Options>`.

Issue #523, opened 2022-05-19
Table definition syntax


max-sixty:
I was just writing the docs for tables — currently we're still using the `=` syntax. Do we want to keep that, or convert to function's new syntax of `->`:

```diff
-table a = (
+table a -> (
    from employees
    take 50
    aggregate [s"count(*)"]
)
from a
```

max-sixty:
@aljazerzen what are your thoughts on this pre-0.2?

aljazerzen:
My opinion is that it makes more sense to have = here.

But i think that eventually, we won't need table keyword because this will be possible to achieve with a function that takes no arguments and returns a table. Then, we will have -> here.

So it would be more consistent to change table syntax to ->...

In short, I dont have a strong opinion and even think that it will not matter in the future.

aljazerzen:
This issue getting pinged reminded me I actually wanted to change the syntax as:
```diff
-table a = (
+let a = (
    from employees
    take 50
    aggregate [s"count(*)"]
)
from a
```

It signals that you could assign any value to the variable, which would be useful in some cases, like declaring constants.

It also leaves doors open for this lambda function syntax:
```elm
let a = (4.7 | round | x -> x + 3)

# these two would be equivalent
let say_hello = x -> f"Hello {x}!"
func say_hello x -> f"Hello {x}!"
```

snth:
I've started a related discussion in #1323 .

max-sixty:
> ```elm
> let say_hello = x -> f"Hello {x}!"
> ```

I'll respond to the fuller discussion later, but there's a reasonable case for having this as the canonical form. Maybe it becomes a bit "functional-y", and clearly demarcating things with `func` and `table` is helpful for new users. But it would be both complete and coherent to have this form, given the `->` syntax

max-sixty:
I recently came across @chris-pikul's go-prql again; he makes an interesting point in https://github.com/chris-pikul/go-prql/blob/main/SYNTAX-NOTES.md#function-as-a-constant, which arguably supports the case for a general `let` syntax

max-sixty:
Circling back on this from Discord. For what I would want as a user, I'd vote for just using `let`:

#### Table
(and potentially other variables in the future

```elm
let x = (
  from employess
  filter [country == "USA]
)
```

#### Function

```elm
let add = a, b -> a + b
```

[...]

PR #1315, opened 2022-12-21
feat: use let for table def

aljazerzen:
Ref #523

### DIFF
--- a/book/src/examples/cte.md
+++ b/book/src/examples/cte.md
@@ -1,11 +1,11 @@
 ```prql
-table newest_employees = (
+let newest_employees = (
   from employees
   sort tenure
   take 50
 )
 
-table average_salaries = (
+let average_salaries = (
   from salaries
   group country (
     aggregate average_country_salary = (average salary)
--- a/book/src/examples/misc.md
+++ b/book/src/examples/misc.md
@@ -9,7 +9,7 @@ some prospect list in a MariaDB of a CRM system.
 ```prql
 # TODO: this table should have a column `part` with values 1..5,
 # but such data declaration is not yet supported, see #286
-table parts = (
+let parts = (
     from seq_1_to_5
 )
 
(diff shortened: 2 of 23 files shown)


max-sixty:
I find this quite reasonable.

One downside is that we only allow tables here — and probably only ever will, since everything can be represented by a table, and having a consistent type there simplifies things.

Will folks try doing:

```elm
let x = 5

from employees
filter id = x
```

...which looks like it should work, but doesn't?

I guess one option would be to force typing, too verbose but makes the point:

```elm
let x<table> = ...
```

---

I also think it's worth considering whether we should align the `func` and `table` definition syntax. With the proposed code, they don't look that similar:

```elm
let foo_table = (from employees)
func add a b -> a + b
```

Arguably the corollary of the proposed code would be to define funcs like:

```elm
let add = func a b -> a + b
# or
let add = a b -> a + b
# or
let add<func> = a b -> a + b
```

---

I recognize those suggestions widen the aperture of options rather than focus us — I don't mean to reopen every discussion about these. But in this case I do think it's worth ensuring the whole language is coherent. What do others think?

aljazerzen:
Actually, signaling that you could define constants was my intention. Currently it (should) throw an error saying it's expecting a table, but this is a feature that does have use-cases.

See the linked issue for my comment about that.

snth:
I agree with @max-sixty 's point that we should try to ensure the language remains coherent.

I'm in the process of writing up a bigger discussion around "Tables, transforms and pipelines" which I'll post in the discussions section. I've been trying to formulate my thought around that for months and I think I actually had a breakthrough insight yesterday (for me).

The points raised about functions and constants open the scope even further so can we hold on this for a bit until we've had time to discuss this in more depth?

aljazerzen:
Agreed. Let's leave this PR open until then.

snth:
Thanks. The discussion is in #1323 .

max-sixty:
Great! I'm a +0.7.

We should update the docs and the grammars too, and do a release, so the docs aren't out of date for something so fundamental. I'm happy to own those.

snth:
As mentioned in #523 , I had a discussion with Aljaz on this and had misunderstood an aspect of this. Very happy with this feature now.

Thanks!

Issue #172, opened 2022-03-11
an exclude clause, that would select all columns except whats specified


qharlie:
Big Query has a `SELECT * EXCEPT`  , but otherwise I don't think it's a widely supported feature.

But it would be a neat feature when and if it's more supported.

One method might be to select into a temp table then alter that table , would take some exploring to see how wildly supported that is.

max-sixty:
+1!

We could use something like https://dba.stackexchange.com/a/1958 for other DBs.

We could represent this as a new transformation:
```elm
from employees
drop salary
```

Or a named arg to `select`
```elm
from employees
select except:[salary] []   # Given current function specs, we'd need to still supply the positional arg
```

danjones1618:
The `drop` clause feels more natural to me however I think we should call it something like `exclude` so it doesn't get confused with an actual SQL `DROP`!

aljazerzen:
I agree that this would be useful, but the problem here is that we cannot (easily) transpile that to SQL. If we have a query like:
```
from example
select [a, b, c, d, e, f]
exclude [d, b]
```
That's easy. No problem.

But
```
from example
exclude [d, b]
```
requires knowledge about what columns there are in the database.

As max wrote, there are queries that allow you to `SELECT * EXCEPT d b`, but they require lookups into `information_schema.columns` or equivalent. We can do that, but we cannot do that reliably for all databases.

aljazerzen:
Wait, no, I found something:

```
CREATE TEMPORARY TABLE TempStudents AS SELECT * FROM Students;
ALTER TABLE TempStudents DROP COLUMN comment;
SELECT * FROM TempStudents; 
```

it [says that it works for MySQL](https://sebhastian.com/mysql-select-all-except/), but it may work for others.

max-sixty:
As discussed on discord:

> I think we could do this, and then just fail if we don't know the column names (unless it's a dialect, like BigQuery, which supports this natively).

> It would be a bit unpredictable — `select *` and `select {list of all columns}` wouldn't be equivalent, but I think it would outweigh the benefits

---

I would hesitate to create a temporary table, I worry it might have quite a bad performance cliff, but we could assess if that's correct.

max-sixty:
One options here would be to do:

```elm
from employees
select [-salary, -dob]
```

...meaning "select the columns apart from `salary` & `dob`. 

Though possibly the `-` should be outside the list, given it needs to apply to everything, since `select [-salary, dob]` wouldn't resolve to anything:

[...]

PR #1317, opened 2022-12-22
feat: Parsing for negative select

max-sixty:
Based on discussion at #172

(this is a very small share of the work obv, though hopefully helpful to see that it's possible from the parser side)


aljazerzen:
When the `select -[foo, bar]` was suggested, I did an experiment if this parsing would be possible. We did exactly the same thing!

058f7c3

Issue #1286, opened 2022-12-17
switch / case / match semantics


max-sixty:
As discussed on Discord, we have a fully functioning `switch` statement from #1278 thanks to @aljazerzen, and there are a couple of options on how this is done:

### Conditions

```elm
switch [
  score <= 10 -> "low",
  score <= 30 -> "medium",
  score <= 70 -> "high",
  true -> "very high",
]
```

- Advantage — anything is representable in a reasonable way
- This is the current approach in `main`

### Compare against a value

```elm
switch score [
  <= 10 -> "low",
  <= 30 -> "medium",
  <= 70 -> "high",
  _ -> "very high",
]
```

- Advantage — visually simpler and more concise (personally I really like the thinking process of "OK we care about `foo`, no need to learn that on each subsequent line")
  - Comparing a single expression to multiple possible values in a very common use of a logic branch like this
- Disadvantages:
  - We probably need to adjust our parser to allow `<= foo` as a valid expression — does everything become a `UnOp`?
  - Representing a full condition is possible but inelegant ("OK we care about `true`"??)
    ```
    switch true [
      score <= 10 -> "low",
      status == "exempt" -> "exempt",
    ]
    ```
    ...since `status == "exempt"` is compared to `true`.

### Allow an optional condition

Allow either of these. This seems quite compelling, and there's a track-record from other languages. But there are some tradeoffs:
- We'd need to use a named parameter
  ```elm
  switch on:score [
    <= 10 -> "low",
    <= 30 -> "medium",
    <= 70 -> "high",
    _ -> "very high",
  ]
  ```
- We still pay the costs of adjusting the parser to alow for `<= 10` as a valid expression.
  - ...or allow for optional positional parameters (which we don't currently do). I'm not entirely opposed to this at some point, though it does make our parameter logic more complicated. Allowing "optional positional parameters with a default of `null`" doesn't complicate the language much, but isn't sufficient here, since the optional parameter would come first, meaning we'd need something like multiple dispatch. Type-checking and autocomplete could become much more difficult.

## Naming

- Most languages call this `switch`, e.g. C / Java / etc
- [Python calls this `match`](https://benhoyt.com/writings/python-pattern-matching/), though also does some level of pattern matching
- [Kotlin calls this `when`](https://kotlinlang.org/docs/control-flow.html#when-expression)

SQL calls this `case`...`when`...`then`, quite the coalition...

## Defaults

[...]

Issue #643, opened 2022-06-23
Should compiler strive to evaulate expressions?


aljazerzen:
We could make compiler smarter and try to compute some of data transformations at the compile time. But it would reduce similarity between input PRQL and output SQL, so I'm not sure if it's worth it.

There is much we can do with compile-time evaluation (we could basically write a turing-complete interpreter), but simple cases start with:
```
derive [5 * 15 - 3]
```
```sql
SELECT 42
```

My use case would be:
```elm
let fuel_consumption = 4.8 # L per 100km
let fuel_price = 1.95 # EUR per L

from trips
derive fuel_cost = distance_km * fuel_consumption / 100 * fuel_price
```
```sql
SELECT distance_km * 0.0936 as fuel_cost FROM trips
```
Also there is [this issue](https://github.com/prql/prql/issues/50).

This would be possible for any values that are known at compile-time; thus mostly constants or function parameters.

Now, I know that the amount of work we could offload with this would be minuscule, because majority of the work happens when iterating over rows - which we cannot compute in advance, since we don't know the contents of the rows. Even further, there would probably be no performance improvement on real databases, because they (probably) already evaluate scalar expressions before applying them to all the rows.

But in compiler, this would be very easy to do - we already have to infer types of binary operations so would only have to:
- check if an operation uses only constant operands (they don't reference any columns or s-strings)
- evaluate the operation

As PRQL functions are already materialized during the compilation, that would also be included.

But the question is not whether we can, but whether we *should* we implement that? Arguably, having output be similar to input is a big help with debugging. If implemented, it must have a `--no-eval` flag to disable it.

The only upside I can think of is that some databases may not support all operations PRQL does, for example date arithmetic:
```elm
let today_started = @2022-06-23T04:36+02
let today_finished = @2022-06-23T11:44+01

from hikes
filter duration_min < (to_min today_finished - today_started)
```
```sql
SELECT * FROM hikes WHERE duration_min < 488 
```

So, as I'm done writing this, I realize that this is quite an unnecessary feature and there is little point investing time into it. Let's just leave it here for future generations.

max-sixty:
I agree with both the question and conclusion — I think we should try and do the minimum possible, and offload everything else to the underlying DB. As you say — it's easier to debug — and we also compile on every keystroke, but only execute the query once, so we want to do as little as possible when compiling.

> The only upside I can think of is that some databases may not support all operations PRQL does, for example date arithmetic

[...]

PR #1324, opened 2022-12-22
feat: static analysis

aljazerzen:
Basic static analysis (i.e. expression evaluation during compilation).

Even though we decided to "offload as much work to database as possible" in #643, it turns out that we will need it for implementation of match and `in` built-in function #1286.


max-sixty:
> it turns out that we will need it for implementation of match and `in` built-in function #1286.

Ah interesting point, that makes sense. I was thinking eventually we'd find ourselves implementing something a full language! (https://wiki.c2.com/?GreenspunsTenthRuleOfProgramming :) )

aljazerzen:
If we are persistent enough, yes :smile: 

The other day I was thinking how I'd extend PRQL so it could replace bash. So yeah, common lisp, here we come.

max-sixty:
Ha! There are actually some interesting similarities to https://www.nushell.sh/ around pipelines, maybe there's some potential integration there.

They even had a kinda custom DB query lang which is no longer (https://github.com/nushell/nushell/pull/7040)

prql-compiler/src/utils/mod.rs
```diff
@@ -85,3 +85,31 @@ impl<T> BreakUp<T> for Vec<T> {
         (self, after)
     }
 }
+
+/// Common operations on iterators over [Option<bool>]
+pub trait IterOfOptBool {
+    /// If any of the items is `None`, this returns `None`.
+    /// If any of the items is `Some(false)`, this returns `Some(false)`.
+    /// Otherwise it returns `Some(true)`.
```
max-sixty, line 93: To what extent is this just `iter.all(|x| matches!(x, Some(true)))`? I guess the proposed code gives us `Some(true)` rather than the bool.

aljazerzen: Your example does not do this:
```
/// If any of the items is `None`, this returns `None`.
```

aljazerzen: This trait is unused now anyway, because I realized I don't need the `is_const` code.

fdncred:
@max-sixty as one of the maintainers of nushell, I've had many thoughts about integrating prql somehow. I think it would be cool.

max-sixty:
@fdncred we're huge fans of nushell! Would it be worth talking? I'm on m at maximilianroos.com — I'd enjoy a conversation even if we don't find an immediate way of collaborating.

fdncred:
@max-sixty Cool! We have an open forum where we talk with collaborators in our [Discord](https://discord.gg/NtAbbGn) in #external-collab. It might be good to start there sometime so others can jump in. I'm one voice out of a half-dozen or so and we're not all online at the same time.

PR #1329, opened 2022-12-23
feat: select exclude

aljazerzen:
Just opening this PR to give advice on how to do this

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,6 +28,13 @@
   union other_employees
   ```
 
+- _Experimental:_ Excluding columns (@aljazerzen, #1329)
+
+  ```prql
+  from employees
+  select ![title, composer]
+  ```
+
 - Numbers can now contain underscores, which can make reading long numbers
   easier (@max-sixty, #1467):
 
--- a/prql-compiler/src/semantic/mod.rs
+++ b/prql-compiler/src/semantic/mod.rs
@@ -80,6 +80,136 @@ mod test {
     }
 
     #[test]
+    fn test_resolve_01() {
+        assert_yaml_snapshot!(parse_and_resolve(r###"
+        from employees
+        select ![foo]
+        "###).unwrap(), @r###"
+        ---
+        def:
+          version: ~
+          other: {}
+        tables:
+          - id: 0
+            name: employees
+            relation:
+              kind:
+                ExternRef:
+                  LocalTable: employees
+              columns:
+                - Single: foo
+                - Wildcard
+        relation:
+          kind:
+            Pipeline:
+              - From:
+                  source: 0
+                  columns:
+                    - - Single: foo
+                      - 0
+                    - - Wildcard
+                      - 1
+                  name: employees
+              - Select:
+                  - 1
+              - Select:
+                  - 1
+          columns:
+            - Wildcard
+        "###)
+    }
+
+    #[test]
+    fn test_resolve_02() {
+        assert_yaml_snapshot!(parse_and_resolve(r###"
+        from foo
+        sort day
+        window range:-4..4 (
+            derive [next_four_days = sum b]
+        )
+        "###).unwrap(), @r###"
+        ---
+        def:
+          version: ~
+          other: {}
+        tables:
+          - id: 0
+            name: foo
+            relation:
+              kind:
+                ExternRef:
+                  LocalTable: foo
+              columns:
+                - Single: day
+                - Single: b
+                - Wildcard
+        relation:
+          kind:
+            Pipeline:
+              - From:
+                  source: 0
+                  columns:
+                    - - Single: day
+                      - 0
+                    - - Single: b
+                      - 1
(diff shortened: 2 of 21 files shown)


aljazerzen:
https://github.com/PRQL/prql/pull/1329/files#diff-d0a5d98a5eeaa51d37d7fbb753111872d244f7586131dda4fcf35c5d1f071426R14-R17
This may cause problems / will have to be handled in quite a few places.

Instead I suggest leveraging this:
https://github.com/prql/prql/blob/0dbb4b5e24e6990e56db0157a8447a2d5f130fe9/prql-compiler/src/ast/rq/mod.rs#L52-L53

If we declare all mentioned column in the ExternTable relation, the wildcard will now mean "all other columns". That combined with other known columns not mentioned in the exclude, will achieve the select we want.

aljazerzen:
For example:

```
from albums          # at this point we know only of [albums.*]
                     # (this wildcard means "there are other columns we don't know about)
filter album_id > 10 # here we know [albums.*, albums.album_id]
select -[title]      # and now we   [albums.*, albums.album_id, albums.title]
```

For this, we'd emit RQ looking like:
```yaml
tables: 
  - id: 0
    name: albums
    columns: [Wildcard, album_id, title]
    relation: ...
relation:
  Pipeline:
  - From:
      source: 0
      columns: [/*wildcard*/1, /*album_id*/2, /*title*/3]
  - Filter:
      ...
  - Select: [1, 2] # excludes title
```

aljazerzen:
Translator will try to exclude the columns, but in this last example, it won't be able to and will just report a warning here:
https://github.com/prql/prql/blob/0dbb4b5e24e6990e56db0157a8447a2d5f130fe9/prql-compiler/src/sql/translator.rs#L492

So this should be extended to either:
- emit EXCLUDE if dialects support it,
- emit an error, saying "need more information, consider adding a select that lists all columns right after from"

max-sixty:
Awesome, thanks a lot for adding the comments — you probably saw that I had pushed this as a quick experiment to see if I could continue moving through the compiler.

[This comment](https://github.com/PRQL/prql/pull/1329#issuecomment-1363850625) is v clear — thanks, I can visualize that working nicely.

prql-compiler/src/semantic/mod.rs
```diff
@@ -79,6 +79,136 @@ mod test {
...
+              - Select:
+                  - 0
+                  - 1
+                  - 2
+                  - 3
+          columns:
+            - Single: day
+            - Single: b
+            - Wildcard
+            - Single: next_four_days
```
max-sixty, line 208: I know this isn't new at all, but another note to say this format for the AST is really great, much easier to read for the uninitiated than it used to be. I had plans to do something like this a few months ago, this is better than I could have done

prql-compiler/src/test.rs
```diff
@@ -2583,3 +2590,111 @@ fn test_hint_missing_args() {
...
+      bytes
+    FROM
+      tracks
+    "###
+    );
+
+    assert_display_snapshot!(compile(r#"
+    from tracks
+    select [track_id, title, composer, bytes]
+    group ![title, composer] (aggregate count)
```
max-sixty, line 2632: Interesting! Cool generality!

prql-compiler/src/test.rs
```diff
@@ -2583,3 +2590,111 @@ fn test_hint_missing_args() {
...
+    );
+
+    assert_display_snapshot!(compile(r#"
+    prql target:sql.snowflake
+    from tracks
+    select ![milliseconds,bytes]
+    "#).unwrap(),
+        @r###"
+    SELECT
+      * EXCLUDE (milliseconds, bytes)
```
max-sixty, line 2682: Really nice, great work!

prql-compiler/src/sql/dialect.rs
```diff
@@ -56,55 +60,74 @@ pub struct MsSqlDialect;
...
 
+pub struct SnowflakeDialect;
+pub struct DuckDbDialect;
+
+pub enum ColumnExclude {
+    Exclude,
+    Except,
+}
 pub trait DialectHandler {
-    fn dialect(&self) -> Dialect;
```
max-sixty, line 60: Cool! I thought we had this for a reason but now can't find it...

CHANGELOG.md
```diff
@@ -28,6 +28,13 @@
   union other_employees
   ```
 
+- _Experimental:_ Excluding columns (@aljazerzen, #1329)
+
+  ```prql
+  from employees
+  select ![title, composer]
+  ```
```
max-sixty, line 36: (it's worth putting the result here as it doesn't compile here, but I can do that when releasing)

prql-compiler/src/sql/dialect.rs
aljazerzen: It was to check if this is bigquery dialect for quoting.

PR #1331, opened 2022-12-23
fix: unused assigns of list arguments

### LINKED ISSUE #1308: Confusing behavior around derive & lists?
While writing https://github.com/prql/prql/issues/648#issuecomment-1360070136, I found a case where I think we should raise an error:

```elm
from employees
derive x = [foo]
```

Currently we add `foo`, but just drop the alias:

```sql
SELECT
  *,
  foo   # No x!
FROM
  employees
```

aljazerzen:
Closes #1308

### DIFF
--- a/prql-compiler/src/test.rs
+++ b/prql-compiler/src/test.rs
@@ -1940,6 +1940,25 @@ fn test_filter_and_select_changed_alias() {
 }
 
 #[test]
+fn test_unused_alias() {
+    // #1308
+    assert_display_snapshot!(compile(r###"
+    from account
+    select n = [account.name]
+    "###).unwrap_err(), @r###"
+    Error:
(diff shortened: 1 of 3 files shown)


prql-compiler/src/test.rs
```diff
@@ -1939,6 +1939,25 @@ fn test_filter_and_select_changed_alias() {
...
+    select n = [account.name]
+    "###).unwrap_err(), @r###"
+    Error:
+    ╭─[:3:12]
+    │
+    3 │     select n = [account.name]
+    ·            ─────────┬────────
+    ·                     ╰────────── unexpected assign to `n`
+    ·
+    · Help: move assign into the list: `[n = ...]`
```
max-sixty, line 1956: That's really great! Both the error message and the test!

PR #1332, opened 2022-12-23
feat: match

### LINKED ISSUE #504: `if` or `case` expressions

### LINKED ISSUE #1286: switch / case / match semantics

aljazerzen:
Closes #1286 and #504

### DIFF
--- a/prql-compiler/src/test.rs
+++ b/prql-compiler/src/test.rs
@@ -2189,6 +2189,32 @@ fn test_switch() {
 
     assert_display_snapshot!(compile(
         r###"
+    from sales
+    select result = match sales.count [
+        eq 1 -> 'one',
+        in 2..5 -> 'a few',
+        in ..10 -> 'many',
+        eq null -> 'null',
+        in .. -> 'out of bounds'
+    ]
+        "###).unwrap(),
+        @r###"
+    SELECT
+      CASE
+        WHEN count = 1 THEN 'one'
+        WHEN count BETWEEN 2
+        AND 5 THEN 'a few'
+        WHEN count <= 10 THEN 'many'
+        WHEN count IS NULL THEN 'null'
+        ELSE 'out of bounds'
+      END AS result
+    FROM
+      sales
+    "###
+    );
+
+    assert_display_snapshot!(compile(
+        r###"
     from employees
     derive display_name = switch [
         nickname != null -> nickname,
--- a/prql-compiler/src/ast/pl/expr.rs
+++ b/prql-compiler/src/ast/pl/expr.rs
@@ -60,6 +60,7 @@ pub enum ExprKind {
     SString(Vec<InterpolateItem>),
     FString(Vec<InterpolateItem>),
     Switch(Vec<SwitchCase>),
+    Match(Box<Expr>, Vec<SwitchCase>),
     BuiltInFunction {
         name: String,
         args: Vec<Expr>,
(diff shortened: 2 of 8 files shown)


prql-compiler/src/test.rs
```diff
@@ -2187,6 +2187,32 @@ fn test_switch() {
     "###
     );
 
+    assert_display_snapshot!(compile(
+        r###"
+    from sales
+    select result = match sales.count [
+        eq 1 -> 'one',
```
max-sixty, line 2194: What's the objection to `==1` or just `1`?

Personally I find having operators-as-words bash-y and not so intuitive nor elegant... But open-minded ofc.

PR #1336, opened 2022-12-25
fix: resolving pipelines

aljazerzen:


prql-compiler/src/semantic/resolver.rs
```diff
@@ -721,15 +752,14 @@ mod test {
     }
 
     #[test]
-    #[ignore]
```
max-sixty, line 724: 👏  (I think this was the final one!)

prql-compiler/src/test.rs
```diff
@@ -2291,3 +2291,26 @@ fn test_static_analysis() {
...
+    assert_display_snapshot!(compile(
+        r###"
+    func addthree<column> a b c -> s"{a} || {b} || {c}"
+    func arg myarg myfunc -> ( myfunc myarg )
+
+    from y
+    select x = (
+        addthree "apples"
+        arg "bananas"
+        arg "citrus"
```
max-sixty, line 2306: It's very impressive it can resolve this.

So I understand, it first gets to `addthree`, and captures the whole expression:

```
        (addthree "apples"
        arg "bananas"
        arg "citrus")
```

Then it gets to arg and captures the next two terms:

```
        (addthree "apples"
        (arg "bananas"
        arg) "citrus")
```

Then it resolves the inner into:

```
        (addthree "apples"
        (arg "bananas")
        "citrus")
```

And then resolves the inner again into:

```
        (addthree "apples"
        ("bananas")
        "citrus")
```

...and then that gives it the result.

---

Is that right? If it is, the part I don't yet understand is why
- it captures the whole expression when evaluating `addthree` — i.e. `(addthree "apples" arg "bananas" arg "citrus")` and not `(addthree "apples" arg "bananas") (arg "citrus")`
- But only two terms when evaluating the first `arg` — i.e. `arg ("bananas" arg) "citrus"`, and not the full expression `arg ("bananas" arg "citrus")`

My guess is I have the logic slightly wrong, and actually it's simpler than that. But somehow `arg "citrus"` is being resolved despite `arg` having two parameters...

prql-compiler/src/ast/pl/fold.rs
```diff
@@ -122,7 +122,7 @@ pub fn fold_stmt_kind<T: ?Sized + AstFold>(fold: &mut T, stmt_kind: StmtKind) ->
     Ok(match stmt_kind {
         FuncDef(func) => FuncDef(fold.fold_func_def(func)?),
         TableDef(table) => TableDef(fold.fold_table(table)?),
-        Pipeline(expr) => Pipeline(Box::new(fold.fold_expr(*expr)?)),
+        Main(expr) => Main(Box::new(fold.fold_expr(*expr)?)),
```
max-sixty, line 125: Makes sense!

prql-compiler/src/semantic/resolver.rs
aljazerzen: It was. But it turns out that I only had to remove ignore - it was fixed sometime before.

max-sixty: Ah, I coincidentally tried last week and it wasn't there quite yet!

prql-compiler/src/test.rs
aljazerzen: This is not quite right. It goes like this:

1. convert pipeline to function calls
   ```elm
   arg "citrus" (arg "bananas" (addthree "apples"))
   ```

2. evaluate inner-most function `addthree`. 
   Because addthree does not get enough args, it gets wrapped into a lambda function.
   ```elm
   arg "citrus" (arg "bananas" <func b c -> addthree "apples" b c>)
   ```

2. evaluate inner-most function `arg`.
   This will apply "bananas" to the lambda function.
   But because the lambda function does not get enough args, it get wrapped into another lambda function.
   ```elm
   arg "citrus" <c -> addthree "apples" "bananas" c>
   ```

2. evaluate inner-most function `arg`.
   now, the lambda function gets enough args, so it can be resolved.
   ```elm
   addthree "apples" "bananas" "citrus"
   ```   

2. evaluate inner-most function `addthree`.
   ```elm
   s"{"apples"} || {"bananas"} || {"citrus"}"
   ```

max-sixty: OK right! I'll go over it and add this to the docs (maybe under internals for the moment.

Issue #1334, opened 2022-12-24
Panic error with invalid take range 


armanm:
Unfortunately I don't know enough Rust to submit an acceptable PR but I spotted this error while poking at prql-compiler.

This query

```
from foo | take 1.8
```

Will throw this error

```
thread '<unnamed>' panicked at 'not implemented: `take` range: 1.8', /Users/arman/.cargo/git/checkouts/prql-c68707e55b45942e/9856b29/prql-compiler/src/semantic/transforms.rs:79:22
```

https://github.com/PRQL/prql/blob/9856b2970ec5db9d4fe2d7b691754ea71740660e/prql-compiler/src/semantic/transforms.rs#L79

I've tested this on the main branch.

aljazerzen:
Thanks for the report.

This is a symptom of a bigger problem I've caused with semantic branch and 0.3 where I was too lazy to gracefully handle many corner cases...

I could go and fix just this one, but let's rather dedicate this issue to eliminating all `.unwrap`s and `unimplemented!`s. Let's also link any future issues with examples that panic to  here, so they can be converted into tests when we decide to fix this.

My priority for sorting this out is low, because many of these call sites will get refactored out in the near future anyway, so it's not worth spending the effort now.

max-sixty:
I'll close this given the immediate issue is solved and there's no immediate path to the general issue.

I started a discussion on Discord on whether there were areas we could have a pipeline of small issues for folks to start on, in the same way that Ruff had been so successful with. So maybe this could be one of them....

PR #1338, opened 2022-12-25
fix: Don't panic on invalid take

max-sixty:


max-sixty:
This fixes the report in https://github.com/PRQL/prql/issues/1334, though possibly we use that issue to manage removing all conditions that can panic...

PR #1347, opened 2022-12-26
style: Add md config

max-sixty:
Was hoping to try adding this as a default linter, but didn't quite get there, seems like there's actually lots of layers here (https://github.com/DavidAnson/markdownlint/issues/689). A very small issue for this repo so stopping early.

### DIFF
--- a/.markdownlint-cli2.yaml
+++ b/.markdownlint-cli2.yaml
@@ -1,0 +1,10 @@
+config:
+  # We use prettier for line length & wrapping
+  MD013: false
+# markdownItPlugins:
+#
+# Doesn't seem to help with some issues — if someone wants to help resolve then
+# great, but it's no huge stress to have some false positives.
+#
+# https://github.com/DavidAnson/markdownlint/issues/689
+#   - ["markdown-it-footnote"]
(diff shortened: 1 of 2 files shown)


eitsupi:
Is there a reason you used `.markdownlint-cli2.yaml` as the name of the config file? (this is not the default name, is it?)

max-sixty:
> (this is not the default name, is it?)

It is! https://github.com/DavidAnson/markdownlint-cli2#command-line

eitsupi:
> > (this is not the default name, is it?)
> 
> It is! [DavidAnson/markdownlint-cli2#command-line](https://github.com/DavidAnson/markdownlint-cli2#command-line)

Thanks for pointing this out, I didn't know that.

Since something like `.markdownlint.yaml` seems to be more common and supported by more tools (e.g. markdownlint-cli), perhaps it would be better to switch to it?
https://github.com/DavidAnson/markdownlint/blob/eb2a3c092223fbce7b01d02399de5f9a3152c667/schema/.markdownlint.yaml

max-sixty:
I tried both markdownlint-cli and markdownlint-cli2, and I remember the latter working better, but:
- I don't have a strong view — if someone tries it again and finds differently, I'm completely fine to switch
- We don't use it by default, because it reports errors that it doesn't fix and are not dangerous, so fails our test for including by default

FWIW markdownlint2 is written by the author of markdownlint, so is no pretender :)

eitsupi:
Thanks, I just thought a more generic configuration file name might be better.
I am using vscode-markdownlint, which seems to support `.markdownlint-cli2.yaml` as well.

PR #1359, opened 2022-12-29
docs: fix alignment issues on homepage

EArazli:
Hi, 

the cards on the homepage had some alignment issues on mobile:

Before:
image
 
After:
image

The problem was that the grid row containing the cards was nested within the grid row containing the title and all that was needed was to un-nest them.

This PR also tries to improve the alignment of the Github star button in the navbar:

Before:
image

After:
image

The Github button fix is a bit hacky and not perfect (it can't be fixed this way as either mobile or desktop would be a tiny bit misaligned, this is because the main styling is provided by the buttons.js script) but I feel like this solution is better than what was there before.

### DIFF
--- a/website/themes/prql-theme/layouts/_default/home.html
+++ b/website/themes/prql-theme/layouts/_default/home.html
@@ -48,23 +48,23 @@
               <div class="section-title">
                 <h2>{{ .title | markdownify }}</h2>
               </div>
-              <div class="row row-cols-1 row-cols-lg-2 justify-content-center">
-                {{ range .items }}
-                  <div class="col g-3">
-                    <div class="card">
-                      <h4>{{ .title | markdownify }}</h4>
-                      {{/* Probably there's a more general way of formatting
-                        this padding; the default is 2em, which looked too indented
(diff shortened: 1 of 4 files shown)


max-sixty:
The star button is much better! I had spent some time trying to get that to look decent — thanks for doing it successfully!

And welcome @EArazli as a contributor!

max-sixty:
@EArazli I don't think this warning matters much, let me know if you disagree

image

EArazli:
@max-sixty I'll look at it again, but in any case, the unsafe code was there before my change, I only changed the nesting of some <div> tags so whatever this is was present beforehand. I'll probably open a PR adding noopener and noreferrer to this code & other occurrences once I read up on it a bit more. But as long as we only link to trusted pages this shouldn't be a security risk, especially because the DOM of prql-lang.org can't really contain any sensitive user information 😅

max-sixty:
> especially because the DOM of prql-lang.org can't really contain any sensitive user information 😅

Yes this was my thought too! :)

PR #1362, opened 2022-12-30
fix: Fix dialect() of MsSqlDialect

boramalper:


aljazerzen:
Good find. How can you even spot something like that?

boramalper:
@aljazerzen I was looking to add a dialect for [Trino](https://trino.io). :)

aljazerzen:
And how's that going? Are there any major incompatibilities?

PR #1365, opened 2022-12-30
fix: tweak website code css

aljazerzen:


max-sixty:
This makes the code blocks look much nicer — thanks!

PR #1366, opened 2022-12-30
docs: static tweets for website

### LINKED ISSUE #1343: Improve Tweet display on homepage
We have tweets on the homepage to signal some social value. (And fwiw we could add some more now there's been more tweeting about PRQL).

I saw an interesting tweet about this, from @mhevery — https://twitter.com/mhevery/status/1606438382561026049. I also noticed that we now get a huge number of cookie warnings on the homepage[^1]. (FWIW I'm fairly confident that it didn't used to be this many; possibly there are some changes on Twitter's backend)

[^1]: image

To the extent we can build the tweet locally, that would reduce the need for JS calls on the homepage. I see some approaches for doing this, some found through the [HN discussion of that tweet](https://news.ycombinator.com/item?id=34127675):
- https://swizec.com/blog/build-privacy-focused-blazing-fast-tweet-embeds-codewithswiz-30/
- https://github.com/brycewray/hugo_site/blob/main/.deprecated/content/posts/2022/06/static-tweets-hugo-update/index.md — hugo (our static site builder), but says it's deprecated
- https://vercel.com/templates/next.js/static-tweets / https://github.com/lfades/static-tweet — next.js

If anyone who is familiar with web & JS wants to take this on, that would be great, and happy to answer any questions!

aljazerzen:
Closes #1343

This replaces embedded tweets with simple HTML + a little CSS to display the tweets. 0 javascript and 0 cookie warnings.

### DIFF
--- a/website/content/_index.md
+++ b/website/content/_index.md
@@ -430,8 +430,7 @@ comments_section:
         created_at: "2022-05-06T13:03:21.000Z"
         favorite_count: 2
         conversation_count: 0
-        text:
-          very excited for prql!
+        text: very excited for prql!
 
     - tweet:
         user_id: "16080017"
@@ -444,8 +443,11 @@ comments_section:
         favorite_count: 20
         conversation_count: 2
         text: >
-          A few years ago, I started working on a language, called "dsql", short for declarative SQL, and a pun on "the sequel (to SQL)". I kinda chickened out of it then, the amount of study and research I needed was massive.
-          prql here is better than I imagined: github.com/max-sixty/prql
+          A few years ago, I started working on a language, called "dsql", short
+          for declarative SQL, and a pun on "the sequel (to SQL)". I kinda
+          chickened out of it then, the amount of study and research I needed
+          was massive. prql here is better than I imagined:
+          github.com/max-sixty/prql
 
     - quote:
         text:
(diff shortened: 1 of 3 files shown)


max-sixty:
I deployed this to see what it would look like (though I see we merged it 👍 )

I think it's decent — less good than the actual tweets, but agree re the JS & warnings & flickering. Specifically, it only links to the tweet in the "Twitter" link on the bottom right  (though weirdly it does seem to in the preview at https://www.tweetic.io/docs), and doesn't make actual links in the tweet (e.g. the infoworld article)

image

Those are small refinements though!

aljazerzen:
I've seen you've deployed it, so I considered this an approved PR :D

I don't use twitter and don't understand what's important in a tweet/what people click, so it does need some refinements and I didn't know if you'd even approve the approach.

It can be reverted, but I do think that it looks and functions better than before.

It looks a bit different than http://tweetic.io, because it's not using their HTML, but a handcrafted one I've committed to our website (it is based on theirs though).

We could make non-twitter quotes look more like twitter ones (link on the bottom, name on the top), but it was looking good enough for me.

max-sixty:
> I've seen you've deployed it, so I considered this an approved PR :D

Ha! You don't need to wait for me regardless! :) 

No no, I think it's good, let's def keep it, I hope it's OK for me to list small refinements without coming across as negative. 

"Incremental progress" etc etc 

(So if anyone sees this and wants to make it link to Twitter by clicking on the card, then great)

PR #1370, opened 2022-12-31
docs: Make s-strings more prominent in the book

max-sixty:
Based on feedback from @viraptor at https://news.ycombinator.com/item?id=34192486


max-sixty:
(also refines some file names, which then creates a bunch of diffs, probably should have been a separate PR)

PR #1377, opened 2022-12-31
docs: fix website link for dialects

allurefx:

PR #1378, opened 2022-12-31
docs: Fix link

max-sixty:
This wasn't caught in our linter since that only tests changed files, and here the target changed. It was caught in our main & nightly tests, which failed yesterday. And @rjnn caught it :clap: !


max-sixty:
Actually already caught by #1377 !

Issue #1171, opened 2022-11-28
Better error message for missing args


aljazerzen:
```elm
from film
select [film_id, lag film_id]
```
... produces:
```
Error: 
   ╭─[:2:18]
   │
 2 │ select [film_id, lag film_id]
   ·                  ─────┬─────  
   ·                       ╰─────── function std.select, param `columns` expected type `column`, but found type `func infer -> column`
───╯
```

We should add a hint here, saying "Have you forgot an argument to function `std.lag`?"

This error is thrown here: https://github.com/prql/prql/blob/9831c244d4bf3331744de9536d5639bbaab361bc/prql-compiler/src/semantic/type_resolver.rs#L73-L78

TODO:
- check if `found_ty` is a `Ty::Function` & check that `expected_ty` is not,
- add hint via calling `Error::with_help(...)`.

Name of the function called can be obtained by destructing `found.kind` and matching it to `ExprKind::Closure`, which contains `Closure::name`. If expr does not match or closure does not specify a name, fallback to a message "Have you forgot an argument in this function call?"

max-sixty:
Good onboarding issue! I can take a swing at this soon if no one gets to it (I wouldn't start before midweek though)

PR #1385, opened 2023-01-01
feat: Better error message for missing args

JettChenT:
Addresses #1171

### DIFF
--- a/prql-compiler/src/semantic/type_resolver.rs
+++ b/prql-compiler/src/semantic/type_resolver.rs
@@ -3,7 +3,7 @@ use std::cmp::Ordering;
 use anyhow::Result;
 
 use crate::ast::pl::*;
-use crate::error::{Error, Reason};
+use crate::error::{Error, Reason, WithErrorInfo};
 
 pub fn resolve_type(node: &Expr) -> Result<Ty> {
     if let Some(ty) = &node.ty {
@@ -91,12 +91,22 @@ where
         Some(Ordering::Equal | Ordering::Greater)
     );
     if !expected_is_above {
-        return Err(Error::new(Reason::Expected {
+        let e = Err(Error::new(Reason::Expected {
             who: who(),
             expected: format!("type `{}`", expected),
             found: format!("type `{}`", found_ty),
         })
         .with_span(found.span));
+        if matches!(found_ty, Ty::Function(_)) && !matches!(expected, Ty::Function(_)){
+            return e.with_help(match &found.kind {
+                ExprKind::Closure(closure) => match &closure.name {
+                    Some(name) => format!("Have you forgot an argument to function `{}`", name.name),
+                    None => "Have you forgot an argument in this function call?".to_string()
+                },
+                _ => "Have you forgot an argument in this function call?".to_string()
+            });
+        };
+        return e;
     }
     Ok(found_ty)
 }
--- a/prql-compiler/src/semantic/type_resolver.rs
+++ b/prql-compiler/src/semantic/type_resolver.rs
@@ -101,11 +101,11 @@ where
             return e.with_help(match &found.kind {
                 ExprKind::Closure(closure) => match &closure.name {
                     Some(name) => {
-                        format!("Have you forgot an argument to function `{}`", name.name)
+                        format!("Have you forgotten an argument to function `{}`", name.name)
                     }
-                    None => "Have you forgot an argument in this function call?".to_string(),
+                    None => "Have you forgotten an argument in this function call?".to_string(),
                 },
-                _ => "Have you forgot an argument in this function call?".to_string(),
+                _ => "Have you forgotten an argument in this function call?".to_string(),
             });
         };
         return e;
--- a/prql-compiler/src/semantic/type_resolver.rs
+++ b/prql-compiler/src/semantic/type_resolver.rs
@@ -101,7 +101,7 @@ where
             return e.with_help(match &found.kind {
                 ExprKind::Closure(closure) => match &closure.name {
                     Some(name) => {
-                        format!("Have you forgotten an argument to function `{}`", name.name)
+                        format!("Have you forgotten an argument to function `{}`?", name.name)
                     }
                     None => "Have you forgotten an argument in this function call?".to_string(),
                 },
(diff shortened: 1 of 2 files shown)


aljazerzen:
Looks good to me.

Can you add a test, similar to [this](https://github.com/prql/prql/blob/main/prql-compiler/src/test.rs#L1923-L1940)?

JettChenT:
Hi! Are there any methods for pasting the output into the test program?

Here's my current code snippet for testing:
```rust
#[test]
fn test_hint_missing_args(){
    assert_display_snapshot!(compile(r###"
    from film
    select [film_id, lag film_id]
    "###).unwrap_err(), @r###"
    Error:
       ╭─[:3:22]
       │
     2 │  select [film_id, lag film_id]
       ·                   ─────┬─────
       ·                        ╰─────── function std.select, param `columns` expected type `column`, but found type `func infer -> column`
       ·
       · Help: Have you forgot an argument to function `lag`
    ───╯
    "###)
}
```

Result of the test:
```
────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
-old snapshot
+new results
────────────┬───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
    0     0 │ Error:␊
    1     1 │    ╭─[:3:22]␊
    2     2 │    │␊
    3       │- 2 │  select [film_id, lag film_id]␊
    4       │-   ·                   ─────┬─────␊
    5       │-   ·                        ╰─────── function std.select, param `columns` expected type `column`, but found type `func infer -> column`␊
          3 │+ 3 │     select [film_id, lag film_id]␊
          4 │+   ·                      ─────┬─────␊
          5 │+   ·                           ╰─────── function std.select, param `columns` expected type `column`, but found type `func infer -> column`␊
    6     6 │    ·␊
    7     7 │    · Help: Have you forgot an argument to function `lag`␊
    8       │-───╯
          8 │+───╯␊
────────────┴───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
```

I have tried directly copy and pasting from terminal, as well as piping the results to an external file.

aljazerzen:
Yes there are. We are using [cargo-insta](https://insta.rs/docs/cli/). In your case, you need:
```
$ cargo install cargo-insta
$ cargo insta review
```

max-sixty:
Thanks @JettChenT ! Great first PR!

Re accepting the snapshots — the lines that @aljazerzen wrote are the closest way. There are also some pre-written tasks if you want to use those, as discussed in https://github.com/PRQL/prql/blob/main/DEVELOPMENT.md

prql-compiler/src/semantic/type_resolver.rs
```diff
@@ -91,12 +91,24 @@ where
...
             who: who(),
             expected: format!("type `{}`", expected),
             found: format!("type `{}`", found_ty),
         })
         .with_span(found.span));
+        if matches!(found_ty, Ty::Function(_)) && !matches!(expected, Ty::Function(_)) {
+            return e.with_help(match &found.kind {
+                ExprKind::Closure(closure) => match &closure.name {
+                    Some(name) => {
+                        format!("Have you forgot an argument to function `{}`", name.name)
```
max-sixty:
```suggestion
                        format!("Have you forgotten an argument to function `{}`", name.name)
```

prql-compiler/src/semantic/type_resolver.rs
```diff
@@ -91,12 +91,24 @@ where
...
-        return Err(Error::new(Reason::Expected {
+        let e = Err(Error::new(Reason::Expected {
             who: who(),
             expected: format!("type `{}`", expected),
             found: format!("type `{}`", found_ty),
         })
         .with_span(found.span));
+        if matches!(found_ty, Ty::Function(_)) && !matches!(expected, Ty::Function(_)) {
+            return e.with_help(match &found.kind {
+                ExprKind::Closure(closure) => match &closure.name {
```
max-sixty, line 102: If you prefer (but no need to change, let's merge regardless), we can encapsulate the initial conditions in a single statement, I think, and then only have two cases:

```suggestion
                ExprKind::Closure(Some(name)) => {
```

(this isn't a full suggestion tbc, we'd need to remove the `Some` term and the `None` condition)

...though possibly the references don't quite work and we need the three conditions.

aljazerzen: I don't think this is possible because `ExprKind::Closure` contains a `Box` which prevents further pattern matching.

max-sixty: Ah thanks @aljazerzen. 

Sorry for the noise @JettChenT ! Ignore my comment. 

(We can revisit on https://doc.rust-lang.org/beta/unstable-book/language-features/box-patterns.html 😄  )

JettChenT:
@aljazerzen @max-sixty Thanks for the help! I have updated the test file and fixed the grammar issue.

prql-compiler/src/semantic/type_resolver.rs
```diff
@@ -101,11 +101,11 @@ where
             return e.with_help(match &found.kind {
                 ExprKind::Closure(closure) => match &closure.name {
                     Some(name) => {
-                        format!("Have you forgot an argument to function `{}`", name.name)
+                        format!("Have you forgotten an argument to function `{}`", name.name)
```
max-sixty:
```suggestion
                        format!("Have you forgotten an argument to function `{}`?", name.name)
```

prql-compiler/src/test.rs
```diff
@@ -2366,3 +2366,21 @@ fn test_errors() {
...
+    select [film_id, lag film_id]
+    "###).unwrap_err(), @r###"
+    Error:
+       ╭─[:3:22]
+       │
+     3 │     select [film_id, lag film_id]
+       ·                      ─────┬─────
+       ·                           ╰─────── function std.select, param `columns` expected type `column`, but found type `func infer -> column`
+       ·
+       · Help: Have you forgotten an argument to function `lag`
```
max-sixty:
```suggestion
       · Help: Have you forgotten an argument to function `lag`?
```

max-sixty:
Thanks a lot @JettChenT ! And welcome to PRQL!

prql-compiler/src/semantic/type_resolver.rs
JettChenT: I originally tried to use the if let guard (https://github.com/rust-lang/rust/issues/51114), but it seems as though the feature is still experimental.

PR #1401, opened 2023-01-02
feat: Use `target` rather than `dialect`

### LINKED ISSUE #1388: `sql_dialect` / `dialect`/ `target`?
Previously our header used `dialect`, like:

```elm
prql dialect:postgres
```

Now we use `sql_dialect`
```elm
prql sql_dialect:postgres
```

I was thinking through the semantics of how something like https://github.com/PRQL/prql/issues/1369 would work. 
It could be `prql python_dialect:xarray`. But then we have two parameters that can't be used together, rather than just one parameter. Should it instead be more general, something like:

```elm
prql target:sql.postgres
```

and then #1369 would use

```elm
prql target:python.xarray
```

While I don't expect we'll have any target apart from SQL soon (#1369 is highly speculative!), 0.4 is a breaking change, so if we want to change it, doing so now would avoid another breaking change later.

max-sixty:
Closes #1388.

Includes a changelog entry!


aljazerzen:
It's not that important now, but within sql module and sql::Options, dialect is more appropriate than target, I think.

max-sixty:
> It's not that important now, but within sql module and sql::Options, dialect is more appropriate than target, I think.

Yes I think you're probably right. I won't do the refactor now in case there are more changes coming there, but happy to step up in the future assuming it's still needed.

PR #1418, opened 2023-01-03
style: Parentheses added.

### LINKED ISSUE #1414: `table` `Display` misses parentheses
This is the output of a query's `.to_string`, from https://github.com/PRQL/prql/pull/1413/files#diff-238f7d8cbd582e19ac0d1da76d1d7ec3985c0204c0140c571967815ba42abec8R6 . 

The s-string needs surrounding parentheses; this wouldn't be valid.

```elm
table grouping = s"
  SELECT SUM(a)
  FROM tbl
  GROUP BY
    GROUPING SETS
    ((b, c, d), (d), (b, d))
"
```

This would be a good first issue — it should be quite reasonable to add parentheses to `Table`'s `Display` trait output.

Rishav1707:
## This PR closes the following issue:
closes #1414 

## Changes made:
- Added `parentheses` to `Table's` `Display` trait output.

### DIFF
--- a/book/tests/snapshots/snapshot__run_display_reference_prql@queries__tables-1.prql.snap
+++ b/book/tests/snapshots/snapshot__run_display_reference_prql@queries__tables-1.prql.snap
@@ -3,13 +3,13 @@ source: book/tests/snapshot.rs
 expression: prql_to_pl(&prql).and_then(pl_to_prql).unwrap()
 input_file: book/tests/prql/queries/tables-1.prql
 ---
-table grouping = s"
+table grouping = (s"
   SELECT SUM(a)
   FROM tbl
   GROUP BY
     GROUPING SETS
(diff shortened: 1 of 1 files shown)


aljazerzen:
First of all, thank you for contribution.

The issue is not descriptive enough. It is about changing PRQL code generation, such that table definitions always emit parenthesis, even if the expression is not a pipeline.

The file you've changed is an auto generated from PL AST that has been parsed from an example in the book, which means that it will get overridden when someone runs `cargo test`.

I'll prepare better descriptions of what is to be done.

max-sixty:
@Rishav1707 this was my fault for not offering a better description. We'd love your contributions in another PR, check out the ones in https://github.com/PRQL/prql/labels/good%20first%20issue and we'd be happy to help you.

I'll close this for now, given #1433

PR #1428, opened 2023-01-04
test: Test explicitly for table reference errors

max-sixty:


max-sixty:
@BlurrechDev this is an example of a small test that now explicitly tests for the error message (maybe we don't strictly need both here though)

PR #1432, opened 2023-01-04
fix: JS readme error and new syntax

### LINKED ISSUE #1367: prql-js syntax suggestion/sugar
It sounds very exciting that you are revamping the outdated SQL language with a pipeline-based approach. Perhaps at the beginning of the project, there is still an opportunity to incorporate modern syntax into the language. I have two suggestions:

First is the
```const sql = compile(`from employees | select first_name`);```
instead of the
```const sql = prql`from employees | select first_name`;```
use, which immediately shows what it is about.

The second is that expressions can be formatted much better with line breaks:
```
const sql = prql`
   from employees
   select first_name
`;
```

### PR DESCRIPTION
**Pull request by BCsabaEngine**
I have found an error in JS readme, fixed it. And decorated readme with some new style.

### DIFF
--- a/prql-js/README.md
+++ b/prql-js/README.md
@@ -26,6 +26,7 @@ function rq_to_sql(rq_json: string): string;
 ### From NodeJS
 
 Direct usage
+
 ```javascript
 const prqljs = require("prql-js");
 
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: prql-js/README.md [hunk 1]
```diff
@@ -25,11 +25,36 @@ function rq_to_sql(rq_json: string): string;
...
-console.log(sql);
+const sql = prqljs.compile(`from employees | select first_name`);
+console.log(sql.sql);
+```
+
+Template literal
+
+```javascript
+const prqljs = require("prql-js");
+const prql = (string) => prqljs.compile(string[0] || "");
```
aljazerzen, line 41: I really like that you've just provided an example for template function instead of trying to squeeze it into our package that generated by wasm-pack anyway.

aljazerzen (approved):
Look good to me. Thank you for making PRQL better!

prql-js/README.md
BCsabaEngine: I did not want to break your code with this decoration. Anyone can use this format in her project if likes it. I wanted to present only the possibilities.

max-sixty:
Thanks @BCsabaEngine !

(If you want to do another PR to add this to Changelog then great, otherwise I'll do it when we do our release)

BCsabaEngine:
> Thanks @BCsabaEngine !
> 
> (If you want to do another PR to add this to Changelog then great, otherwise I'll do it when we do our release)

Ready: https://github.com/PRQL/prql/pull/1437

PR #1444, opened 2023-01-06
feat: Make s-string select req case insensitive

### LINKED ISSUE #1439: S-string table creation case sensitivity
This should work, but doesn't:
```elm
from s"select first_name, id, age from employees order by age asc"
```

...while this does:
```elm
from s"SELECT first_name, id, age from employees order by age asc"
```

```
s-strings representing a table must start with `SELECT `
```

SQL keywords are case-insensitive, so to avoid confusing people, it'd make sense for PRQL to match this behaviour.

Related to me finding this but unrelated to the point of this issue: I was today years old when I realised that the web playground shows certain errors only on hover (and not in the red box at the bottom of the editor)

BlurrechDev:
Attempts to close https://github.com/PRQL/prql/issues/1439

Uses eq_ignore_ascii_case to confirm the prefix is correct.
Adds two tests for "select" and "sElEcT". (We have plenty of "SELECT" coverage already.)
Very new to Rust so if anything I do is contrived please let me know so I can learn 👍

Take 2 with formatting corrected.

### DIFF
--- a/prql-compiler/src/sql/codegen.rs
+++ b/prql-compiler/src/sql/codegen.rs
@@ -293,38 +293,47 @@ pub(super) fn translate_query_sstring(
     context: &mut Context,
 ) -> Result<sql_ast::Query> {
     let string = translate_sstring(items, context)?;
-    if let Some(string) = string.trim().strip_prefix("SELECT ") {
-        Ok(sql_ast::Query {
-            body: Box::new(sql_ast::SetExpr::Select(Box::new(sql_ast::Select {
-                projection: vec![sql_ast::SelectItem::UnnamedExpr(sql_ast::Expr::Identifier(
-                    sql_ast::Ident::new(string),
-                ))],
-                distinct: false,
-                top: None,
-                into: None,
-                from: Vec::new(),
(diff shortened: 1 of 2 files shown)


prql-compiler/src/sql/codegen.rs
```diff
@@ -293,38 +293,47 @@ pub(super) fn translate_query_sstring(
...
-            }))),
-            with: None,
-            order_by: Vec::new(),
-            limit: None,
-            offset: None,
-            fetch: None,
-            locks: vec![],
-        })
+
+    let prefix = if let Some(string) = string.trim().get(0..7) {
```
max-sixty, line 297: Yeah, I think we probably do need this `0..7` here, even though it's a bit awkward. The alternative would be upper case the whole string and feed it into the existing function, but this is more than fine.

max-sixty (approved):
Great! Thanks a lot @BlurrechDev 

If you're up for adding an entry in changelog as a follow-up, then great, otherwise I can add one when I do a release.

(BTW no need to close the previous one and reopen a new one, it's always fine to push more commits. But zero stress!)

PR #1447, opened 2023-01-07
fix: playground now highlights window, switch, concat, union, prql

BlurrechDev:
I have:

- Added window, switch, concat, union and prql to match the book highlighting.
- Split the keyword list up to be more consistent with prql.js and highlight-prql.js
- Left out as/in for now as they've not really landed yet AFAIK.

### DIFF
--- a/playground/src/workbench/prql-syntax.js
+++ b/playground/src/workbench/prql-syntax.js
@@ -14,20 +14,13 @@ const TRANSFORMS = [
 ];
 const BUILTIN_FUNCTIONS = ["switch"]; // "in", "as"
 const KEYWORDS = ["func", "table", "prql"];
-const LITERALS = [
-  "and",
-  "or",
-  "not",
-  "null",
-  "true",
-  "false",
-];
+const LITERALS = ["and", "or", "not", "null", "true", "false"];
 
 const def = {
   // Set defaultToken to invalid to see what you do not tokenize yet
   // defaultToken: 'invalid',
 
-  keywords:  [...TRANSFORMS, ...BUILTIN_FUNCTIONS, ...KEYWORDS, ...LITERALS],
+  keywords: [...TRANSFORMS, ...BUILTIN_FUNCTIONS, ...KEYWORDS, ...LITERALS],
 
   operators: [
     "-",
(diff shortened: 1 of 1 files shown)


playground/src/workbench/prql-syntax.js
```diff
@@ -1,26 +1,26 @@
...
+  "join",
+  "aggregate",
+  "group",
+  "window",
+  "concat",
+  "union",
+];
+const BUILTIN_FUNCTIONS = ["switch"]; // "in", "as"
+const KEYWORDS = ["func", "table", "prql"];
+const LITERALS = ["and", "or", "not", "null", "true", "false"];
```
max-sixty: Technically `and` & `or` are operators rather than Literals

Also `not` is actually not a thing in PRQL, I realize it was there before

If you want to make those changes then great, otherwise fine to merge as-is

max-sixty (approved):
Thanks @BlurrechDev !

playground/src/workbench/prql-syntax.js
BlurrechDev: Good points - updated.

The literal thing was only because highlight-prql.js passes them to that key, but there's no need to copy the hljs fudge over.

playground/src/workbench/prql-syntax.js
```diff
@@ -14,13 +14,14 @@ const TRANSFORMS = [
 ];
 const BUILTIN_FUNCTIONS = ["switch"]; // "in", "as"
 const KEYWORDS = ["func", "table", "prql"];
-const LITERALS = ["and", "or", "not", "null", "true", "false"];
+const LITERALS = ["null", "true", "false"];
+const OPERATORS = ["and", "or"]; // "not"
```
max-sixty: Thanks for changing this. Shall we merge with the `operators` here? https://github.com/PRQL/prql/pull/1447/files/80d7c57262bda9e0d3d6997561e99c6e55111e71..28cd9b487b0fc87d34bf75dd7d36e91247a09b3a#diff-90def8ff0b4c8f49895df5b4685919259d365a97481f30bdda177b2242066cc4R26

BlurrechDev: I just tested this locally and it does not work. There must be some setMonarchTokensProvider related quirk.

EDIT: Scratch that, it does but my test case was wrong.

max-sixty:
Nice improvement, thanks @BlurrechDev, hitting the button

(+ as ever about the changelog...)

PR #1448, opened 2023-01-07
test: Raise on unreferenced snapshots

max-sixty:
Since reverting #1449, this is now repurposed for re-reverting that without breaking windows tests


max-sixty:
So the break is actually that the new test calls `cargo insta test` rather than `cargo test`. 

I'm surprised that can change the test result...

PR #1450, opened 2023-01-07
fix: simplify playground operators highlighting list.

BlurrechDev:
setMonarchTokensProvider refuses to recognise "and"/"or" in the operators array so keep passing them as keywords so we get the desired highlighting behaviour.


max-sixty:
Ah, sorry for being trigger-happy on the merge button! Completely fine to have lots of PRs for something like this, thanks for iterating

PR #1456, opened 2023-01-07
test: Enable force-update in insta

max-sixty:
Expect it to fail on Windows but provides a test case


max-sixty:
> Expect it to fail on Windows but provides a test case

FYI this now passes because of https://github.com/PRQL/prql/pull/1456/files#diff-b73931d39044edca7c78f83713af7326a8d63dac151088b8d905e39f060ed8a6R69-R72, when removed it'll fail again

max-sixty:
> > Expect it to fail on Windows but provides a test case
> 
> FYI this now passes because of https://github.com/PRQL/prql/pull/1456/files#diff-b73931d39044edca7c78f83713af7326a8d63dac151088b8d905e39f060ed8a6R69-R72, when removed it'll fail again

Now upgraded to 1.28, removed this, and it passes!

max-sixty:
I think waiting to resolve https://github.com/mitsuhiko/insta/issues/351 may make sense; otherwise folks may be getting large confusing diffs locally

PR #1458, opened 2023-01-07
docs: Update website syntax

max-sixty:
I think something I did prior broke strings.


max-sixty:
Old: 

image

New:

image

Feedback welcome as ever

PR #1460, opened 2023-01-08
docs: Update contributing sections on website

max-sixty:
This moves the contributing copy to the bottom, re-does the content & spacing.

I think it's important to have Contributing very prominent. But the very first section on the Hero part might be reducing the cleanliness of that. Very open to feedback

### DIFF
--- a/website/config.yml
+++ b/website/config.yml
@@ -32,22 +32,38 @@ menu:
       weight: 5
 
 params:
-  help_the_project:
+  join_us:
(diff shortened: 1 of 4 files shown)


max-sixty:
Old:

image

image


New:

image


image

---

Still not sure the "footer" is the best way of doing this — maybe it should be a section?

We also need a way of keeping this up to date — I haven't been good at that so far...

PR #1464, opened 2023-01-08
docs: Refine Contributing

max-sixty:
Some idiomization and a bit of reorganization.

@aljazerzen I left a comment in there. And feel free to refine / revert anything I changed.


aljazerzen:
> What do you think about moving into the Roadmap "Long term"?

Good idea, it is too much for this page.

> I also think we could suggest that people find a tool and fork it (I know we discussed the tradeoffs)

:+1:

PR #1465, opened 2023-01-08
docs: website touchup

aljazerzen:


website/content/examples.md
```diff
@@ -1,5 +1,6 @@
 ---
 title: "Examples"
+layout: article
```
max-sixty, line 3: This is much nicer now!!

Issue #1451, opened 2023-01-07
Underscores in numbers — `5_000`


max-sixty:
As suggested by @snth on Discord. 

I like this feature a lot in Python, I think it makes sense to add — I don't really see any downside.

Here's the python spec, with references to other langs https://peps.python.org/pep-0515/

I had thought it could only be permitted every three digits (i.e. `5_000` or `50_000`, but not `5_00_000`), but that makes it more difficult to edit numbers, and we can autoformat to every three digits.

If folks agree, I think this could be a good first issue — it would be:
- editing the pest file to allow underscores (but ensuring that a single underscore isn't a number) at https://github.com/PRQL/prql/blob/a138cdaa246adc043b36783cf872477be82046ea/prql-compiler/src/prql.pest#L93 to allow underscores
- removing the underscores in `parser.rs` when parsing to numbers: https://github.com/PRQL/prql/blob/a138cdaa246adc043b36783cf872477be82046ea/prql-compiler/src/parser.rs#L251

A follow-up would be editing the grammar files, but that wouldn't need to be coupled (indeed, could even happen prior if anyone is more comfortable editing those)

snth:
Thanks for creating the issue @max-sixty . I got the idea for it from the recent Hackernews thread: https://news.ycombinator.com/item?id=34266078

One of the comments mentioned that the reason for allowing the underscores in arbitrary positions in numbers is that other countries have different conventions for grouping, e.g. India might group every two digits or something like that.

Therefore I think it might be good to allow underscores in arbitrary positions in numbers.

PR #1467, opened 2023-01-09
feat: Support underscores in numbers

max-sixty:
Completes https://github.com/PRQL/prql/issues/1451

I haven't updated all the various grammars, I think that doesn't need to block (there will be some false errors but no false successes)

The pest is actually kinda difficult; expressing "can't *end* with an underscore".


snth:
Very cool!

I was wondering if for your small example you might want to use `0.000_000_1` rather than `1.000_000_1`? Of course that could also be expressed as `1e-7` so perhaps your example is better.

max-sixty:
That's what I had — but unfortunately `sqlformat-rs` messes up the formatting... (Something we should work out separately)

max-sixty:
I've now updated the grammars — the only one not updated is the textmate grammar in `prql-vscode`

Issue #1405, opened 2023-01-02
Union doesn't seem to work well with `derive`


max-sixty:
(I realize I'm putting more issues in than I'm fixing — the curse of writing docs!)

This returns an error:

```elm
from employees
derive x = 5
union foo
```

```
cannot concat two relations with non-matching number of columns.⏎
```

...which seems overly conservative; we don't know that `foo` doesn't have an `x` column.

FWIW we also don't get a nice error message, just the message above

Issue #1363, opened 2022-12-30
Transform row ordering guarantees


felixge:
Just learned about this project from [HN](https://news.ycombinator.com/item?id=34181319), it looks really great.

While playing with some of the examples, I noticed that the row ordering guarantees of transforms are unclear to me. In particular I don't understand which transforms (if any?) are guaranteed to leave the order of rows unchanged.

For example, does the query below guarantee that the output rows are ordered by salary?

```
from employees
sort salary
take 3
filter salary > 0
```

What about this one?

```
table x = (
  from employees
  sort salary
)
from x
```

From the generated SQL it appears that the answer is NO in both cases (selecting from a CTE has no ordering guarantees in theory). This implies that the following two queries may also produce different row ordering results:

```
from employees
sort salary
filter salary > 0
```

```
from employees
filter salary > 0
sort salary
```

Maybe that's reasonable, but it would certainly feel surprising. Either way, the exact ordering guarantees of transforms should be documented. Apologies if this is already the case and I missed it.

aljazerzen:
Standard definition of database relation specifies "a set of ordered tuples", which is what PRQL also uses.

So all of your queries **should** result in relations ordered by `salary`.

But, compiler (or rather me) doesn't know that selecting from a CTE has no ordering guarantees. It's another SQL thing I'd call bullshit and something that PRQL must to better. And regarding the compiler, this is a bug.

Thanks for pointing this out, this is why opensource works.

aljazerzen:
Can you find any resources talking about this? I cannot find any in postgres docs.

Also, does this apply to subqueries? Will this always produce an ordered relation:
```sql
SELECT * FROM (SELECT * FROM x ORDER BY a) t;
```

felixge:
> Standard definition of database relation specifies "a set of ordered tuples", which is what PRQL also uses.

Which standard are you referring to? SQL? AFAIK Relational algebra didn't have a concept of ordered tuples. Everything was a set or a set of sets with not inherent order. SQL tables by default also don't have an inherent ordering AFAIK.

> Can you find any resources talking about this? I cannot find any in postgres docs.

The best quote I can give you is this part of PostgreSQL's [SELECT](https://www.postgresql.org/docs/current/sql-select.html) documentation:

[...]

Issue #407, opened 2022-04-24
`WITH RECURSIVE` based iteration


AntonioL:
At my shop we do lot of *ELT* where the *T* happens in database.
A typical use case is the following:
- We get JSON files and we insert them inside the database, this is our *E* and *L*
- We write materialized views/queries to process those JSON files, this is our *T*

This has the benefit that everything happens in database, and if there is an error in the business logic we only need to refresh the materialized views. We love materialized views. Our *E* are literally scripts which do a single request and a single INSERT query. And when we need to dump the database, the only content are the JSON files content + definition, so our dumps are very light.

In our context some times we need iteration, for example, we have a feed of events, a single JSON document, we need to process those events accumulating some state, to process the message T[n], we need the state derived from T[1],...,T[n-1]. 
In functional programming languages this is called a `fold`.

In postgres we use `WITH-RECURSIVE` queries, https://www.postgresql.org/docs/14/queries-with.html#QUERIES-WITH-RECURSIVE
```sql
WITH RECURSIVE t(n) AS (
    VALUES (1)
  UNION ALL
    SELECT n+1 FROM t WHERE n < 100
)
SELECT sum(n) FROM t;
```

I can see a world where I would write the body of materialized views in `PRQL`.

aljazerzen:
That's an interesting use of a database. The `WITH-RECURSIVE` could maybe be expressed with our s-string, but in place of pipeline during a table definition:

```
table t = s"VALUES (1) UNION ALL SELECT n+1 FROM t WHERE n < 100"
```
which could be packed into a function for easier use.

This would require #376 and a table flag to add the `RECURSIVE` keyword.

max-sixty:
> This would require #376 and a table flag to add the `RECURSIVE` keyword.

Possibly we could also have an s-string that doesn't bind to anything; it's just:

```sql
s"""
WITH RECURSIVE t(n) AS (
    VALUES (1)
  UNION ALL
    SELECT n+1 FROM t WHERE n < 100
)
"""
```

...and then there's always an escape hatch.

---

I would probably put this feature into the "ensure compatible" bucket, rather than "support natively", but if there were lots of 👍 s on the issue as PRQL becomes more used, then we would work on supporting it in PRQL itself.

AntonioL:
What about adding the concept of a table-valued function which can be marked as explicitly recursive?
In FSharp if a function is recursive is defined with `let rec` 

Example:
```fsharp
[...]

PR #1470, opened 2023-01-09
docs: website article

aljazerzen:


aljazerzen:
pre-commit.ci autofix

max-sixty:
> pre-commit.ci autofix

I don't get why this was required — this should be the default. Possibly because it was failing on a commit from itself, and pre-commit doesn't know whether it might be in a loop, so disables that?

max-sixty:
Excellent! I'll take a pass later.

Would you prefer this is from "the project" or from you? I notice it doesn't have your name at the moment. I would think it's the sort of piece that could come from an individual (rather than some official project documentation), but no strong view. If we have it from the project, we can adjust some of the language from "I" 

(I went through a phase of doing that through PRQL's docs as PRQL became a proper project, lmk if you still see any remnants!)

aljazerzen:
> Would you prefer this is from "the project" or from you?

Actually this can remain as coming from me - a few times I speak in first person and give some personal opinions may not apply to the whole prql team.

website/content/functional-database.md
```diff
@@ -0,0 +1,298 @@
...
+```
+mod default_db {
+    let albums = ...
+    let artists = ...
+    let tracks = ...
+}
+
+mod std {
+    func from tbl -> ...
+    func select tbl -> ...
```
snth: Does this not need a `cols` argument between `select` and `table`?

It's not really relevant to article but it's one of these nits that bothered me while reading it.

website/content/functional-database.md
```diff
@@ -0,0 +1,298 @@
...
+
+## Running real relations
+
+Up to this point, we were talking language design without a clear justification.
+Language being cool is fine, but it is much more important that the language is
+actually the right tool for the job.
+
+The job, in the case of PRQL, is querying databases.
+
+Core unit of data here is a relation: an ordered set of rows, each of which
```
snth: > Core unit of data here is a relation: an ordered set of rows,

Is a relation an *ordered set of rows* or a *set of ordered tuples*?

I don't really know the answer but to me the statement seemed to imply that there was an inherent order in the data whereas my impression has always been that there are no order guarantees in SQL until you introduce an `ORDER BY` clause (cf #1363).

website/content/functional-database.md
```diff
@@ -0,0 +1,298 @@
+---
+title: Functional databases
```
snth: I'm wondering whether you want to reference things like [Functional Database Model](https://en.wikipedia.org/wiki/Functional_database_model) or [The Functional Database - Rich Hickey](https://www.youtube.com/watch?v=tRoVyblAGrs) and [Datomic](https://www.datomic.com/)?

Otherwise **A Functional Query Language** could be an alternative title?

Btw, I recall the Rich Hickey talk being very good but I've completely forgotten all the content now.

snth (commented):
Some comments from me.

As I said on Discord, I really like this and would like to use this as a base to reason out more of the theoretical foundations of PRQL.

website/content/functional-database.md
aljazerzen: Oh, you're right, thanks

website/content/functional-database.md
aljazerzen: Well there is an inherent order, it's just that for tables it's unknown and that all SQL SELECT statements without ORDER BY may change it.

So yes, in practice this is almost the same as relations being unordered, but with that definition, ORDER BY does not make much sense. As expressed in #1363, I'd prefer that PRQL does this better.

Also, I just realized that rows and tuples seem to be another ambiguous pair that is almost the same, but we never really define it, so we all use it a little differently...

website/content/functional-database.md
aljazerzen: Yes, Rich's talk is somewhat related, thanks for linking.

I'm open to changing the title, I'd want it be not only descriptive but also a bit provocative. "A Functional Query Language" may put off people that are fed up with new languages :D

snth: The title is completely up to you of course. I just felt that we're not really changing the functional nature of the database (as opposed to what Rich Hickey is talking about IIRC) but rather how you query or retrieve data.

* A functional paradigm for querying databases
* A functional approach to data queries
* ?

aljazerzen: That's true - it's not the database, but just the query language...

aljazerzen:
pre-commit.ci autofix

website/content/functional-database.md
```diff
@@ -0,0 +1,299 @@
...
+toc: true
+---
+
+I believe that many of the ideas from functional programming are very suitable
+for a query language. To make a case for it, I'll briefly explain a few
+features, adapt their syntax and show how they are used in PRQL.
+
+## Fancy functional features
+
+Functional programming is an old school programming paradigm, that is slowly
```
max-sixty:
```suggestion
Functional programming is an old school programming paradigm that is gradually
```

website/content/functional-database.md
```diff
@@ -0,0 +1,299 @@
...
+gaining traction within a few common languages that are designed with procedural
+paradigm in mind.
+
+The many aspects of paradigm have been discussed extensively, so I'll try to be
+brief. If you've already read then all, I suggest you skip to the next section,
+but if you haven't, I think you'll be glad that I decided to showcase the
+features using pseudo code that looks like JavaScript. I know it has problem,
+but it's something most of us can read.
```
max-sixty: for brevity

```suggestion
brief. If you've already read then all, I suggest you skip to the next section.
```

website/content/functional-database.md
```diff
@@ -0,0 +1,299 @@
...
+
+### Pure functions
+
+This is the core aspect that allows us to do most of the magic later. When we
+say "pure function", we just mean the mathematical function and not "a
+procedure" or "a method". To be more precise:
+
+- Pure functions have no side effects. This means that no global state is
+  altered and nothing is logged or printed. The function result is the only this
```
max-sixty: clarity / extraneous

```suggestion
- Pure functions have no side effects. This means that no global state is
  altered. The function result is the only the
```

website/content/functional-database.md
```diff
@@ -0,0 +1,299 @@
...
+  output that the function produces.
+- Pure functions produce the same output given the same args. The output does
+  not depend on any external information, so no random can be called, no user
+  input requested and no system calls made.
```
max-sixty: clarity

```suggestion
  not depend on any external information; for example the current time or a random number generator
```

website/content/functional-database.md
```diff
@@ -0,0 +1,299 @@
...
+
+```js
+res1 = my_function(arg);
+res2 = res1;
+res3 = res2;
+```
+
+Because `my_function` is guaranteed to return the same result when given `arg`,
+we can skip calling `my_function` for the second and the third time, because we
+(or even better: the compiler) already know what the output will be.
```
max-sixty: brevity
```suggestion
already know what the output will be.
```

website/content/functional-database.md
```diff
@@ -0,0 +1,299 @@
...
+
+This may seem like a minor point, but we just getting started.
```
max-sixty:
```suggestion
This may seem like a minor point, but we're just getting started.
```

website/content/functional-database.md
```diff
@@ -0,0 +1,299 @@
...
+
+### First class citizens
+
+Nearly a buzz word, the saying "functions are first class citizens" means that
```
max-sixty: idioms (though we can replace it with something else if helpful)

```suggestion
Functions being first class citizens" means that
```

website/content/functional-database.md
```diff
@@ -0,0 +1,299 @@
...
+// yields [8, 10, 12]
+```
+
+Here, we've stored function `double` in variable `preprocess` and then passed
+that to method `Array.map`, which applies a given function to each of the array
+elements.
+
+### Currying
+
+Named after Haskell Curry, currying is an implicit act of converting a function
```
max-sixty:
```suggestion
Named after Haskell Curry, currying is an implicitly converts a function
```
idioms

website/content/functional-database.md
```diff
@@ -0,0 +1,299 @@
...
+  return x + y;
+}
+let add_one = add(1);
+```
+
+Because we didn't specify `y` parameter of `add` function, the result is a new
+function that is still waiting for the last argument. It is equivalent to
+defining add_one like this:
+
+function add_one(y) { return add(1, y) }
```
max-sixty:
````suggestion
```js
function add_one(y) { return add(1, y) }
```
````

File: website/content/functional-database.md [hunk 13]
```diff
@@ -0,0 +1,299 @@
...
+Because it is a pure function, it cannot depend on anything other than its
+non-existing arguments. Another way to phrase it, this is a constant expression.
+
+And as such, it might as well be defined as one:
+
+```js
+let my_var = 42;
+```
+
+In conventional programming languages[1], there is a difference between using
```
max-sixty:
```suggestion
In conventional programming languages[^1], there is a difference between using
```

website/content/functional-database.md
```diff
@@ -0,0 +1,299 @@
...
+
+To make this work, let's say that `my_function` is "implicitly invoked" just as
+it would have been expressed as `my_function()`.
+
+Also, if we declare all functions to be pure it doesn't matter exactly _when_
+they are evaluated. So let's give the compiler the authority to make an informed
+guess about _when_ and make `my_var` and `my_function()` semantically
+equivalent.
+
+[1]: Let's say that conventional languages are first 15 from this list:
```
max-sixty:
```suggestion
[^1]: Let's say that conventional languages are first 15 from this list:
```

(unless I'm wrong about markdown and this isn't needed??)

website/content/functional-database.md
```diff
@@ -0,0 +1,299 @@
...
+let my_var = 42;
+```
+
+In conventional programming languages[1], there is a difference between using
+`my_function()` and `my_var`. The first one evaluates the expression at the call
+site, while the second one evaluates it at the declaration site. It is also
+possible to express just `my_function`, a reference to a function that can be
+called without arguments.
+
+But if all your functions are pure, we generalize by saying that all three cases
```
max-sixty:
```suggestion
But if all functions are pure, we generalize by saying that all three cases
```

singular voice

website/content/functional-database.md
```diff
@@ -0,0 +1,299 @@
...
+## Syntax, surly shinier
+
+First of all, lets change the function call to this:
+
+```
+(my_function arg1 arg2 arg3)
+```
+
+It may look strange because function name is within the parenthesis and there is
+no commas between arguments. But trust me, there are benefits to this syntax.
```
max-sixty:
```suggestion
no commas between arguments. But there are some benefits to this syntax...
```

website/content/functional-database.md
```diff
@@ -0,0 +1,299 @@
...
+
+## Syntax, surly shinier
+
+First of all, lets change the function call to this:
+
+```
+(my_function arg1 arg2 arg3)
+```
+
+It may look strange because function name is within the parenthesis and there is
```
max-sixty:
```suggestion
It may look strange because the function name is within the parenthesis and there are
```

website/content/functional-database.md
```diff
@@ -0,0 +1,299 @@
...
+no commas between arguments. But trust me, there are benefits to this syntax.
+
+Firstly, when the call has no arguments, the parenthesis can be omitted:
+
+```
+(my_function) == my_function
+```
+
+which feels very natural because of the similar behavior with expressions. As
+intended, function call with no argument and a plain reference to the function
```
max-sixty:
```suggestion
intended, a function call with no argument and a plain reference to the function
```

website/content/functional-database.md
```diff
@@ -0,0 +1,299 @@
...
+are expressed with the same syntax.
+
+Let's extend this behavior and allow allow bare function calls in a few places
```
max-sixty:
```suggestion
Let's extend this behavior and allow bare function calls in a few places
```

website/content/functional-database.md
```diff
@@ -0,0 +1,299 @@
...
+where they cannot become ambiguous:
```
max-sixty:
```suggestion
where they're unambiguous:
```

website/content/functional-database.md
```diff
@@ -0,0 +1,299 @@
...
+
+```
+# a list:
+[my_function arg1 arg2 arg3, my_function arg4 arg5 arg6]
+# a declaration:
```
max-sixty: I actually like this rather than `assign` maybe we should rename internally...

website/content/functional-database.md
```diff
@@ -0,0 +1,299 @@
...
+
+```
+let curry = my_function arg1 arg2
+let res = curry arg3
+
+# or inline:
+let res = (my_function arg1 arg2) arg3
+```
+
+Now let's go one step further and introduce a "pipe" operator. It applies left
```
max-sixty:
```suggestion
Now let's go one step further and introduce a "pipe" operator. It applies its left
```

website/content/functional-database.md
```diff
@@ -0,0 +1,299 @@
...
+operand as an argument to the right operand:
+
+```
+arg3 | my_function arg1 arg2
+```
+
+This is especially useful for chaining function calls. If we declare `div`,
+`floor` and `mul` as common arithmetic functions, the function call syntax
+really starts to make sense:
```
max-sixty: I don't think `div` and `mul` are great examples — `/` & `*` are simple too! It would be better to compare nested function calls. Indeed PRQL pipelines expressed as nested function calls would really look terrible, whereas as pipelines they're beautiful

website/content/functional-database.md
```diff
@@ -0,0 +1,299 @@
...
+
+```
+12 | div 5 | floor | mul 5
+```
+
+## Running real relations
+
+Up to this point, we were talking language design without a clear justification.
```
max-sixty:
```suggestion
Up to this point, we've discussed language design in the abstract
```

website/content/functional-database.md
```diff
@@ -0,0 +1,299 @@
...
+Language being cool is fine, but it is much more important that the language is
+actually the right tool for the job.
```
max-sixty:
```suggestion
Writing in an elegant language is all good, but we need to show it's the right tool for the job.
```

website/content/functional-database.md
```diff
@@ -0,0 +1,299 @@
...
+
+The job, in the case of PRQL, is querying databases.
+
+Core unit of data here is a relation: an ordered set of rows, each of which
```
max-sixty:
```suggestion
The core unit of data here is a "relation": an ordered set of rows, each of which
```

website/content/functional-database.md
```diff
@@ -0,0 +1,299 @@
...
+contains entries for each of the columns.
```
max-sixty:
```suggestion
contains entries for each of the columns. (More people are likely to be familiar with a "table", which is relation stored in a DB.)
```

website/content/functional-database.md
```diff
@@ -0,0 +1,299 @@
...
+    func sum col -> ...
+    func average col -> ...
+}
+```
+
+The exact structure and naming may change, but the important part is that we
+have global immutable variables that can be referenced from pure functions.
+
+PRQL currently supports most of the features described above, focused around the
+unconventional function call syntax.
```
max-sixty:
```suggestion
innovative function call syntax.
```

WDYT? I don't think it's _that_ unconventional, it's more like we're applying something good to this area...

website/content/functional-database.md
```diff
@@ -0,0 +1,299 @@
...
+The job, in the case of PRQL, is querying databases.
+
+Core unit of data here is a relation: an ordered set of rows, each of which
+contains entries for each of the columns.
+
+Queries operate in a static environment that contains references to database
+tables and a library of standard functions.
+
+```
+mod default_db {
```
max-sixty: I find this a bit confusing, but maybe it's just me.

I think the article is strong because it has the potential to persuade data people that:
a) functional languages are good
b) functional languages can be applied to relational data transformations
c) functional languages can work well with SQL, by operating at the relational level — there's no need to write a whole Scala framework with its own data processing engine...

That's already a lot! So anything else, like `default_db` we could try and minimize? I don't think folks will be confused by where `albums` come from?

website/content/functional-database.md
```diff
@@ -0,0 +1,299 @@
...
+    let albums = ...
+    let artists = ...
+    let tracks = ...
+}
+
+mod std {
+    func from tbl -> ...
+    func select cols tbl -> ...
```
max-sixty: I reckon we should either link to the function docs, or explain that this is a function named `select` which takes columns and a table...

website/content/functional-database.md
```diff
@@ -0,0 +1,299 @@
...
+
+PRQL currently supports most of the features described above, focused around the
+unconventional function call syntax.
+
+A basic operation on a relation would be:
+
+```
+(take 3 default_db.albums)
+# of with a pipeline
+(default_db.albums | take 3 )
```
max-sixty: As above, I think the `default_db` makes it less clear.

One option would be focus just on `take` above (rather than lots of fucns), explain what it does, and then apply it here?

website/content/functional-database.md
```diff
@@ -0,0 +1,299 @@
...
+We focused on the core features and left these out because they can be worked
+around.
+
+Also, PRQL may not ever get all of these features, because ideas in this article
+are only my own and not necessarily of the whole PRQL core team.
+
+#### Many functions in SQL are not pure
+
+And we don't have a plan on
+[how to deal with that](https://github.com/PRQL/prql/issues/1111).
```
max-sixty: I think this would work best as a footnote

website/content/functional-database.md
```diff
@@ -0,0 +1,299 @@
...
+
+#### In math, function call syntax is ambiguous
```
max-sixty: I also this this could be cut for brevity / relegated to a footnote

max-sixty (commented):
Overall I think it's good!

At the moment it's maybe 70/30 on functional programming in general / functional programming applied to relations. 

We obv don't have to do this now, and I might be wrong, but swapping that for 30/70 would add more to what's out there — lots of people have written about functional programming in general, but there's not much out there on how it could apply to relations / SQL.

Great work for writing it — I have nothing but admiration for taking the time and writing something so compelling. (I hope this feedback is perceived as encouraging rather than critical)

website/content/functional-database.md
snth: I'm inclined to agree on the default_db point. It was very interesting for me personally to see how the `from tbl` translates into `default_db.tbl` but it does seem a bit orthogonal to the rest of the argument (from what I recall right now, would have to reread to be sure).

website/content/functional-database.md
aljazerzen: I'm pretty sure I wrote it as `[^1]`... Strange

website/content/functional-database.md
aljazerzen: I'm not sure if I understand correctly, but this is a declaration of a variable that also contains an assign?

website/content/functional-database.md
aljazerzen: I wanted to start with simple functions that don't operate on some complex relation types, but plain numbers.

website/content/functional-database.md
aljazerzen: I was searching for a word and couldn't find anything...

website/content/functional-database.md
aljazerzen: Maybe you are right - it's not the main point of the article and but code block attracts too much attention.

snth:
@aljazerzen I just found this, have a look! [Functional Programming with SQL Query lambdas, functions and pipelines #6
](https://github.com/DavidPratten/sql-fp/discussions/6)

The whole repo seems to be about this: [SQL Functional Programming](https://github.com/DavidPratten/sql-fp)

I haven't done more than glance at it, mostly because I find the SQL really doesn't lend itself to taking a quick look and getting a good idea of what it's about (a great motivation for PRQL), but it seems very similar in aim and spirit (if not syntax) to PRQL and your article in particular.

aljazerzen:
Oh my god - that's exactly what I'm talking about. A lot of useful material! Thanks!

snth:
From that article:

> In SQL-99 every query is, at it's base, hard coded to a global namespace of tables and columns. This means that there is no mechanism to cleanly separate the query transformation from the actual data that is being processed.

For me this is a key point, and from my understanding, still an issue with our current name resolution. See for example @aljazerzen 's [comment here](https://github.com/PRQL/prql/discussions/1323#discussioncomment-4476695) where it is said:
> ... we'd be looking up rel.salary in rel. In cases you provided, no relation had a column rel.salary, so it would be also produce an error.

I believe that creating function scope local table names and then restricting column name resolution to that scope is essential to enabling reusable functions in PRQL.

I also think that this links back to the Monadic structure of our transformation pipelines and the actions of the bind operator, which I unfortunately cannot elucidate properly yet and remains a hunch for me at this stage.

The rest of the **Benefits** section of that article are basically the same motivations that are shared by PRQL.

website/content/functional-database.md
```diff
@@ -180,94 +180,246 @@ let res = curry arg3
...
+# and with a pipeline:
 (from albums | select [title, artist_id])
 ```
 
 All these queries can be simplified to an expression of relations and scalars.
-In PRQL, we call such an expression a Relational Query. It is an intermediate
-representation of prql-compiler and can be translated into SQL to be executed on
-basically any relational database.
+In PRQL, we call such expressions "Relational Queries" or RQ for short. It is an
+intermediate representation of prql-compiler and can be translated to SQL and
```
snth: *intermediate representation of **the** prql-compiler*

website/content/functional-database.md
```diff
@@ -180,94 +180,246 @@ let res = curry arg3
...
+other uses than `aggregate`.
+
+But to separate these core relational operations, we need a way to express
+aggregate as a function. This is why the functional paradigm fits the relational
+data model.
+
+### Distinct
+
+A common question when learning SQL is "how do I select the row where column x
+is min?". It has many variations, but there are two ways of doing it:
```
snth: I would say "where column x is **smallest**" or "where column x is **minimised**".

website/content/functional-database.md
```diff
@@ -180,94 +180,246 @@ let res = curry arg3
...
+operations into orthogonal transforms[^3]. These transforms are in most cases
+pure functions that are easiest to express in a functional language.
+
+<!-- SQL, on the other hand, uses many keywords and syntactic constructs to
+express the most commonly used combinations of these transforms. This falls short
+when trying to express uncommon operations while not providing significantly
+shorter queries. -->
+
+If you want to see more of what PRQL is capable of, come and check out
+[the project](https://github.com/PRQL/prql). It may not have monads (yet), it's
```
snth: "It may not have monads (yet), **but** it's"

website/content/functional-database.md
```diff
@@ -180,94 +180,246 @@ let res = curry arg3
...
+This is the gist of how to express SQL queries with a functional language. At
+this stage a curious reader might ask "can PRQL express any SQL query?" to which
+I'd say: almost. Another might be wondering "but is it less cumbersome and more
+consistent?" and I'd reply "very much, yes", but it depends on who you ask. And
+someone might say "why functional?" to which I say "exactly the question I was
+waiting for!".
+
+There are a few quick benefits I want to get out of the way:
+
+- Pipelines make the flow of the query "top-to-bottom". An earlier part of a the
```
snth: This is a great point!

snth (commented):
Great stuff. Love the new section! Just some typo/phrasing type comments from me.

aljazerzen:
Thank you both for all the feedback. It really was too skewed toward functions and not enough toward relations.

I consider this finished, but as always leave comments even after is merged. We can link to it in the next release.

DavidPratten:
Hi team, nice to know that someone has some value from my reflections on functional programming with SQL!  I don't love SQL syntax either - so for me, PRQL is an exciting development.

It is a bit hard to follow the thread above to understand where your function definition/ reference discussions landed. Could someone sketch out how a reference to ```summarise``` and the definition of it would land in PRQL?

## Usage

```sql
SELECT *
from summarise((SELECT country_id summary_id, state_id id FROM states),
               (SELECT state_id id, population amount FROM population_by_states)
              );
```

## Definition
```sql
CREATE VIEW summarise(summary_id, amount) AS
    GIVEN (mapping(summary_id, id), rawdata(id, amount))
    SELECT mapping.summary_id, sum(rawdatea.amount) amount
    FROM rawdata rawdata
             JOIN mapping mapping ON rawdata.id = mapping.id
    GROUP BY mapping.summary_id;
)
```

I'm keen to catch up with the conversation.

David

snth:
Hi @DavidPratten ,

Welcome and great to have you part of the conversation!

I did promise to raise a new issue for you, apologies for not getting to that yet. We're trying to tie up a couple of loose ends in order to get the 0.4 release out this weekend. There will be actually a number of changes that impact the "table" (aka CTE) and transform definitions which has bearing on what you are interested in.

Our functions actually work amazingly great for the most part and many things that appear to be something different on the surface, e.g. relations and transforms, are actually functions underneath, following the functional programming paradigm.

What I see as an outstanding issue for us, is exactly as you described so well in the piece that I quoted from your article:
> In SQL-99 every query is, at it's base, hard coded to a global namespace of tables and columns. This means that there is no mechanism to cleanly separate the query transformation from the actual data that is being processed.

My understanding is the compiler currently needs to be able to resolve names in the global namespace of tables and columns at function definition time which in some ways leaves us in the same position as SQL. @aljazerzen is really our compiler expert and mage though so he would have to confirm that.

I advise though that we put a temporary hold on this discussion until 0.4 is released so that the public documentation matches the current state of the compiler that the core team are referencing in the discussions. 0.4 should come out soon (this weekend probably) so we can pick things up after that.

I hope you find that acceptable.

P.S. You can read @aljazerzen 's article that this thread is about here: [A functional approach to relational queries](https://github.com/PRQL/prql/blob/main/website/content/functional-relations.md)

snth:
@DavidPratten , since we found each other through Torsten Grust's twitter post about "Making Recursive CTEs more lovable", you may also want to take a look at the current discussion around the proposed Recursive CTE syntax for PRQL and weigh in on that.

That discussion can be found here: https://github.com/PRQL/prql/issues/407#issuecomment-1380921570

aljazerzen:
> It is a bit hard to follow the thread above to understand where your function definition/ reference discussions landed. Could someone sketch out how a reference to summarise and the definition of it would land in PRQL?

Yeah, this thread does not give good context. Your function `summarize` would look like this:

```
func summarise mapping rawdata -> (
    from rawdata = _param.rawdata
    join mapping [==id]
    group mapping.summary_id (
        aggregate [amount = sum rawdata.amount]
    )
)


summarise (
    from states | select [country_id, summary_id, state_id]
) (
    from population_by_states | select [id = state_id, amount = population]
)
```
Don't mind the `_param` param thing, name resolving is a work in progress.

DavidPratten:
Thanks @aljazerzen nice.   A couple of follow-ups to start with.

## Polish notation
```
summarise (
    from states | select [country_id, summary_id, state_id]
) (
    from population_by_states | select [id = state_id, amount = population]
)
```
This looks a bit like Polish Notation where parameters are not grouped by ```(``` ```)``` -  is this a pattern used in other parts of PRQL, or just here?

## Declaring column names and types
In the PRQL example the column names are being drawn from the underlying table in global name space.

To disconnect the function body from the global namespace and have a function that is "write-once use anywhere" the function declaration needs to include column names and types just like any other relation (table) declaration.

In my article, I ignored types, but if they were included the ```GIVEN``` clause might look this:
```
GIVEN (mapping(summary_id varchar, id bigint), rawdata(id bigint, amount float))
```
What is the PRQL equivalent of this?

## On invocation are we matching names or ordinal position?
On invocation of the function then there is a choice, match columns by ordinal position or by name.  e.g. SQL UNION uses ordinal position match. Alternatively, the matching could be on column name.  This would force explicit renaming e.g. ```id = state_id``` of input relations to use the names in the function's column declaration. 

David

aljazerzen:
> Polish notation

This is our function call notation that starts with the function name and continues with arguments separated by whitespace. It's explained [here](https://prql-lang.org/book/internals/functional-lang.html) and [here](https://prql-lang.org/functional-relations/).

----

> Declaring column names and types

At the moment we don't have functioning type annotations, but they are [on the map](https://prql-lang.org/roadmap/).

Functions are meant to be "write once, use anywhere" and this is achieved by expecting the input relation to have a certain structure. Right now, this contract expressed implicitly by just using references to the columns, but the plan is to 1) allow adding type annotation to the functions 2) automatically infer the columns (maybe types) from the function body.

----

> On invocation are we matching names or ordinal position?

Relations in PRQL have columns ordered and optionally named. This is reflected in their type signature. For now, we only had to do matching for `append` (which is UNION ALL) and we haven't explicitly decided on how to do matching exactly. As a first iteration, I implemented matching based on position, but [we will have to revisit](https://github.com/PRQL/prql/issues/1405) that some time soon.

PR #1471, opened 2023-01-09
docs: showcase video for the website

### LINKED ISSUE #675: Video / gif
Do we want some sort of video / gif of typing in the playground? I think that could be very engaging, both for the GH Readme, and particularly on platforms like Reddit which allow for embedding a video without another step.

I don't think there's enough context to do a full voiceover video, but could be persuaded if people think there is.

aljazerzen:
Closes #675

### DIFF
--- a/website/content/_index.md
+++ b/website/content/_index.md
@@ -325,6 +325,12 @@ principles_section:
         and readability. We de-emphasize other SQL features such as inserting
         data or transactions.
 
+videos_section:
+  enable: true
+  title: "Pipelines in action"
+  items:
+  - youtube_id: IQRRsfavEic
+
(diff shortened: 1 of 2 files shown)


max-sixty:
This is really cool!!

Nits: it's quite small at the moment; maybe that's a limitation of YouTube embedding? Maybe we show less text by zooming in if that's immovable. I guess we could do a gif, but maybe that's too large / too poor quality?

Ultra nit: having it centered might look better...

aljazerzen:
The small text would be fixed by re-recording. But I think that video player size can be changed by adjusting `<iframe width height>` attributes.

> centered might look better

I tend to disagree :D

PR #1480, opened 2023-01-10
ci: Try running clippy last in CI

max-sixty:
I doubt this will make a difference but https://www.reillywood.com/blog/rust-faster-ci/ claims it does...


max-sixty:
Seems a bit slower!

Issue #1420, opened 2023-01-03
PRQL code generation & formatting


aljazerzen:
> This is an onboarding issue, see instructions on [how to get started](https://github.com/PRQL/prql/blob/main/CONTRIBUTING.md#compiler).

For debugging purposes and for the `prqlc fmt` command, we have implemented a basic PRQL code generation from PL AST. It is done via `impl std::fmt::Display for X`, which is convenient, but lacks a few key features:
- indentation can only be done ad-hoc, when outputting nested pipelines or lists,
- we have no control over maximum line width,
- we have no control on tab size.

Instead, I suggest we follow implementation of [rustfmt](https://github.com/rust-lang/rustfmt/blob/master/Contributing.md#a-quick-tour-of-rustfmt). I've started work on [codegen branch](https://github.com/PRQL/prql/compare/main...codegen).

- Start by implementing `WriteSource` for `pl::StmtKind`.
- You should be able to reuse majority of [`impl Display for StmtKind`](https://github.com/PRQL/prql/blob/9a024566c11f2fc3658f00d08f87af22d7e54f55/prql-compiler/src/ast/pl/stmt.rs#L94-L125) and adapt it to use string concatenation, similar to existing code in `src/codegen.rs`. Whit this I mean to change things like this:
  ```
  write!(f, "prql")?;
  if let Some(version) = &query.version {
      write!(f, " version:{}", version)?;
  }
  Ok(())
  ```
  ... into:
  ```
  let mut res = String::new();
  res += "prql";
  if let Some(version) = &query.version {
    res += &format!(" version:{}", version);
  }
  Some(res)
  ```

- When you use `write!(f, "{some_var}")` or `format!("{some_var}")`, Rust will call `std::fmt::Display` on `some_var`. We used this to do formatting and this issue is about migrating away from that to our own trait `WriteSource`. So when reusing the code from current impl, it's fine to use write! and format! for now, but the goal is to replace this with calls to `WriteSource::write()`.

- To test it, uncomment `r += &stmt.kind.write(opt.clone()).unwrap();` in `codegen.rs` and `cargo run fmt a_test_file.prql` should be working.

Here are a few test files:
```
prql    version:   "^0.4"    dialect:  sql.postgres

```

```
prql

from       my_table
select   [    some_column
]
```

```
prql    version:   "^0.4"    dialect:  sql.postgres

let my_table   =    (from base_table | select a)

from my_table | select [some_column]
```

[...]

PR #1481, opened 2023-01-10
refactor: switch

aljazerzen:


max-sixty:
Thanks!

When we move https://github.com/PRQL/prql/issues/1420 forward, we can have full test coverage for this, will be good!

PR #1482, opened 2023-01-10
refactor: revert target back to dialect within sql backend, minor picks

aljazerzen:
- refactor: rename sql::Resolver::context
- refactor: revert target back to dialect within sql backend
- error message


max-sixty:
Thanks!

Issue #648, opened 2022-06-24
Inconsistent behavior of a single item in `aggregate`


max-sixty:
Currently this works:

```elm
from employees
aggregate (sum salary)
```

...but not with an assignment...

```elm
from employees
aggregate (x = sum salary)
```

Instead, `[ ]` are required:

```elm
from employees
aggregate [x = sum salary]
```

aljazerzen:
Well, this is syntax level issue: we allow assigns in lists but not in nested pipelines. It may be possible to allow an assign as a first expression in a pipeline?

I can see into that.

max-sixty:
Ah, good point. 

We removed `parenthesized_expr` in favor of `(...)` always being an inline pipeline. I remember thinking that was a good change!

Though it does mean we now have this inconsistency. I'm not sure what the best thing to do is; do we allow for assigns in pipeline — that might make for some odd error messages.

Alternatively we could always ask for a list there. I rather liked the equivalence between a list and a single item; but it's probably worse to have it _partly_ equivalent. WDYT?

aljazerzen:
I don't think this is even inconsistently - lists and parenthesis are not on the same "level".

Lists pass multiple items into functions. These inner items can be parenthesized:
```elm
from employees
derive (sum salary)
derive [(sum salary)]
```

Equivalence for assigns:
```elm
from employees
derive x = (sum salary)

derive [x = (sum salary)]
```

We could add assigns in parenthesis but i'll have to think about the types of such expressions.

max-sixty:
I was originally envisaging that we could pass a single item where we could pass in a list, that those would be interchangeable.

Totally agree that they mean different things at the semantic level — so maybe that's an assumption we should drop...

max-sixty:
I think I should clarify the docs at https://prql-lang.org/book/examples/list-equivalence.html

In particular, we get some confusing error messages if we forget to put parentheses around things that work in lists. 

So the rule is something like:

Expressions need to be one of :
- An item in a list
- A single expression
- In parentheses (e.g. `select (foo + bar)

Here's an example of the currently confusing error message:

```
from employees
filter foo | in 5..10
```

```

max-sixty:
I think I should clarify the docs at https://prql-lang.org/book/examples/list-equivalence.html, and possibly we should think about whether the rules are clear here.

In particular, we get some confusing error messages if we forget to put parentheses around things that work in lists. 

Atm the rules are something like:

---

Assignments need one of these to be true:
[...]

PR #1483, opened 2023-01-11
docs: Try and explain parentheses...

max-sixty:
Helps with https://github.com/PRQL/prql/issues/648

I find this the most confusing comparison:

```elm
# Requires parentheses, because it's a function call
derive total_distance = (sum distance)
derive [
  # Doesn't require parentheses, because it's in a list!
  average_distance = average distance,
]
```

...this is because we can't allow a full function call in an assign outside of a list, because it might be an alias:

```elm
join e=employees [==id]
```

...where we don't want `employees [==id]` to be parsed as a function employees taking an argument `[==id]`

(Unfortunately this creates a huge diff, because it adds a query to the Syntax page, which does n+1 to the filename of all subsequent queries.)

### DIFF
--- a/book/src/queries/syntax.md
+++ b/book/src/queries/syntax.md
@@ -104,10 +104,10 @@ other languages.
 Parentheses are required around:
 
 - Any nested function call containing a pipe, either the `|` symbol or a new
-  line. "nested" means within a transform; i.e. not just the main pipeline.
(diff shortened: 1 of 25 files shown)


max-sixty:
I'm not that happy about this, but I'll merge, and we can iterate.

The biggest issue is the point above; apart from that it's confusing but tolerable.

PR #1484, opened 2023-01-11
fix: wildcard names

aljazerzen:
Followup for except


prql-compiler/src/ast/pl/expr.rs
```diff
@@ -41,6 +45,10 @@ pub struct Expr {
 #[derive(Debug, EnumAsInner, PartialEq, Clone, Serialize, Deserialize, strum::AsRefStr)]
 pub enum ExprKind {
     Ident(Ident),
+    All {
+        within: Ident,
+        except: Vec<Expr>,
+    },
```
max-sixty, line 51: Nice design!

PR #1487, opened 2023-01-12
docs: Add docs for excluding select

max-sixty:


book/tests/snapshots/snapshot__run_display_reference_prql@transforms__select-4.prql.snap
```diff
@@ -0,0 +1,17 @@
+---
+source: book/tests/snapshot.rs
+expression: prql_to_pl(&prql).and_then(pl_to_prql).unwrap()
+input_file: book/tests/prql/transforms/select-4.prql
+---
+prql target:sql.bigquery
+
+
+
```
max-sixty, line 9: FYI looks like it's putting in additional linebreaks here

book/src/transforms/select.md
```diff
@@ -35,6 +37,37 @@ from e=employees
...
 
+### Excluding columns
+
+We can use `!` to exclude a list of columns. This can operate in two ways:
+
+- We use `SELECT * EXCLUDE` or `SELECT * EXCEPT` for the supplied columns in
+  dialects which support it.
+- Otherwise, the columns much have been defined prior in the query (unless all
+  of a table's columns are excluded); for example in another `select` or a
+  `group` transform.
```
aljazerzen:
```suggestion
We can use `!` to exclude a list of columns.

Note that in some cases, the columns will actually be excluded in the output SQL. This is because it is not possible to express such queries in SQL, due to limited support for `SELECT * EXCLUDE/EXCEPT` in many dialects. We can workaround these cases by selecting all of the columns of an input table prior in the pipeline.
```

aljazerzen (approved):

book/src/transforms/select.md
```diff
@@ -35,6 +37,37 @@ from e=employees
 select [e.first_name, e.last_name]
 ```
 
+### Excluding columns
+
+We can use `!` to exclude a list of columns. This can operate in two ways:
+
+- We use `SELECT * EXCLUDE` or `SELECT * EXCEPT` for the supplied columns in
+  dialects which support it.
+- Otherwise, the columns much have been defined prior in the query (unless all
```
BlurrechDev: "much have been defined" is a typo.

max-sixty: Thanks!

book/src/transforms/select.md
max-sixty: Thanks, added a modified version of this to the bullets; feel free to adjust further

PR #1496, opened 2023-01-13
test: Remove unreferenced snapshots

max-sixty:
This was causing tests to fail. I'll look at why the tests passed in the PR


max-sixty:
It passed previously because it was on an outdated version of `main`; this test was old: https://github.com/PRQL/prql/actions/runs/3903018074/jobs/6666814326

PR #1497, opened 2023-01-13
feat: expose PR and RQ in playground, take 2

aljazerzen:


max-sixty:
Thanks!

aljazerzen:
My mistake - i didn't have prql-js built to the latest API changes

max-sixty:
No prob at all. FYI adding a `pr-test-all` label will run all the tests, if you want to avoid the revert.

I'll add an issue to conditionally build the playground on changes to the playground; this is one we should have caught (sometimes we change something in a different place and so we can't catch it, but not the case here)

Issue #1424, opened 2023-01-03
Simplify GHA, move actions into workflows


max-sixty:
GHA now allows for a depth of four workflows, up from a depth of two: https://docs.github.com/en/actions/using-workflows/reusing-workflows#nesting-reusable-workflows

As described in [this footnote](https://github.com/prql/prql/blob/main/DEVELOPMENT.md#user-content-fnref-6-2023e5f0a169d67b6f944f92bb253e5a), there's currently some complexity in our workflows because of the prior limitation of two — we have steps in Actions such as https://github.com/prql/prql/blob/0.3.1/.github/actions/test-rust/action.yaml, because it was a way around that limitation.

So a welcome change would be to:
- Remove copypasta
  - `test-js.yaml` could be called from `test-all.yaml`, rather than copy&pasting the text
  - same for `test-java`
- Move the actions steps back into workflows
  - We could start with `test-python`; there is no need for that to be an action now, and it has an existing workflows which just call the action. (Done, #1554)
  - `test-rust` could become a workflow which gets called from `pull-request.yaml` & `test-all.yaml`
  - We may want to leave the two `build` actions as actions — they produce artifacts we use later in the workflows that call them — and maybe `time-compilation`. It's probably a bit cleaner to have them as workflows, but much more marginal than the others.

While this is less "critical path" than a rust contribution, it would be a good way of contributing to PRQL without writing rust, or if someone is interested in learning more about GitHub workflows.

max-sixty:
Tangentially related — building the playground when that changes. With the proposed design here, that could be its own workflow which the release steps calls. ref https://github.com/PRQL/prql/pull/1497#issuecomment-1382442379

fool1280:
Hi Max, can I take a look at this?

fool1280:
Hi Max, I would like to work on this :)

max-sixty:
Great @fool1280 ! Feel free to ask questions or PR something. Feel free to start with something small; no need to do it all in one PR.

fool1280:
Hi @max-sixty, please correct me if I'm wrong. I'm still a bit stuck on this part: 

> Move the actions steps back into workflows
We could start with test-rust & test-python; there is no need for these to be actions now, and they each have existing workflows which just call the actions.

For test-python, are we moving some actions to test_all or are we moving all of the test_python action into the test_python workflow, then call it? Sorry, it is the first time I worked with GitHub actions

max-sixty:
[...]

PR #1500, opened 2023-01-13
feat: Elixir Bindings for PRQL

kasvith:
# Bring PRQL to Elixir world ⚗️🧙 

- Added Elixir NIF bindings for PRQL 🔥 
- Added unit tests 🧪 
- Added doc tests 📜 
- Docs are WIP 🚧 

TODO:

- [ ] Provide precompiled binaries for release with `RustlerPrecompiled`
- [x] Automated Testing
- [ ] Release on [Hex.pm](https://hex.pm)


prql-elixir/native/prql/src/lib.rs
```diff
@@ -0,0 +1,161 @@
...
+    to_result_tuple(
+        Ok(prql_query)
+            .and_then(prql_compiler::prql_to_pl)
+            .and_then(prql_compiler::pl_to_rq)
+            .and_then(|rq| {
+                prql_compiler::rq_to_sql(rq, Some(options).map(prql_compiler::sql::Options::from))
+            })
+            .map_err(|e| e.composed("", prql_query, false)),
+    )
+}
```
max-sixty: For these functions, would it be possible to instead just defer to the equivalent function upstream?

We'd need to convert the `CompileOptions` struct at the start, and the output at the end, but that would reduce the duplication for most of this, and mean that upstream changes were reflected here immediately.

prql-elixir/native/prql/src/lib.rs
```diff
@@ -0,0 +1,161 @@
...
+            format: o.format,
+            dialect: Some(dialect_from_atom(o.dialect)),
+            signature_comment: o.signature_comment,
+        }
+    }
+}
+
+#[derive(Clone, NifStruct, Debug)]
+#[module = "PRQL.Native.CompileOptions"]
+pub struct CompileOptions {
```
max-sixty, line 78: I guess because this needs to derive `NifStruct`, we need to copy & paste everything from upstream. IIRC we have to do the same in other bindings. 

(Would be nice to have a macro like `make_nif!(CompileOptions)` which would do the copying for us, and include upstream updates...)

prql-elixir/native/prql/src/lib.rs
```diff
@@ -0,0 +1,161 @@
...
+      ansi,
+      big_query,
+      click_house,
+      generic,
+      hive,
+      mssql,
+      mysql,
+      postgres,
+      sql_lite,
+      snow_flake
```
max-sixty: Nit: these are generally one word, so wouldn't have a `_` separator

(though I guess `ClickHouse` has two upper case characters, but `Snowflake` doesn't. Regardless most consistent would be to just use a single word)

prql-elixir/native/prql/Cargo.toml
```diff
@@ -0,0 +1,16 @@
+[package]
```
max-sixty, line 1: Generally we've put the rust crate at the root of `prql-{lang}`. Is that viable here or is it better to be in a subdir?

(no great stress, easy to change later)

prql-elixir/native/prql/Cargo.toml
```diff
@@ -0,0 +1,16 @@
...
+name = "prql"
```
max-sixty, line 2: I know you asked this on discord and I said `prql` was fine. But I _think_: 
- the _rust crate_ should be called `prql-elixir`
- we should reference that from the root `Cargo.toml` so it builds with everything else
- we should also take the `publish` & `version.workspace` from https://github.com/prql/prql/blob/main/prql-js/Cargo.toml#L6-L8

Then to the extent the `elixir` package can be called something different, then we can follow the elixir convention.

But if we can't have different names between languages and `prql-elixir` is inconsistent with convention, then we can reassess.

prql-elixir/native/prql/.gitignore
```diff
@@ -0,0 +1 @@
+/target
```
max-sixty: This is ignored at the root, no need for this

max-sixty (commented):
Great @kasvith ! This already looks fairly close.

I added a few comments. I don't have all the context, so please push back on anything that doesn't seem correct.

As you say above, let's add some tests. We can model that on https://github.com/prql/prql/blob/main/.github/workflows/test-taskfile.yaml (there's no need to make a separate Action like in the other bindings, those are going away soon anyway ref https://github.com/PRQL/prql/issues/1424)

Thank you!

prql-elixir/native/prql/Cargo.toml
kasvith: this has to be a subdir for rustler.

prql-elixir/native/prql/Cargo.toml
kasvith: This rust crate only interface with elixir, generally we use the same naming conversation for the crate and elixir otp_app

kasvith: so essentially prql in this crate means, its referring to otp_app called prql. Thats the elixir package name

max-sixty: OK. I guess this depends on whether the crate should be in the prql workspace in https://github.com/prql/prql/blob/main/prql-js/Cargo.toml#L6-L8, or just independent in the repo.

Having it independent in the repo is functionally the same as having it in a different repo from a packaging POV, I think.

I think either is fine. If it's independent, it's more difficult to test and release them together. But that's no great problem.

Maybe we try leaving it independent — don't add it to the workspace, and we see how that goes? (or if others have views, please comment...)

prql-elixir/native/prql/Cargo.toml
max-sixty: OK

prql-elixir/native/prql/Cargo.toml
kasvith: its independent, I had to turn workspace to empty to run it.

prql-elixir/native/prql/src/lib.rs
kasvith: @max-sixty i didn't get what you meant by `upstream` changes

prql-elixir/native/prql/src/lib.rs
kasvith: is `BigQuery` from google also considered as a single word?

prql-elixir/native/prql/.gitignore
kasvith: Removed

prql-elixir/native/prql/Cargo.toml
kasvith: Sorry I was on mobile all the time for previous replies.

1. This rust crate acts as the NIF for elixir repo, its not related to the workspace project
2. Elixir Rustler handles the building and linking the NIF without the workspace(even works with a remote rust crate)
3. The standard practice is to name your NIF crate with the same name as the OTP app(We can config this, but this crate name makes sense in my opinion). Read More [here](https://hexdocs.pm/rustler/Rustler.html#module-configuration-options)
4. Since Elixir purely handles the building/testing/releasing of this crate, we can make it a part of Elixir package building process which is independent from others

prql-elixir/native/prql/src/lib.rs
kasvith: Yes, Nifs are responsible for handling the binding between Rust and BEAM VM. 
For example, in here we have to derive the `NifStruct` for a struct in Rust to be compatible with [`PRQL.Native.CompileOptions`](https://github.com/kasvith/prql/blob/9b4a17645461f13b61606b694cbc20b0e885528c/prql-elixir/lib/prql/native.ex#L14) struct in elixir.

kasvith:
@max-sixty I've written some tests in elixir, i will add them to the `Taskfile.yml`

prql-elixir/native/prql/src/lib.rs
max-sixty: Sorry, by `upstream`, I mean the code in `prql-compiler`.

So rather than:

```rust
pub fn compile(prql_query: &str, options: CompileOptions) -> NifResult<Response> {
    to_result_tuple(
        Ok(prql_query)
            .and_then(prql_compiler::prql_to_pl)
            .and_then(prql_compiler::pl_to_rq)
            .and_then(|rq| {
                prql_compiler::rq_to_sql(rq, Some(options).map(prql_compiler::sql::Options::from))
            })
            .map_err(|e| e.composed("", prql_query, false)),
    )
}
```

could we do something like

```rust
pub fn compile(prql_query: &str, options: CompileOptions) -> NifResult<Response> {
    let rust_options = rust_options_of_ex_options(options);
    let result = prql_compiler::compile(prql_query, rust_options);
    to_result_tuple(result)
}
```

(this is approximate, but I hope it communicates the point; please let me know if it's unclear!)

That way, if the code like `.and_then(prql_compiler::prql_to_pl)` changes upstream / in `prql-compiler`, this library takes the changes without having to change its code.

prql-elixir/native/prql/src/lib.rs
max-sixty: Ha, I think it's reasonable to have them all as a single word. If I look at examples like https://pypi.org/project/google-cloud-bigquery/ or https://clickhouse.com/docs/en/integrations/language-clients/python/intro/, they're all a single word....

prql-elixir/native/prql/src/lib.rs
kasvith: Got it, i was following the JS example but this makes sense

max-sixty:
> @max-sixty I've written some tests in elixir, i will add them to the `Taskfile.yml`

> As you say above, let's add some tests. We can model that on https://github.com/prql/prql/blob/main/.github/workflows/test-taskfile.yaml (there's no need to make a separate Action like in the other bindings, those are going away soon anyway ref https://github.com/PRQL/prql/issues/1424)

I think maybe my message was a bit unclear. I meant to make a file at `.github/workflows/test-elixir.yaml`, starting from the `test-taskfile.yaml` as an example.

Though we can consider adding those to the actual `Taskfile.yml` too; but the GitHub action will be the one that runs in CI.

prql-elixir/native/prql/src/lib.rs
kasvith: Shall i also use a single word for `ClickHouse`? `:clickhouse` would be elixir representation

kasvith:
> > @max-sixty I've written some tests in elixir, i will add them to the `Taskfile.yml`
> 
> > As you say above, let's add some tests. We can model that on https://github.com/prql/prql/blob/main/.github/workflows/test-taskfile.yaml (there's no need to make a separate Action like in the other bindings, those are going away soon anyway ref #1424)
> 
> I think maybe my message was a bit unclear. I meant to make a file at `.github/workflows/test-elixir.yaml`, starting from the `test-taskfile.yaml` as an example.
> 
> Though we can consider adding those to the actual `Taskfile.yml` too; but the GitHub action will be the one that runs in CI.

Aha, got it. I will proceed with a new workflow

prql-elixir/native/prql/src/lib.rs
max-sixty: Yes, I think for all of them

prql-elixir/native/prql/src/lib.rs
max-sixty: Ah you're right, that's not great. 

Let me add a TODO to that code to change it. Sorry for the misdirection.

prql-elixir/native/prql/Cargo.toml
max-sixty: OK, that's very helpful, thanks.

We lose some cohesion, like integrated versioning and build/testing; but that's not essential. If that becomes inconvenient, we can explore whether having it part of the workspace is even possible.

prql-elixir/native/prql/src/lib.rs
max-sixty: Thanks @kasvith https://github.com/PRQL/prql/pull/1504

prql-elixir/native/prql/Cargo.toml
kasvith: yes, as I can see all other bindings are built using the workspace. But for elixir this is not the case

prql-elixir/native/prql/src/lib.rs
kasvith: Awesome, on it

Taskfile.yml
```diff
@@ -253,6 +254,14 @@ tasks:
       - npm run build
       - npm cit
 
+  test-prql-elixir:
+    # These are mostly copied from the tasks in `test-all.yaml`
+    dir: prql-elixir
+    cmds:
+      - mix deps.get
+      - mix compile
+      - mix test
```
max-sixty, line 263: Re the tests, these are great too! Thanks for adding them. We can add this `task: test-prql-elixir` to this file's `test-all` at some point

kasvith: Do we need some other test cases also?

max-sixty: These work for me locally! Very nice

prql-elixir/native/prql/src/lib.rs
kasvith: @max-sixty shall we continue in discord? i ve some doubts to be cleared out

Taskfile.yml
max-sixty: I'm not sure re other test cases. If you're comfortable with the tests like this to start, then that's fine.

Probably we should add the current status to the Readme — i.e. "It's new, has minimal tests, and doesn't yet publish to Hex". That would let us merge now without writing lots of tests, and then we can add more robustness & publishing in the next PRs?

max-sixty: (but, in case I wasn't clear — we do still need to add the tests to the GitHub Actions path — otherwise they don't run in CI, which is important)

prql-elixir/native/prql/src/lib.rs
kasvith: Fixed

kasvith:
@max-sixty for releasing the package you may need to create an account [here](https://hex.pm/signup)

Taskfile.yml
kasvith: > I'm not sure re other test cases. If you're comfortable with the tests like this to start, then that's fine.

This has some other hidden tests, in elixir we have [doc tests](https://github.com/kasvith/prql/blob/9b4a17645461f13b61606b694cbc20b0e885528c/prql-elixir/lib/prql.ex#L49) which are founded in documentation comments(so we already have 6 tests) 

>(but, in case I wasn't clear — we do still need to add the tests to the GitHub Actions path — otherwise they don't run in CI, which is important)

I think i ve accidentally added it to the `Taskfile.yml`, it seems running the tests. I will add to github workflow also

prql-elixir/native/prql/src/lib.rs
kasvith: Updated

prql-elixir/native/prql/src/lib.rs
```diff
@@ -0,0 +1,155 @@
...
+    /// result string
+    result: String,
+}
+
+#[rustler::nif]
+/// compile a prql query into sql
+pub fn compile(prql_query: &str, options: CompileOptions) -> NifResult<Response> {
+    let opts = prql_compiler::sql::Options::from(options);
+    let result = prql_compiler::compile(prql_query, Some(opts));
+    to_result_tuple(result)
```
max-sixty: FYI now you have the `From` implemented, this will even work (but the current one is also fine):

```suggestion
    let result = prql_compiler::compile(prql_query, Some(options.into()));
    to_result_tuple(result)
```

kasvith: Yeah its working, pushed

prql-elixir/native/prql/src/lib.rs
```diff
@@ -0,0 +1,154 @@
...
+
+#[rustler::nif]
+/// Convert RQ to SQL
+pub fn rq_to_sql(rq_json: &str) -> NifResult<Response> {
+    to_result_tuple(
+        Ok(rq_json)
+            .and_then(prql_compiler::json::to_rq)
+            .and_then(|x| prql_compiler::rq_to_sql(x, None)),
+    )
+}
```
max-sixty, line 149: Can we do the same thing with these functions that we did with `compile`? Maybe they can all be one line each...

prql-elixir/test/prql_test.exs
```diff
@@ -0,0 +1,23 @@
...
+    """
+
+    excepted_result = "SELECT\n  *\nFROM\n  customers"
+
+    assert PRQL.compile(prql_query, @compile_opts) == {:ok, excepted_result}
+  end
+
+  test "return errors on invalid query" do
+    excepted_result =
+      "{\"inner\":[{\"reason\":\"Unknown name invalid\",\"hint\":null,\"span\":{\"start\":0,\"end\":7},\"display\":\"Error: \\n   ╭─[:1:1]\\n   │\\n 1 │ invalid\\n   · ───┬───  \\n   ·    ╰───── Unknown name invalid\\n───╯\\n\",\"location\":{\"start\":[0,0],\"end\":[0,7]}}]}"
```
max-sixty, line 19: To the extent elixir has multiline strings, this would be much easier to read as one! (but no stress)

prql-elixir/native/prql/src/lib.rs
kasvith: Sure, will do

prql-elixir/test/prql_test.exs
kasvith: Sure i will add multi lines

Taskfile.yml
```diff
@@ -253,6 +254,14 @@ tasks:
       - npm run build
       - npm cit
 
+  test-prql-elixir:
```
```suggestion
  test-elixir:
```

max-sixty:
Everything looks great! I added one small comment re the functions.

Can we add a summary of the current state to the Readme? i.e. that it currently requiring cloning this repo and following the instructions in the Rustler docs, including having a rust compiler, but we're planning to add precompilation etc?

Then we can hit the green button!

kasvith:
> Everything looks great! I added one small comment re the functions.
> 
> Can we add a summary of the current state to the Readme? i.e. that it currently requiring cloning this repo and following the instructions in the Rustler docs, including having a rust compiler, but we're planning to add precompilation etc?
> 
> Then we can hit the green button!

Sure, i will add all the required information

prql-elixir/native/prql/src/lib.rs
kasvith: @max-sixty 

```rust
pub fn prql_to_pl(prql_query: &str) -> NifResult<Response> {
    let pl = prql_compiler::prql_to_pl(prql_query);
    let result = prql_compiler::json::from_pl(Some(pl));
    to_result_tuple(result)
    // to_result_tuple(
    //     Ok(prql_query)
    //         .and_then(prql_compiler::prql_to_pl)
    //         .and_then(prql_compiler::json::from_pl),
    // )
}
```

like this?

prql-elixir/test/prql_test.exs
kasvith: Its bit difficult to do with new lines, shall we do it later?

max-sixty: Yes completely fine

prql-elixir/native/prql/src/lib.rs
kasvith: As per discussion in DC lets do this in a later phase

.github/workflows/test-elixir.yaml
```diff
@@ -0,0 +1,93 @@
...
+    working-directory: prql-elixir
+
+env:
+  MIX_ENV: test
+
+jobs:
+  test:
+    strategy:
+      matrix:
+        os: [ubuntu-latest, windows-latest]
```
max-sixty, line 29: No problem if this doesn't work on Mac yet — it's even fine to merge if it only works on Linux at the moment.

But let's put a TODO so that doesn't get lost, and possibly include it in the description in the Readme

kasvith: Actually, it should work on mac, but the action to setup elixir doesn't have Mac support yet

kasvith: Added TODO

max-sixty:
@pre-commit autofix

max-sixty:
pre-commit.ci autofix

kasvith:
@max-sixty can you give a final review :)

prql-elixir/README.md
```diff
@@ -0,0 +1,33 @@
...
+
+
+  iex> PRQL.compile("from customers\ntake 10", dialect: :mssql)
+  {:ok, "SELECT\n  TOP (10) *\nFROM\n  customers\n\n-- Generated by PRQL compiler version 0.3.1 (https://prql-lang.org)\n"}
+```
+
+## Development
+
+We are using `Rustler` to provide rust bindings for `prql-compiler`. So you need
+to install `Rust` toolchain in order to develop this project.
```
max-sixty:
```suggestion
We are in the early stages of developing Elixir bindings. 

We're using `Rustler` to provide rust bindings for `prql-compiler`.

Currently using the bindings in an Elixir project requires compiling the
rust crate from this repo:
```

prql-elixir/README.md
```diff
@@ -0,0 +1,33 @@
...
+
+- Install dependencies with `mix deps.get`
+- Compile project `mix compile`
+- Run tests `mix test`
```
max-sixty, line 37:
```suggestion
- Run tests `mix test`

Future work includes publishing pre-compiled artifacts, so Elixir projects can run PRQL without needing a rust toolchain.
```

prql-elixir/native/prql/README.md
```diff
@@ -0,0 +1,21 @@
...
+
+  # When your NIF is loaded, it will override this function.
+  def add(_a, _b), do: :erlang.nif_error(:nif_not_loaded)
+end
+```
+
+## Examples
+
+[This](https://github.com/rusterlium/NifIo) is a complete example of a NIF
+written in Rust.
```
max-sixty, line 21:
```suggestion
```

max-sixty:
I made a couple of edits to the Readme — check them out

I'll merge and we can continue with the other items in future PRs!

max-sixty:
We can also and an entry to the Changelog in a follow-up PR

kasvith:
Sure, lets add a changelog entry in a follow up pr

max-sixty:
Done! Congratulations on getting this in @kasvith . I'm looking forward to seeing this develop!

kasvith:
Thanks for the support @max-sixty 
This was my first NIF ⚗️ 

Looking forward to working with you.

prql-elixir/native/prql/src/lib.rs
```diff
@@ -0,0 +1,154 @@
...
+        D::Generic
+    } else if a == atoms::hive() {
+        D::Hive
+    } else if a == atoms::mssql() {
+        D::MsSql
+    } else if a == atoms::mysql() {
+        D::MySql
+    } else if a == atoms::postgres() {
+        D::PostgreSql
+    } else if a == atoms::sqllite() {
```
ruslandoga, line 56: 👋

I think it should be `sqlite` :)

https://github.com/PRQL/prql/pull/1628

PR #1506, opened 2023-01-14
fix Cargo.lock to not include two versions of clap

aj-bagwell:


max-sixty:
@aj-bagwell thank you very much! Sorry to take the blame that dependabot deserved and float it anywhere near `clio`...

(I had to make another change as there were _three_ version of `clap`, ref https://github.com/PRQL/prql/pull/1508#issuecomment-1382887860, but now working!)

PR #1509, opened 2023-01-14
refactor: Inherit logic in prql-js

max-sixty:
@kasvith actually the logic in the other functions was as simple as it could be — it was only the compile function that was overly versbose. So there's no work in this specific area in `prql-elixir` either.

### DIFF
--- a/prql-elixir/native/prql/src/lib.rs
+++ b/prql-elixir/native/prql/src/lib.rs
@@ -113,8 +113,7 @@ pub struct Response {
 #[rustler::nif]
 /// compile a prql query into sql
 pub fn compile(prql_query: &str, options: CompileOptions) -> NifResult<Response> {
-    let result = prql_compiler::compile(prql_query, Some(options.into()));
-    to_result_tuple(result)
+    to_result_tuple(prql_compiler::compile(prql_query, Some(options.into())));
 }
 
 #[rustler::nif]
--- a/prql-js/src/lib.rs
+++ b/prql-js/src/lib.rs
@@ -4,21 +4,10 @@ mod utils;
 
 use wasm_bindgen::prelude::*;
 
-// TODO: these functions could be smaller and automatically inherit
-// `prql-compiler` behavior if we replaced this logic with a call directly to
-// the prql-compiler functions, and these functions just did the JS<>Rust
-// conversions. Ref this discussion re `prql-elixir`
-// https://github.com/PRQL/prql/pull/1500/files#r1070213774
-
 #[wasm_bindgen]
 pub fn compile(prql_query: &str, options: Option<SQLCompileOptions>) -> Option<String> {
     return_or_throw(
-        Ok(prql_query)
-            .and_then(prql_compiler::prql_to_pl)
-            .and_then(prql_compiler::pl_to_rq)
-            .and_then(|rq| {
-                prql_compiler::rq_to_sql(rq, options.map(prql_compiler::sql::Options::from))
-            })
+        prql_compiler::compile(prql_query, options.map(|x| x.into()))
             .map_err(|e| e.composed("", prql_query, false)),
     )
 }


prql-elixir/native/prql/src/lib.rs
```diff
@@ -113,8 +113,7 @@ pub struct Response {
 #[rustler::nif]
 /// compile a prql query into sql
 pub fn compile(prql_query: &str, options: CompileOptions) -> NifResult<Response> {
-    let result = prql_compiler::compile(prql_query, Some(options.into()));
-    to_result_tuple(result)
+    to_result_tuple(prql_compiler::compile(prql_query, Some(options.into())));
```
kasvith, line 116: Need to remove `;`

kasvith:
Yeah, i think they were also in the most simple form

prql-elixir/native/prql/src/lib.rs
kasvith: @max-sixty i think build for elixir is failing because of the `;`

kasvith: Should be `to_result_tuple(prql_compiler::compile(prql_query, Some(options.into())))`

max-sixty: Sorry!

kasvith: :)

max-sixty: I'm pushing a fix.

Because this crate isn't part of the workspace, it doesn't get built with the usual tests, so is liable to fail in this sort of error. But:
- we'll see the error after merge, when `test-elixir.yaml` runs
- we can look into making it part of the workspace, on reviewing the docs I think it might be possible (though maybe not a good idea)

kasvith: Im not sure whether we can make it a part of the workspace as elixir bindings are totally different from other bindings we have. 

Can we add `test-elixir.yaml` for the default test suite to avoid this kind of errors?

max-sixty: I'm out of my depth so I won't push on this — I don't even know whether it's overall a good idea.

But this does pass the tests!

```diff
diff --git a/Cargo.toml b/Cargo.toml
index ed29d034..96610687 100644
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -6,6 +6,7 @@ members = [
   "prql-lib",
   "prql-js",
   "prql-python",
+  "prql-elixir/native/prql",
   "book",
 ]
 # Note we don't have a `default-members = ["prql-compiler"]`, since that causes
diff --git a/prql-elixir/native/prql/Cargo.toml b/prql-elixir/native/prql/Cargo.toml
index 0fab6646..a3636472 100644
--- a/prql-elixir/native/prql/Cargo.toml
+++ b/prql-elixir/native/prql/Cargo.toml
@@ -1,10 +1,10 @@
 [package]
 name = "prql"
 description = "Elixir NIF bindings for prql-compiler"
-version = "0.1.0"
 license = "Apache-2.0"
 authors = ["Kasun Vithanage <kasvith.me@gmail.com>"]
 edition = "2018"
+version.workspace = true
 
 [lib]
 name = "prql"
@@ -14,5 +14,3 @@ crate-type = ["cdylib"]
 [dependencies]
 rustler = "0.26.0"
 prql-compiler = {path = "../../../prql-compiler", default-features = false}
-
-[workspace]

```

max-sixty: Re adding `test-elixir.yaml` — if it's very fast we could add it. As a project we really focus on having the tests on the PR very fast, so it would have to be <1.5-2minutes.

kasvith: If we add a single OS and a Single OTP version, we can get builds pretty fast(check ubuntu test)

kasvith: > I'm out of my depth so I won't push on this — I don't even know whether it's overall a good idea.
> 
> But this does pass the tests!
> 
> ```diff
> diff --git a/Cargo.toml b/Cargo.toml
> index ed29d034..96610687 100644
> --- a/Cargo.toml
> +++ b/Cargo.toml
> @@ -6,6 +6,7 @@ members = [
>    "prql-lib",
>    "prql-js",
>    "prql-python",
> +  "prql-elixir/native/prql",
>    "book",
>  ]
>  # Note we don't have a `default-members = ["prql-compiler"]`, since that causes
> diff --git a/prql-elixir/native/prql/Cargo.toml b/prql-elixir/native/prql/Cargo.toml
> index 0fab6646..a3636472 100644
> --- a/prql-elixir/native/prql/Cargo.toml
> +++ b/prql-elixir/native/prql/Cargo.toml
> @@ -1,10 +1,10 @@
>  [package]
>  name = "prql"
>  description = "Elixir NIF bindings for prql-compiler"
> -version = "0.1.0"
>  license = "Apache-2.0"
>  authors = ["Kasun Vithanage <kasvith.me@gmail.com>"]
>  edition = "2018"
> +version.workspace = true
>  
>  [lib]
>  name = "prql"
> @@ -14,5 +14,3 @@ crate-type = ["cdylib"]
>  [dependencies]
>  rustler = "0.26.0"
>  prql-compiler = {path = "../../../prql-compiler", default-features = false}
> -
> -[workspace]
> ```

Does it pass `mix test` also?

max-sixty: It passes `test-elixir`!

max-sixty: > If we add a single OS and a Single OTP version, we can get builds pretty fast(check ubuntu test)

Ubuntu comes in at 1:57. So possibly. Let's run it for a while on `main` and then decide.

kasvith: awesome, i will also double check the NIF build with workspace settings 

Then we can have greenlight

kasvith: @max-sixty green 🟢 from me. I tested locally with workspace settings and everything works fine.
We can use this 

Thanks for the solution

PR #1513, opened 2023-01-15
feat: added alternative exception based functions to call prql methods

kasvith:
Added alternative functions to call the same API with exceptions

Added functions:

- `compile!/2` 
- `prql_to_pl!/1`
- `pl_to_rql!/1`
- `rq_to_sql!/1`

All functions works same as their counterparts, but they raises `PRQL.PRQLError` exception in case of an error


richb-hanover:
@kasvith Would you add a couple sentences to the CHANGELOG that gives a quick overview of how to use these new functions? (Or link to an issue/discussion that describes these?)  That will make rolling out the release much less work when the time comes. Thanks.

kasvith:
@richb-hanover sure, will do

kasvith:
I've also added relevant documentations about using all functions. They will be visible once we publish to hex

max-sixty:
Awesome, thanks @kasvith !

I also see the tests are passing on windows, I guess you solve the line endings issue last time?

kasvith:
Yep i did it with CI in last pr

PR #1514, opened 2023-01-15
fix: Renames from_csv to from_text (#286)

snth:
I'd like to add a `format` parameter as well but that can probably come later.

This should probably also get a mention in the docs and the changelog. I could take a stab at those in maybe 9 hours time.

I also haven't run any tests as my env is not set up. Will making this a PR automatically trigger a CI run?

### DIFF
--- a/prql-compiler/src/semantic/std.prql
+++ b/prql-compiler/src/semantic/std.prql
@@ -58,4 +58,4 @@ func _eq<list> a<list> -> null
 func _is_null a -> a == null
 
 # Misc
-func from_csv<table> csv<string> -> null
+func from_text<table> text<string> -> null
--- a/prql-compiler/src/semantic/transforms.rs
+++ b/prql-compiler/src/semantic/transforms.rs
@@ -295,20 +295,20 @@ pub fn cast_transform(resolver: &mut Resolver, closure: Closure) -> Result<Resul
             return Ok(Ok(res));
         }
 
-        "std.from_csv" => {
+        "std.from_text" => {
             // yes, this is not a transform, but this is the most appropriate place for it
 
-            let [csv_expr] = unpack::<1>(closure);
+            let [text_expr] = unpack::<1>(closure);
 
-            let ExprKind::Literal(Literal::String(csv)) = csv_expr.kind else {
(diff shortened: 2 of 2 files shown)


aljazerzen (approved):

max-sixty:
Were we also going to have a named-arg for `csv`? 

We could merge this, have `csv` be the default, and then add in a named arg if we ever add json?

max-sixty:
I won't put this in the release notes yet given the open question, we can do 0.4.1 in a few days

aljazerzen:
This function is for parsing string *literals* - I'm not sure why anyone would prefer writing json instead of csv here...

snth:
Not writing, but pasting. I was thinking that you might get some JSON data back from some API and you might quickly just want to run a query over it. It would be convenient if you could just paste it (maybe even in the playground) and then test your query, rather than having to save the data to a file and then run it against that.

aljazerzen:
I see.

Do you think this signature is ok:
```
from_text csv:"a,b,c\n1,2,3\n4,5,6"
from_text json:'[{"a": 1, "b": 2, "c": 3, }, {"a": 4, "b": 5, "c": 6 }]'
from_text json:'{ "columns": ["a", "b", "c"], "data": [[1, 2, 3], [4, 5, 6]] }'
```

snth:
Interesting. In #286 we discussed having the signature as

```elm
from_text format:csv "a,b,c\n1,2,3\n4,5,6"
from_text format:json '[{"a": 1, "b": 2, "c": 3, }, {"a": 4, "b": 5, "c": 6 }]'
```

but your new suggestion is shorter.

Having different formats as different named parameters almost makes it a kind of enum and might make autodiscovery simpler and improve errors.

I think I might quite like that. Let me think on it a bit. 

Any pros and cons for either approach that you could see on the code side?

max-sixty:
Interesting! I was also thinking about it like `from_text format:csv`.

I'm not strongly opposed to the `csv:` approach. It's less consistent with the language, though we're at an early stage to start weighing that highly.

For example, we have `join [foo, bar] side:left`, not `join left:[foo, bar]`

I'm +0.5 for the `format:` approach, but as ever happy to be see it go either way, and the do-er decides...

snth:
Recent thought on this: to make the API uniform across `from_text` and `from_file`, the `format:csv` will be better since the data won't come as a string.

Therefore I'm now +1 on `format:`.

aljazerzen:
It seems that we have a consensus and this can be merged.

PR #1515, opened 2023-01-15
build: prql-compiler examples, docs, prqlc & macros

### LINKED ISSUE #1106: `prql-compiler` name?
I remember this was discussed a while ago, around possibly renaming to `prqlc`. I don't have a view but wanted to ensure I wasn't dropping something others were expecting. 

The others I'm familiar with are `rustc` and `ocamlc` (but also `ocamlopt`), though go is just `go` IIUC. Any others?

I can understand not just calling it `prql` given it is more "core" and probably not going to be called directly by many people.

We should ensure the brew references reflect the correct name, whether that's `prqlc` or `prql-compiler`. https://github.com/prql/prql/issues/866

aljazerzen:
Closes #1106


prql-compiler/src/lib.rs
```diff
@@ -31,6 +31,44 @@
...
+//!     Call this crate from `build.rs`. See
+//!     [this example project](https://github.com/PRQL/prql/tree/main/prql-compiler/examples/compile-files).
+//!
+//! - ... compile, format & debug PRQL from command line.
+//!
+//!     ```sh
+//!     $ cargo install ripgrep
+//!     $ prqlc compile query.prql
+//!     ```
+//!
```
max-sixty, line 71: Nice docs!

prql-compiler/examples/compile-files/.gitignore
```diff
@@ -0,0 +1,2 @@
+/target
```
max-sixty, line 1: FYI this is ignored at root

max-sixty:
FYI running `cargo run -p prql-compiler` doesn't work any longer, since `cli` is a required feature not enabled by default. Maybe that's better though...

prql-compiler/Cargo.toml
```diff
@@ -39,10 +36,6 @@ sqlformat = "0.2.0"
 sqlparser = {version = "0.30.0", features = ["serde"]}
 strum = {version = "0.24.0", features = ["std", "derive"]}# for converting enum variants to string
 strum_macros = "0.24.0"
-
-# These bring in `errno`, which fails under wasm. We disable the cli in wasm
-# anyway, so they're not required.
-[target.'cfg(not(target_family="wasm"))'.dependencies]
```
max-sixty, line 45: I did so much yak shaving for this a month or so ago; seems like they changed the dependency tree so it's no longer a problem. Thanks for removing.

max-sixty:
I'm unclear why cargo doesn't seem to find the example — any ideas? I'm getting `No examples available.`...

```
❯ cargo run --example
error: "--example" takes one argument.
No examples available.
```

max-sixty:
> I'm unclear why cargo doesn't seem to find the example — any ideas? I'm getting `No examples available.`...

Slight adjustment in #1515

prql-compiler/Cargo.toml
max-sixty: I now get why it could be removed — because `prqlc` wasn't being compiled with the rest of the workspace...

Issue #1498, opened 2023-01-13
Can I use file paths as table names for DuckDB?


max-sixty:
### Discussed in https://github.com/PRQL/prql/discussions/1493

<div type='discussions-op-text'>

<sup>Originally posted by **eitsupi** January 13, 2023</sup>
DuckDB can accept the path to a Parquet file as the table name.
For example, `FROM 'test/*.parquet'`.
Is there any way to use this in PRQL?
https://duckdb.org/docs/data/parquet.html

It seems that s-string cannot be used for this purpose.</div>

I think this should™ be possible with:

```elm
from `test/*.parquet`
```

but unfortunately this returns:

```sql
Error:
   ╭─[:1:6]
   │
 1 │ from `test/*.parquet`
   ·      ────────┬───────
   ·              ╰───────── Unsupported feature: advanced wildcard column matching
───╯
```

Maybe we need to give it a namespace, and so we should require:

```elm
from X=`test/*.parquet`
```

(currently this fails with the same error)

---

And it should also be possible with an s-string:

```elm
from s"test/*.parquet"
```

though we currently require `SELECT` here:

```
s-strings representing a table must start with `SELECT `
```

Instead, in `from`, we can take a table name without a `SELECT`?

> Probably this needs someone who really cares about s-strings to do the work here. (me?)

---

For those who want to use this in the meantime, currently this works:

```elm
from s"select * from 'test/*.parquet'"
```

max-sixty:
#1516 would (kinda) solve this, and contains the broader tradeoff + suggestion for a broader change

max-sixty:
Closed by https://github.com/PRQL/prql/pull/1516, follow https://github.com/PRQL/prql/issues/1535 for the broader issue

Issue #822, opened 2022-07-18
Schema name in Postgres `from` is incorrectly quoted


mklopets:
This behaviour was introduced in https://github.com/prql/prql/pull/809

Assume you have a Postgres database with a schema `some_schema` (this could even be `public`)
```
prql dialect:postgres
from some_schema.tablename
```

compiles to
```
SELECT
  some_schema.tablename.*
FROM
  "some_schema.tablename"
```

which is an invalid query – the schema name shouldn't be inside quotes (postgres will complain with `relation "some_schema.tablename" does not exist`.

If we want to keep the quotes, a valid Postgres query would be:

```diff
-  "some_schema.tablename"
+  "some_schema"."tablename"
```

or just

```diff
-  "some_schema.tablename"
+  some_schema.tablename
```

max-sixty:
Ah, this is unfortunate, #809 was quite elegant. We can make it split by `.` I think.

Does this block you? We can revert #809 if it does (we have a [policy](https://github.com/prql/prql/blob/main/CONTRIBUTING.md) of fast merges and fast reverts!)

mklopets:
it's non-blocking until prql-js has a new version we need to update to for other reasons (the latest published version, `v0.2.2` is fine)

splitting by `.` would be neat, though I'm not sure which other dialects this applies to

max-sixty:
OK. Probably we should revert rather than block a release, but let me see whether it's an easy fix

blackary:
Similar issue with BigQuery

```
prql dialect:bigquery
from db.schema.table
``` 
-> 

```sql
SELECT
  db.schema.table.*
FROM
  `db.schema.table`
 ```
 
 Which causes this error:
 `Unrecognized name: db at [2:3]`
 
 In this case the backticks after the FROM are fine, but they are missing after the SELECT
 
 ```diff
 -  db.schema.table.*
+  `db.schma.table`.*
```

huw:
This is also breaking Spark SQL (not a supported dialect, I know)—in this case, the `FROM` clause needs a table name of either `some_schema.tablename` or 
```
`some_schema`.`tablename`
```
(Seems like it would be ideal to split on `.` in all dialects, regardless of quote marks).

max-sixty:
Closed by #838. Thanks for everyone's patience

PR #1516, opened 2023-01-15
feat: Parse idents with `*` as quoted

max-sixty:
I'm not sure this is a good idea. It solves https://github.com/PRQL/prql/issues/1498 is a hacky way.

The tradeoff is that:

```elm
from `schema.table`
```

...needs to compile to 

```sql
SELECT * FROM 'schema'.'table' -- (or just `schema.table`, no backticks, but *not* 'schema.table' with quotes)
```

But:

```elm
from `dir/*.parquet`
```

...needs to compile to

```sql
SELECT * FROM 'dir/*.parquet' -- not 'dir/*'.'parquet'
```

And:

```elm
from schema.table
```

will be parsed as a namespace `schema` and a column name `table`, which then breaks.

So this implements a hack where anything with `*` in gets compiled to a single ident, and so fixes the case in #1498. But the semantics are quite confusing.

---

I think ideally — but it might be a lot of work — we would instead 
- parse the argument in `from` and `join` as namespaces
- then `from schema.table` would be parsed as a single namespace `schema.table`, not a namespace and column
- but `select table.column` would still be parsed as a namespace `table` & column `column`

And then any ident in backticks could be opaque — we wouldn't need to split by `.`, we'd take `from schema.table`, not ```from `schema.table` ``` and produce the correct thing.

---

In the meantime, I'm +0.3 on merging this, since it solves the immediate problem.

Past discussions on this include https://github.com/PRQL/prql/issues/822 & https://github.com/PRQL/prql/issues/852


prql-compiler/src/semantic/context.rs
```diff
@@ -233,10 +242,6 @@ impl Context {
     }
 
     fn resolve_ident_wildcard(&mut self, ident: &Ident) -> Result<Ident, String> {
-        if ident.name != "*" {
-            return Err("Unsupported feature: advanced wildcard column matching".to_string());
-        }
-
         let (mod_ident, mod_decl) = {
```
aljazerzen, line 245: The intention with this was to do wildcard column matching:
```
from albums
select [a*_id]
# this selects artist_id and album_id
```
We'd need full knowledge of the column for that, so let's leave it for now.

aljazerzen: But this feature does interfere with this PR

aljazerzen: Oh wait, but if you use backticks, wildcard matching shouldn't be used...

aljazerzen:
I quite dislike the idea that `dir/*.parquet` can be an ident, but hey we support backticks, so it should work...

I'm +0.

snth:
I can't comment on the code or the parsing implications but just to say that in #286 we discussed possibly adding a `from_file` transform in the future. See for example https://github.com/PRQL/prql/issues/286#issuecomment-1331930214.

Would that not maybe make the compiler/parser code simpler?

While DuckDB is quite flexible and allows `SELECT * FROM 'dir/*.parquet'`, they also have `read_csv` and `read_parquet` functions which can take additional parameters which could be added to `from_file`.

You could then keep your existing identifier rules for use with `from`.

`from_file` could then also error out when compiling to a target other than DuckDB or one that doesn't allow reading csv files.

prql-compiler/src/semantic/context.rs
max-sixty: To confirm, would the wildcard column matching be a PRQL or DB feature?

max-sixty:
I see `from_file` as a PRQL function which reads a file and uses its data in a query, like `from_text` does. Otherwise why not just defer to the DB?

Regardless, we still have a somewhat complex system for parsing idents — I think the middle section in the issue would solve all of these, but be some work to implement.

I'll merge this and have opened a new issue for that, in #1535

prql-compiler/src/semantic/context.rs
aljazerzen: PRQL of course. I don't know any DB that support this.

max-sixty: https://duckdb.org/docs/sql/query_syntax/select.html

```sql
SELECT COLUMNS('number\d+') FROM addresses;
```

!

PR #1517, opened 2023-01-16
refactor: Make `compile-files` example compile with prql 

max-sixty:


prql-compiler/examples/compile-files/build.rs
```diff
@@ -1,8 +1,3 @@
-// This extern is probably not needed in the general case,
-// but this is a crate references a create in a different workspace, so
-// cargo does not link it auto-magically.
-extern crate prql_compiler;
```
max-sixty, line 4: @aljazerzen FYI I think this was just an `edition="2021"` issue

PR #1527, opened 2023-01-16
fix: prqlc & docs

aljazerzen:


prql-compiler/src/lib.rs
```diff
@@ -65,7 +65,7 @@
 //! - ... compile, format & debug PRQL from command line.
 //!
 //!     ```sh
-//!     $ cargo install ripgrep
+//!     $ cargo install prql-compiler --features=cli
```
max-sixty, line 68: WDYT about having this be a default feature? Atm installing from crates.io breaks.

I think I'll do a PR and then we can decide

PR #1528, opened 2023-01-16
feat: change name inference

### LINKED ISSUE #1503: Strictness over supplying namespaces
The compiler is very strict about when it needs namespaces and when it doesn't — i.e. it's not that they're always required, it's very context-sensitive.

I'll start by saying I realize that we've had discussions on this in the past, and that @aljazerzen has done basically all the work around these features, so his view should be weighed particularly highly. And a project we decided on "Namespace required" in https://github.com/PRQL/prql/issues/1193 & https://github.com/PRQL/prql/issues/1168, even though I dissented, and I'm not necessarily trying to rehash that.

But if we're doing that, could we then allow namespaces after a `select` or `join`?

Here are the two cases:

#### Namespace required

```sql
from problem
join sol [problem.solution==sol.id]
select foo
```

...though we do get a decent error message:

```
Error: 
   ╭─[:3:8]
   │
 3 │ select foo
   ·        ─┬─  
   ·         ╰─── Ambiguous name. Could be from any of _frame.problem._infer, _frame.sol._infer
───╯
```

In SQL, we can leave it ambiguous

```sql
SELECT
  foo
FROM
  problem
  JOIN sol ON problem.solution = sol.id
```

#### Original namespace not allowed

OTOH, once a column has been resolved, it's no longer part of a namespace, and supplying the original namespace raises an error: https://prql-lang.org/book/transforms/select.html#admonition-note

```elm
from e=employees
select e.first_name
filter e.first_name == "Fred"
```

```
Error: 
   ╭─[:4:8]
   │
 4 │ filter e.first_name == "Fred"
   ·        ──────┬──────  
   ·              ╰──────── Unknown name e.first_name
───╯
```

I totally get why the resolver does this. But it makes moving lines around in a query difficult, which is a central feature of pipelines. There is no semantic ambiguity there.

I would hold a different view if the line were instead `select first_name=e.first_name` — then we have a new item, with a new namespace. 

But if column hasn't been redefined, what do we think about allowing the original namespace to be optionally included?

aljazerzen:
Closes #1503

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -605,7 +605,7 @@ dependencies = [
 
 [[package]]
 name = "compile-files"
-version = "0.3.1"
+version = "0.4.0"
 dependencies = [
  "prql-compiler",
 ]
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,17 @@
 # PRQL Changelog
 
+## 0.4.1 — (unreleased)
+
+**Features**:
+
+- Change column name inference to also include relation name.
+
+  ```
+  from albums
+  select title # name used to be inferred as title only
+  select albums.title # so using albums was not possible here
+  ```
+
 ## 0.4.0 — 2022-01-15
 
 0.4.0 brings lots of new features including `switch`, `select ![]` and numbers
--- a/prql-compiler/src/semantic/mod.rs
+++ b/prql-compiler/src/semantic/mod.rs
@@ -121,6 +121,19 @@ mod test {
     }
 
     #[test]
+    fn test_resolve_04() {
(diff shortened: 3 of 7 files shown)


Cargo.lock
```diff
@@ -605,7 +605,7 @@ dependencies = [
 
 [[package]]
 name = "compile-files"
-version = "0.3.1"
+version = "0.4.0"
```
max-sixty, line 608: Hmmm, not sure why this got through before...

max-sixty: Oh, because it merged after the release, all good

max-sixty:
Awesome, thanks!

Issue #131, opened 2022-02-26
Potentially using Chumsky


max-sixty:
There are great readability advantages to having a parser grammar, and we're still iterating between the language design and the parser, so we're not going to rewrite it at this stage.

But this project looks really impressive: https://github.com/zesterer/chumsky, from @zesterer (who seems to be actively supporting & developing it, unlike `pest` & `nom`)

We're already hitting the limits of pest — e.g. on https://github.com/pest-parser/pest/issues/271 has been frustrating, and my current work refactoring how functions are handled is not efficient — so If we find the existing error messages aren't good enough, it's worth considering.

zesterer:
I'm interested to hear how you get on if you give it a go!

AlexRiedler:
Totally unaware of this issue, I actually did an incomplete proof-of-concept in chumsky last night. https://github.com/AlexRiedler/prql-chumsky (just the lexer and parser for now). I am hoping by the end of the month I will have the complete lexer and parser done.

max-sixty:
Awesome @AlexRiedler !

Copying what I wrote on [Discord](https://discord.com/channels/936728116712316989/945832046121607208/995408215741304952):

> If someone wanted to do the work — maybe they feel like a big parser project — I would vote to make the change. 

Though I'm not the only vote — CC @aljazerzen — so if you're doing this with an expectation it would merge, we should discuss first.

A couple of thoughts:
- If there's any way of using it in part, to start — e.g. for the later part of the parsing, such as something like https://github.com/prql/prql/blob/6b374b176b02af92b454c2ff279eaad5500b9f6d/prql-compiler/src/parser.rs#L167-L200, that would allow for a more gradual on-ramp — we could make incremental progress rather than aiming for a single switch-over. But I'm not sure how viable that is.
- We have lots of tests in parser.rs that aren't coupled to our current parsing which you could use (though they are coupled to the AST, but that would probably need to be invariant anyway)
- We try to be abstract over Transforms in the early stage of parsing, rather than parse `select` any differently to `derive`

AlexRiedler:
Swamped more than I hoped the last month, ended up going on vacation so no real update on the state of my experiment.

>  we could make incremental progress rather than aiming for a single switch-over. But I'm not sure how viable that is.

I think this should be possible, let me try this out and see where I can get in a relatively short period of time.

> We try to be abstract over Transforms in the early stage of parsing, rather than parse select any differently to derive

[...]

PR #1529, opened 2023-01-16
feat: start transition to chumsky parser

aljazerzen:
I've refactored parser into it's own dir and converted tests from snapshot of pest parse trees to snapshots of PL AST.

The plan for chumsky transition is:
- bottom-up rewrite: for terminals (e.g. idents, types), we can write modular chumsky parsers. Pest parsing remains in place (because pest does not support declaring a combinator parser), but when we do conversion from pest pair to PL AST, we instead convert pest pair to string and parse that with chumsky. If chumsky parser is written correctly, it will never fail.
  
  This does mean that we have duplicated parsing definitions and that parsing is ran twice. The upside is that it's easy and you can rewrite one term at a time.

- top-down rewrite: if we manage to implement custom chumsky parser that takes a string, run a pest parser, construct PL AST and returns that, we could split PEST parser into two or more smaller ones and rewrite each one of them. I'm not sure if this is possible, but it would parsing only once and having only one parsing implementation for syntactic construct.

Ref #131


max-sixty:
Super!

I can see the bottom-up rewrite going well. I can't visualize it as much with the top-down rewrite.

For the bottom-up rewrite, I was imagining that for some items, we could remove them from the pest parser and add them to the Chumsky parser, and achieve a gradual shift that way. Items would need to be fully encapsulated to remove them. f-strings & s-strings come to mind — i.e. pest just sees `f"<blank>"` and chumsky can manage the interpolation etc.

But possibly there aren't many of those encapsulated things. For example, if we want to parse a list, we actually need to parse the whole language, because a list can contain anything. Maybe it's as small as those + dates&times (on seeing a `@`)?

PR #1530, opened 2023-01-16
fix: js docs

aljazerzen:


max-sixty:
Thanks!

I have no context on whether we should be returning JS objects vs JSON. If it's the former, feel free to add a TODO / issue (as ever, don't feel you have to carry it all yourself...)

PR #1542, opened 2023-01-17
build: Inherit more from the workspace in Cargo

max-sixty:
I saw how @epage did things in `typos`; this approach is simpler, allows for less variance.

Unsupported by `wasm-pack` though

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -27,4 +27,3 @@ opt-level = "s"
 [workspace.metadata.release]
 consolidate-commits = true
 pre-release-commit-message = "chore: Release {{version}}"
-shared-version = true
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -13,12 +13,16 @@ members = [
 # commands like `cargo test` to only run tests from the default package. And
 # `cargo insta test` doesn't have a `--workspace` flag.
 
+[workspace.package]
+edition = "2021"
+license = "Apache-2.0"
+repository = "https://github.com/PRQL/prql"
+rust-version = "1.60.0"
+version = "0.4.0"
+
 [profile.release.package.prql-js]
 # Tell `rustc` to optimize for small code size.
 opt-level = "s"
-
-[workspace.package]
-version = "0.4.0"
 
 [workspace.metadata.release]
 consolidate-commits = true
--- a/prql-js/Cargo.toml
+++ b/prql-js/Cargo.toml
@@ -1,11 +1,17 @@
 [package]
 description = "Javascript bindings for prql-compiler"
-edition = "2021"
-license = "Apache-2.0"
 name = "prql-js"
 publish = false
-repository = "https://github.com/PRQL/prql"
+
+edition.workspace = true
+rust-version.workspace = true
 version.workspace = true
+
+# Unsupported by `wasm-pack` (which seems kinda unmaintained now...)
+# license.workspace = true
+# repository.workspace = true
+license = "Apache-2.0"
+repository = "https://github.com/PRQL/prql"
 
 [lib]
 crate-type = ["cdylib", "rlib"]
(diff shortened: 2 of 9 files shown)


prql-js/Cargo.toml
```diff
@@ -1,12 +1,18 @@
...
+
+edition.workspace = true
+rust-version.workspace = true
 version.workspace = true
 
+# Unsupported by `wasm-pack` (which seems kinda unmaintained now...)
+# license.workspace = true
+# repository.workspace = true
+license = "Apache-2.0"
+repository = "https://github.com/PRQL/prql"
```
epage, line 14: I'm curious, what is it wasm-pack is doing?  Looking at the dependencies, it doesn't look like it should be directly parsing these files; it uses `cargo-metadata` which relies on the `cargo-metadata` command, so the relevant version of cargo is used to parse the files.

prql-js/Cargo.toml
```diff
@@ -1,12 +1,18 @@
...
 description = "Javascript bindings for prql-compiler"
-edition = "2021"
-license = "Apache-2.0"
 name = "prql-js"
 publish = false
-repository = "https://github.com/PRQL/prql"
+
+edition.workspace = true
+rust-version.workspace = true
 version.workspace = true
```
epage, line 8: btw
```toml
[package.metadata.release]
shared-version = true
```
is implicit when you use `version.workspace = true`

Cargo.toml
```diff
@@ -13,13 +13,17 @@ members = [
 # commands like `cargo test` to only run tests from the default package. And
 # `cargo insta test` doesn't have a `--workspace` flag.
 
+[workspace.package]
```
epage, line 16: FYI you can also share dependency versions in `workspace.dependencies`.  If you were to use it, it seems like your dev-dependencies are the ones that would most likely benefit (like `insta`) but unsure if that is enough to justify it.  Some projects take the approach of having every dependency be in the workspace.  I've not gone to that extreme yet.

prql-js/Cargo.toml
max-sixty: Great, thanks, changed!

Cargo.toml
max-sixty: Thanks. dependabot makes this fairly maintainable, and not everything needs even insta atm, but if that changes we can promote things

prql-js/Cargo.toml
max-sixty: Hmmm, I'm not sure. This is the error though:

```
# from prql-js

❯ wasm-pack build --target bundler
Error: failed to parse manifest: .../prql/prql-js/Cargo.toml
Caused by: invalid type: map, expected a string for key `package.license`
```

max-sixty:
Thanks for the review @epage !

Cargo.toml
epage: Yes, dependabot helps and I don't think it has support for `workspace.dependencies`.  As a side tip, I'd recommend checking out Renovate.  Dependabot seems dead in the water for feature development while [Renovate does have workspace.dependencies support](https://github.com/renovatebot/renovate/pull/18831) and its easier to have more complex policies ([clap](https://github.com/clap-rs/clap/blob/master/.github/renovate.json5) vs [typos](https://github.com/crate-ci/typos/blob/master/.github/renovate.json5)).

The one remaining benefit for `workspace.dependencies` for something like `insta` is making it easier to keep the feature flags aligned.  Granted, if you aren't really adding new workspace members, that might not be much of an issue.

prql-js/Cargo.toml
epage: Ah, I missed `toml` in their dependency list.  Yup, they are manually processing things.

Cargo.toml
max-sixty: Great, yes I think renovate also does grouped updates, so we could do that on a schedule rather than manually (we don't want loads of PRs from each patch change).

I probably won't get to this soon — but if anyone sees this and wants to add Renovate, that would be welcome!

epage: Depending on your situation, my clap (library, keep build/regular versions within major low) and typos (binary, use latest and greatest) configs are good starting points.

PR #1545, opened 2023-01-17
build: Add `cli` as a default feature

max-sixty:
This reverts a change from @aljazerzen, so I'll wait for his review. It's required for `cargo install` to work without a `--features` flag.

Even if we don't want this, I would vote to merge this, release 0.4.1, and then do the work to the docs & brew recipe so that works again

### DIFF
--- a/prql-compiler/Cargo.toml
+++ b/prql-compiler/Cargo.toml
@@ -10,19 +10,19 @@ version.workspace = true
 
 [features]
 cli = ["clio", "atty", "clap", "color-eyre"]
+default = ["cli"]
 
 [[bin]]
 name = "prqlc"
 path = "src/cli.rs"
 required-features = ["cli"]
 
+[lib]
+doctest = false
+
 [dependencies]
 anyhow = {version = "1.0.57", features = ["backtrace"]}
 ariadne = "0.1.5"
-atty = {version = "0.2.14", optional = true}
-clap = {version = "4.1.1", optional = true, features = ["derive"]}
-clio = {version = "0.2.4", features = ['clap-parse'], optional = true}
-color-eyre = {version = "0.6.1", optional = true}
 csv = "1.1.6"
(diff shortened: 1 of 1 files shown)


max-sixty:
This actually isn't trivial, since the way `cli.rs` vs `main.rs` works has been changed. And I see a docstring that shows @aljazerzen knew that `--features=cli` was required. So I'll wait to hear from him. But we should decide / fix since currently the brew recipe is broken.

One option is to add add another crate `prqlc` which contains the CLI. When thinking through this before, I thought having a single crate with the compiler was easier for users though. In other projects, I've found it a bit annoying and unnecessary that running `cargo install foo` doesn't work, and actually someone has to run `cargo install foo-cli`...

aljazerzen:
I don't feel like adding `--features=cli` is too burdensome. The primary intention of the crate is to be used as a library (which should not include cli module) so it makes sense to say "hey I want the CLI flavor".

In initial intention here was to remove cli module from docs, but that was not possible, even if I removed cli from default features. So I changed the structure (and removed main.rs) so cli binary now essentially acts as its own crate, apart from dependencies, which have to be controlled via features.

And if we don't want CLI dependencies in the library build, installing cli will never be possible with just `cargo install prql-compiler`. Maybe it would make most sense to has `cargo install prql-compiler --features=prqlc` or `cargo install prqlc`.

max-sixty:
OK, that's quite reasonable. I'll change the brew recipe instead.

> Maybe it would make most sense to has `cargo install prql-compiler --features=prqlc` or `cargo install prqlc`.

Yes these are reasonable; any thoughts on which?

aljazerzen:
I prefer `cargo install prqlc`

snth:
+1 for `cargo install prqlc`

max-sixty:
Closing in favor of #1549

PR #1548, opened 2023-01-17
docs: website theme

aljazerzen:
My wife said that the website looks like the barber shop thing and suggested this instead:

image

I quite like it, but I may be a bit biased :D

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -605,7 +605,7 @@ dependencies = [
 
 [[package]]
 name = "compile-files"
-version = "0.3.1"
+version = "0.4.0"
 dependencies = [
  "prql-compiler",
 ]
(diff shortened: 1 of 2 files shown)


snth:
You mean like this?
image

I mean I kinda see where she's coming from but I don't see it as a big issue.

Your theme is too dark for my liking. I'd happily defer to a designer on this though.

richb-hanover:
I like the barber pole design. It's distinctive and looks "bright". It's easy to read the text. I think @aljazerzen's proposal is too dark (as is the pane at the bottom of the home page - I noticed that it didn't seem to belong.)

Do we have a designer in mind? I also would defer to them.

max-sixty:
I trust @aljazerzen 's wife more than me! FWIW I rather like the barbershop background. "Cutting SQL into shape"...

That said, if someone who has contributed to the website (or their partner...) has a designer in mind and would like to manage a refresh of the page, let me know, and I can sort out the funding.

aljazerzen:
It seems that current design is good enough, we can link back to this PR when needed.

PR #1549, opened 2023-01-17
build: prqlc crate

aljazerzen:
Ref #1545

@max-sixty Can you setup releasing to crates.io for prqlc & prql-compiler-macros?


.gitignore
```diff
@@ -8,7 +8,7 @@ target
 **/.idea
 
 /*.prql
-*/*.prql
+**/_*.prql
```
max-sixty: nit: implicit `**`

```suggestion
_*.prql
```

.github/actions/test-rust/action.yaml
```diff
@@ -46,17 +47,7 @@ runs:
...
         command: insta
         args: test --unreferenced=auto ${{ inputs.target_option }}
-    - name: 📋 Test
-      uses: richb-hanover/cargo@v1.1.0
-      if:
-        ${{ ! (inputs.target_option != '--target=wasm32-unknown-unknown' &&
-        runner.os == 'ubuntu') }}
-      with:
-        command: insta
-        args: test ${{ inputs.target_option }}
```
max-sixty: This requires ` --unreferenced=auto` on wasm, to not delete a bunch of snapshots. We do this so we can test there are no unreferenced snapshots around (but not all tests run on wasm)

.github/workflows/pull-request.yaml
```diff
@@ -26,7 +26,7 @@ jobs:
       matrix:
         include:
           - target_option: ""
-          - target_option: "--target=wasm32-unknown-unknown"
+          - target_option: "--exclude=prqlc --target=wasm32-unknown-unknown"
```
max-sixty: Better would be to promote this setting to the crate, so that `cargo build --target=wasm32-unknown-unknown` works on a fresh clone.

I'm not sure whether there's an easy way of doing that. We can do what we used to do — have the whole thing behind an attribute, so it compiles almost nothing with `wasm` (just a main that says "This doesn't work on `wasm`"!)

(I can take these btw)

.github/actions/test-rust/action.yaml
```diff
@@ -37,7 +38,7 @@ runs:
       uses: richb-hanover/cargo@v1.1.0
       with:
         command: test
-        args: ${{ inputs.target_option }} --no-run --locked
+        args: --workspace ${{ inputs.target_option }} --no-run --locked
```
max-sixty: I think this implicit with a virtual manifest, no?

max-sixty:
> @max-sixty Can you setup releasing to crates.io for prqlc & prql-compiler-macros?

Yes. + I can take some of the other todos here.

Can `prql-compiler` re-export `prql-compiler-macros`? Or does that not work with proc macros? It would be nice to have a single interface.

.github/actions/test-rust/action.yaml
aljazerzen: I was having problems with wasm because insta does not have a nice way of excluding a crate like other cargo commands.

If I understand correctly, we already run `cargo test` for wasm and check for unreferenced snapshots in the first insta step?

.github/actions/test-rust/action.yaml
aljazerzen: I don't think so. It says this:
```
➜  prql git:(prqlc) cargo test --exclude=prqlc
error: --exclude can only be used together with --workspace
```

aljazerzen:
> Can prql-compiler re-export prql-compiler-macros? Or does that not work with proc macros? It would be nice to have a single interface.

`prql-compiler` is a dependency of `prql-compiler-macros`, so I don't know if that would work.

.github/workflows/pull-request.yaml
aljazerzen: Yes, please do take this :D

.github/actions/test-rust/action.yaml
max-sixty: Hmm, from `cargo help test`:

> The default members of a workspace can be set explicitly with the workspace.default-members key in the root manifest. If this is not set, a virtual workspace will include all workspace members (equivalent to passing --workspace), and a non-virtual workspace will include only the root crate itself.

...though makes sense re `--exclude`; hopefully we can avoid that though

.github/actions/test-rust/action.yaml
max-sixty: The first step (i.e. this one: https://github.com/PRQL/prql/pull/1549/files/71e685695451e0c61645e0b327c578088f220265#diff-a748c5034df2d4fd42ef14e41d511155b668bdb490ff772175bf3fd40b4e2c9aR48) doesn't run for wasm, because it's conditioned on `inputs.target_option == ''`, which is false for wasm. So this in an XOR with the two steps labeled "Test".

Does that make sense? Tell me if I've made this unclear...

.gitignore
max-sixty: Nit^2 — the `**` is implicit! So we can just have

```suggestion
_*.prql
```

max-sixty:
I'm not confident the release step is working yet, but this code is reasonable-ish, so I think OK to merge.

Is the `lib.rs` / `main.rs` split a reasonable way of incorporating wasm without requiring CLI exclusions?

PR #1554, opened 2023-01-17
refactor: update test-python workflow

fool1280:
Issues: https://github.com/PRQL/prql/issues/1424
- Move all actions in `./.github/actions/test-python` to `./github/workflows/test-python`


max-sixty:
Super! Before merging, we then need to replace the reference in `test-all.yaml` on line 42 (searching the repo for `actions/test-python` is a way of finding this)

Very small, but this line can also go: https://github.com/PRQL/prql/pull/1554/files#diff-5405a6a14181df246119f41da914c6b91501ad8fa55720ad5e0dbdcea40de723L11

fool1280:
> Very small, but this line can also go: https://github.com/PRQL/prql/pull/1554/files#diff-5405a6a14181df246119f41da914c6b91501ad8fa55720ad5e0dbdcea40de723L11

I'm so sorry, are you referring to the `checkout`?

max-sixty:
> > Very small, but this line can also go: https://github.com/PRQL/prql/pull/1554/files#diff-5405a6a14181df246119f41da914c6b91501ad8fa55720ad5e0dbdcea40de723L11
> 
> I'm so sorry, are you referring to the `checkout`?

Sorry, I think the link doesn't work. I meant line 11 in `test-python.yaml` with `      - ".github/actions/test-python/**"` — it doesn't need to watch for that path any longer, so that line can be removed. Does that make sense?

.github/workflows/test-all.yaml
```diff
@@ -39,7 +39,7 @@ jobs:
     steps:
       - name: 📂 Checkout code
         uses: actions/checkout@v3
-      - uses: ./.github/actions/test-python
+      - uses: ./.github/workflows/test-python
```
```suggestion
      - uses: ./.github/workflows/test-python.yaml
```

fool1280:
I have updated the PR accordingly :D Apparently reusable workflow has to be called by a job itself: https://github.com/orgs/community/discussions/27362.

Let me know what you think, and thank you so much for the guidance!

.github/dependabot.yml
```diff
@@ -45,13 +45,6 @@ updates:
     schedule:
       interval: "daily"
 
-  - directory: ".github/actions/test-python"
```
max-sixty, line 48: Great find, thanks

fool1280: thank you so much, it was the first time I contributed so the PR looks kinda messy, will improve more in the future

max-sixty:
I can't work out why these don't start. I changed one thing re `steps`, but these tasks are in `pull-request.yaml` which isn't even changed in this PR...

image

.github/dependabot.yml
max-sixty: Not at all!

max-sixty:
(I'm also not sure my change to `steps` was necessary, feel free to undo if it's not)

max-sixty:
@fool1280 if it's OK, I'm going to see if reverting all the code here makes the workflow run, then I'll re-revert it. That will help us understand whether it's an issue with the code.

max-sixty:
> @fool1280 if it's OK, I'm going to see if reverting all the code here makes the workflow run, then I'll re-revert it. That will help us understand whether it's an issue with the code.

OK, that does make it work...

image

fool1280:
Hmm, this is so weird 🥲

max-sixty:
Very weird! This made it work, so it's something in `test-python` there.... https://github.com/PRQL/prql/pull/1554/commits/a65ec39e3122a7b2b4414f2ef720158a26cbaba6

max-sixty:
Ah OK — it does give an error message! If we click through to "checks" and then "pull-request": https://github.com/PRQL/prql/actions/runs/3945154253. Otherwise it was like searching around in the dark...

image

This is very fiddle-y, thanks for your patience...

I'll give it a couple more tries

fool1280:
Thank you so much for helping me! This is so quick on your side; I'm still trying to figure it out locally 😂

max-sixty:
So this works. There were two issues:
- The original issue. I couldn't understand why your attempt didn't work for ages; it looked exactly the same as https://docs.github.com/en/actions/using-workflows/reusing-workflows#using-a-matrix-strategy-with-a-reusable-workflow. But I think the issue was that `test-python.yaml` doesn't take an `os` input! So then it fails, and because I forgot where the error message was, it wasn't easy to debug...
- The concurrency issue, which I disabled so we could get this working. We can come back to this another time.

Thanks a lot @fool1280 ! These are helpful changes, and you did 90% of this.

If you're up for doing this with `test-rust` & `time-compilation`, those would be good candidates too...

For reference, here's the full diff of the changes I made on top of yours.

```diff
diff --git a/.github/workflows/test-all.yaml b/.github/workflows/test-all.yaml
index 9005b25c..87dc6212 100644
--- a/.github/workflows/test-all.yaml
+++ b/.github/workflows/test-all.yaml
@@ -32,12 +32,7 @@ jobs:
           target_option: ${{ matrix.target_option }}
 
   test-python:
-    strategy:
-      matrix:
-        os: [ubuntu-latest, macos-latest, windows-latest]
     uses: ./.github/workflows/test-python.yaml
-    with:
-      os: ${{ matrix.os }}
 
   test-java:
     runs-on: ${{ matrix.os }}
diff --git a/.github/workflows/test-python.yaml b/.github/workflows/test-python.yaml
index de6f1e1e..9abf4df3 100644
--- a/.github/workflows/test-python.yaml
+++ b/.github/workflows/test-python.yaml
@@ -10,9 +10,15 @@ on:
       - ".github/workflows/test-python.yaml"
   workflow_call:
 
-concurrency:
-  group: ${{ github.workflow }}-${{ github.ref }}
-  cancel-in-progress: true
+# TODO: we needed to remove this, because if there's a) a change in the path
+# such that this is run on a PR and b) a label on the PR such that `test-all` is
+# triggered, then this deadlocks. I'm sure there's a way of designing this so it
+# doesn't deadlock — either adding something to the key so it knows the "parent"
+# workflow, or (worse but acceptable) skipping the job if there's a
+# `pr-test-all` label on the PR. Or some other way...
+#
+# concurrency: group: ${{ github.workflow }}-${{ github.ref }}
+#   cancel-in-progress: true
 
 jobs:
   test:
```

fool1280:
Yes, I'd like to continue doing the same thing for `test-rust` & `time-compilation` :) Thank you for debugging this

max-sixty:
Great! And FYI I adjusted the description in #1424 a bit

PR #1560, opened 2023-01-18
refactor: add test-rust workflow

fool1280:
Issue: https://github.com/PRQL/prql/issues/1424
- Move `test-rust` actions to its separate workflow
- Update `test-all.yaml` and `pull-request.yaml` accordingly


max-sixty:
Super @fool1280 ! 

I'm seeing an error over at https://github.com/PRQL/prql/actions/runs/3949909631 (I clicked on "Checks" on this page, and then the workflow "pull-request").

I'll add one comment re the OSs too.

.github/workflows/test-rust.yaml
```diff
@@ -0,0 +1,75 @@
...
+  test:
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        include:
+          - os: ubuntu-latest
+          - os: windows-latest
+          - os: macos-latest
+          - os: ubuntu-latest
+            target_option: "--target=wasm32-unknown-unknown"
```
max-sixty: For PRs, we want these to run really quickly, so we only run on ubuntu, which catches all but the rarest errors (maybe 1/1000 errors are only on windows). Then on `test-all` we want to run all the OSs.

So I think it might be worth going back to defining this list of OSs in `test-all` and passing the values to `test-rust`.
`pull-request` already defines the two options it wants for `target_option`, so that should work.

Does that make sense? Lmk if it's confusing...

fool1280: I understand it now, let me fix it

fool1280:
I'm having the same problems triggering the workflow in `pull-request`. Will look into more tomorrow.

.github/workflows/test-all.yaml
```diff
@@ -22,21 +22,14 @@ concurrency:
...
-          - os: ubuntu-latest
-            target_option: "--target=wasm32-unknown-unknown"
-    steps:
-      - name: 📂 Checkout code
-        uses: actions/checkout@v3
-      - uses: ./.github/actions/test-rust
-        with:
-          target_option: ${{ matrix.target_option }}
+        os: [ubuntu-latest, windows-latest, macos-latest, ubuntu-latest]
+        target_option: [--target=wasm32-unknown-unknown]
```
```suggestion
        target_option: ["", --target=wasm32-unknown-unknown]
```

.github/dependabot.yml
```diff
@@ -45,13 +45,6 @@ updates:
     schedule:
       interval: "daily"
 
-  - directory: ".github/actions/test-rust"
```
max-sixty, line 48: Can we also remove the file at `".github/actions/test-rust"`?

.github/workflows/test-rust.yaml
```diff
@@ -0,0 +1,76 @@
...
+      - ".github/workflows/test-rust.yaml"
+  workflow_call:
+    inputs:
+      os:
+        required: false
+        default: ubuntu-latest
+        type: string
+      target_option:
+        required: false
+        default: --target=wasm32-unknown-unknown
```
max-sixty:
```suggestion
        default: ""
```

(and to the extent there was a description of this field in the original, would be great if we can copy it here as a comment)

max-sixty:
Well done for getting it working! These workflows can be frustrating because often the only way is pushing...

I added a couple of small comments, and then we can merge!

max-sixty:
FYI if you want to do more... the quickest wins are the ones in `test-all` that are copy-pasted; I think `test-java` & `test-js` — those can now just reference the underlying workflow

.github/workflows/test-rust.yaml
fool1280: Can I ask more context about which comments? I'm seeing the description of the target-option or this one: 

`Without inputs.target, it seems to select linux x64 in the wasm32 tests, though in the docs it says it uses the target as the key?`

max-sixty:
@fool1280 this was excellent!

I made a couple of small additions — fixed this weird deadlock issue, and changed `test-rust` to only be a workflow call because of the conflict over `inputs`. Your design was nicer, but I'm not sure GH can support it — it didn't seem to use the default `worklow_call` inputs in a `type: pull-request`.

fool1280:
Thank you so much @max-sixty, I was going to look into the deadlock issue later in the day. I'm not sure if I'm understanding correctly the issues, so please correct me if I'm wrong.

We have to add the suffix in the `test-all` and `pull-request` because those two both trigger `test-rust` which can be a problem?

max-sixty:
> We have to add the suffix in the `test-all` and `pull-request` because those two both trigger `test-rust` which can be a problem?

I think so. The problem comes from the concurrency group — generally this is used to free up resources — if we push a new commit and a new set of jobs start, everything within that concurrency group that's still running from a previous commit gets stopped, which is good since we don't care about the old result any longer.

My guess is that by starting jobs from multiple workflows, we have multiple jobs from the same commit with the same concurrency group value, which GH doesn't know how to handle, so calls that a deadlock.

We solve it by adding a suffix to the concurrency group, so they're unique across workflows.

But I don't have a tight mental model of exactly what's going on, and it doesn't seem possible to query for a concurrency group to understand what value each job has. And weirdly, I can't find anything online about it — we can't be the first people to be doing things like this (can we?!) — so it's possible we're making some other mistake...

But most importantly — this works for now, and isn't going to compound into something great or terrible — so I think it's fine to leave, and just add a suffix onto any concurrency group def if we hit it somewhere else...

PR #1567, opened 2023-01-18
docs: Update rust bindings example for PRQL compiler v0.4.0

nkicg6:
Updated the Rust Bindings example from the PRQL book so it compiles by including sql::Options.
The previous example wouldn't compile.

### DIFF
--- a/book/src/bindings/rust.md
+++ b/book/src/bindings/rust.md
@@ -16,11 +16,17 @@ cargo add prql-compiler
 
 ```rust
 use prql_compiler::compile;
+use prql_compiler::sql;
 
 fn main() {
     let prql = "from employees | select [name,age]  ";
-    let sql = compile(prql).unwrap();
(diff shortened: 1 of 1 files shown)


max-sixty:
Thank you very much @nkicg6 !

---

Jotting down a couple of thoughts for the future:
- It would be better if we could have this in our docstring, have it tested, and then `include_str` into the book
- We could add something like [this](https://github.com/PRQL/prql/blob/5c5f274fcef489a194d0dc90ad495df222d874b4/prql-js/Cargo.toml#L41-L45) to the `book/Cargo.toml` to update the version on every release.

nkicg6:
Thanks for the tip @max-sixty, good idea. I'll take a swing at that and tag you in the next few days

max-sixty:
> Thanks for the tip @max-sixty, good idea. I'll take a swing at that and tag you in the next few days

That would awesome @nkicg6

PR #1576, opened 2023-01-19
docs: added elixir-prql to book

kasvith:
- Added elixir bindings to the book


book/src/bindings/elixir.md
```diff
@@ -0,0 +1,40 @@
+# PRQL
```
max-sixty: If this is just the readme, let's include with `{{ include` like in https://github.com/PRQL/prql/blob/main/book/src/language-features/standard-library.md?plain=1#L20, and then it'll stay updated?

kasvith: aa wait

kasvith: @max-sixty can you do a build on docs? i cant do it in my PC due to the python error i mentioned before

max-sixty: Yeah, I just labeled it so it'll deploy to the main docs, easy to undo if it looks bad

kasvith: https://github.com/PyO3/pyo3/issues/2731 this issue keeps hitting me :(

kasvith: image

max-sixty: You have a 32bit python interpreter?! :)

kasvith: Yeah, need to reinstall but im bit lazy :D

max-sixty:
Thanks @kasvith !

kasvith:
Shall we do this for all bindings?

max-sixty:
> Shall we do this for all bindings?

Yes that's a good idea!

BTW we can also add prql-elixir here, maybe when we publish to Hex? 
image

PR #1579, opened 2023-01-19
fix: group by switch

### LINKED ISSUE #1564: Grouping by column derived from switch function generates invalid SQL
Repoducable in playground.
```
# from playground
from tracks
select [
  avg_length = average milliseconds,
  length = milliseconds,
  category = switch [
    length > avg_length -> 'long',
    true -> 'short'
  ]
]
group [category] (
  aggregate (
    count
  )
)
```
Grouping by a column derived from switch function (**that uses window functions**) generates SQL that with window functions in group by.
```
SELECT
  CASE
    WHEN milliseconds > AVG(milliseconds) OVER () THEN 'long'
    ELSE 'short'
  END AS category,
  COUNT(*)
FROM
  tracks
GROUP BY
  CASE
    WHEN milliseconds > AVG(milliseconds) OVER () THEN 'long'
    ELSE 'short'
  END

-- Generated by PRQL compiler version 0.4.0 (https://prql-lang.org)
```

Error thrown:
```
ERROR:  window functions are not allowed in GROUP BY
````

aljazerzen:
Closes #1564

### DIFF
--- a/prql-compiler/src/sql/anchor.rs
+++ b/prql-compiler/src/sql/anchor.rs
@@ -472,7 +472,9 @@ pub fn infer_complexity_expr(expr: &Expr) -> Complexity {
 
 #[derive(Default)]
 pub struct CidCollector {
-    cids: HashSet<CId>,
+    // we could use HashSet instead of Vec, but this caused nondeterministic
+    // results downstream
+    cids: Vec<CId>,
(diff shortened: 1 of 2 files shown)


max-sixty:
Thanks a lot!

PR #1581, opened 2023-01-19
feat!: change switch syntax to =>

aljazerzen:
Ref #1286

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,12 @@
 # PRQL Changelog
 
+## 0.5.0 — [unreleased]
+
+**Features**:
+
+- Change switch syntax to use `=>` instead of `->`
+  due to clash with planned lambda function syntax.
+
 ## 0.4.2 — [unreleased]
 
 **Features**:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,8 +4,8 @@
 
 **Features**:
 
-- Change switch syntax to use `=>` instead of `->`
-  due to clash with planned lambda function syntax.
+- Change switch syntax to use `=>` instead of `->` due to clash with planned
+  lambda function syntax.
 
 ## 0.4.2 — [unreleased]
 
--- a/book/src/language-features/switch.md
+++ b/book/src/language-features/switch.md
@@ -9,8 +9,8 @@ PRQL uses `switch` for both SQL's `CASE` and `IF` statements. Here's an example:
 ```prql
 from employees
 derive distance = switch [
-  city == "Calgary" -> 0,
-  city == "Edmonton" -> 300,
+  city == "Calgary" => 0,
+  city == "Edmonton" => 300,
 ]
 ```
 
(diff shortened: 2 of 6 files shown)


snth:
Great!

snth:
We'll need to release soon then though because the switch examples with `->` already don't work in the playground anymore.

CHANGELOG.md
```diff
@@ -1,5 +1,12 @@
 # PRQL Changelog
 
+## 0.5.0 — [unreleased]
```
snth, line 3: This can't be left for 0.5.0 and has to be included in 0.4.2 because currently the playground doesn't match otherwise.

aljazerzen: Ah. But it's a breaking change, it cannot go into 0.4.2.

I think we need to freeze the playground & docs to the latest release.

snth: Totally agree with you on both points.

How would 0.4.2 work though because it's already merged onto main so wouldn't it be included in 0.4.2 anyway?

How much work do you think it would be to maintain release branches and have those rebased automatically with each merged PR?

aljazerzen: Many projects do this: it's called backporting.

You have the main/master branch with all new commits.

For each release (i.e. 0.4.0) you create a branch and to release 0.4.1 you cherry-pick commits that are not breaking changes. Sometimes it's easy, but if you get merge conflicts, it's easiest to abort and release the commit only in the next major release (0.5.0).

Let's say you have this commit chain, where B is 0.4.0 release and D is a breaking change:
```
     0.4.0                         
A ---> B ---> C ---> D! ---> E ----> F
```

To release you can cherry-pick C, E & F to `0.4.0-backports` branch. Let's assume that E somehow depended on D (which was not picked) and is causing a merge conflict. We just skip it and pick just C, F
```
     0.4.0                        0.5.0
A ---> B ---> C ---> D! ---> E ---> F
       \
        \           0.4.1
          ---> C ---> F
```

But this all includes more work than just releasing often.

max-sixty:
I think we should discuss, I don't see the logic of this yet. 

Not nixing it, and I've often be voted down, but let's have the discussion...

aljazerzen:
@max-sixty do you mean switch syntax or backporting?

snth:
I had the same question.

RE: switch syntax I think we should revert the change for now because it's broken things in the playground.

aljazerzen:
> it's broken things in the playground

What's broken?

CHANGELOG.md
max-sixty: I think this is viable. 

I also think we can just do 0.5.0 next if we have a breaking change, rather than 0.4.2?

max-sixty:
> @max-sixty do you mean switch syntax or backporting?

The switch syntax...

snth:
> What's broken?

It's broken in the sense that the playground doesn't match the docs or at least doesn't match the 0.4.x released code in the wild.

I went to the playground to test the code examples in #1564 and it didn't work because the playground no longer accepted the `->` syntax, only the `=>` syntax. From a UX perspective I think that's broken as richb also pointed out in #1582 .

Issue #1507, opened 2023-01-14
Book & playground don't match the latest published version


mklopets:
Because the website (book + playground) get published from `main`, they are frequently out-of-sync with the reality that's been published & taken into use by others. This can cause a lot of confusion when "known good" queries suddenly start breaking in the playground, or when docs don't match the playground.

We've discussed this briefly in the past, but I wanted to open an issue since I've seen my team run into this multiple times in the past few weeks.

As the project gains popularity, I think we need to solve this in one way or another. Otherwise, this will make for a lot of frustrated would-be users. IMO, the published docs (+playground) should match the latest published version of the library, and they should also explicitly mention which version they're for.

max-sixty:
I agree this isn't ideal. I can see a couple of ways of improving the situation:

1. Release faster. That would involve
  - Mandating Changelogs & docs
    - Basically anything that makes @aljazerzen slower or less keen to write code is a negative for the project, so this depends on him mostly!
  - Release every couple of days, either an automated workflow, a button anyone can press, or someone just does it.
  - "Holding" breaking changes until we bump non-patch version
    - Another approach would be to implement versioning and have a feature flag for `nightly`. That would be a fun project, but a difficult one. (And some things would be very difficult / not worth doing, like changing the `pest` grammar.) Most of the payoff of that work is beyond this request
2. An option in the playground for which version is running. That would involve
  - Having multiple wasm files when building the playground, either built from switching to a tag, or publish artifacts to GitHub and use those
  - A UI for selecting them
  - I'm not sure how difficult the JS side would be

3. Publish dev releases to npm. This would let folks like @mklopets work with a more up-to-date version. This is probably the easiest but least conclusive option?

@mklopets I totally get that having the playground & book match the currently released version would make development easier for you, and I really want development to be easy for you! But having it match the current `main` also makes development easier for the project, so I'd be a -0.5 on just making that switch, assuming one of the above is viable. That said, others should weigh in. Is that reasonable?

richb-hanover:
Today, PRQL has conflicting priorities: 

1. We need people to try out PRQL (because it's now "worth criticizing") to find the next most important improvement, notice the sharp edges, and discover actual bugs
2. Those people "crash up on the rocks" when the language changes out from under known-good queries. 

My perspective: We're still very much in an experimental phase. There are lots of known "language issues" that need to be ironed out that will break queries. We should target intrepid users who can tolerate this breakage, all the while minimizing their pain. How can we do that?

[...]

PR #1582, opened 2023-01-19
docs: Update CHANGELOG.md to link to updated `switch` example

richb-hanover:
I was surprised to see the `=>` syntax described as 0.5.0, yet have it implemented in the current Playground.

I would suggest that we use the "0.x.y - Unreleased" section to contain _everything_ that is currently implemented, even if it arrives prior to its "stated delivery date/release" in the Roadmap.


snth:
I agree with you and just commented on this in #1581.

While I'm very much for the change, we need to figure something out about the release process and website updates.

max-sixty:
I think we can add a section to the template with `**Breaking changes**`, WDYT?

TaxFairness:
> I think we can add a section to the template with `**Breaking changes**`, WDYT?

I have seen the separate "Breaking Changes" section, but I have a sense that we should keep all the "features" (new or altered) together. They're what most people want/need to see. 

As I look at the 0.3.0 CHANGELOG entries, I would have placed all seven (three breaking, four new) items in the same group so they're seen as the "Big News" part of the release. All the rest of the 0.3.0 entries are careful reporting of other stuff that happened.  By marking each of those three as **Breaking change**, that information stays with the feature description (and won't get lost if items get re-ordered).

richb-hanover:
Uh oh... @TaxFairness (my alter-ego) responded. It was really me :-)

max-sixty:
> Uh oh... @TaxFairness (my alter-ego) responded. It was really me :-)

I agree with your alter ego!

aljazerzen:
This change was placed in 0.5.0 section because it is a breaking change, which cannot go into a patch release.

The fact that playground & book immediately contained the change is a separate problem from #1507

aljazerzen:
Disregard that: semver states that "Major version zero (0.y.z) is for initial development. Anything MAY change at any time. The public API SHOULD NOT be considered stable."

TaxFairness:
I'm glad to hear a citation for the rule (which makes complete sense). Thanks.

aljazerzen:
Since the breaking change has been reverted, I removed the entry in the changelog, but kept the extended entry about `from_text`.

Thanks for contributing.

max-sixty:
Good changes, thank you!

PR #1584, opened 2023-01-19
docs: use `prqlr::prql_compile` instead of `prqlr::prql_to_sql`

eitsupi:
I changed prqlr's function name to match other language bindings.
`prql_to_sql` continues to work but is deprecated.
https://github.com/eitsupi/prqlr/blob/main/NEWS.md


max-sixty:
Thanks @eitsupi ! And sorry for the change without a deprecation cycle on our end — I think even at our stage, we should have done one.

PR #1591, opened 2023-01-20
docs: Added documentation example for prql_compiler::compile

nkicg6:


max-sixty:
Looks good I think! 
image

@aljazerzen any thoughts?

aljazerzen:
LGTM.

We'd need more examples on other functions.

Issue #761, opened 2022-07-05
Intersect and Difference Operators


snth:
As requested in #656, it would be nice if PRQL could provide `intersect` and `difference` operators along with the `union` operator.

`intersect` and `difference` can be achieved with `semi joins` and `anti joins`. I'm not sure how widely supported those are but there are correlated subquery patterns with EXISTS that can work around this. This will require backend adaptor specific code though so I think we should rather address this in a separate issue in order to keep the union implementation simple. I will open an issue for those.

References:

https://en.wikipedia.org/wiki/Relational_algebra#Semijoin_(%E2%8B%89)(%E2%8B%8A)

https://en.wikipedia.org/wiki/Relational_algebra#Antijoin_.28.E2.96.B7.29

https://link.springer.com/chapter/10.1007/978-1-4302-3229-2_11

max-sixty:
I think we can do these with standard `INTERSECT` operators — along with https://github.com/prql/prql/issues/656? Or are these not well-supported?

e.g. Postgres https://www.postgresqltutorial.com/postgresql-tutorial/postgresql-intersect/

pbsds:
Leaving this here, for performance reasons.
https://www.crunchydata.com/blog/rise-of-the-anti-join

aljazerzen:
I was thinking how to best implement `union` and came to realization, that many of the operations can be translated down a small set of operators in relational algebra.

For example our current approach for `distinct`/`unique` is:
```
func unique tbl -> group [identity] (take 1) tbl
```
This representation with `group + take` is just intermediate and is unpacked into actual `SELECT DISTINCT` when translating. But the point is that we keep RQ (the intermediate stage) simple so it's debuggable and easy to implement for new backends.

That said `union`, `intersect` and `difference` can all be implemented only by adding `concat` operator (dplyr calls it `bind_rows`, pandas calls it `concatenate`):

```
concat a b
-> UNION ALL


union a b
 = concat a b | unique
-> UNION


difference a b =
 = join a [identity] b side:left | filter b.identity == null | select a.*
-> SELECT a.* FROM a LEFT JOIN b ON ... WHERE b.id IS NULL


intersection a b
 = join a [identity] b side:inner | select a.*
-> SELECT a.* FROM a INNER JOIN b ON ...
```

`union`, `difference` and `intersection` can be defined in std lib as normal functions that resolve to using derivations above. The best things is that when someone writes:

```
from a
select x, y
concat b
group [x, y] (take 1)
```
[...]

PR #1592, opened 2023-01-20
feat: remove & intersect with EXCLUDE & INTERSECT

aljazerzen:
Ref #761

### DIFF
--- a/prql-compiler/src/semantic/lowering.rs
+++ b/prql-compiler/src/semantic/lowering.rs
@@ -2,7 +2,7 @@ use std::collections::hash_map::RandomState;
 use std::collections::{HashMap, HashSet};
 use std::iter::zip;
 
-use anyhow::{bail, Result};
+use anyhow::Result;
 use enum_as_inner::EnumAsInner;
 use itertools::Itertools;
 
@@ -313,7 +313,8 @@ impl Lowerer {
                 self.declare_as_columns(assigns, false)?;
             }
             pl::TransformKind::Select { assigns, .. } => {
-                self.declare_as_columns(assigns, false)?;
+                let cids = self.declare_as_columns(assigns, false)?;
+                self.pipeline.push(Transform::Select(cids));
             }
             pl::TransformKind::Filter { filter, .. } => {
                 let filter = self.lower_expr(*filter)?;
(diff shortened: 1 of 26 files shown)


prql-compiler/tests/integration/main.rs
```diff
@@ -25,15 +25,33 @@ mod tests {
...
+                let opts = prql_compiler::sql::Options::default()
+                    .with_dialect(prql_compiler::sql::Dialect::PostgreSql)
+                    .no_format()
+                    .some();
+                let sql = prql_compiler::compile(&prql, opts).unwrap();
+                let pg_out = postgres::query_csv(pg_client, &sql);
+                assert_eq!(sqlite_out, pg_out);
             }
+
+            assert_snapshot!(sqlite_out);
```
max-sixty: Very nitty comment: to the extent we want the source expression in the insta snapshot, we can pass it with the three argument form.

prql-compiler/src/semantic/lowering.rs
```diff
@@ -558,8 +559,40 @@ impl Lowerer {
...
+                    match self.node_mapping.get(target_id) {
+                        Some(LoweredTarget::Compute(cid)) => targets.push(*cid),
+                        Some(LoweredTarget::Input(input_columns)) => {
+                            targets.extend(input_columns.values().map(|(c, _)| c))
+                        }
+                        _ => {}
+                    }
+                }
+
+                // this is terrible code
```
max-sixty, line 575: lol, actually underrated comment — we should have more of these! It's also a nice way for someone to make a very small rust contribution!

prql-compiler/src/semantic/std.prql
```diff
@@ -33,17 +33,16 @@ func aggregate<table> a<column> tbl<table> -> null
...
-)
 func group<table> by pipeline tbl<table> -> null
 func window<table> rows:0..0 range:0..0 expanding:false rolling:0 pipeline tbl<table> -> null
 
-func intersection<table> bottom<table> top<table> -> (
-    from t = _param.top | join b = _param.bottom (all (map _eq (zip t.* b.*))) | select t.*
+func append<table> `default_db.bottom`<table> top<table> -> null
+func intersect<table> `default_db.bottom`<table> top<table> -> (
+    from t = _param.top
+    join b = _param.bottom (all (map _eq (zip t.* b.*)))
```
max-sixty, line 42: What's this? :)

aljazerzen: What do you mean? the `(all (map _eq (zip t.* b.*)))` part?

`t.*` and `b.*` each evaluate to a list of columns.

`(zip t.* b.*)` evaluates to `[[t.col1, b.col1], [t.col2, b.col2], ...]`

`(map _eq (zip t.* b.*))` evaluates to `[t.col1 == b.col1, t.col2 == b.col2, ...]`

`(all (map _eq (zip t.* b.*)))` evaluates to `(t.col1 == b.col1 and t.col2 == b.col2) and ...`

... it's a join condition over all columns of both tables, abusing our resolver.

The helper functions can remain an implementation details. It'd be nice if we could hide them from the public interface.

prql-compiler/tests/integration/main.rs
aljazerzen: I tried that, but the macro understood first argument as snapshot name.

PR #1601, opened 2023-01-22
docs: Move development & contributing docs to book

max-sixty:
At the moment, they're somewhat scattered in the home path of the repo, which makes them more difficult to navigate and clutters the home path

This ended up being a bit of a "whack-a-link" exercise. I think I'll have to do another pass before merging.


max-sixty:
I can't publish to the web from my fork (which makes sense as then anyone could). I've manually checked the links. I'm going to merge and then resolve any links which are still broken...

PR #1604, opened 2023-01-22
feat(prql-compiler): make `prql_compiler::PRQL_VERSION` public

eitsupi:
When embedding prql-compiler in a package such as prqlr (own version does not match prql-compiler version), it would be useful to be able to easily check the version of the embedded prql-compiler. (eitsupi/prqlr#51)

If the release cycle is fast (related to #1507), downstream software may not be able to keep up with the latest version, so users will want to check the version of PRQL that is supported.

### DIFF
--- a/prql-compiler/src/lib.rs
+++ b/prql-compiler/src/lib.rs
@@ -92,7 +92,7 @@ pub use utils::IntoOnly;
 use once_cell::sync::Lazy;
 use semver::Version;
 
-static PRQL_VERSION: Lazy<Version> =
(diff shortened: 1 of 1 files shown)


aljazerzen:
Sure.

eitsupi:
Thank you for quick response and merging this!

max-sixty:
@eitsupi I added you with write permissions on the repo, so feel free to merge these sorts of things yourself (easy to revert if anything goes wrong!).

eitsupi:
@max-sixty Thank you very much. I am honored.
I will do so when needed.

PR #1605, opened 2023-01-22
build: postgres integration tests

aljazerzen:


prql-compiler/Cargo.toml
```diff
@@ -35,16 +35,12 @@ strum_macros = "0.24.0"
 
 [dev-dependencies]
 cfg-if = "1.0"
-chrono = "0.4"
-globset = "0.4.8"
 insta = {version = "1.26", features = ["colors", "glob", "yaml"]}
-pulldown-cmark = "0.9.1"
-pulldown-cmark-to-cmark = "10.0.1"
-similar = "2.2.0"
-walkdir = "2.3.2"
```
max-sixty, line 44: Good find! Did you use a tool to get these?

max-sixty:
Awesome! Shall I try and add this to GHA? Or should we put that into a Good First Issue?

prql-compiler/Cargo.toml
aljazerzen: No, it just seemd suspicious that compiler required markdown.

max-sixty: OK cool, I added https://github.com/PRQL/prql/pull/1609 to do this systematically

Issue #1583, opened 2023-01-19
Link bindings pages in book to `Readme`s


eitsupi:
It doesn't seem to reflect the API changes in 0.4.

Perhaps it's better to copy the contents of the README like Elixir's page does?

https://github.com/PRQL/prql/blob/8a85eb12b65a8218a732ea2184cda2b4521acde7/book/src/bindings/elixir.md?plain=1#L1

aljazerzen:
Yup, it'd be better if we were including, otherwise there just too much duplicated text that needs to be maintained.

max-sixty:
The js one is at https://github.com/PRQL/prql/pull/1608

eitsupi:
I think we have done.

PR #1607, opened 2023-01-22
docs: Fix footnote in prql-js readme

max-sixty:


max-sixty:
Actually this suffers from https://github.com/PRQL/prql/issues/1583, let's do that one first

Issue #1602, opened 2023-01-22
Mention backtick quoting in book page on `from`


snth:
The section on [from](https://prql-lang.org/book/transforms/from.html) in the Book should mention that table names containing schemas or needing to be quoted for other reasons need to be contained in backticks.

This should also link to the relevant section on the [syntax page](https://prql-lang.org/book/syntax.html):
* [Quoted Identifiers](https://prql-lang.org/book/syntax.html#quoted-identifiers)
* [Quoting Schemas](https://prql-lang.org/book/syntax.html#quoting-schemas)

veenaamb:
I just made the change and linked to quoted identifiers, quoting schemas and ran it on my localhost. However, I'm not quite sure if this is how you want the change. Could you please let me know?

[
<img width="1439" alt="Screenshot 2023-01-23 at 10 36 42 AM" src="https://user-images.githubusercontent.com/47256853/214096947-1b22d046-3dfc-412c-a71c-385bd14dd47a.png">
](url)

snth:
Hi @veenaa014 and welcome!

That looks good. Could you please just move the text you added to just above the **Examples** heading?

Then please go ahead and open a PR and reference this issue. Looking forward to have you as a contributor!

veenaamb:
Sounds good. Will do that!

PR #1614, opened 2023-01-23
docs: Update from.md in book

veenaamb:
Updated documentation on From page in the book to include conditions where table names need to be contained within backticks


veenaamb:
Closes #1602

max-sixty:
Thanks a lot @veenaa014 ! Happy to have you as a contributor.

PR #1618, opened 2023-01-24
build: Add wheel `task`, use in CI

max-sixty:
This lets us build & test locally


.github/workflows/test-python.yaml
```diff
@@ -49,28 +52,29 @@ jobs:
...
-          path: .nox
-          key:
-            nox-${{ hashFiles('prql-python/noxfile.py') }}-${{
-            hashFiles('prql-python/requirements.txt') }}
-      - name: Run tests
-        shell: bash
-        run: nox -s tests -f bindings/prql-python/noxfile.py
+      - run: pip install maturin
+      - name: Build wheel
+        run: task build-python-wheel
```
eitsupi, line 60: The wheels have already been built.

https://github.com/PRQL/prql/blob/8d2a55d23a81e17c8fd9a632a267ae95f21083a3/.github/workflows/test-python.yaml#L16-L44

max-sixty:
Thanks for reviewing — I should probably either finish this or merge it, sorry for leaving it hanging for so long...

max-sixty:
Note to future self (or if someone else finishes this) — we should add the ability to use `maturin develop`, since that installs the python package in a single step. 

CI etc uses `maturin build` + `pip install`, which is a more complete installation path and better simulates how it'll be used, but requires an extra step + a path change

max-sixty:
Most of this was added in a separate PR

PR #1621, opened 2023-01-25
fix: export constructor for SQLCompileOptions

### LINKED ISSUE #1620: [prql-js] null pointer passed to rust when using object created from `new SQLCompileOptions`
Hey, a exception is thrown when trying to construct and using the `SQLCompileOptions`. Reproduce code:

```typescript
import { Dialect, SQLCompileOptions } from "prql-js/dist/bundler";

function test() {
  const opts = new SQLCompileOptions;
  opts.dialect = Dialect.Sqlite // this call throws error `null pointer passed to rust`
}
```

This is because the `SQLCompileOptions` doesn't export the constructor object by default.

Ref: https://github.com/rustwasm/wasm-bindgen/issues/166

bcho:
This fixes #1620

### DIFF
--- a/prql-js/package-lock.json
+++ b/prql-js/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "prql-js",
-  "version": "0.3.1",
+  "version": "0.4.1",
   "lockfileVersion": 2,
   "requires": true,
   "packages": {
     "": {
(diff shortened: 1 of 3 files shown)


aljazerzen (approved):
Looks good to me. Thanks for contributing!

PR #1632, opened 2023-01-26
docs: Correcting the name of the python method

RalfNorthman:
The name of the compile method seems to have changed since the python bindings page was written:
https://pypi.org/project/prql-python/
This PR aims to correct that.


max-sixty:
Thanks @RalfNorthman !

PR #1644, opened 2023-01-30
fix(prql-compiler): ensure the formatted output end with a new line

eitsupi:
I noticed that with or without signature comment, it switches with or without a newline at the end of SQL string.
I think it is better to have a new line at the end, so I add a new line at the end even without signature comment.

I have noticed this behavior on R (prqlr).

```r
> "from a" |> prqlr::prql_compile(signature_comment = FALSE)
[1] "SELECT\n  *\nFROM\n  a"

> "from a" |> prqlr::prql_compile(signature_comment = TRUE)
[1] "SELECT\n  *\nFROM\n  a\n\n-- Generated by PRQL compiler version 0.4.1 (https://prql-lang.org)\n"
```

ToDo

- [x] Add test


aljazerzen:
LGTM

PR #1645, opened 2023-01-30
fix(prql-elixir): fix tests to match prql-compiler update

eitsupi:
Follow up #1644

I was not aware that the update in #1644 would cause the test to fail.
Perhaps tests for subprojects that depend on prql-compiler should also be run if the prql-compiler is updated in PRs......

### DIFF
--- a/prql-elixir/lib/prql.ex
+++ b/prql-elixir/lib/prql.ex
@@ -48,12 +48,12 @@ defmodule PRQL do
 
   Using default `Generic` dialect:
       iex> PRQL.compile("from customers", signature_comment: false)
-      {:ok, "SELECT\n  *\nFROM\n  customers"}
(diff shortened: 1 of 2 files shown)


max-sixty:
Thanks!

Yes, there's a tradeoff with completeness vs latency for the PR tests. We can always add the `pr-test-all` label to run all tests if we suspect there might be a failure. And all tests always run on merge, so we can revert the commit if something breaks we didn't expect...

eitsupi:
> We can always add the `pr-test-all` label to run all tests if we suspect there might be a failure.

Wow, that's convenient! I will make use of this next time, thanks!

max-sixty:
Great! Full details at https://prql-lang.org/book/contributing/development.html#how-we-test !

PR #1646, opened 2023-01-30
fix(playground): table -> let

eitsupi:
image

I didn't know where the playground is currently deployed from, so I send a PR to main for now.
Please cherry pick if needed.


max-sixty:
Thanks a lot @eitsupi 

---

Separately, we should take these from some canonical examples that are tested. It's not great that we had an example broken for so long

Issue #1588, opened 2023-01-19
Escaping curly braces inside S-strings


mklopets:
```elm
from x
derive y = = s"asd{blah"
```
for reasons(tm), one might need something like this to work, but I wasn't able to figure out how. Instinctively, I tried a backslash, but that didn't work.

This is related to #1225 but different, as that one technically has a known workaround (a very large number of quotes around S-strings 😄)

max-sixty:
Hmmm, I was excited to finally have an immediate answer for you @mklopets , but yet I disappoint myself — this works with matching braces, but weirdly not with unmatched braces: https://prql-lang.org/book/language-features/s-strings.html#braces

Let me know if you were indeed looking for non-matching braces, I can prioritize if so

max-sixty:
To confirm @mklopets — this works:

```elm
from employees
derive [
  has_valid_title = s"regexp_contains(title, '([a-z0-9]*-){{2,}}')"
]
```

...because it uses double braces in `{{2,}}` — does that suffice?

mklopets:
In our uber-specific case, our S-string replaced things inside of a string. X got replaced with `{` and Y got replaced with `}`. It's technically possible I could have added double curly braces and ordered these two replacements accordingly, but that would have made it extremely confusing, and does in fact only work if you need both of these, not one.

FWIW, the super-duper-specific workaround was to use the curly braces' ASCII codes to creates those strings (`chr(123)`)

I'm a huge proponent for us having a generic `\escape` pattern

Issue #1284, opened 2022-12-17
Don't format s-strings


max-sixty:
From #965, broadening that issue to being able to avoid the auto-formatting. For context, this is can be important with non-standard SQL, like Snowflake's semi-structured data.

A couple of options for how we could do this:
- Implement an "ignore" option upstream, in https://github.com/shssoichiro/sqlformat-rs/issues/15. I'm not sure whether upstream would accept a PR (though it seems reasonable, and worst case we could maintain a fork)
- Replace s-strings after the auto-formatting, using `$s_string_n`. This requires passing s-strings all the way through the compiler, reducing the modularity of the compiler phases, but otherwise being fairly simple.
- Relying on the new `Options` struct to disable formatting entirely. This has the disadvantage of being an all-or-nothing option, but might be an acceptable temporary solution.

aljazerzen:
Yeah, we should **never** format s-strings. I think the second in is the best option and we can fallback to using the third for the time being.

BlurrechDev:
I'm going to look at this. I will explore a fix for this based on the second option above to see how feasible this is.

max-sixty:
Great @BlurrechDev !

If that becomes unwieldy (i.e. we're passing a huge struct along), then we can reassess.

Feel free to post half-complete code — either for feedback or to merge something initial.

aljazerzen:
Great. If you need some pointers, this is how I'd do it:
- the SQL is generated [here](https://github.com/prql/prql/blob/main/prql-compiler/src/sql/mod.rs#L22-L24), so before that, I'd replace all s-strings in RQ AST with generated s-strings.
- for that, I'd implement [rq::RqFold](https://github.com/prql/prql/blob/main/prql-compiler/src/ast/rq/fold.rs#L23), similar to [CidRedirector](https://github.com/prql/prql/blob/main/prql-compiler/src/sql/anchor.rs#L461-L481).
- generated s-strings should be something that will never appear in actual queries. Max's suggestion of `$s_string_1, $s_string_2, $s_string_3, ...` is ok.
- the s-string-extractor must return the new AST and something like a `Vec<(String, String)>` (vec of pairs of generated and actual s-strings).
- after SQL is generated and formatted, replace the generated placeholders with actual s-strings using basic string replacement.

max-sixty:
This issue prevents folks from using the escape hatch, so bumping this to "Priority"...

aljazerzen:
[...]

PR #1650, opened 2023-01-30
feat: Support asymmetrical double brackets in s-strings

max-sixty:


max-sixty:
This should close https://github.com/PRQL/prql/issues/1588. But https://github.com/PRQL/prql/issues/1284 will still affect it a lot of the time

PR #1652, opened 2023-01-30
feat: WIP: parse dates with chumsky

max-sixty:


prql-compiler/src/parser/chumsky.rs
```diff
@@ -78,3 +85,35 @@ pub fn ident() -> impl Parser<char, Ident, Error = Simple<char>> {
...
+    use insta::assert_yaml_snapshot;
+    #[test]
+    fn test_date() {
+        let date = just('@').ignore_then(
+            digits(4)
+                .then_ignore(just('-'))
+                .then(digits(2))
+                .then_ignore(just('-'))
+                .then(digits(2)),
+        );
```
aljazerzen: This is arguably more readable than PEST derivations.

max-sixty: Yes it's not too bad!

prql-compiler/src/parser/chumsky.rs
```diff
@@ -78,3 +85,35 @@ pub fn ident() -> impl Parser<char, Ident, Error = Simple<char>> {
...
+        // - What's the type signature to extract `date` to a function?
+        // - Separately from this, but we could be much more liberal about
+        //   dates, this has been discussed on Twitter as a helpful feature.
+        //   (and then we could lint to a standard format, so we're coaching
+        //   users into simplicity)
+        assert_yaml_snapshot!(date.parse("@1984-01-01").unwrap(), @r###"
+        ---
+        - - "1984"
+          - "01"
+        - "01"
```
max-sixty: Does anyone know how we could get a sequence of `("1984", "01", "01")`?

Currently I can either get it nested with `.then` (i.e. now), or a vec of char with `.chain`, but I'm guessing I'm doing something wrong? Maybe it's collecting at the wrong place?

aljazerzen: I think of it this way:
- then combines `before: B` and `after: A` into tuples of `(before, after): (B, A)`,
- chain combines `before: T` and `after: Vec<T>` into vec `result: Vec<T>`.

If you have `just(a).then(b).then(c).then(d)`, the result will be `(((a, b), c), d)`.

If you have `just(a).then(just(b).then(just(c).then(d)))`, you get `(a, (b, (c, d))`

aljazerzen: I hope this is at least a little helpful.

max-sixty: Yes great. 

I had thought there might be a general solution, i.e. without the `.map(|((y, m), d)| (y, m, d));`. Something like `sequence(vec![digits(2), digits(4), digits(2)])` (ignoring the `-` for a second). 

But totally fine without one obv. Thanks

max-sixty: ...because if we get years, months, days... all the way to microseconds, then that `map` is going to be quite the closure...

aljazerzen: If AST will contain this as string anyway, we can join it back as soon as possible.

I've pushed a commit that does this.

max-sixty: Yes, if the whole thing is a string, then `.chain` seemed to work well. This approach too.

And at the moment we do just store it as a string in the AST.

WDYT about moving the date parsing to this? I can do that soon-ish; then we'll have our first bit of logic in chumsky.

aljazerzen: Yeah, I think we could simplify pest definitions to something like:
```
datetime = { "@" ~ (ASCII_DIGIT | ":" | "T" | "+" | "-" | ".")* ~ &end_expr }
```

But it may be prone to errors...

max-sixty: Right — make the pest definition less strict (maybe almost `datetime = { "@" ~ [any] }`, and then pass it to Chumsky to parse?

prql-compiler/src/parser/chumsky.rs
```diff
@@ -78,3 +89,37 @@ pub fn ident() -> impl Parser<char, Ident, Error = Simple<char>> {
...
+
+    use insta::assert_yaml_snapshot;
+    #[test]
+    fn test_date() {
+        let date = just('@')
+            .ignore_then(digits(4))
+            .then_ignore(just('-'))
+            .then(digits(2))
+            .map(join::<'-'>)
+            .then_ignore(just('-'))
```
max-sixty, line 105: Not a huge deal — maybe we should just proceed and then come back and refine — but it's seems repetitive to have the `.then_ignore(just('-'))` and then `join::<'-'>`.

@zesterer feel free to ignore, but let us know if you see this + there's a canonical chumsky solution...

prql-compiler/src/parser/chumsky.rs
```diff
@@ -78,3 +89,37 @@ pub fn ident() -> impl Parser<char, Ident, Error = Simple<char>> {
...
+
+#[cfg(test)]
+mod test {
+    use super::*;
+
+    use insta::assert_yaml_snapshot;
+    #[test]
+    fn test_date() {
+        let date = just('@')
+            .ignore_then(digits(4))
```
zesterer, line 101: To be clear, `digits(4)` parses any number of base 4 digits (0/1/2/3), not 4 digits. You'll probably want to use `filter(|c| c.is_digit()).repeated().exactly(4)` instead (in the up-coming 0.9 version you'll also be able to do `digits(10).exactly(4)` also).

prql-compiler/src/parser/chumsky.rs
zesterer: There's not really a more canonical solution. For homogenous items you can do `item.separated_by(separator)`, but in this case you have heterogeneous items.

That said, it *looks* like the overall pattern is to preserve the original string, yes? If so, you may prefer something akin to the `chain` approach, such as is used here: https://github.com/zesterer/chumsky/blob/master/examples/json.rs#L29-L37.

Once `chumsky/zero-copy` is merged (probably chumsky 0.10) you'll also be able to use `.map_with_slice` to get access to the original section of the input that corresponds to the parser too.

prql-compiler/src/parser/chumsky.rs
max-sixty: (this refers to https://github.com/PRQL/prql/pull/1652/files#diff-bb3391b38da46712662bacac63ebceb0265e1070a736620862193eed84fd1640R71-R76 — I think maybe this conflicts with a `chumsky` function name :) )

prql-compiler/src/parser/chumsky.rs
max-sixty: Super, thanks a lot @zesterer

prql-compiler/src/parser/chumsky.rs
zesterer: Ah, ignore me then :grinning:

PR #1662, opened 2023-02-02
feat: Small playground improvements

### LINKED ISSUE #1533: Playground: "Copy to clipboard" button copies output.sql and not tab currently selected
It would be nice if that button could copy the currently selected tab, e.g. output.arrow, output.pl.json, or output.rq.json, instead of always copying output.sql.

This is not a high priority though.

### LINKED ISSUE #1534: Playground: Replace output.{pl,rq}.json with output.{pl,rq}.yaml
I tried to inspect the output.pl.json and output.rq.json output for some queries to get know the PL and RQ structures and I found the JSON output very difficult to parse (for at least this)/a human. Once I passed it through https://www.json2yaml.com/ it was much more readable.

I think the playground output is probably for human consumption so there YAML would make more sense. 

For the APIs it makes sense to keep it as JSON since those will probably be consumed by machines.

Not a priority.

nicot:
This improves a couple small things in the playground:
- Increase the editor font size to be the same size as the output tab to make it easier to read (they are still different fonts, so it doesn't look perfect). I want to follow up with a PR to change the output to be rendered in a readonly Monaco window.
- Copy the contents of the active tab (fixes #1533).
- Swap the output from JSON to YAML (fixes #1534).

Screenshot 2023-02-02 at 8 56 39 AM

This is my first contribution, so please let me know if there's anything I'm missing. I can also update the changelog if it'd be helpful.

### DIFF
--- a/playground/package-lock.json
+++ b/playground/package-lock.json
@@ -20,7 +20,8 @@
         "react-dom": "^18.2.0",
         "react-syntax-highlighter": "^15.5.0",
         "wasm-react-scripts": "5.0.3",
-        "web-vitals": "^3.1.0"
+        "web-vitals": "^3.1.0",
+        "yaml": "^2.2.1"
       },
       "devDependencies": {
         "npm-watch": "^0.11.0"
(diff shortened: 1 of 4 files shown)


max-sixty:
This is great @nicot ! Thank you. I'm impressed you did it so succinctly.

snth:
Thank you very much for these @nicot ! 🙏

PR #1667, opened 2023-02-03
docs: Use bold font for the website header

max-sixty:
I realize I'm behind on stuff, so making cosmetic changes is probably not the most productive thing I could be doing.

But I thought the original website header was cleaner, what do others think? I don't trust myself here, so we can default to closing unless others agree. 

image

### DIFF
--- a/website/themes/prql-theme/static/style.css
+++ b/website/themes/prql-theme/static/style.css
@@ -167,6 +167,7 @@ pre ::-webkit-scrollbar-thumb:hover {
 #header a:not(.btn-icon) {
   color: var(--main-color);
   transition: 0.2s;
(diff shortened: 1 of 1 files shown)


aljazerzen:
Bold looks better, but it's missing an underline. It's a link.

max-sixty:
Is that the norm for headers though? I'm as far from an expert as it gets, though I'm looking at GitHub or Amazon.com...

aljazerzen:
https://www.rust-lang.org/

It's was also a design rule at my previous company. It has to look like a button (with borders and stuff) or it has to be a link.

max-sixty:
Ok! I'll close

PR #1668, opened 2023-02-03
build: publish website from website branch

aljazerzen:


max-sixty:
Weren't we planning to use `stable` as the name?

(I'm fine with this though if you have a preference)

aljazerzen:
Stable implies that it contains the latest stable version of the language. But the main point of the branch is to be the head for website releases.

Issue #1666, opened 2023-02-03
Write playground query examples using chinook data


max-sixty:
Currently we have two sets of examples in the playground:

1. Full queries
    - Three canonical queries — `introduction`, `employees`, `cte`, which aren't based on real tables
    - An example which uses the chinook tables — `artists`
2. Some `take 10` tables which show the tables

It would be great to move the canonical queries to tables that can use the chinook data — then folks can see the actual output. That would involve changing the current queries to instead use the tables, while retaining all the language features that the queries demonstrate. 

We don't necessarily need four queries — if we can demonstrate language features with one or two, great. Another benefit of this is that a "music" database is arguably friendlier than an "employees" database. And possible it would help us clean up the playground UI.

So this would involve (fine to do partial parts ofc):
- Writing queries that are based on the chinook data and demonstrate PRQL language features
- Replace the canonical queries in the playground
- Replace the canonical queries on the website / readme / tests

PR #1676, opened 2023-02-04
docs: adapt example queries to chinook

aljazerzen:
Start on #1666

### DIFF
--- a/playground/src/examples.js
+++ b/playground/src/examples.js
@@ -1,69 +1,57 @@
 const examples = {
   "introduction.prql": [
-    "sql",
-    `from employees
-filter country_code == "USA"   # Each line transforms the previous result.
-derive [                       # This adds columns / variables.
-  gross_salary = salary + payroll_tax,
-  gross_cost = gross_salary + benefits_cost  # Variables can use other variables.
+    "arrow",
+    `from invoices
+filter billing_country == "USA" # Each line transforms the previous result
+derive [                        # This adds columns
+  transaction_fees = 0.8,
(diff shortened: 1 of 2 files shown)


website/content/_index.md
```diff
@@ -12,24 +12,28 @@ hero_section:
     link: https://prql-lang.org/book/
     label: "Reference"
   prql_example: |
-    from employees
-    filter start_date > @2021-01-01
```
max-sixty, line 16: WDYT about keeping a date one? I think that's a nice PRQL feature

playground/src/examples.js
```diff
@@ -1,69 +1,57 @@
...
-select [dept_name, title, avg_salary]
+# anti-join soundtracks
+join side:left s=soundtracks [==track_id]
+filter s.track_id == null
+
+# limit to kicker genres
+join g=high_energy [==genre_id]
+
+# format output
+select [t.track_id, track = t.name, genre = g.name]
```
max-sixty, line 53: Should our standard formatting have a space here? I generally find having no space keeps the lvalue & rvalue visually bound in a list of lots of items like this (minor point obv!)

max-sixty (approved):
Very nice! Thanks

website/content/_index.md
aljazerzen: Yup, good point.

playground/src/examples.js
aljazerzen: We can have an example showing off different styles we allow. Sometimes it's useful to segment things you are doing into understandable chunks.

PR #1679, opened 2023-02-04
docs: Added doctest and use readme for prql rust bindings in book

nkicg6:
@max-sixty this set of changes is related to what we discussed in #1567. 

1. To keep the PRQL Rust bindings page in the book in sync with the prql-compiler readme, I mimicked the js bindings approach (see [bindings/javascript.md](https://github.com/PRQL/prql/blob/2f6df0c2c2d6e9d1aa1b4311ad1e002b3ac9b474/book/src/bindings/javascript.md?plain=1#L1)) and simply `{#include ...}` the prql-compiler readme. I also made a few detail and stylistic changes to the readme. 
2. To ensure the readme and book example don't break with future changes, the readme is pulled into `lib.rs` and run as a doctest (using [this rustdoc feature](https://doc.rust-lang.org/stable/rustdoc/write-documentation/documentation-tests.html#include-items-only-when-collecting-doctests) see [lib.rs](https://github.com/nkicg6/prql/blob/a980af2b72040c3fe99946d5530e781d24c385cb/prql-compiler/src/lib.rs#L130-L132)).
3. I made the rustdoc example from #1567 an [actual doctest](https://github.com/nkicg6/prql/blob/a980af2b72040c3fe99946d5530e781d24c385cb/prql-compiler/src/lib.rs#L107-L118)

One thing I don't love about this is that change 2 duplicates the [rustdoc doctest](https://github.com/nkicg6/prql/blob/a980af2b72040c3fe99946d5530e781d24c385cb/prql-compiler/src/lib.rs#L130-L132) in change 3. I think it is useful to have that example in the rust docs page, and I couldn't figure out how to include the readme example and have it also show up in the rust docs, so I left it as is.


prql-compiler/src/lib.rs
```diff
@@ -128,6 +127,10 @@ pub fn compile(prql: &str, options: Option<sql::Options>) -> Result<String, Erro
         .map_err(|e| e.composed("", prql, false))
 }
 
+#[doc = include_str!("../README.md")]
```
max-sixty, line 130: That's clever, I didn't know this was possible

prql-compiler/README.md
```diff
@@ -1,20 +1,51 @@
...
+  };
+
+  let sql = compile(&prql, Some(opt)).unwrap();
+  assert_eq!("SELECT name, age FROM employees", sql);
+```
+
 ## Terminology
 
-Relation = Standard definition of a relation in context of databases:
+[_Relation_](<https://en.wikipedia.org/wiki/Relation_(database)>): Standard
```
max-sixty, line 38: Great!

prql-compiler/README.md
```diff
@@ -1,20 +1,51 @@
...
+```
+
+## Examples
+
+Compile a PRQL string to a SQLite dialect string.
+
+**src/main.rs**
+
+```rust
+  use prql_compiler::{compile, sql};
```
max-sixty: Very minor but should this block be indented?

prql-compiler/README.md
```diff
@@ -1,20 +1,51 @@
-# PRQL compiler
+# Rust PRQL compiler
```
max-sixty: Maybe just:

```suggestion
# PRQL compiler
```

As there's only one? Or if you think the currently proposed version is clearer then fine to leave

prql-compiler/src/lib.rs
```diff
@@ -105,18 +105,17 @@ pub static PRQL_VERSION: Lazy<Version> =
...
-///     format: true,
+///     format: false,
 ///     dialect: Some(sql::Dialect::SQLite),
-///     signature_comment: true
+///     signature_comment: false
 /// };
 /// let sql = compile(&prql, Some(opt)).unwrap();
-/// println!("PRQL: {}\nSQLite: {}", prql, sql);
+/// println!("PRQL: {}\nSQLite: {}", prql, &sql);
+/// # assert_eq!("SELECT name, age FROM employees", sql)
```
max-sixty: Is this commented out intentionally?

max-sixty (approved):
Very nice, thanks a lot @nkicg6 !

I added a couple of small comments; but depending on your appetite it's fine to merge now. You can also add a line to the Changelog if you're up for it

prql-compiler/src/lib.rs
max-sixty: > > One thing I don't love about this is that change 2 duplicates the [rustdoc doctest](https://github.com/nkicg6/prql/blob/a980af2b72040c3fe99946d5530e781d24c385cb/prql-compiler/src/lib.rs#L130-L132) in change 3. I think it is useful to have that example in the rust docs page, and I couldn't figure out how to include the readme example and have it also show up in the rust docs, so I left it as is.

Ah, so it tests it, but it doesn't appear in the documentation itself? It's still much better even though it's duplicated.

nkicg6: right! I like it too, I stumbled on it earlier today in the rustdoc docs. I think it fits this use case well by linking the readme, library, and book bindings to a test. Thanks for your edits and review I'll incorporate those shortly.

prql-compiler/src/lib.rs
nkicg6: The `#` comment just [hides that line](https://doc.rust-lang.org/rustdoc/write-documentation/documentation-tests.html#hiding-portions-of-the-example) in the generated rustdocs or if you were to include the example it in another file. I added it when I was trying to just include this example directly in the book, but I couldn't get it to work properly in with the `{#rustdoc_include ...}` so it isn't needed anymore. I'll update.

max-sixty:
Awesome, thanks a lot @nkicg6 !

PR #1684, opened 2023-02-05
refactor!: use target instead of dialect in public API

aljazerzen:


max-sixty:
Thanks a lot!

This might have been one to add the `pr-test-all` label — let me see if I can fix the Elixir test quickly

max-sixty:
I'll revert to fix `prql-elixr` and then re-revert

prql-compiler/src/lib.rs
```diff
@@ -105,28 +106,83 @@ pub static PRQL_VERSION: Lazy<Version> =
...
 /// };
-/// let sql = compile(&prql, Some(opt)).unwrap();
+/// let sql = compile(&prql, opt).unwrap();
 /// println!("PRQL: {}\nSQLite: {}", prql, &sql);
 /// assert_eq!("SELECT name, age FROM employees", sql)
 ///
 /// ```
 /// See [`sql::Options`](sql/struct.Options.html) and [`sql::Dialect`](sql/enum.Dialect.html) for options and supported SQL dialects.
-pub fn compile(prql: &str, options: Option<sql::Options>) -> Result<String, ErrorMessages> {
+pub fn compile(prql: &str, options: Options) -> Result<String, ErrorMessages> {
```
max-sixty, line 123: @aljazerzen I don't have a strong view on whether we take an `Option` or force folks to provide a `Default`, but was this intentional?

aljazerzen: I realized that this was a bad api design, because None here has no semantic value since we have `Options::default()` anyways.

max-sixty:
CC @eitsupi 

hopefully this doesn't cause much churn in `prqlr` — it was quite small changes in the in-repo bindings.

eitsupi:
@max-sixty Thank you for your care and concern.
I noticed this yesterday and have already completed the prqlr update 👍(eitsupi/prqlr#71)

Since prqlr specifies the target as a string, I changed the method of specifying the target name from something like `"sqlite"` to something like `"sql.sqlite"`, as mentioned https://github.com/PRQL/prql-vscode/issues/48#issuecomment-1416887856.
Please let me know if you need to make any changes, as I would like to unify the API with other languages as much as possible.

max-sixty:
Very impressive you saw this and already made changes!

We'll be more deliberate about these designs in the future — I think we're now calibrating that these have more downstream impacts than just an internal code change. I'm going to add a note to our docs.

PR #1688, opened 2023-02-05
build: Add `prql-elixir` to workspace

max-sixty:
(tests will fail because of #1684, but this makes working in the crate easier, including rust-analyzer support)


max-sixty:
FYI @kasvith I did this! Hope that's OK

kasvith:
LGTM @max-sixty

Issue #1593, opened 2023-01-20
Test examples on website


eitsupi:
The following query, which appears on the top page of the website, will cause an error with the latest prql-compiler.

https://github.com/PRQL/prql/blob/18524c343bd1e8e39399e26a7747c8ffc4e60a67/website/content/_index.md?plain=1#L234-L249

Perhaps files that have been tested externally need to be imported here?

max-sixty:
Thanks for the spot @eitsupi .

I fixed that example. I agree these could be parsed and tested, possibly in a similar way to how we do this for the book.

If it's OK, I'll change the title of the issue to one about testing and leave it open. 

If anyone wants a modest issue to work on without much complex rust, this could be a good one — parse the YAML file, extract the PRQL, check it compiles — then that would be very welcome.

nkicg6:
@max-sixty I took a very rough swing at this using string comparisons and println! debugging (still a WIP, haven't written the actual tests yet) to get an idea of the current status and how it could work. You can find it in my fork under the [website branch](https://github.com/nkicg6/prql/tree/website/website/website-tests).

The results I got were:
```
Passed: 4, Failed: 8
```

Just looking through the failures, some were stylistic and could be fixed by editing the expected SQL, for example the `orthogonal` test:
```
Failed on ID: 'orthogonal'
Expected: SELECT country, MAX(salary) AS max_salary FROM employees WHERE start_date > DATE '2021-01-01' GROUP BY country HAVING MAX(salary) > 100_000
Got:      SELECT country, MAX(salary) AS max_salary FROM employees WHERE start_date > DATE '2021-01-01' GROUP BY country HAVING MAX(salary) > 100000
```
will pass if you drop the stylistic `_` from the SQL. 

Others look like they might be a SQL style difference, and one caused a prql_compiler::compile error (`s-string`). A few questions:
1. What SQL dialect should we use to check the output? For these tests I didn't pass a dialect, but it might make sense to specify one for reproducibility. 
2. How do you want to handle stylistic differences (e.g., `orthogonal`) do we change the expected SQL if it is equivalent?
3. How do you envision this running? To get started I made a new binary in the website directory, but I think it makes more sense to have it as a library in that directory and write tests that run with `cargo tests`. But should we add it to run before the website build in the taskfile, or add it to the website CI... thoughts?

Happy for any feedback or advice here, let me know if I'm on the right track.

max-sixty:
> will pass if you drop the stylistic `_` from the SQL.

[...]

PR #1691, opened 2023-02-05
docs: Fix SQL number in showcase

max-sixty:


max-sixty:
Thanks to @nkicg6 for finding this: https://github.com/PRQL/prql/issues/1593#issuecomment-1418043890

PR #1707, opened 2023-02-06
fix: MacOS test for prql-elixir

kasvith:
Fixes tests for macos

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -2,6 +2,7 @@
 runner = 'wasm-bindgen-test-runner'
 
 [target.'cfg(target_os = "macos")']
+# Required for prql-elixir on Mac
 rustflags = [
     "-C", "link-arg=-undefined",
     "-C", "link-arg=dynamic_lookup",
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -1,2 +1,8 @@
 [target.wasm32-unknown-unknown]
 runner = 'wasm-bindgen-test-runner'
+
+[target.'cfg(target_os = "macos")']
+rustflags = [
+    "-C", "link-arg=-undefined",
+    "-C", "link-arg=dynamic_lookup",
+]
--- a/.github/workflows/test-all.yaml
+++ b/.github/workflows/test-all.yaml
@@ -41,6 +41,9 @@ jobs:
   test-js:
     uses: ./.github/workflows/test-js.yaml
 
+  test-elixir:
+    uses: ./.github/workflows/test-elixir.yaml
+
   measure-code-cov:
     # Currently disabled due to https://github.com/actions-rs/tarpaulin/issues/21
     if: false
--- a/.github/workflows/test-elixir.yaml
+++ b/.github/workflows/test-elixir.yaml
@@ -55,8 +55,7 @@ jobs:
           cache-name: cache-elixir-deps
         with:
           path: prql-elixir/deps
-          key:
-            ${{ runner.os }}-mix-${{ env.cache-name }}-${{
+          key: ${{ runner.os }}-mix-${{ env.cache-name }}-${{
             hashFiles('**/mix.lock') }}
           restore-keys: |
             ${{ runner.os }}-mix-${{ env.cache-name }}-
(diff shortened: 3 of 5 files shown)


kasvith:
@max-sixty precommit failing again :(

prql-elixir/native/prql/Cargo.toml
```diff
@@ -1,7 +1,6 @@
 [package]
 authors = ["Kasun Vithanage <kasvith.me@gmail.com>"]
 description = "Elixir NIF bindings for prql-compiler"
-# TODO: should this be `prql-elixir`?
```
kasvith, line 4: crate name is used by the elixir rustler for integration

kasvith:
There seems to be a problem with `build-all` with apple silicon

.github/workflows/test-elixir.yaml
```diff
@@ -55,8 +55,7 @@ jobs:
           cache-name: cache-elixir-deps
         with:
           path: prql-elixir/deps
-          key:
-            ${{ runner.os }}-mix-${{ env.cache-name }}-${{
+          key: ${{ runner.os }}-mix-${{ env.cache-name }}-${{
```
max-sixty: Because pre-commit's CI can't fix files in the `workflows` path, this change in formatting is causing it to fail.

Does `pip install pre-commit` or `brew install pre-commit` not work for you? Running locally is unfortunately the only way of fixing the files in the `workflows` path.

(I fixed it myself this time, I'm fine to keep doing it! I ask so you can be independent :) )

max-sixty:
This is looking excellent @kasvith ! Let's see if it passes

.cargo/config.toml
```diff
@@ -1,2 +1,8 @@
 [target.wasm32-unknown-unknown]
 runner = 'wasm-bindgen-test-runner'
+
+[target.'cfg(target_os = "macos")']
+rustflags = [
```
max-sixty, line 6:
```suggestion
# Required for prql-elixir on Mac
rustflags = [
```

.github/workflows/test-all.yaml
```diff
@@ -41,6 +41,9 @@ jobs:
   test-js:
     uses: ./.github/workflows/test-js.yaml
 
+  test-elixir:
+    uses: ./.github/workflows/test-elixir.yaml
```
max-sixty, line 45: Great!

.github/workflows/test-elixir.yaml
```diff
@@ -97,3 +97,65 @@ jobs:
...
+    name: macos-latest test
+    runs-on: macos-latest
+    steps:
+      - name: Checkout sources
+        uses: actions/checkout@v3
+
+      - name: Install Erlang/Elixir
+        run: |
+          brew install elixir
+          mix local.hex --force
```
max-sixty: Is this the only difference between Mac & the others? Would you be up for using an `if` statement like we do here instead of copying & pasting the rest of the block? https://github.com/PRQL/prql/blob/ff7ac7a880e3f18c7bec4b4e78aed0a7ae2b52fa/.github/workflows/test-rust.yaml#L28

.github/workflows/test-elixir.yaml
kasvith: Oh i forgot the command to do this, thanks max

.github/workflows/test-elixir.yaml
kasvith: this workflow installs the elixir using a different method. other version with a matrix uses a strategy which can use multiple elixir versions

I used this from rustler https://github.com/rusterlium/rustler/blob/80f6c4b926ce15dea4fac5389a67decdc9732ba8/.github/workflows/main.yml#L135

.github/workflows/test-elixir.yaml
```diff
@@ -41,12 +40,19 @@ jobs:
         uses: actions/checkout@v3
 
       # Step: Setup Elixir + Erlang image as the base.
-      - name: Set up Elixir
+      - name: Set up Elixir on Windows or Linux
+        if: "{{ runner.os != 'macOS' }}"
```
max-sixty: @kasvith I made a change here; I think this should reduce the duplication, lmk if that makes sense

.github/workflows/test-elixir.yaml
max-sixty: I adjusted in another commit

max-sixty:
Thanks @kasvith ! Feel free to add an entry to the changelog in another PR. Merging!

.github/workflows/test-elixir.yaml
kasvith: Thanks max, makes sense

kasvith:
@max-sixty done in #1726

PR #1708, opened 2023-02-06
feat(prqlc): watch

aljazerzen:
- refactor: remove prqlc lib
- feat(prqlc): watch


prql-compiler/prqlc/Cargo.toml
```diff
@@ -8,10 +8,6 @@ repository.workspace = true
 rust-version.workspace = true
 version.workspace = true
 
-[lib]
-doctest = false
```
max-sixty, line 12: The reason I added these around is that the creating the binaries & linking them takes the majority of the time on incremental builds — in particular when running tests when watching the path. And so following https://matklad.github.io/2021/02/27/delete-cargo-integration-tests.html advice, it makes sense to disable these if we're not using them

max-sixty:
Very cool indeed @aljazerzen ! I'd been doing something similar with `watchexec`; much nicer to have it built in!

Issue #1712, opened 2023-02-06
Formal definition of operators and implict casting


max-sixty:
Found by @nkicg6 in #1593 :

```elm
from weather
select temp_f = temp * (9/5)
```

```sql
SELECT
  temp * 1 AS temp_f
FROM
  weather
```

...I guess this is because we're parsing them as ints and doing int division in the complier?

aljazerzen:
Is this not expected behavior?

`(9/5)` is equivalent to `1` in PRQL, right?

max-sixty:
> `(9/5)` is equivalent to `1` in PRQL, right?

I definitely hadn't thought so!

Are you thinking that `/` should be int division, like python2? Or that we shouldn't silently convert ints to floats?

I think this would be quite unexpected from most users...

aljazerzen:
```
➜  ~ python3.11
Python 3.11.1 (main, Dec  7 2022, 01:11:44) [GCC 9.4.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> 9/5
1.8
>>> 9//5
1
>>>
➜  ~ node
Welcome to Node.js v19.4.0.
Type ".help" for more information.
> 9/5
1.8
>
➜  ~ echo 'fn main() { println!("{}", 9 / 5); }' > a.rs && rustc a.rs && ./a
1
➜  ~
➜  ~ sqlite3 db.sql "SELECT 9/7"
1
➜  ~ psql -U postgres -d chinook -c "SELECT 9/7"
 ?column?
----------
        1
(1 row)
```

There doesn't seem to be consensus here, so the question is: do we want implicit type conversions?

This should probably be decided in broader context of how we want to define semantics of operators and do we want to support operator overloading?

So I'd suggest we postpone this issue until typing system is decided.

nkicg6:
Interesting, yes Python changed integer division from `/` to `//` in the python2->python3 move to make `/` represent "true" division (see [pep 238](https://peps.python.org/pep-0238/)). This matches the behavior of most dynamic languages in my experience, so most dynamic language users likely "expect" this behavior (see your python and node examples), so I'd suspect most users would be surprised by the prql behavior here despite it being technically correct. 

But databases (postgres and sqlite at least) will default to integer division unless you are explicitly dividing floats. 

```
sqlite> CREATE TABLE weather (temp REAL);
sqlite> INSERT INTO weather (temp) VALUES (20),(25.5),(30);
[...]

PR #1719, opened 2023-02-07
test: Add website query tests for prql-python and update expected sql

nkicg6:
This is a work in progress, related to #1593. Added tests for prql-python to exercise all the example queries from the website. As discussed in #1593, two of these tests are failing: `functions` and `s-string`.
[`functions`](https://github.com/PRQL/prql/blob/18524c343bd1e8e39399e26a7747c8ffc4e60a67/website/content/_index.md?plain=1#L182-L193) is a likely a prql bug related to #1712, but [`s-string`](https://github.com/PRQL/prql/blob/18524c343bd1e8e39399e26a7747c8ffc4e60a67/website/content/_index.md?plain=1#L224-L232) isn't compiling for me and I couldn't figure out how to fix the query to match (maybe @max-sixty can help?). 

I added one dependency to python (pyyaml) to parse the hugo header (`_index.md`) with all the examples (`pip install -r requirements.txt`). Besides the two failing tests, this will likely also break the [python CI](https://github.com/PRQL/prql/blob/main/.github/workflows/test-python.yaml) since it is referencing a file outside of prql-python and uses unix-y paths. 

Rust version is on the way later this week, but we probably have to think about how to reference the website from outside the crate to test there in CI as well.


website/content/_index.md
```diff
@@ -101,9 +101,9 @@ showcase_section:
         sort [-amount]           # sort order
       sql: |
         SELECT
-          order.*
+          *
         FROM
-          order
+          "order"
```
max-sixty: This would be clearer as `orders`, if you'd be up for changing it in the PRQL...

prql-python/python/tests/test_website_queries.py
```diff
@@ -0,0 +1,144 @@
...
+        "top-n",
+        "s-string",
+        "joins",
+        "null-handling",
+        "dialects",
+    ].sort()
+    assert EXPECTED_EXAMPLES == website_examples
+
+
+def test_basics(example_queries):
```
max-sixty: We can probably do these with a loop! That would cut down the amount of code a lot.

prql-python/python/tests/test_website_queries.py
```diff
@@ -0,0 +1,144 @@
...
+        "orthogonal",
+        "f-strings",
+        "windows",
+        "functions",
+        "top-n",
+        "s-string",
+        "joins",
+        "null-handling",
+        "dialects",
+    ].sort()
```
max-sixty: We could probably generate this list dynamically rather than list it explicitly — i.e. in a loop, run something like

```python
for showcase_item in example_queries:
    compiled = prql.compile(item["prql"])
    compiled_normalized = normalize(compiled)
    truth_normalized = normalize(item["sql"])
    assert truth_normalized == compiled_normalized
```

prql-python/requirements.txt
```diff
@@ -1,3 +1,4 @@
 pytest==7.0.0
 pytest-xdist==2.5.0
 pytest-clarity==1.0.1
+PyYAML==6.0
```
max-sixty, line 5:
```suggestion
# Required for website example test
PyYAML==6.0
```

prql-python/python/tests/conftest.py
```diff
@@ -0,0 +1,13 @@
+import pytest
+
+import yaml
+
+
+@pytest.fixture()
+def example_queries():
+    book_path = "../website/content/_index.md"
```
max-sixty: Technically these aren't actually in the book; they're only on the website...

```suggestion
    website_path = "../website/content/_index.md"
```

max-sixty:
Thanks a lot! I think this is good! No need to do a rust version for sure.

I made a couple comments, including one to cut down the code a bit; let me know if you agree.

For the s-string one, would you be up for just adding a `from`, like this:

```
from x
derive s"version()"
```

That should fix it.

For the functions one — so we can merge this, do you want to change `9/5` to `1.8`? 

The CI seems to work well! I think it should be fine to reference something outside the path...

prql-python/python/tests/test_website_queries.py
nkicg6: I went back and forth on this, the first version was exactly as you described, one function that looped through all the cases. But when some queries were broken, the tests would quick as soon as it failed. I wanted an overview of which ones were failing, so I used this verbose separate test for each query method.
 
But now that we've worked through these I can switch back to the all in a loop version.

nkicg6:
I made all the suggested changes, it looks like the pre-commit CI check is looking for the type annotations for pyyaml for the mypy check. I can look into that this evening but feel free to point me in the right direction if you have time

prql-python/requirements.txt
```diff
@@ -1,3 +1,6 @@
 pytest==7.0.0
 pytest-xdist==2.5.0
 pytest-clarity==1.0.1
+# Required for website example tests
+# Required for website example test
+PyYAML==6.0
```
max-sixty, line 5:
```suggestion
PyYAML==6.0
types-PyYAML==6.0.12.4
```

prql-python/requirements.txt
```diff
@@ -1,3 +1,7 @@
 pytest==7.0.0
 pytest-xdist==2.5.0
 pytest-clarity==1.0.1
+# Required for website example tests
+# Required for website example test
```
max-sixty:
```suggestion
```

nkicg6:
had to add the types via the pre-commit `additional_dependencies` (https://github.com/pre-commit/mirrors-mypy)

prql-python/python/tests/test_website_queries.py
max-sixty: Yes I empathize! 

For future pytest endeavors — it is possible to generate the list as fixture values, and then the tests can run independently. But the current state is great.

max-sixty:
@nkicg6 I'll merge — thank you very much; this is a great contribution.

max-sixty:
(We'll release soon, so the conflict won't matter. This is one downside of the backport branch approach — if you forget one, the next one / ones won't work without resolving conflicts locally)

PR #1721, opened 2023-02-07
docs: opensource website fonts

### LINKED ISSUE #1713: Font files in website
It looks like the font files are included in the source code of the website, which is probably not compatible with the Apache-2.0 license.
(As far as I could find, I couldn't even tell if redistribution of this is allowed)

https://github.com/PRQL/prql/blob/7c96513a2144671ca90e6328ac01b0d4f518e427/website/themes/prql-theme/static/fonts/AvenirNext.woff
https://github.com/PRQL/prql/blob/7c96513a2144671ca90e6328ac01b0d4f518e427/website/themes/prql-theme/static/fonts/AvenirNext.woff2

Perhaps we need to remove these from the source code?

aljazerzen:
Closes #1713


max-sixty:
I'll publish this and we can see how it looks, let's not merge yet

richb-hanover:
I'm always curious: what's the difference between the two fonts? Not much...
Inter Extra Bold (in `main`; open-source) on the left; Avenir (in the currently`web` branch; not open-source) on the right.

image

aljazerzen:
There is a difference, which typographers will see, but we coders won't :smile: 

I think Avenir is more "geometric", which I understand to mean that all lines have the same width and use simpler geometric curves (circles instead of elipses, look at 'o' for example).

Inter is also more condensed.

It's also has a hidden curse of knowledge, which makes you notice bad font choices everywhere, and you can never go back.

richb-hanover:
Yes. And XKCD always has a relevant observation:

https://xkcd.com/1015/

Enjoy!

PR #1722, opened 2023-02-07
feat(prqlc)!: preprocess Jinja templates

aljazerzen:
See demo repo: https://github.com/aljazerzen/jaffle_shop

### Method

We start with Jinja-PRQL:
```
from {{ ref('my_model') }}
```
... parse Jinja and extract interpolations:
```
from _jinja_0
```
... compile PRQL:
```
SELECT * FROM _jinja_0
```
... replace anchor with original Jinja expressions:
```
SELECT * FROM {{ ref('my_model') }}
```

### Minijinja

I used [minijinja](https://github.com/mitsuhiko/minijinja) to lex the input into tokens. For that I had to [fork](https://github.com/aljazerzen/minijinja) the project so it exposes the lexer. 

### Problems

This approach does not work any jinja expression that would not produce exactly one output token. This includes for loops and if statements.

We could do full Jinja evaluation and provide all functions that are available in dbt, but set them to evaluate to anchor tokens just as I do now. This would make for loops work, but could not fully emulate evaluation that depends on any state that is provided by dbt.

I don't know much about dbt, should I'd ask for people to check it out and try to translate a few dbt projects.

### Ideal solution

Have access to dbt state, i.e. compile PRQL **after** dbt evaluates the Jinja template.

### DIFF
--- a/prql-compiler/prqlc/Cargo.toml
+++ b/prql-compiler/prqlc/Cargo.toml
@@ -20,7 +20,7 @@ itertools = "0.10.3"
 notify = "^5.1.0"
 minijinja = {git = "https://github.com/aljazerzen/minijinja"}
 prql-compiler = {path = '..', version = "0.5.2"}
-regex = {version = "1.7", features = ["std", "unicode", "pattern"]}
+regex = {version = "1.7.1", features = ["std", "unicode"]}
 serde_json = "1.0.81"
 serde_yaml = "0.9.1"
 serde = "^1"
@@ -28,4 +28,3 @@ walkdir = "^2.3.2"
 
 [target.'cfg(not(target_family="wasm"))'.dev-dependencies]
 insta = {version = "1.28", features = ["colors", "glob", "yaml"]}
-similar-asserts = "1.4.2"
--- a/prql-compiler/prqlc/Cargo.toml
+++ b/prql-compiler/prqlc/Cargo.toml
@@ -18,10 +18,14 @@ color-eyre = "0.6.1"
 env_logger = {version = "0.9.1", features = ["termcolor"]}
 itertools = "0.10.3"
 notify = "^5.1.0"
-prql-compiler = {path = '..', version = "0.5.2" }
+minijinja = {git = "https://github.com/aljazerzen/minijinja"}
+prql-compiler = {path = '..', version = "0.5.2"}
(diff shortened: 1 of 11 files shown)


aljazerzen:
Because of mentioned limitations, I'm leaning toward not merging this. If anyone does find any value in using this workflow on dbt projects, we can merge.

max-sixty:
I admittedly share your perspective re jinja. 

Ideally we won't need jinja, because we have a structured language. That's also the perspective of the dbt folks.

Maybe this could be an advanced feature of s-strings? But outside of that, I think we should hold the language itself to the standard of producing the abstractions...

aljazerzen:
> But because prql-compiler does not (and should not) know how to handle Jinja, ...

With this I meant not only for-loops and other evaluation, but even parsing Jinja templates.

```
from {{ ref('table') }}
```
... should be syntactically invalid PRQL code

max-sixty:
> ... should be syntactically invalid PRQL code

Oh yes I agree; we can remove Jinja, either now or when we get the proper dbt plugin up. 

IIRC, in the end we don't even use it in dbt-prql — PRQL actually gets called _after_ dbt has processed the jinja (which is why it doesn't work with all DBs). Though my memory is hazy and I'd need to confirm, depending on how quickly you think we should remove it!

max-sixty:
Let's discuss this on the next dev call. In some ways I think this is a very clever implementation. OTOH it does require everyone using it to have `prqlc`, which is a big disadvantage for teams, and have `prqlc watch` running.

It is also quite a nice demo of `prqlc watch`!

max-sixty:
Should we remove it from the pest grammar?

prql-compiler/prqlc/Cargo.toml
```diff
@@ -18,9 +18,12 @@ color-eyre = "0.6.1"
 env_logger = {version = "0.9.1", features = ["termcolor"]}
 itertools = "0.10.3"
 notify = "^5.1.0"
-prql-compiler = {path = '..', version = "0.5.2" }
+minijinja = {git = "https://github.com/aljazerzen/minijinja"}
```
max-sixty, line 21: @aljazerzen we should release this as soon as possible; as at the moment we can't release PRQL...

PR #1724, opened 2023-02-07
docs: Remind people to update Issue #1

richb-hanover:
That way, people who have subscribed to PRQL will get a notification


max-sixty:
Yeah, I haven't done a good job of this obviously, you're right to point us towards the failure. 

Let's merge this, and I'll re-evaluate if we should instead suggest folks subscribe to releases — a couple more clicks but otherwise better...

PR #1727, opened 2023-02-08
feat(prql-compiler): implement Target::names() and Dialect::names() to show available target names or dialect names

eitsupi:
~~Expose `DIALECT_NAMES` so that downstream software can get serialize representations of the available dialect names.~~
-> Changed to `Dialect::names()`

Like this (from eitsupi/prqlr#77)

```rust
pub fn dialects() -> Vec<&'static str> {
    prql_compiler::sql::DIALECT_NAMES.to_vec()
}
```

### DIFF
--- a/prql-compiler/src/sql/mod.rs
+++ b/prql-compiler/src/sql/mod.rs
@@ -10,6 +10,7 @@ mod preprocess;
 mod std;
 
 pub use dialect::Dialect;
+pub use dialect::DIALECT_NAMES;
 
 use anyhow::Result;
 
--- a/prql-compiler/src/sql/mod.rs
+++ b/prql-compiler/src/sql/mod.rs
@@ -10,7 +10,6 @@ mod preprocess;
 mod std;
 
 pub use dialect::Dialect;
-pub use dialect::DIALECT_NAMES;
 
 use anyhow::Result;
 
--- a/prql-compiler/src/lib.rs
+++ b/prql-compiler/src/lib.rs
@@ -134,6 +134,15 @@ pub enum Target {
     Sql(Option<sql::Dialect>),
 }
 
+impl Target {
+    pub fn names() -> Vec<String> {
+        sql::Dialect::names()
+            .into_iter()
+            .map(|d| format!("sql.{d}"))
+            .collect()
+    }
+}
+
 /// Compilation options for SQL backend of the compiler.
 #[derive(Debug, Clone, Serialize, Deserialize)]
 pub struct Options {
--- a/prql-compiler/src/sql/dialect.rs
+++ b/prql-compiler/src/sql/dialect.rs
@@ -13,7 +13,6 @@
 //! constructs. The upside is much less complex translator.
 
 use core::fmt::Debug;
-use once_cell::sync::Lazy;
 use serde::{Deserialize, Serialize};
 use strum::{EnumMessage, IntoEnumIterator};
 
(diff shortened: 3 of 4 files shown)


eitsupi:
I am wondering if it might be better to change the serialization expressions from `name` to `sql.name` (e.g. `ansi` -> `sql.ansi`) as well, since it is actually what is used in the PRQL script header.

@aljazerzen What do you think about renaming serialized expressions?
From my reading of the source code, it seems to me that the internal use is to remove the `sql.` so there seems to be no problem with adding `sql.` to the serialized expressions.

max-sixty:
Looks reasonable to me.

I'm guessing this is because bindings would prefer strings to the Enum variants?

I don't have much experience on whether this is the best pattern to do that, but no objections for sure.

eitsupi:
> I'm guessing this is because bindings would prefer strings to the Enum variants?

Yes, this matches the target names that can be used in the headers of the actual PRQL scripts, so users should be able to easily find out what target names are available regardless of the version of prql-compiler they are using.

eitsupi:
For example, we could use this to automatically update the following lines.

https://github.com/PRQL/prql/blob/e4bc3e2d83794142c027a526f848cf8a6326ce34/book/src/language-features/target.md?plain=1#L34-L44

eitsupi:
> I am wondering if it might be better to change the serialization expressions from `name` to `sql.name` (e.g. `ansi` -> `sql.ansi`) as well, since it is actually what is used in the PRQL script header.

I made this change in d11b3bceeebb3249a7208e751db5170bc372944f.

We can revert this or separate this into other PR.
I would like to hear your thoughts on this, thanks.

prql-compiler/src/sql/mod.rs
```diff
@@ -10,6 +10,7 @@ mod preprocess;
 mod std;
 
 pub use dialect::Dialect;
+pub use dialect::DIALECT_NAMES;
```
aljazerzen: Can we expose something like:
```
impl Dialect {
    pub fn names() -> Vec<&'static str> {
        ...
    }
}
```

aljazerzen (approved):
Looks good

eitsupi:
Thank you for your review, I changed it from `DIALECT_NAMES` to `Dialect::names()`.

RandomFractals:
can we drop `sql.` prefix from the supported dialect names?

They are all some form of sql and seem redundant.

eitsupi:
> can we drop `sql.` prefix from the supported dialect names?
> 
> They are all some form of sql and seem redundant.

This was recently done to set up a target for something other than SQL in the future.

https://github.com/PRQL/prql/blob/e4bc3e2d83794142c027a526f848cf8a6326ce34/CHANGELOG.md?plain=1#L189-L199

aljazerzen:
Dialect is now just the last part of target. In other words, target = execution.dialect

So actually it does not make sense to have `sql.` prefix in the `sql::Dialect`.

What you probably want to expose list of targets not dialect, right? There is also `Target` enum that only has Sql variant, so you can probably do something like this:

```rust
impl Target {
    fn values() -> Vec<String> {
         Dialect::values().into_iter().map(|d| format!("sql.{d}")).collect()
    }
}
```

RandomFractals:
I still feel like maybe we drop `sql.`. Even if you add compile to new target system type, it will be the new platform name, like `polars` or something. I believe it's clear that the other dialects are databases and the compiled result will be sql.

max-sixty:
> I still feel like maybe we drop `sql.`. Even if you add compile to new target system type, it will be the new platform name, like `polars` or something. I believe it's clear that the other dialects are databases and the compiled result will be sql.

I empathize with this. I do think it has some benefits though — it's a hierarchy, it's easy to understand, it's extensible.

We also _just_ made this change, so we should live with it for a while before we consider switching it again....

PR #1729, opened 2023-02-08
refactor!: prql-js options.target

aljazerzen:
Ref https://github.com/PRQL/prql-vscode/issues/48

### DIFF
--- a/prql-js/README.md
+++ b/prql-js/README.md
@@ -37,7 +37,7 @@ Options
 
 ```javascript
 const opts = new prql.CompileOptions();
-opts.target = 'sql.mssql';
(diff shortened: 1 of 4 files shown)


RandomFractals:
desired simple string target option JS API sig.:

```js
prqljs.compile(`from employees | take 10`, { target: 'duckdb'});
``` 

I am not sure the updates above support it.

aljazerzen:
It does not, but I can make it so.

A moment.

max-sixty:
Super! Thanks.

(and agree, think it makes sense to do strings rather than enums in languages that prefer strings...)

PR #1750, opened 2023-02-09
feat(prql-compiler): implement Target::from_str and Target::default

eitsupi:
Currently, the `sql.` prefix have to be removed from the target name and pass to `Dialect::from_str`.
This PR implements `Target::from_str` and does this on the `Target::from_str` side, eliminating the need for downstream software to implement this process.

Example of Use: eitsupi/prqlr#78

### DIFF
--- a/prql-compiler/src/lib.rs
+++ b/prql-compiler/src/lib.rs
@@ -86,7 +86,7 @@ pub mod sql;
 mod test;
 mod utils;
 
-pub use error::{downcast, ErrorMessage, ErrorMessages, SourceLocation, Span};
+pub use error::{downcast, Error, ErrorMessage, ErrorMessages, Reason, SourceLocation, Span};
 pub use utils::IntoOnly;
 
 use once_cell::sync::Lazy;
@@ -151,15 +151,18 @@ impl Target {
 }
 
 impl FromStr for Target {
-    type Err = ErrorMessages;
+    type Err = Error;
 
     fn from_str(s: &str) -> Result<Target, Self::Err> {
         if s.starts_with("sql.") {
             let maybe_dialect = s.strip_prefix("sql.").unwrap_or(s);
-            let dialect = sql::Dialect::from_str(maybe_dialect).unwrap_or(sql::Dialect::Generic);
+            let dialect = sql::Dialect::from_str(maybe_dialect).unwrap_or_default();
             Ok(Target::Sql(Some(dialect)))
         } else {
-            Err(downcast(anyhow::anyhow!("Unknown target: {}", s)))
+            Err(Error::new(Reason::NotFound {
+                name: format!("{s:?}"),
+                namespace: "target".to_string(),
+            }))
         }
     }
 }
--- a/prql-compiler/src/lib.rs
+++ b/prql-compiler/src/lib.rs
@@ -155,9 +155,10 @@ impl FromStr for Target {
 
     fn from_str(s: &str) -> Result<Target, Self::Err> {
         s.strip_prefix("sql.")
-            .map(|dialect| sql::Dialect::from_str(dialect).unwrap_or_default())
+            .map(sql::Dialect::from_str)
+            .unwrap()
             .map(|d| Target::Sql(Some(d)))
-            .ok_or_else(|| {
+            .map_err(|_| {
                 Error::new(Reason::NotFound {
                     name: format!("{s:?}"),
                     namespace: "target".to_string(),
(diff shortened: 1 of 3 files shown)


prql-compiler/src/lib.rs
```diff
@@ -134,6 +135,12 @@ pub enum Target {
     Sql(Option<sql::Dialect>),
 }
 
+impl Default for Target {
+    fn default() -> Self {
+        Self::Sql(Some(sql::Dialect::default()))
```
eitsupi: I wonder this should be `Sql(None)` or `Sql(Some(sql::Dialect::default())`?
If not `Sql(None)`, the default behavior is `sql.generic` and the target specified in the script's prql header is ignored.

```suggestion
        Self::Sql(None)
```

eitsupi:
As far as trying in eitsupi/prqlr#78, I think this is working fine.

But obviously tests on these should be added here.
I am not familiar with Rust testing and would like to add tests at another time...

prql-compiler/src/lib.rs
```diff
@@ -143,6 +150,23 @@ impl Target {
...
+        if s.starts_with("sql.") {
+            let maybe_dialect = s.strip_prefix("sql.").unwrap_or(s);
+            let dialect = sql::Dialect::from_str(maybe_dialect).unwrap_or_default();
+            Ok(Target::Sql(Some(dialect)))
+        } else {
+            Err(Error::new(Reason::NotFound {
+                name: format!("{s:?}"),
+                namespace: "target".to_string(),
+            }))
+        }
```
max-sixty: It seems like we have two identical logic branches here. Could we do something like:

```suggestion
            s.strip_prefix("sql.")
                      .map(|d| Target::Sql(Some(d)))
                      .map_err(
                         Err(Error::new(Reason::NotFound {
                         name: format!("{s:?}"),
                         namespace: "target".to_string(),
            )?
```

(I may be missing something, and I imagine this won't quite be correct, but I _think_ it should be close)

max-sixty (commented):
Thanks a lot @eitsupi — great that you're diving into the rust!

Would it be possible to add a short test in the file? Let me know if an example would be helpful.

max-sixty:
> I am not familiar with Rust testing

It's surprisingly easy!

We can just have something like this, in the file:

```rust
#[cfg(test)]
mod test {
use insta::assert_debug_snapshot;

#[test]
fn test_target_from_str() {

    assert_debug_snapshot!(Target::from_str("sql.postgres".to_string()), @"");

    assert!(Target::from_str("sql.poostgres".to_string().is_err());
}
}
```

...then running `cargo insta test --accept` will fill out the `@` string.

prql-compiler/src/lib.rs
eitsupi: Thank you for reviewing.
My intention was to branch the process for each prefix, but so far it is unnecessary.

eitsupi: I think it got better thanks to your advice, but after a lot of trial and error the if statement remained. (I couldn't pass the tests otherwise)

prql-compiler/src/lib.rs
```diff
@@ -143,6 +150,27 @@ impl Target {
...
+        // We have a closure here because we can't create the error in the
+        // pipeline, since it needs to be in two places, and we'd need to clone.
+        // (Though possibly it's too optimize-y.)
+        let not_found_error = |s| {
+            Error::new(Reason::NotFound {
+                name: format!("{s:?}"),
+                namespace: "target".to_string(),
+            })
+        };
+        s.strip_prefix("sql.")
```
max-sixty, line 166: @eitsupi this wasn't actually as simple as I thought. I made a small tweak to remove the extra condition. It's fine to revert if you thought the existing one was clearer.

Whenever you're happy, let's merge!

prql-compiler/src/lib.rs
```diff
@@ -242,3 +270,49 @@ pub mod json {
...
+
+#[cfg(test)]
+mod tests {
+    use crate::Target;
+    use insta::assert_debug_snapshot;
+    use std::str::FromStr;
+
+    #[test]
+    fn test_target_from_str() {
+        assert_debug_snapshot!(Target::from_str("sql.postgres"), @r###"
```
max-sixty, line 282: Perfect tests!

prql-compiler/src/lib.rs
eitsupi: Thank you for the update. I will merge this.

prql-compiler/src/lib.rs
max-sixty: (just saw this — yes I was a bit optimistic! Thanks for making it better!)

PR #1774, opened 2023-02-10
build: Add hugo to dockerfile

richb-hanover:
This PR adds `hugo` to the Dockerfile so that it's possible
   to work on the `website` directory from the Docker container
It also adds instructions for running `hugo` and `mdbook` in Docker
It also updates the Fortnightly Dev Call `.ics` so it seems to work with GCalendar


book/README.md
```diff
@@ -36,3 +36,13 @@ As described in [**book.toml**](book.toml), we have a few preprocessors which
 convert the markdown into the code displayed on the site. Some of these are
 quite hacky, and will likely not work on Windows. If this is a problem, please
 post an issue and we'll try and find a workaround.
+
+## Dockerfile
```
max-sixty:
```suggestion
## Docker
```

book/README.md
```diff
@@ -36,3 +36,13 @@ As described in [**book.toml**](book.toml), we have a few preprocessors which
...
+
+To work on the Book, start the Docker container and then use:
```
max-sixty:
```suggestion
To work on the Book using Docker, start the Docker container and then use:
```

max-sixty (commented):
Thanks @richb-hanover , appreciate you adding this in!

It's important to have a test — maybe we try running `task run-web` in the docker test in CI? Or a custom command if that blocks?

Dockerfile
```diff
@@ -50,6 +50,14 @@ COPY Taskfile.yml .
 # ========= Install cargo-tools =========
 RUN task install-cargo-tools
 
+# ========= Install hugo =========
+# https://stackoverflow.com/a/75330596/1827982 and
+# https://www.docker.com/blog/faster-multi-platform-builds-dockerfile-cross-compilation-guide/
+
+ARG BUILDARCH
+RUN curl -L "https://github.com/gohugoio/hugo/releases/download/v0.110.0/hugo_0.110.0_linux-$BUILDARCH.deb" -o hugo.deb
+RUN apt-get install ./hugo.deb
```
eitsupi, line 59: Some comments

- Do you need to fix the version? GitHub have URLs that specifies the latest version release so you can always install the latest version if you do that. (Hugo is released fairly frequently)
  - If you want to fix the version, it would be convenient to make the version a variable so that you do not have to edit two places.
- Since each use of the RUN clause adds a layer of images, it is sometimes better to perform the process within a single RUN clause, e.g. `RUN curl ... && apt-get install ...`
- Since Hugo is a stand-alone executable, there is also a way to download the executable instead of the deb package.

I apologize for not being able to offer specific suggestions for modifications, but I think the following script may be helpful.
https://github.com/devcontainers/features/blob/main/src/hugo/install.sh

richb-hanover:
@max-sixty @eitsupi Thanks for these comments. You help me realize this PR is not ready for prime time and will continue to refine it.

richb-hanover:
Closing for now - I'll come back with a new one

max-sixty:
Ok! A reminder that it doesn't need to be perfect, just needs to make incremental progress and be verifiable (e.g. with tests)...

PR #1780, opened 2023-02-11
docs(web): Use external book link

max-sixty:
Because `book_integrated` doesn't reflect the pages within the book, and we get confusions like https://github.com/PRQL/prql/issues/286#issuecomment-1426383971


max-sixty:
@richb-hanover this mistakenly includes your generous change to the ical. I'll leave it in if that's OK!

richb-hanover:
> @richb-hanover this mistakenly includes your generous change to the ical. I'll leave it in if that's OK!

Please feel free to "make it right". Thanks.

PR #1807, opened 2023-02-12
feat(prql-python): expose CompileOptions

### LINKED ISSUE #1792: [python] CompileOptions?
There does not seem to be a way to specify options currently?
`CompileOptions` do not seem to be exposed on the Python side?

https://github.com/PRQL/prql/blob/741c2b30c32d77206b2a167f1bcedd23398da365/prql-python/src/lib.rs#L5-L13

It would be useful to be able to set options here so that the target can be fixed on the pyprql side.

eitsupi:
Close #1792

Expose `CompileOptions` to allow compile options to be specified from the Python side.

Note: This is a basic implementation only.
For example, I am not sure if the behavior when an invalid target name is specified falling back to `sql.generic` silently is appropriate.
It would be worth changing later.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,8 +4,6 @@
 
 **Features**:
 
-- Compilation options can now be specified from Python. (@eitsupi, #1807)
-
 **Fixes**:
 
 - Delegate dividing literal integers to the DB. Previously integer division was
@@ -20,6 +18,8 @@
 **Web**:
 
 **Integrations**:
+
+- [prql-python] Compilation options can now be specified from Python. (@eitsupi, #1807)
 
 **Internal changes**:
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,7 +19,8 @@
 
 **Integrations**:
 
-- [prql-python] Compilation options can now be specified from Python. (@eitsupi, #1807)
+- [prql-python] Compilation options can now be specified from Python. (@eitsupi,
+  #1807)
 
 **Internal changes**:
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@
 
 **Features**:
 
+- Compilation options can now be specified from Python. (@eitsupi, #1807)
+
 **Fixes**:
 
 - Delegate dividing literal integers to the DB. Previously integer division was
--- a/prql-python/python/tests/test_all.py
+++ b/prql-python/python/tests/test_all.py
@@ -33,4 +33,8 @@ def test_all():
         )
     """
 
+    options = prql.CompileOptions(
+        format=True, signature_comment=True, target="sql.postgres")
+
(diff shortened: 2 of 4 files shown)


CHANGELOG.md
```diff
@@ -4,6 +4,8 @@
 
 **Features**:
 
+- Compilation options can now be specified from Python. (@eitsupi, #1807)
```
eitsupi: Ideally, it would be necessary to make it easier to understand which subproject the change relates to, but so far nothing has been done.

prql-python/src/lib.rs
```diff
@@ -69,19 +72,33 @@ pub struct CompileOptions {
...
+            format,
+            target,
+            signature_comment,
+        }
+    }
+}
+
 impl From<CompileOptions> for prql_compiler::Options {
     fn from(o: CompileOptions) -> Self {
+        let target = Target::from_str(&o.target).unwrap_or(Target::Sql(Some(Dialect::Generic)));
```
eitsupi, line 97: I think it is better to fall back to `Target::Sql(None)` (default of Target), but for now I have it matched to prql-js.

https://github.com/PRQL/prql/blob/6adbba316688be9beb3e3c7a4513b55d4218683c/prql-js/src/lib.rs#L106

max-sixty: I definitely agree, particularly if we start raising a warning / error when this differs from what's in the PRQL header.

CHANGELOG.md
max-sixty: I was thinking we'd put these in the "Integrations" (which maybe should be "Integrations / Bindings" section. I don't think we'll have enough that we'll need multiple ones... And then the "Features" section would be for language features.

What do you think?

max-sixty:
Excellent, thanks a lot @eitsupi !

CHANGELOG.md
eitsupi: Sorry, I missed the integration section...

eitsupi:
Thank you for your review. Minor fixes have been made to the documentation.

I would appreciate it if you could merge it if everything is ok.
In that case, I think we can create an issue about the issue not generating the appropriate error or warning.

prql-python/python/tests/test_all.py
```diff
@@ -33,4 +33,21 @@ def test_all():
...
+
+def test_compile_options():
+    """
+    Test the CompileOptions
+    """
+    query_mssql = "prql target:sql.mssql\nfrom a | take 3"
+    options = prql.CompileOptions(format=False, signature_comment=False, target="foo")
+
+    assert prql.compile(query_mssql).startswith("SELECT\n  TOP (3) *\nFROM\n  a")
+    # TODO: This should be unknown target error?
```
max-sixty, line 52: Nice, +1 to adding the TODO and merging what we have...

max-sixty:
Looks great! I'll hit the button since you asked

PR #1809, opened 2023-02-12
docs: fix link (follow up #1805)

eitsupi:

PR #1814, opened 2023-02-12
docs: s/Tables/Variables/

richb-hanover:
Found another place that refers to the name change from "Tables" to "Variables"


richb-hanover:
Also: #1805 & #1809

No good deed goes unpunished :-)

eitsupi:
Thanks!

PR #1818, opened 2023-02-12
feat: rewrite parser with chumsky

aljazerzen:
Rewrite most of the parsing with Chumsky.

Aim of this commit is to replace Pest with Chumsky entirely. I wanted to continue partial bottom-to-top approach I described in #1529, but it turned out that when I written parser for expr, I'd have to replace like 90% of PEST implementation with chumsky. So let's just do it.

TODO:
- [x] multi-quoted strings (I'm not sure how we'll do this),
- [x] ranges
- [x] dates (@max-sixty  can I just cherry-pick #1652 ?)
- [x] finish s-strings
- [x] ~disallow trailing `_` in numbers~
- [x] add raw string that don't escape

Incompatibilities:
- we now parse escapes `" \" \n \\ "`. This is because I plagiarized examples and have based our strings on implementation of JSON strings,
- subtle differences in what is considered a statement (this should not have any effects of things that previously compiled),
- s-strings and f-string interpolations cannot contain `{`. This is no longer valid: `s"text{'}'}text"`,
- operator precedence on main is right-to-left (`1 + 2 - 3 - 4 == 1 + (2 - (3 - 4))`. This is [very unconventional](https://en.cppreference.com/w/cpp/language/operator_precedence), so I'm changing it to left-to-right,
- operator precedence for `and` and `or` on main is right-to-left `a or b and c or d == a or (b and (c or d))`. All languages I tested give `and` precedence over `or`, so I'm changing it to that.


prql-compiler/src/parser/expr.rs
```diff
@@ -0,0 +1,233 @@
...
+
+        binary_op_parser(expr_coalesce, operator_logical())
+    })
+}
+
+pub fn pipeline<E>(expr: E) -> impl Parser<Token, ExprKind, Error = Simple<Token>>
+where
+    E: Parser<Token, Expr, Error = Simple<Token>>,
+{
+    // expr is a param, so I can be either a normal expr() or
```
max-sixty:
```suggestion
    // expr is a param, so it can be either a normal expr() or
```

prql-compiler/src/parser/lexer.rs
```diff
@@ -0,0 +1,253 @@
...
+        .or(just("!="))
+        .or(just(">="))
+        .or(just("<="))
+        .or(just("and")) // TODO: negative lookahead for whitespace
+        .or(just("or")) // TODO: negative lookahead for whitespace
+        .or(just("??"))
+        .map(|x| x.to_string())
+        .map(Token::Control);
+
+    let control = one_of("></%=+-*[]().,:|!")
```
max-sixty: Is this copied from somewhere or do we actually need `%`? (fine if so, mostly curious, and maybe we comment)

prql-compiler/src/parser/mod.rs
```diff
@@ -1022,23 +1104,19 @@ Canada
...
-        for bad_number in bad_numbers {
-            assert!(stmts_of_string(bad_number).is_err());
-        }
-        Ok(())
+        // Underscores at the beginning are parsed as ident
+        expr_of_string("_2").unwrap().kind.into_ident().unwrap();
+        expr_of_string("_").unwrap().kind.into_ident().unwrap();
+
+        expr_of_string("_2.3").unwrap_err();
+        // expr_of_string("2_").unwrap_err(); // TODO
```
max-sixty, line 808: All OK with these? They actually took a while in pest, happy to help with them in chumsky...

prql-compiler/src/parser/expr.rs
```diff
@@ -0,0 +1,233 @@
...
+                    .allow_trailing(),
+            )
+            .then_ignore(whitespace().or(new_line()).repeated())
+            .then_ignore(ctrl("]"))
+            .map(ExprKind::List)
+            .labelled("list");
+
+        let pipeline = ctrl("(")
+            .ignore_then(pipeline(func_call(expr)))
+            .then_ignore(ctrl(")"));
```
max-sixty: Can we use `delimited_by` for these?

prql-compiler/src/parser/mod.rs
```diff
@@ -688,76 +782,69 @@ mod test {
...
+            .unwrap();
+        assert_eq!(escaped_string, " \nU S A ");
 
         // Currently we don't allow escaping closing quotes — because it's not
         // trivial to do in pest, and I'm not sure it's a great idea either — we
         // should arguably encourage multiline-strings. (Though no objection if
         // someone wants to implement it, this test is recording current
         // behavior rather than maintaining a contract).
-        let escaped_quotes = expr_of_string(r#"" Canada \""#, Rule::string)?;
+        let escaped_quotes = expr_of_string(r#"" Canada \"""#).unwrap();
```
max-sixty: FWIW I think we can delete this block; it's not needed now we're escaping quotes

prql-compiler/src/parser/lexer.rs
```diff
@@ -0,0 +1,253 @@
...
+
+    let number = just('+')
+        .or(just('-'))
+        .or_not()
+        .chain::<char, _, _>(number_part)
+        .chain::<char, _, _>(frac.or_not().flatten())
+        .chain::<char, _, _>(exp.or_not().flatten())
+        .try_map(|chars, span| {
+            // pest is responsible for ensuring these are in the correct place,
+            // so we just need to remove them.
```
max-sixty: I can't actually see how pest is still involved — is it?

max-sixty (commented):
It looks great! The chumsky code is more readable than I expected.

Even if there are optimizations — and hints on how to recover from errors — I'm guessing the errors are already better than the pest errors, and so we can merge now and iterate later. Maybe improving errors would be a good place for new contributors?

I think escaping is fine. Maybe we need a raw string format; no need to gate on that though. If we can add a TODO to the docs I'm happy to take that...

Re dates — I don't think we can quite cherry-pick that — it doesn't do times — but to match the current behavior is actually much simpler than that (I was predominantly using it for exploring chumsky...)


> s-strings and f-string interpolations cannot contain `{`. This is no longer valid: `s"text{'}'}text"`.

I'm guessing it's `}` that is forbidden, is that right? So `f"{{foo}}" should still work, since that will evaluate to `"{foo}"`

Does this no longer rely on pest at all?

CC @AlexRiedler if you have any feedback!

prql-compiler/src/parser/mod.rs
aljazerzen: Ah yeah, this is also todo

prql-compiler/src/parser/expr.rs
aljazerzen: It seems so :D

prql-compiler/src/parser/lexer.rs
aljazerzen: :D I've copied the comment along with the code without reading it

prql-compiler/src/parser/expr.rs
aljazerzen: Wait, no `delimited_by` is for things that are iterable (I may be reading the error wrong).

prql-compiler/src/parser/lexer.rs
aljazerzen: Yeah, this is the mod operator, we have it in pest.

aljazerzen:
> I'm guessing it's } that is forbidden, is that right?

Yeah, my typo.

> `f"{{foo}}"` should still work, since that will evaluate to "{foo}"

Well, with my impl it won't - it will parse `{{foo}` as the interpolation and `}` as the following string.

> Does this no longer rely on pest at all?

Currently not, but if we wanted to merge now, we could do a little plumbing and transform this into the bottom-up. That's why I haven't removed pest yet - a fallback option is we cannot finish the TODO list.

aljazerzen:
(I just want to document a decision I did here.)

Initially, I started with parser only. The problem with this that generated errors are saying
`expected one of 'f', 't', 'n', got '('` instead of `expected one of true, false, null, got '('`. This is because parser is consuming one character at the time.

So I extracted the lexer to convert stream of char into stream of tokens, which are then consumed by the parser. It's also a nice distinction between high and low level parsing.

prql-compiler/src/parser/expr.rs
```diff
@@ -0,0 +1,233 @@
...
+use super::lexer::{InterpolItem, Token};
+use crate::ast::pl::*;
+
+pub fn expr_call() -> impl Parser<Token, Expr, Error = Simple<Token>> {
+    func_call(expr())
+}
+
+pub fn expr() -> impl Parser<Token, Expr, Error = Simple<Token>> + Clone {
+    recursive(|expr| {
+        let literal = select! { Token::Literal(lit) => lit }.map(ExprKind::Literal);
```
zesterer: The rhs of the `select!` branch is an expression, so you can just do

```rs
select!{ Token::Literal(lit) => ExprKind::Literal(lit) }
```

prql-compiler/src/parser/expr.rs
```diff
@@ -0,0 +1,233 @@
...
+                    .then_ignore(ctrl("=").padded_by(whitespace().or_not()))
+                    .or_not()
+                    .then(func_call(expr.clone()))
+                    .map(|(alias, expr)| Expr { alias, ..expr })
+                    .padded_by(whitespace().or(new_line()).repeated())
+                    .separated_by(ctrl(","))
+                    .allow_trailing(),
+            )
+            .then_ignore(whitespace().or(new_line()).repeated())
+            .then_ignore(ctrl("]"))
```
zesterer: Instead of

```rs
let list = ctrl("[")
    .ignore_then(...)
    .then_ignore(ctrl("]"));
```

you can do

```rs
let list = ...
    .delimited_by(ctrl("["), ctrl("]"));
```

(the same applies to other places below that use this pattern)

prql-compiler/src/parser/expr.rs
```diff
@@ -0,0 +1,233 @@
...
+
+        let s_string = select! { Token::Interpolation('s', string) => string }
+            .map(|s| {
+                s.into_iter()
+                    .map(|i| match i {
+                        InterpolItem::String(s) => InterpolateItem::String(s),
+                        InterpolItem::Expr(s) => InterpolateItem::String(s),
+                    })
+                    .collect_vec()
+            })
```
zesterer: The bodies of these `map`s can be flattened into `select!` as mentioned before.

prql-compiler/src/parser/expr.rs
```diff
@@ -0,0 +1,233 @@
...
+        // TODO: range
+
+        let term = literal
+            .or(list)
+            .or(pipeline)
+            .or(s_string)
+            .or(f_string)
+            .or(ident_kind)
+            .map_with_span(into_expr)
+            .boxed();
```
zesterer: This is fine (I do this in my own parser), but be aware that `or` chains like this unfortunately produce exponential behaviour in Rust's trait solver. If you find that compilation times are an issue, replacing these cases with [`choice`](https://docs.rs/chumsky/latest/chumsky/primitive/fn.choice.html), which seems to only have close-to-linear complexity, is probably the first thing to do.

prql-compiler/src/parser/expr.rs
```diff
@@ -0,0 +1,233 @@
...
+}
+
+pub fn pipeline<E>(expr: E) -> impl Parser<Token, ExprKind, Error = Simple<Token>>
+where
+    E: Parser<Token, Expr, Error = Simple<Token>>,
+{
+    // expr is a param, so it can be either a normal expr() or
+    // a recursive expr called from within expr()
+
+    (new_line().or(whitespace()).repeated())
```
zesterer: To check: is `whitespace` here the same as `chumsky::text::whitespace`? That parser is permitted to accept no input, so `... .or(whitespace()).repeated()` can potentially accept no input an unbounded number of times, meaning that the parser will loop infinitely. I'm currently developing an introspection mode for a future version of chumsky that will catch cases like this before they happen, but it's worth considering.

Edit: Based on the usage below, it seems like they're not the same and I was mistaken.

prql-compiler/src/parser/expr.rs
```diff
@@ -0,0 +1,233 @@
...
+        .ignore_then(
+            expr.padded_by(whitespace().or_not())
+                .separated_by(ctrl("|").or(new_line().repeated().at_least(1).to(())))
```
zesterer: `.to(())` is the same as `.ignored()`

prql-compiler/src/parser/expr.rs
```diff
@@ -0,0 +1,233 @@
...
+        .map(|(name, args)| {
+            if args.is_empty() {
+                return name.kind;
+            }
+
+            let mut named_args = HashMap::new();
+            let mut positional = Vec::new();
+            for (name, arg) in args {
+                if let Some(name) = name {
+                    named_args.insert(name, arg);
```
zesterer, line 250: You might want to check that two arguments aren't the same here by checking the result of `HashMap::insert`. If you want to report an error in here, then `.validate` instead of `.map` will allow you to do this.

prql-compiler/src/parser/lexer.rs
```diff
@@ -0,0 +1,322 @@
...
+        .or(just("=>"))
+        .or(just("=="))
+        .or(just("!="))
+        .or(just(">="))
+        .or(just("<="))
+        .or(just("and")) // TODO: negative lookahead for whitespace
+        .or(just("or")) // TODO: negative lookahead for whitespace
+        .or(just("??"))
+        .map(|x| x.to_string())
+        .map(Token::Control);
```
zesterer: I'd recommend [`choice`](https://docs.rs/chumsky/latest/chumsky/primitive/fn.choice.html) for `or` chains like this. It's faster to compile and a bit easier to maintain.

prql-compiler/src/parser/lexer.rs
```diff
@@ -0,0 +1,322 @@
...
+
+    let control = one_of("></%=+-*[]().,:|!")
+        .map(|c: char| c.to_string())
+        .map(Token::Control);
+
+    let ident = ident_part().map(Token::Ident);
+
+    let literal = literal().map(Token::Literal);
+
+    let comment = just('#').then(filter(|c: &char| *c != '\n').repeated());
```
zesterer: `filter(|c: &char| *c != '\n')` is the same as `none_of('\n')`.

prql-compiler/src/parser/lexer.rs
```diff
@@ -0,0 +1,322 @@
...
+
+    // s-string and f-strings
+    let interpol_string = filter(|c| *c != '"' && *c != '{')
```
zesterer: `filter(|c| *c != '"' && *c != '{')` is the same as `none_of("\"{")`

prql-compiler/src/parser/lexer.rs
```diff
@@ -0,0 +1,322 @@
...
+        .repeated()
+        .collect::<String>()
+        .map(InterpolItem::String);
+    let interpolation = one_of("sf")
+        .then_ignore(just('"'))
+        .then(
+            interpol_string.chain(
+                just('{')
+                    .ignore_then(filter(|c| *c != '}').repeated().collect::<String>())
+                    .then_ignore(just('}'))
```
zesterer: This pattern can use `.delimited_by(just('{'), just('}'))` instead.

prql-compiler/src/parser/lexer.rs
```diff
@@ -0,0 +1,322 @@
...
+    let interpol_string = filter(|c| *c != '"' && *c != '{')
+        .repeated()
+        .collect::<String>()
+        .map(InterpolItem::String);
+    let interpolation = one_of("sf")
+        .then_ignore(just('"'))
+        .then(
+            interpol_string.chain(
+                just('{')
+                    .ignore_then(filter(|c| *c != '}').repeated().collect::<String>())
```
zesterer: `none_of('}')`

prql-compiler/src/parser/stmt.rs
```diff
@@ -0,0 +1,168 @@
...
+            if let Some(param) = param {
+                Ty::Parameterized(Box::new(ty), Box::new(param))
+            } else {
+                ty
+            }
+        });
+
+        ctrl("<")
+            .ignore_then(type_term.separated_by(ctrl("|").padded_by(whitespace().or_not())))
+            .then_ignore(ctrl(">"))
```
zesterer: `.delimited_by(ctrl("<"), ctrl(">"))`

zesterer (commented):
Looks nice overall! I don't have a very in-depth knowledge of SQL/PRQL syntax though, so I can't say for sure that it parses the grammar you think it parses, of course.

prql-compiler/src/parser/expr.rs
aljazerzen: Apparently I need:
```rust
    .delimited_by(ctrl("[").boxed(), ctrl("]").boxed())
 ```
... as `DelimitedBy` requires `Clone` on start and end, which I'm not sure is required.

Comment (aljazerzen) [reply prql-compiler/src/parser/expr.rs]: Oh, I see the comment:
```
// TODO: Don't clone!
```

File: prql-compiler/src/parser/expr.rs (hunk 10 above)
Comment (aljazerzen) [reply prql-compiler/src/parser/expr.rs]: Yeah, I ran into this issue when I listed all control tokens with `.or()` :D

File: prql-compiler/src/parser/expr.rs (hunk 11 above)
Comment (aljazerzen) [reply prql-compiler/src/parser/expr.rs]: Not they are not, my whitespace is this:
```
    pub fn whitespace() -> impl Parser<Token, (), Error = Simple<Token>> + Clone {
        just(Token::Whitespace).repeated().at_least(1).to(())
    }
```
... and even during lexing, we only count \t and ' ' as whitespace.

File: prql-compiler/src/parser/expr.rs (hunk 4 above)
Comment (max-sixty) [reply prql-compiler/src/parser/expr.rs]: Hmmm, I'll try this out; atm I'm getting an error building

**Comment by max-sixty** [comment]
Thanks for the review @zesterer ! Very grateful!

File: prql-compiler/src/parser/expr.rs (hunk 4 above)
Comment (zesterer) [reply prql-compiler/src/parser/expr.rs]: In general, `a.ignore_then(inner).then_ignore(b)` can be replaced with `inner.delimited_by(a, b)`.

**Comment by aljazerzen** [comment]
Closes #131 #1815

File: prql-compiler/src/parser/lexer.rs [hunk 20]
```diff
@@ -63,17 +64,32 @@ pub fn lexer() -> impl Parser<char, Vec<(Token, std::ops::Range<usize>)>, Error
...
         .separated_by(new_line.then(whitespace.or_not()))
         .at_least(1)
         .ignored();
 
-    token
-        .map_with_span(|tok, span| (tok, span))
-        .padded_by(comments.or(whitespace).repeated())
+    let range = whitespace
+        .or_not()
+        .then(just(".."))
```
zesterer: Perhaps `.then_ignore` given that the `..` isn't used in the `map` below?

max-sixty:
I thought the tests might be running slower locally so I ran a performance test — we had an old one with `carge bench`. 

I think something might be a bit off in the current implementation...

- main: `time:   [581.55 µs 583.69 µs 585.94 µs]`
- parser: `time:   [30.697 ms 30.786 ms 30.881 ms]`

We got 600x slower! I'm guessing this is something quite small...

zesterer:
That sounds like you might have part of the parser that's branching off and checking a pattern on both branches, causing a reparse. Something like this:

```rs
let prefix = ... ;
let a = prefix.then(...);
let b = prefix.then(...);

a.or(b) // `prefix` gets parsed in both branches, so correctly parsing `b` requires parsing `prefix` twice: if `prefix` is deeply nested or recursive, this can produce exponential behaviour
```

The solution to cases like this is usually to separate out the prefix, like so:

```rs
let a = ...;
let b = ...;

prefix.then(a.or(b))
```

aljazerzen:
I did some tests and it really is 100x slower.

Here is is [a flamechart](prql-compiler/flamegraph.svg) of only lexing, no parsing. I cannot say what's the problem.

It may be the prefix issue @zesterer mentioned, or it may be all the copying from the input string into Token, which contains `Token::Ident(String)`.

@zesterer would it be possible to have this:
```rust
enum Token<'a> {
    ...
    Ident(&'a str),
    ...
}
```
I tried it, but rustc is really not happy about the extra lifetime when it comes to `choice` and `Parser::or`.

Also, how would I get &str from `filter(|c: &char| ...)` which produces `Vec<char>`. There should a way to get reference to the underlying string, right?

zesterer:
> I tried it, but rustc is really not happy about the extra lifetime when it comes to `choice` and `Parser::or`.

Allowing this pattern is the primary goal of the `zero-copy` rewrite (unfortunately, it's quite a complex thing to do post-hoc, hence the rewrite) (see [here](https://github.com/zesterer/chumsky/blob/zero-copy/examples/zero-copy.rs) for the sort of things you can do with it).

That said, it's an even better idea to intern your strings during the lexing stage so that comparing strings is just an integer comparison and your strings become copyable. There are [several](https://crates.io/crates/internment) [crates](https://crates.io/crates/lasso) that are very good at this sort of thing.

zesterer:
Additionally, out of interest, have you tried using an alternative error type? `Rich` uses an internal `HashSet` to keep track of alternative characters, but [`Cheap`](https://docs.rs/chumsky/latest/chumsky/error/struct.Cheap.html) does not and should be substantially faster.

aljazerzen:
I've removed all nested data within `Token`:

```rust
#[derive(Clone, PartialEq, Debug)]
pub enum Token {
    NewLine,

    Ident,
    Keyword,
    Literal,

    Interpolation,

    Control,
}
```

Just lexing of the benchmark file is still taking `[3.9285 ms 3.9305 ms 3.9335 ms]`, which is basically the same as before, which means that copying is not the problem here.

I also realized that we would have to copy idents to AST anyway, so it doesn't matter if we do it sooner rather than later (idents are ~90% of Tokens).

aljazerzen:
> Allowing this pattern is the primary goal of the zero-copy rewrite

Ah nice, I've heard about the rewrite but I though it's referring not cloning parsers when they are built.
 
> Additionally, out of interest, have you tried using an alternative error type? 

I haven't, I'll do that now.

zesterer:
I just did a rough 'n ready swap of `Simple` to `Cheap` (in practice I needed to comment some things out because `Cheap`'s API is simpler, but the results should be pretty much comparable).

Here are the results for the `variables-query` benchmark:

`main`: 942.96 µs

`parser`: 5.6257 ms

`parser` (after replacing `Simple` with `Cheap`): 1.2105 ms

So still a slowdown, but a very marginal one (and once `zero-copy` is released, that should improve dramatically: we're seeing 3-4x improvements in tests)

For a project as large as PRQL, I recommend that you actually just write your own error type from scratch and implement `chumsky::Error` for it: it'll be faster and you'll get more precise control over how you report errors with it (such as using enum variants for errors rather than shunting around strings within the parser).

aljazerzen:
I'm also seeing improvements when replacing the error with `Cheap`.

On which commit did you run the bench for parser branch? Right now there are a few commits here that disable parser and only do lexing. The 5ms on parser seems suspicious.

But in anycase, errors are the culprit it seems, thanks for the tip!

zesterer:
> On which commit did you run the bench for parser branch?

My bad, I ran on a commit after you disabled parsing. Now I get:

`parser`: 44.990 ms

`parser` (after replacing Simple with Cheap): 10.748 ms

I imagine making `Token` copyable (by interning strings or similar) will represent another substantial knock off that last value too, perhaps bringing it within 2-3x that of `pest` with some luck and a little tweaking of the parser impl.

That said, I think it's worth considering how fast is acceptable to you: you're coming from `pest`, which is known for being extremely quick, largely because it's doing its own code generation from the syntax you provide it (i.e: it is itself an optimising compiler, which then feeds its output into LLVM, also an optimising compiler). Because of this, `pest` is in a position to perform all sorts of magic that `chumsky` isn't (currently, although we do have plans) able to such as flattening certain parsers together into optimised jump tables and the like.

In comparison, `chumsky` relies on the Rust compiler's native ability to inline and optimise accordingly. As we've [demonstrated](https://github.com/zesterer/chumsky/pull/104#issuecomment-1066174238) this *can* be very quick, often as fast as hand-written recursive descent parsers - but it's unlikely that LLVM's optimiser alone, as wonderful as it is, will be able to wholly outpace a DSL + compiler such as `pest`.

My usual approach when writing compilers is to aim for a specific proportion of total compilation time spent on parsing, usually ~20%. If parsing is under that goal, I move on to something else (for the languages I work on, type-checking is usually by far and away the biggest offender). What goal you set depends on the constraints the compiler has and the workload it usually needs to deal with.

aljazerzen:
Update:

I have improved the parser & lexer implementations quite a lot.

The two cruxes I improved were:
- cheap errors in lexer, were don't need expeceted and found tokens,
- parsing of ranges.

Benchmarks:
```
main                    time:   [590.68 µs 591.37 µs 592.20 µs]
parser                  time:   [3.0680 ms 3.0722 ms 3.0790 ms]
parser-cheap-errors     time:   [1.7351 ms 1.7372 ms 1.7398 ms]
parser-lexer-only       time:   [725.98 µs 726.90 µs 727.92 µs]
```

Last two are experimental commits:

```
* d93e594 (parser-lexer-only) disable parser
* c0b96b6 (parser) cleanup
* 100c306 cheap errors for lexer
| * 13862d4 (parser-cheap-errors) cheap errors
|/  
* c7869f0 marginal improvement of lexer
```

`parser-cheap-errors` is the best performance we can achieve 
if we optimize error in parser.
As you can see, we can improve for factor 2x.

`parser-lexer-only` shows that lexer is now pretty darn fast,
not much room for improvement - at least in our implementation.
It may benefit from the zero-copy chumsky rewrite.

aljazerzen:
@max can you look into why cargo audit started failing now?

It seems to be a problem with time crate, dependency of chrono, which has a feature oldtime

aljazerzen:
It looks like it's failing on main too.

Otherwise this can be merged.

CHANGELOG.md
```diff
@@ -5,6 +5,18 @@
 **Features**:
 
 - `loop`, which translates to `WITH RECURSIVE` (#1642, @aljazerzen)
+- Convert parser from PEST to Chumsky (@aljazerzen, #1818)
```
max-sixty: Great list! All improvements or reasonable tradeoffs...

CHANGELOG.md
```diff
@@ -5,6 +5,18 @@
...
+  - Error recovery in some cases and more readable errors in general.
```
max-sixty: We could sell the success of this more!

```suggestion
  - Improved error messages, and the potential to make even better in the future. Many of these improvements come from error recovery.
```

CHANGELOG.md
```suggestion
- Convert parser from pest to Chumsky (@aljazerzen, #1818)
```

CHANGELOG.md
```diff
@@ -13,6 +25,8 @@
 
 **Documentation**:
 
+- Operator precedence
```
```suggestion
- Operator precedence (@aljazerzen, #1818)
```

book/src/language-features/s-strings.md
```diff
@@ -36,10 +36,13 @@ join s=salaries side:left [
 ]
 ```
 
-For those who have used Python, s-strings are similar to Python's f-strings, but
-the result is SQL code, rather than a string literal. For example, a Python
-f-string of `f"average{col}"` would produce `"average(salary)"`, with quotes;
-while in PRQL, `s"average{col}"` produces `average(salary)`, without quotes.
+For those who have used python, s-strings are similar to python's f-strings, but
```
max-sixty: I have zero view here. This was recently changed to Title Case, which I merged, since it does appear in the docs as Python rather than python; even though the logo is lower case. But I hesitate to comment, please ignore if you have any actual view

book/src/language-features/strings.md
```diff
@@ -13,7 +13,7 @@ select x = 'hello world'
 ```
 
 To quote a string containing quotes, either use the "other" type of quote, or
-use three-or-more quotes, and close with the same number.
+use 3, 4, 5 or 6 quotes, and close with the same number.
```
max-sixty, line 16: Ah interesting. This seems fine. Because it's easier to make for 3-6 than for `n`?

book/src/syntax.md
```diff
@@ -166,7 +166,24 @@ sort (-distance)
 sort [-distance]
 ```
 
-## Inner transforms
+For a more formal definition, refer to this precedence table. Because function
+call has the lowest precedence, nested function calls or arguments that start or
+end with an operator require parenthesis.
```
max-sixty, line 171: Great!

book/src/syntax.md
```diff
@@ -261,3 +278,29 @@ select [
...
+It may seem that transforms are also keywords, but they are normal function
+within std namespace:
+
+```
+std.from my_table
+std.select [from = my_table.a, take = my_table.b]
+std.take 3
+```
+
+```adonish
```
max-sixty:
```suggestion
```admonish note
```

book/src/syntax.md
```diff
@@ -261,3 +278,29 @@ select [
...
+Note that new keywords will be added before 1.0 release, so your builds may break.
+You can guard against that by using backticks.
+```
```
max-sixty: FWIW I wouldn't have this — it's not really a practical suggestion to use backticks for every identifier, and we've been explicit on the guarantees elsewhere.

book/src/syntax.md
```diff
@@ -261,3 +278,29 @@ select [
...
+- `func`
+- `switch`
+
+If you want to use these names as your columns or relations, use backticks:
+`` `switch` ``.
+
+It may seem that transforms are also keywords, but they are normal function
+within std namespace:
+
+```
```
max-sixty:
```suggestion
```prql
```

book/src/syntax.md
```diff
@@ -261,3 +278,29 @@ select [
...
+## Keywords
+
+At the moment, PRQL uses only four keywords:
+
+- `prql`
+- `let`
+- `func`
+- `switch`
+
+If you want to use these names as your columns or relations, use backticks:
```
max-sixty: I've tried to keep the "voice" consistent through the book, and done that by avoiding the 2nd person (i.e. "you"); so it sounds more descriptive than instructional.

```suggestion
To use these names as columns or relations, use backticks:
```

(not trying to refine your work, if you prefer I do a pass after, or don't change it, totally fine, lmk)

prql-compiler/src/parser/expr.rs
```diff
@@ -0,0 +1,305 @@
...
+
+pub fn pipeline<E>(expr: E) -> impl Parser<Token, ExprKind, Error = Simple<Token>>
+where
+    E: Parser<Token, Expr, Error = Simple<Token>>,
+{
+    // expr has to be a param, because it can be either a normal expr() or
+    // a recursive expr called from within expr()
+
+    new_line()
+        .repeated()
```
max-sixty, line 172: Cool construction!

prql-compiler/src/parser/lexer.rs
```diff
@@ -0,0 +1,399 @@
...
+        bool,
+        null,
+        datetime,
+        date,
+        time,
+    ))
+}
+
+fn quoted_string(escaped: bool) -> impl Parser<char, String, Error = Cheap<char>> {
+    // I don't know how this could be simplified and implemented for n>3 in general
```
max-sixty, line 268: Yeah, whatever is simpler for this code is fine I think...

book/src/language-features/s-strings.md
aljazerzen: oh yeah, that's consequence of a bad merge, whoops

book/src/language-features/strings.md
aljazerzen: Because I don't know how to implement for `n` :D

https://github.com/PRQL/prql/pull/1818/files#diff-ae47c7322a8e7920d5819e0c5a9d2fdf1210e576e808f91a1f4ee72d013e4c2dR267-R283

prql-compiler/src/parser/prql.pest
```diff
@@ -38,19 +38,20 @@ ident_star = { "*" }
...
-func_call = ${ ident ~ WHITESPACE+ ~ (!operator_binary ~ (named_arg | alias | expr) ~ WHITESPACE*)+ }
 
-named_arg   = !{ ident_part ~ ":" ~ !":" ~ expr }
-// alias needs to be distinct from assign, so that in `join s=salaries [==id]`,
-// `s=salaries` is parsed separately from `[==id]`, since aliases allow for an
-// expr as an rvalue, but not a function call.
-alias = !{ ident_part ~ "=" ~ !"=" ~ expr }
+// assign_call needs to be distinct from assign, so that in `derive a = b c`, `b c` is not
+// a function call, but in `derive [a = sum b]`, `sub b` is.
+assign_call = !{ ident_part ~ "=" ~ !"=" ~ expr }
```
max-sixty: Should we remove this file?

book/src/syntax.md
aljazerzen: Yup better

prql-compiler/src/parser/prql.pest
aljazerzen: Yes, we should.

max-sixty:
> can you look into why cargo audit started failing now?
> 
> It seems to be a problem with time crate, dependency of chrono, which has a feature oldtime

~Looks like you fixed it but would have been happy to help with this / fine to comment out for the moment~

I removed the `pr-nightly` label, which runs the `time-compilation` action. But it also runs the `cargo-audit` action, seemingly with stricter settings than it does in nightly, which is passing https://github.com/PRQL/prql/actions/workflows/nightly.yaml

max-sixty:
> The two cruxes I improved were:
> 
>     * cheap errors in lexer, were don't need expeceted and found tokens,
> 
>     * parsing of ranges.
> 
> 
> Benchmarks:

Amazing speed-up! I didn't think we'd get to that much faster! It's down to the point where I think it doesn't matter about small differences vs pest

Sorry the ranges was the thing slowing it down. I know I suggested we treat whitespace differently, maybe that wasn't the correct tradeoff and it created lots of work 😬

aljazerzen:
I actually glad you insisted on keep the rules for whitespace with range - I like current behavior and it was worth spending a bit of effort for it!

All in all, I think this can be merged now (if the tests will allow it). @max-sixty and @zesterer  thank you both for reviews and tips!

max-sixty:
It's weird the Elixir tests are failing with a seg fault... https://github.com/PRQL/prql/actions/runs/4266003467/jobs/7426012687 Maybe something to do with stack space again?

Any thoughts? Should we comment them out? My only worry is that I know @kasvith was planning to publish the libraries to Hex, and won't be able to work on that while it's broken.

aljazerzen:
That's weird - this was failing sometime earlier but then it was not?

max-sixty:
Hmmm, also getting a stack overflow on Windows now: https://github.com/PRQL/prql/actions/runs/4266003467/jobs/7426011971

And the `prql-js` tests are stuck... https://github.com/PRQL/prql/actions/runs/4266003467/jobs/7427880787

kasvith:
@max-sixty not sure if this is something related to elixir though.
It seems the code fails on runtime somehow. Im unsure about the reason

aljazerzen:
You are right - it seems like an infinite recursion.

This could be a bug in the lexer or parser if I constructed a repeating parser that does not consume any tokens. I did a quick check for such a thing but did not find it.

The strange thing is that is happens on windows only. I've injected my files with `\r\n`, but still couldn't reproduce.

Does any have a windows machine and would run `cargo test` from this branch (`parser`)? If it fails, does it work on `parser-only-lexer` (tests on that branch will fail, but do they segfault?). It would really help me debug.

kasvith:
@aljazerzen i will run this and get back to you

kasvith:
Im getting following error for this

```bash
PYO3_PYTHON=py cargo test
```

```bash
     Running unittests src\lib.rs (target\debug\deps\prql_python-918bc678d493d4e2.exe)
error: test failed, to rerun pass `-p prql-python --lib`

Caused by:
  process didn't exit successfully: `C:\Users\Kasun\work\prql\target\debug\deps\prql_python-918bc678d493d4e2.exe` (exit code: 0xc0000135, STATUS_DLL_NOT_FOUND)
C:/Users/Kasun/.cargo/bin/cargo.exe: error while loading shared libraries: ?: cannot open shared object file: No such file or directory
```

I thought this was configured to be a lib, its weird that i ve to pass some arguments

aljazerzen:
Ah, let's just limit it to prql-compiler, i see there are segfaults there in the tests.

```
cargo test -p prql-compiler
```

kasvith:
```
test src\lib.rs - (line 56) ... ignored
test src\lib.rs - ReadmeDoctests (line 245) ... FAILED
test src\lib.rs - compile (line 109) ... FAILED
test src\lib.rs - (line 41) ... FAILED

failures:

---- src\lib.rs - ReadmeDoctests (line 245) stdout ----
Test executable failed (exit code: 0xc00000fd).

stderr:

thread 'main' has overflowed its stack


---- src\lib.rs - compile (line 109) stdout ----
Test executable failed (exit code: 0xc00000fd).

stderr:

thread 'main' has overflowed its stack


---- src\lib.rs - (line 41) stdout ----
Test executable failed (exit code: 0xc00000fd).

stderr:

thread 'main' has overflowed its stack



failures:
    src\lib.rs - (line 41)
    src\lib.rs - ReadmeDoctests (line 245)
    src\lib.rs - compile (line 109)

test result: FAILED. 0 passed; 3 failed; 1 ignored; 0 measured; 0 filtered out; finished in 4.25s
```

aljazerzen:
Does it also fail with segfault on `parser-only-lexer` branch?

kasvith:
@aljazerzen 
```
failures:
    parser::test::test_aggregate
    parser::test::test_allowed_idents
    parser::test::test_assign
    parser::test::test_backticks
    parser::test::test_basic_exprs
    parser::test::test_coalesce
    parser::test::test_dates
    parser::test::test_filter
    parser::test::test_function
    parser::test::test_gt_lt_gte_lte
    parser::test::test_inline_pipeline
    parser::test::test_literal
    parser::test::test_multiline_string
    parser::test::test_number
    parser::test::test_pipeline_parse_tree
    parser::test::test_sort
    parser::test::test_sql_parameters
    parser::test::test_take
    parser::test::test_var_def
    semantic::resolver::test::test_frames_and_names
    semantic::resolver::test::test_functions_1
    semantic::resolver::test::test_functions_nested
    semantic::resolver::test::test_functions_pipeline
    semantic::resolver::test::test_named_args
    semantic::resolver::test::test_variables_1
    semantic::test::check_invalid_version
    semantic::test::test_header
    semantic::test::test_resolve_01
    semantic::test::test_resolve_02
    semantic::test::test_resolve_03
    semantic::test::test_resolve_04
    semantic::transforms::tests::test_aggregate_positional_arg
    semantic::transforms::tests::test_transform_sort
    sql::gen_query::test::test_ctes_of_pipeline
    sql::gen_query::test::test_derive_filter
    sql::gen_query::test::test_filter_windowed
    sql::gen_query::test::test_variable_after_aggregate
    sql::test::test_end_with_new_line
    test::test_append
    test::test_bare_s_string
    test::test_basic_agg
    test::test_casting
    test::test_closures_and_pipelines
    test::test_custom_transforms
    test::test_dates
    test::test_direct_table_references
    test::test_distinct
    test::test_double_aggregate
    test::test_errors
    test::test_exclude_columns
    test::test_f_string
    test::test_filter
    test::test_filter_and_select_changed_alias
    test::test_filter_and_select_unchanged_alias
    test::test_from_text
    test::test_group_all
    test::test_header_target_error
    test::test_hint_missing_args
    test::test_ident_escaping
    test::test_inline_tables
    test::test_intersect
    test::test_interval
    test::test_join
    test::test_literal
    test::test_loop
    test::test_name_inference
    test::test_name_resolving
    test::test_name_shadowing
    test::test_nonatomic
    test::test_nonatomic_table
    test::test_nulls
    test::test_numbers
    test::test_output_column_deduplication
    test::test_pipelines
    test::test_precedence
    test::test_prql_to_sql_1
    test::test_prql_to_sql_2
    test::test_prql_to_sql_table
    test::test_quoting
    test::test_ranges
    test::test_remove
    test::test_rn_ids_are_unique
    test::test_same_column_names
    test::test_sorts
    test::test_sql_of_ast_1
    test::test_sql_of_ast_2
    test::test_static_analysis
    test::test_stdlib
    test::test_strings
    test::test_switch
    test::test_table_alias
    test::test_table_definition_with_expr_call
    test::test_table_names_between_splits
    test::test_table_s_string
    test::test_take
    test::test_target_clickhouse
    test::test_targets
    test::test_toposort
    test::test_unused_alias
    test::test_window_functions_00
    test::test_window_functions_02
    test::test_window_functions_03
    test::test_window_functions_04
    test::test_window_functions_05
    test::test_window_functions_06
    test::test_window_functions_07
    test::test_window_functions_08
    test::test_window_functions_09
    test::test_window_functions_10

test result: FAILED. 26 passed; 109 failed; 1 ignored; 0 measured; 0 filtered out; finished in 2.18s
```

aljazerzen:
This output seems to be missing integration tests that fail with segfault. The failed tests are just the --lib tests (that are inteentionally failing on that branch).

Well, thanks. It looks like there is no other solution than reading the implementation again and finding the bug.

max-sixty:
Thanks a lot for looking @kasvith , appreciate it.

---

@zesterer no stress if you can't respond to every Chumsky question — but if you see this and have ideas, let us know if you have any ideas for finding / handling these stack overflows — thank you!

zesterer:
Does the grammar contain left recursion?

aljazerzen:
I don't think so. There are recursive expressions, but i don't know what you mean exactly.

zesterer:
Left-recursive patterns are those that contain themselves as the first element. For example:

`expr = ident | { expr + '+' + expr }`

Most PEG parsers will stack overflow on cases like this because, in attempting to parse the expression, the `expr` pattern will be repeatedly entered at each level of recursion. It is possible for parsers to handle these cases but this generally requires memoisation, which generally sacrifices speed, or a way for the parser to automatically rewrite such patterns internally as right-recursive, which is complicated to implement correctly.

Thankfully, it is *always* possible to restate left recursion as right recursion, which PEG parsers have an easier job handling. For example, the above could become:

`expr = ident + { '+' + expr }?`

I don't know for certain that the case you're hitting *is* left recursion, but it seems likely that it is.

Debugging cases like this is usually a two-phase process:

1) Find an input that reproduces the problem, remove as many elements of it as possible while maintaining the issue.

2) Mass comment out any sections of the parser that aren't requires to pass the minimised failing input.

Hopefully, any left recursion will make itself 'obvious' at this point.

In the future, `zero-copy` will provide an explicit memoisation combinator that will allow left recursion to 'just work' as well as debugging facilities to identify which patterns are left-recursive and suggest fixes.

aljazerzen:
Ah yes I see. We do have left recursion in the grammar (basically all binary expressions), but I've implemented it as you described - with atoms (that I call term in the source code).

This was actually my suspicion, but I couldn't find anything like it.

Strangely, it is only happening on Windows, so I cannot reproduce it. Which mean that this PR will have to wait until I get the will to spin-up a virtual box.

If anyone is up for a little debugging and would want to find a minimal reproducible example, please go ahead!

zesterer:
Hmm. 'Only happening on Windows' is indeed interesting. On the assumption that the test logic doesn't differ between platforms, it sounds like you're just hitting a regular stack overflow due to some very deep stack usage (this is particularly common in debug mode, where LLVM doesn't get much of a chance to optimise things). Interestingly though, the default stack size on Windows is traditionally larger than on Linux, so perhaps there's some radically different codegen going on by LLVM that produces strange stack use patterns.

Out of interest, is there any reason why you don't have the `spill-stack` feature enabled for chumsky? It uses the `stacker` crate to negate stack overflows by spilling on to the heap. I imagine enabling this feature will effectively solve the problem.

aljazerzen:
Yes, spill-stack does not work on Macos, our tests were failing, due do some strange error (see previous comments).

Maybe we could disable it for macos only.

zesterer:
I think that's sensible. That aside, I'm surprised it didn't work on MacOS: `stacker` is used internally by `rustc` and seems to have fairly broad platform support. I shall look into this.

max-sixty:
Unfortunately I don't have access to a windows machine either. But I did notice that it was overflowing on the readme example, and one item there is a closed range expression (`take 1..10`), so that might be a good starting point for anyone experimenting.

> Yes, spill-stack does not work on Macos,

Just to confirm — I think this was failing with the `wasm32-unknown-unknown` target on MacOS. Ref https://github.com/PRQL/prql/pull/1723. We could make this feature target dependent. I'd be keener to invest into getting this to work in PRQL if we could be confident it worked in Chumsky generally — but not sure how up for some large test matrix in Chumsky's CI you are @zesterer...

zesterer:
> Just to confirm — I think this was failing with the `wasm32-unknown-unknown` target on MacOS. Ref #1723. We could make this feature target dependent.

Ah, right. `stacker` not working in WASM is definitely to be expected then. I assume this test had optimisations disabled? In which case, the compiler is probably unable to do much to optimise stack usage.

PR #1830, opened 2023-02-13
docs: language design notes

aljazerzen:


book/src/internals/language-design.md
```diff
@@ -0,0 +1,35 @@
+# Language design
+
+In a way PRQL is just a transpiler to SQL, this causes its language design to
```
max-sixty:
```suggestion
In a way PRQL is just a transpiler to SQL. This can cause its language design to
```

book/src/internals/language-design.md
```diff
@@ -0,0 +1,35 @@
...
+
+This is flawed because:
+
+- it does not model interactions between features well,
+- SQL behavior can sometimes be misleading (the order of a subquery will not
+  persist in the parent query) or even differs between dialects (set
+  operations).
+
+Instead, we should think of PRQL features in terms of how they affect PRQL
+expressions, which in most cases means relation.
```
max-sixty:
```suggestion
expressions, which in most cases means how they affect relations.
```

book/src/internals/language-design.md
```diff
@@ -0,0 +1,35 @@
...
+                   |
+                   v
+PRQL feature -> relation
+                   |
+                   v
+            relational result
+```
+
+Thinking about SQL comes in only at the last step when relation (or rather
+relational expression) is translated to an SQL expression.
```
max-sixty, line 35:
```suggestion
relational expression) is translated to a SQL expression.

To the extent this causes behavior that's very unfamiliar to SQL users, we can re-evaluate — we still place some weight on producing legible SQL and being comprehensible by those experienced in existing languages. But it should be a secondary thought rather than the root of our language development.
```

WDYT? I do think it's not zero-weight...

book/src/internals/language-design.md
```diff
@@ -0,0 +1,35 @@
...
+SQL.
+
+```
+PRQL feature -> SQL feature -> relational result
+```
+
+This is flawed because:
+
+- it does not model interactions between features well,
+- SQL behavior can sometimes be misleading (the order of a subquery will not
```
max-sixty, line 14:
```suggestion
- SQL's behavior can sometimes be misleading (the order of a subquery will not
```

max-sixty (approved):
I think I agree!

What do you think some areas where we've put too much weight on starting from SQL?

book/src/internals/language-design.md
vanillajonathan: Maybe just currently? Are there further ambitions?
```suggestion
In a way PRQL is currently just a transpiler to SQL, this causes its language design to
```

aljazerzen:
This is aimed toward our discussions about language design and was prompted by me noticing how I do it. I was working on proposal for type system and it really makes a difference there.

vanillajonathan:
I am interested in your proposal for the type system.

PR #1838, opened 2023-02-14
feat(prql-python): implement `get_targets` function

eitsupi:
Same as prql-js.
https://github.com/PRQL/prql/blob/0406d1a4b58defdfaa1bbd0ae2618f047c73420a/prql-js/src/lib.rs#L69-L74

### DIFF
--- a/prql-python/python/tests/test_all.py
+++ b/prql-python/python/tests/test_all.py
@@ -20,7 +20,7 @@ def test_all():
     res = prql.rq_to_sql(res)
     assert res is not None
 
-    assert len(prql.get_targets()) > 0
+    assert len(prql.get_targets())
 
     assert prql.__version__ is not None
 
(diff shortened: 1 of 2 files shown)


prql-python/src/lib.rs
```diff
@@ -104,6 +105,11 @@ impl From<CompileOptions> for prql_compiler::Options {
     }
 }
 
+#[pyfunction]
+pub fn get_targets() -> Vec<String> {
+    prql_compiler::Target::names()
```
eitsupi:
```suggestion
    Target::names()
```

prql-python/python/tests/test_all.py
```diff
@@ -20,6 +20,8 @@ def test_all():
     res = prql.rq_to_sql(res)
     assert res is not None
 
+    assert len(prql.get_targets()) > 0
```
max-sixty: python nit :)
```suggestion
    assert len(prql.get_targets())
```

max-sixty (approved):
Excellent @eitsupi !

PR #1843, opened 2023-02-15
docs: Replace northeast-arrow symbol

richb-hanover:
Substitute `&#8599;` for `&#x1F855;` that seems to work in all browsers


max-sixty:
Thanks!

Issue #1837, opened 2023-02-14
Range parsing precedence


aljazerzen:
Currently, ranges support:

```
0..40
(0)..40
11..
..9
(-5..)
-10..
@2020-01-01..
@2020-01-01..@2021-01-01
```

There a few changes I'd propose to change.

## Only literals

With current grammar, `..` has a rule to only bind with literals:

```
take 1..@2023-02-14  # range from 1 to @2023-02-14
take 1..[]           # two arguments: range from 1 onward and an empty list
take 1..hello        # two arguments: range from 1 onward and an ident
```

Because literals are allowed to be enclosed in parenthesis, ranges are support nested expressions, but only if they contain literals only. 

Current parsing grammar thus contains two types of parenthesis: one for normal nesting and pipeline and the other for the ranges.

This is a weird rule and I propose we eliminate it, by using normal parenthesis for ranges too.

```
(1)..40     # ok
(-1)..40    # ok
(0-1)..40   # not ok currenty, I propose we allow this
(5 | abs | hello [false])..40   # not ok currenty, I propose we allow this
```
There is not a strong case that this would greatly improve some use case, I want to change it in name of consistency.

## Whitespace

In my recent push to remove effects of whitespace from PRQL, ranges proved to be the only breaking point.

```
take 1+1    # a single bin op
take 1+ 1   # a single bin op
take 1..1   # a single range
take 1.. 1  # two arguments: a range and an int
```

Whitespace is not required to bind the literal into range, but if the term is not a literal (i.e. a list) it will not be bound into range even if there is no whitespace.

I propose we remove this distinction and bind ranges the same way we bind all other binary operations.

## Result

```
take (10 - 5 + 2 * 2)..1
take [1, 10]..[5, 15]
take 1 .. 4
take 1 ..
take ( .. 4)  # parenthesis needed
```

aljazerzen:
#1818 implements proposed behavior

max-sixty:
Agree with "Only literals" point, a nice consistency win.

For "Whitespace":
- I do think the ergonomics of `take (..1)` are worse than `take ..1`. I'm not sure how I'd explain it without giving the whole explanation about parsing whitespace, which suggests we're too anchored in the parser relative to the language. 
- But it's also not the biggest deal, and I don't want to slow you down. 
- Do you have a view for why Rust allows for the `..1`? (I'm trying to think of other languages that I've thought of as whitespace-agnostic, this is the only case I can think of...).
[...]

PR #1849, opened 2023-02-15
test: Add two whitespace tests to `parser`

max-sixty:
From the examples in https://github.com/PRQL/prql/issues/1837. One passes, one fails. I can help on this this weekend if that'd be helpful

(Hopefully adding the tests in helpful, not throwing things over the wall at you...)


prql-compiler/src/parser/mod.rs
```diff
@@ -824,6 +824,18 @@ Canada
...
+        ---
+        Binary:
+          left:
+            Literal:
+              Integer: 2
+          op: Add
+          right:
+            Literal:
+              Integer: 2
+        "###);
```
aljazerzen: Shouldn't this be:

```yaml
FuncCall:
  name:
    Ident: add
  args:
  - Literal:
      Float: 1.0
  - Literal:
      Integer: 2
```

max-sixty: Yes! This is the test that's currently failing. I should have left the result blank to make that clear...

prql-compiler/src/parser/mod.rs
```diff
@@ -824,6 +824,18 @@ Canada
...
+        assert_yaml_snapshot!(parse_expr(r#"add 1. 2"#).unwrap(), @r###"
+        ---
+        Binary:
+          left:
+            Literal:
+              Integer: 2
+          op: Add
+          right:
+            Literal:
+              Integer: 2
```
max-sixty:
```suggestion
```

Issue #1855, opened 2023-02-16
[prql-js][prql-python] The default target?


eitsupi:
Related to PRQL/prql-vscode#98

Currently the default target is `Generic`, but a default of `None` may be better to make the target in header available.

https://github.com/PRQL/prql/blob/7c42e5c64b92e9e43b0be82000135663552f0972/prql-js/src/lib.rs#L106
https://github.com/PRQL/prql/blob/7c42e5c64b92e9e43b0be82000135663552f0972/prql-python/src/lib.rs#L98

(Only `None` falls back to the header. if there is nothing in the header after falling back from `None` to the header, it falls back to `Gneric`.)

max-sixty:
Good point, sorry I missed this in the previous reviews.

max-sixty:
And ref https://github.com/PRQL/prql/issues/1686

eitsupi:
Closed by #1856 and #1861

PR #1856, opened 2023-02-16
feat(prql-js)!: use Sql(None) for default Target

eitsupi:
Part of #1855

Change the default to `Sql(None)` to fall back to the target specified in the PRQL query header when the compile option is set.


max-sixty:
Ah, very well observed, thanks for fixing @eitsupi

aljazerzen (approved):

PR #1860, opened 2023-02-17
build: Add PHP binding

### LINKED ISSUE #1844: PHP bindings
PRQL should have bindings for [PHP](https://www.php.net/). PHP have a [FFI](https://www.php.net/manual/en/book.ffi.php), which of particular interest is the [`FFI::cdef`](https://www.php.net/manual/en/ffi.cdef.php) function.

vanillajonathan:
This patch adds a PHP binding that uses the `libprql_lib` through PHP's Foreign Function Interface (FFI) extension.

Closes #1844


prql-php/.gitignore
```diff
@@ -0,0 +1,2 @@
+/vendor
+/.vscode
```
max-sixty: FYI IIRC this exists in the parent
```suggestion
```

prql-php/README.md
```diff
@@ -0,0 +1,25 @@
+# prql-php
+
+`prql-php` offers PHP bindings through FFI.
+
+It provides the `Compiler` class which contains the `toJson` and `toSql`
```
max-sixty, line 5: We'll want to rename `toSql` to `compile` to match the other bindings' APIs, but it's OK to do that later when we update `prql-lib`

max-sixty (commented):
Overall this looks great, thanks a lot @vanillajonathan !

We do need to add a test which runs in GitHub Actions, we add that for everything now. Would you be able to do that? We can use something like `test-python.yaml` as an example.

prql-php/README.md
vanillajonathan: Yes, I would like to do this too, it is `toSql` just for now until the `prql-lib` gets updated, then I plan to update it to use the new API.

vanillajonathan:
> Overall this looks great, thanks a lot @vanillajonathan !
> 
> We do need to add a test which runs in GitHub Actions, we add that for everything now. Would you be able to do that? We can use something like `test-python.yaml` as an example.

I don't know how to do this, but it needs to run [Composer](https://getcomposer.org/) which is a package manager for PHP which will install the the [PHPUnit](https://phpunit.de/) test framework which can be ran as `./vendor/bin/phpunit tests`.

max-sixty:
OK. It does look like there lots of examples out there — e.g. https://github.com/php-actions/example-phpunit/blob/master/.github/workflows/ci.yml — does that help?

I appreciate that it might seem less fun than the code (?) But it does mean your contribution will live on forever! Whenever something changes in PRQL, the test confirms that this code still works. If you can run the test locally, getting it to run in GHA should be 10-15 mins maximum, and happy to help with questions if that's helpful.

max-sixty:
FYI looks like this might be causing the failure: https://github.com/php-actions/composer/issues/95

We can run on just ubuntu in the meantime if that helps

vanillajonathan:
Thanks! I changed to only run on Ubuntu now and this seems to work a little better, it still fails, but at least now I get a better error message.

> Building PHP latest with extensions: FFI ...
Composer could not find a composer.json file in /app
To initialize a project, please create a composer.json file. See https://getcomposer.org/basic-usage
Error: Process completed with exit code 1.

Okay now I managed to get Composer (package manager) working. Now for the unit tests...

max-sixty:
> Thanks! I changed to only run on Ubuntu now and this seems to work a little better, it still fails, but at least now I get a better error message.

Great!

FYI if you can get unit tests running locally, and you can iterate without having to push every change to GitHub, that's generally much more efficient (and fun!). But whatever works — thanks a lot for pushing this forward.

vanillajonathan:
> FYI if you can get unit tests running locally, and you can iterate without having to push every change to GitHub, that's generally much more efficient (and fun!). But whatever works — thanks a lot for pushing this forward.

The unit tests work locally, but I had some problems with getting Composer and PHPUnit to work because they ran from the /app/ directory on the CI and looks for the `composer.json` file in the root of the project directory. There is a `working-directory` that can be set in the YAML for the CI but my understanding is that it only works for commands that you run directly with `run` not for things that get indirectly ran through the `uses` statement. I got around that by specifying where to find the `composer.json` file an the tests through command-line options that get passed by the `args` statement. Luckily both Composer and PHPUnit have support for doing so.

Now Composer and PHPUnit works but the unit tests fail because it cannot find the `libprql_lib` library since it hasn't been built. It expects to find the library in the `prql-php/src/` directory. I could need some help with that.

max-sixty:
Excellent; good progress @vanillajonathan !

Can we build the library in CI? The runner should have a compiler already so shoudn't even require muchsetup. What's the command you run locally to build it?

vanillajonathan:
> Can we build the library in CI? The runner should have a compiler already so shoudn't even require muchsetup. What's the command you run locally to build it?

Yeah, I Think we can do that. Sounds like a plan! I run `cargo build` inside the `prql-lib` directory.

I'll try build the library in the pipeline and then move the `.so` file.

vanillajonathan:
I built the library in the pipeline and move the compiled library file but I am now having issues with PHP not finding the library.

I try move the file:
```yaml
- run: mv ../target/debug/libprql_lib.so ../prql-php/src/libprql_lib.so
  working-directory: prql-lib
```

> Failed asserting that exception of type "FFI\Exception" matches expected exception "InvalidArgumentException". Message was: "Failed loading '/app/prql-php/src/libprql_lib.so'" at
/app/prql-php/src/Compiler.php:150
/app/prql-php/tests/CompilerTest.php:13

The test seems not to find the file and hints at an `/app/` location.

So then I change the YAML to move it to the `/app/` path:
```yaml
- run: mv ../target/debug/libprql_lib.so /app/prql-php/src/libprql_lib.so
  working-directory: prql-lib
```
But it does not want to move there:
> mv: cannot move '../target/debug/libprql_lib.so' to '/app/prql-php/src/libprql_lib.so': No such file or directory
##[error]Process completed with exit code 1.

max-sixty:
> mv: cannot move '../target/debug/libprql_lib.so' to '/app/prql-php/src/libprql_lib.so': No such file or directory

That's likely because the `/app/prql-php/` path doesn't exist yet

max-sixty:
FWIW — maybe because I'm on Mac — I get a `.a` & `.d` & `.dylib` file but no `.so`:

```
❯ ll  ../target/debug/
.rw-r--r--    0 maximilian 18 Feb 14:25 .cargo-lock
drwxr-xr-x    - maximilian 18 Feb 14:25 .fingerprint/
drwxr-xr-x    - maximilian 18 Feb 14:25 build/
drwxr-xr-x    - maximilian 18 Feb 14:25 deps/
drwxr-xr-x    - maximilian 18 Feb 14:25 examples/
drwxr-xr-x    - maximilian 18 Feb 14:25 incremental/
.rw-r--r-- 129M maximilian 18 Feb 14:25 libprql_lib.a
.rw-r--r-- 3.2k maximilian 18 Feb 14:25 libprql_lib.d
.rwxr-xr-x  21M maximilian 18 Feb 14:25 libprql_lib.dylib*
```

Does the test work locally for you? Where do you put the `.so` file locally?

max-sixty:
A couple of points:
- My guess is that this is reasonably close, assuming you have it working locally — just need to reproduce whatever you're doing locally on GHA, likely just moving files around or changing configs.
- If it proves difficult and you want to invest in PHP specifically, I do see options like https://github.com/davidcole1340/ext-php-rs, which would be more like our other bindings, letting a specialized library do the linking. But no need to move onto that now if you're close on this approach.

vanillajonathan:
> That's likely because the `/app/prql-php/` path doesn't exist yet

Then I don't know how to get it inside there.

> WIW — maybe because I'm on Mac — I get a `.a` & `.d` & `.dylib` file but no `.so`

Yes, it is because you're on macOS, I am on Linux so I get a `.so` file, and if you're on Windows then you get a `.dll` file.

Yes, the test works locally for me. The `.so` file I place in `prql-php/src/` directory.
Yes, I believe I am reasonably close.

max-sixty:
> Then I don't know how to get it inside there.

`mkdir -p /app/prql-php` will fix this error

It's unclear why it's looking for `/app/prql-php/`, I'm not sure that's a total solution...

vanillajonathan:
> > Then I don't know how to get it inside there.
> 
> `mkdir -p /app/prql-php` will fix this error
> 
> It's unclear why it's looking for `/app/prql-php/`, I'm not sure that's a total solution...

Just tried that.

>  mkdir: cannot create directory ‘/app’: Permission denied

max-sixty:
> mkdir: cannot create directory ‘/app’: Permission denied

~We need to create the `app` directory first! e.g. `mkdir /app`. Does that make sense?~

It looks like we're still seeing errors because of the `/app` path. I'm not sure why it always goes to the `/app` path. Is this something to do with PHP? Do we see this behavior locally?

We could ask at https://github.com/php-actions/phpunit/issues if we can't figure it out.

max-sixty:
@vanillajonathan do I see it working? Great job!!

Let me know when you're ready and I'll have a final look over

vanillajonathan:
Yes, it is working.

I had some problems, I used the `php-actions/phpunit@v3` and when the unit tests ran it would complain that it couldn't find the library file even though I had confirmed that it indeed existed by running `ls` in the pipeline. It seems `php-actions/phpunit@v3` was executing with the project root directory as context, so it would look for the library in `src/` instead of `prql-php/src/`. So instead of using the `php-actions/phpunit@v3` GitHub Action I now just run the unit tests via the shell instead just as I do on my local machine. This works. The `php-actions/phpunit@v3` would probably work too if `prql-php` was in a separate repository like the `demo-c-program` repository.

prql-php/tests/CompilerTest.php
```diff
@@ -0,0 +1,49 @@
...
+  *
+FROM
+  employees
+
+-- Generated by PRQL compiler version:0.5.0 (https://prql-lang.org)
+EOD;
+        $expected = substr($expected, 0, strpos($expected, "Generated by PRQL compiler"));
+        $prql = new Compiler();
+        $actual = $prql->toSql("from employees");
+        $actual = substr($actual, 0, strpos($actual, "Generated by PRQL compiler"));
```
max-sixty, line 46: Nice workaround for the signature changing. (An alternative would have been `signature_comment=false`, but this is great too)

.github/workflows/test-php.yaml
```diff
@@ -0,0 +1,45 @@
...
+          args: --working-dir=prql-php
+          php_extensions: FFI
+      - name: 🧪 Run tests using PHPUnit
+        run: vendor/bin/phpunit tests
+        working-directory: prql-php
+#      - name: 🧪 Run tests using PHPUnit (php-actions/phpunit)
+#        uses: php-actions/phpunit@v3
+#        with:
+#          args: --bootstrap prql-php/vendor/autoload.php prql-php/tests
+#          php_extensions: FFI
```
max-sixty: Let's remove this?

.github/workflows/test-php.yaml
```diff
@@ -0,0 +1,45 @@
+name: test-php
+
+on:
+  push:
+    branches:
+      - main
```
max-sixty: Can we copy `test-js` or `test-python`? i.e. add an entry to `test-all.yaml` and remove it running itself from `main`. 

Lmk if that's confusing / you want me to do this

max-sixty (commented):
Excellent! A couple of final clean-ups and then we can merge. 

I know this has been quite the journey, thanks @vanillajonathan !

prql-php/tests/CompilerTest.php
vanillajonathan: `signature_comment` is not yet available in the existing libprql library which lacks the `compile` function

.github/workflows/test-php.yaml
vanillajonathan: Yeah, we can remove this, if we don't intend to move the binding into its own repository, like `demo-c-program` is in its own repository.

.github/workflows/test-php.yaml
vanillajonathan: I added `test-php` to `test-all.yaml`. I don't know what you meant by removing it running itself from `main`, I think my the `test-php.yaml` file looks similar the Python and JavaScript ones. It doesn't have that concurrency thing though.

max-sixty: You're correct re `main` — I think that's probably incorrect in the others, but let's leave it here and I'll sort it out.

Can we copy the concurrency thing too?

.github/workflows/test-php.yaml
```diff
@@ -0,0 +1,45 @@
...
+  pull_request:
+    paths:
+      - "prql-php/**"
+      - ".github/workflows/test-php.yaml"
+  workflow_call:
+
```
max-sixty, line 9:
```suggestion

concurrency:
  # See notes in `pull-request.yaml`
  group: ${{ github.workflow }}-${{ github.ref }}-php
  cancel-in-progress: true
  
```

.github/workflows/test-php.yaml
max-sixty: Great, yes we keep the bindings in this repo so they can test & build together. So let's remove it. If there's something that it does that we might want in the future, we can add a comment here.

.github/workflows/test-php.yaml
vanillajonathan: Yeah, sure.

.github/workflows/test-php.yaml
vanillajonathan: Alright

vanillajonathan:
> Excellent! A couple of final clean-ups and then we can merge.
> 
> I know this has been quite the journey, thanks @vanillajonathan !

Yeah, writing the PHP code was quick, fun and easy. The CI stuff, not so much. 😅️

.github/workflows/test-php.yaml
max-sixty:
```suggestion
```

I just merged https://github.com/PRQL/prql/pull/1911

max-sixty:
> Yeah, writing the PHP code was quick, fun and easy. The CI stuff, not so much. 😅️

Yeah, this did end up being quite laborious for that. I'm surprised PHP isn't better supported by the ecosystem — generally these are as easy as adding an action and it works. But here it the action was _worse_ than running a single command...

From our perspective, without the tests, it's very difficult to manage the project, particularly for things we're not familiar with, like PHP. So really appreciate your patience in pushing through.

max-sixty:
Merging! Further work could be to update `prql-lib` to use the latest API, and pushing that change through here.

PR #1861, opened 2023-02-17
feat(prql-python)!: use Sql(None) for default Target and allow CompileOptions without target argument

eitsupi:
Part of #1855

Update to fall back to the target specified in the query header if no target is specified.


eitsupi:
This is the same content as #1856, so merge for now.
Please let me know if there are any problems.

PR #1866, opened 2023-02-17
docs: Add release notes from 0.5.1

max-sixty:


vanillajonathan:
When tagging a release, I think you should tag it with `v` prefix. Example `v0.5.1` not `0.5.1`.

max-sixty:
We just do the bare version @vanillajonathan

vanillajonathan:
Many other projects prefix with `v` which lets you configure CI pipelines to only trigger on tags prefixed with `v`.

PR #1877, opened 2023-02-18
docs: update prqlc and prql-compiler's description in the website index page

eitsupi:
Perhaps it seemed more appropriate to change the positions of prqlc and prql-compiler, so I switched them.

image


max-sixty:
Strong agree, thanks

max-sixty (approved):

Issue #1865, opened 2023-02-17
Allow `from_text` to produce numbers


max-sixty:
Currently `from_text` always produces strings:

```elm
from_text """
a,b,c
1,2,3
"""
```

```sql
WITH table_1 AS (
  SELECT
    '1' AS a,
    '2' AS b,
    '3' AS c
)
SELECT
  a,
  b,
  c
FROM
  table_1 AS table_0

-- Generated by PRQL compiler version:0.5.0 (https://prql-lang.org)
```

`'3' AS c` could instead be just `3 AS c`

That's because here, we always cast to a string:

https://github.com/PRQL/prql/blob/3f637524f4658a81c4c6795820f55496fe679515/prql-compiler/src/semantic/transforms.rs#L935

Two options for selecting the correct type:
- Use our existing parser
- Write a couple lines to decide whether it's a number or not

One other consideration — we may also need to demote _all_ values in a column to strings if one value isn't a number. An alternative to this would be forcing the user to use quotes (`'2'`) for numbers in a string column. Assuming the CSV parser handles the latter option correctly, that would be the closest option so we can make progress.

This would make a good early contribution for anyone who is up for a Rust PR.

fusillicode:
Hi 👋😄 

Thanks a lot for all your work 🙇‍♂️ !

I tried to tackle this issue and end up with the following PR: https://github.com/PRQL/prql/pull/1878

I think the code could be improved and refactored but before doing that I would like to get a feedback around the proposed solution based on your 1st suggested option, i.e. "Use our existing parser".

The proposed solution rationale is: try to get the `Literal`s from the parsed statements and fallback to the already in place `Literal::String` in case of any statements not matching a `Literal` value.

JFYI this PR has been worked on in the context of the Milan Open Source Saturday ![https://img.shields.io/badge/%E2%9D%A4%EF%B8%8F-open%20source%20saturday-F64060.svg](https://img.shields.io/badge/%E2%9D%A4%EF%B8%8F-open%20source%20saturday-F64060.svg) 😄

aljazerzen:
I don't like this, because parsing CSV columns as numbers automatically may produce unexpected results:

This is parsed as int column:
```csv
a
"1"
"2"
```

This is parsed as float column:
```csv
a
"1"
"2.2"
```

This is parsed as string column:
```csv
a
"1"
"2,2"
```

[...]

PR #1878, opened 2023-02-18
feat(prql-compiler): Allow from_text to produce numbers

fusillicode:
This PR tries to tackle the following issue: https://github.com/PRQL/prql/issues/1865

I think the code could be improved and refactored (and a test could be probably added to cover an edge case) but before doing that I would like to get a feedback around the proposed solution based on your 1st suggested option, i.e. "Use our existing parser".

The proposed solution rationale is: try to get the `Literal`s from the parsed statements and fallback to the already in place `Literal::String` in case of any statements not matching a `Literal` value.

JFYI this PR has been worked on in the context of the Milan Open Source Saturday  😄


max-sixty:
Thanks for starting this and appreciate the early PR!

I'll have a look at this later today. There's lots of code there in order to use our parser; to the extent it might be simpler to write a few lines to assess whether it's a number or int specifically for this. But also possibly we can simplify this code a bit.

Given the code is more complicated, it would be worth adding some more test cases — for example columns with mixed int / strings, or columns with only strings. ~I see there's an `.unwrap` there — does that panic if there's an empty column? We really want to avoid panics at this stage; tests would help clarify that.~

My guess is that this requires a small amount more work but not much.

Happy to have you as a contributor @fusillicode !

max-sixty:
@fusillicode I had a look at the code — I do think it's possible to simplify it. But first, to confirm if the approach is sound, could you add some more tests like these?

> Given the code is more complicated, it would be worth adding some more test cases — for example columns with mixed int / strings, or columns with only strings.

fusillicode:
Thank you really a lot for the quick feedback and support! 🙇‍♂️ 

I've added [this test](https://github.com/fusillicode/prql/blob/d125cd3482f7db20e02b1ec214e624bd3013617c/prql-compiler/src/test.rs#L3159-L3162) to check for a mix of different literals but I'm still not sure if we should handle all `Literal`s by returning them or check only for `int`, `null`, and maybe `bool` and fallback to `Literal::String` in all the other `Literal`s cases 🤔 

I'm also thinking in extracting the match on `Stmt` in a method or fn like `fn literal_or_none(stmt: &Stmt) -> Option<Literal>` OR maybe leverage [try_cast](https://github.com/fusillicode/prql/blob/d125cd3482f7db20e02b1ec214e624bd3013617c/prql-compiler/src/ast/pl/expr.rs#L347) somehow 🤔 

For anything I'm at your disposal!
Have a nice Sunday!

max-sixty:
As discussed at #1865, some salient questions:

- Should we use our parser or custom code?
  - I can have a go at simplifying the existing code a bit
- What happens if we pass examples like this to databases:

  ```elm
  from_text """
  a,b
  1,2
  a,b
  """
  ```

  Unfortunately BigQuery raises an error ` Column 2 in UNION ALL has incompatible types: STRING, INT64 at [7:3] `. So I think we'll need to check if _any_ values are strings, and then evaluate all values in that column to strings. FWIW a mix of floats & ints are fine, BigQuery coerces them to floats.

  I think we can implement that with a function that takes the CSV parser and returns which columns should be strings. It's fine it's not computationally efficient — e.g. if it clones the string and does another pass — this is very small, and if we really need, we can optimize later.

Does that sound reasonable to everyone?

prql-compiler/src/semantic/transforms.rs
```diff
@@ -930,18 +932,34 @@ mod from_text {
...
+                        .into_iter()
+                        .exactly_one()?
+                        .kind
+                        .as_main()
+                        .context("Expected main expr")?
+                        .as_ref()
+                        .kind
+                        .as_literal()
+                        .context("Expected literal expr")?
+                        .clone())
```
max-sixty, line 925: @fusillicode I simplified this a bit. It's still not elegant, but it is a bit simpler

fusillicode: @max-sixty definitely simpler and IMO even more elegant! Thanks a lot for working on it and sorry for my absence during the last couple of days 🙏

prql-compiler/src/semantic/transforms.rs
```diff
@@ -930,18 +932,34 @@ mod from_text {
...
+                .try_collect()
         }
 
         Ok(RelationLiteral {
             columns: parse_header(rdr.headers()?),
             rows: rdr
                 .records()
                 .into_iter()
-                .map(|row_result| row_result.map(parse_row))
+                // .map(|row_result| row_result.map(|r| r.map(parse_row)))
```
fusillicode: This comment could be removed right? 🤔

prql-compiler/src/semantic/transforms.rs
```diff
@@ -930,18 +932,34 @@ mod from_text {
...
+                .map(|row_result| row_result.map(parse_row)?)
```
fusillicode, line 935: I'm probably wrong here but...could this become a `flat_map()`? 🤔

prql-compiler/src/semantic/transforms.rs
max-sixty: Yes sorry! Remove away...

prql-compiler/src/semantic/transforms.rs
max-sixty: Yes, I think that's right — feel very feel to have a go...

aljazerzen: I don't think so.

Type here would be `Iterator<Result<Row>>`. 

We want to return an error if any of the rows errors out. If I understand correctly, you'd use `.flat_map` to convert result into iter. That would convert `Result::Err` into empty iterator and just suppress the error.

max-sixty: (good point, @aljazerzen is correct)

max-sixty:
Hi @fusillicode ! Wanted to check in on this. Is there anything I can help with?

I thought the initial effort was great, and it's not that far from completion. And we all agreed it would be a good feature.

fusillicode:
Hi @max-sixty! Sorry for the lateness of my reply. I've removed the unneeded commented code and rebased with `main`. 
I think that the only remaining thing is this:

```
I think we can implement that with a function that takes the CSV parser and returns which columns should be strings. It's fine it's not computationally efficient — e.g. if it clones the string and does another pass — this is very small, and if we really need, we can optimize later.
```

I'll try to give it a shot this weekend or, in the worst case scenario, the next 25th of April during the Open Source Saturday if that's ok.

Thanks really a lot for all your support and your patience ♾️🙇‍♂️

max-sixty:
> I think that the only remaining thing is this:

Yes, agree!

I think this is "moderate difficulty", but we can do it together if it becomes difficult.

I would:
- Run the current `parse_csv` function to get the rows as `Vec<Vec<Literal>>` (already done)
- Then iterate over the _columns_ of that — i.e. we want `rows[0][n]` for all `n`, then `rows[1][n]` for all `n`, etc
- Check whether _all_ values for each column can be cast into a float with `x.to_string().parse::<f32>().is_ok()`
- If they can, we're great.
- If they can't, then we don't want any numbers. So any values which _can_ be cast into a float should be replaced with `Literal::String(x.to_string())`, to force them into a string.

The test for this can be ensuring that 

```elm
    from_text format:csv """
    a,b,c
    '1',null,3.14
    4,'5',false
    """
```

...doesn't have any numbers. Since pasting the current output into [BigQuery](https://console.cloud.google.com/bigquery):

```sql
WITH
  table_1 AS (
  SELECT
    '1' AS a,
    NULL AS b,
    3.14 AS c
  UNION ALL
  SELECT
    4 AS a,
    '5' AS b,
    FALSE AS c )
SELECT
  *
FROM
  table_1
```

...gives:

```
 Column 1 in UNION ALL has incompatible types: STRING, INT64 at [8:7] 
```

---

FWIW, I realize this is a bit more complicated than I'd originally envisioned — to some extent we're doing a workaround for awkward SQL semantics. I do think it would be very valuable for PRQL specifically — i.e. it's worth doing something a bit involved.

max-sixty:
We now have relation literals, which allow expressing small amounts of data as PRQL code.

While the `from_text` CSV & JSON can still be useful, particularly when copying from output elsewhere, it's now fairly easy to get numbers & types by typing out a PRQL array. So I think that covers the case here, and we can close this.

I'm sorry that I underestimated how much work this would be — apologies to @fusillicode. I appreciate the effort — that it didn't end up being merged is on me... I'm still thinking through how we can better cater to new contributors, and I'll reflect on this case.

PR #1880, opened 2023-02-18
docs: Fix misplaced comma

richb-hanover:


website/content/roadmap.md
```diff
@@ -75,7 +75,7 @@ because fast iteration cycle can drastically decrease frustrations caused by
 banal misspellings.
 
 This requires development across multiple dimensions — writing an
-[LSP server](https://langserver.org/), better support for typing in the
+[LSP server,](https://langserver.org/) better support for typing in the
```
max-sixty, line 78: Surely the comma shouldn't be part of the link? It's a separator?

richb-hanover: This avoids a particularly annoying line break in my browser when the window is wide. (See image below.) This is my solution to a long-standing conundrum: placing closing punctuation (commas, periods) inside the `<a> ... </a>` tag to avoid this problem.

image

max-sixty: I think this is caused by the space being added after the external link token, which IIRC happens because the autoformatter adds a newline at the end of the file which adjusts external links.

The solution would be to fix that...

aljazerzen: It looks like this on my screen:

image

website/content/roadmap.md
```diff
@@ -75,10 +75,10 @@ because fast iteration cycle can drastically decrease frustrations caused by
...
 compiler, and possibly database cohesion.
 
-While PRQL compiler will never depend on a database to compile queries, LPS
+While PRQL compiler will never depend on a database to compile queries, an LSP
```
max-sixty, line 81: This was a good change! Would you be up for PR-ing this?

richb-hanover:
It's not worth changing our machinery to fix this single mis-placed comma. I'm just closing this PR

richb-hanover:
Dang - I forgot that. I'll create another PR.

PR #1881, opened 2023-02-18
build: Package`prqlc` as Snap

vanillajonathan:
Package the `prqlc` compiler as a [Snap](https://snapcraft.io/) app for Ubuntu and other Linux distributions using Snap.

The PRQL repository admin needs to add a secret with the name `STORE_LOGIN`. For info see: https://github.com/snapcore/action-publish#store-login

This CI pipeline makes use of:
- [snapcore/action-build](https://github.com/snapcore/action-build)
- [snapcore/action-publish](https://github.com/snapcore/action-publish)


.github/workflows/publish-snap.yaml
```diff
@@ -0,0 +1,26 @@
+name: publish-snap
+on:
+  release:
+    types: [published]
+
+jobs:
+  publish-snap:
```
max-sixty: Could we put this job into `release.yaml`? We can encapsulate all tasks on release there

.github/workflows/publish-snap.yaml
```diff
@@ -0,0 +1,26 @@
...
+      - name: 📂 Checkout code
+        uses: actions/checkout@v3
+      - name: 📦 Build Snap
+        uses: snapcore/action-build@v1
+        with:
+          path: prql-compiler
+      - name: 🆙 Publish Snap
+        uses: snapcore/action-publish@v1
+        env:
+          SNAPCRAFT_STORE_CREDENTIALS: ${{ secrets.STORE_LOGIN }}
```
max-sixty: I will register and add this secret in later today...

prql-compiler/snap/snapcraft.yaml
```diff
@@ -0,0 +1,25 @@
+name: prqlc # you probably want to 'snapcraft register <name>'
```
max-sixty: If possible, can we put this file into `/.config/snap`? 

If that's not possible somewhere in `prql-compiler/prqlc`, since we're publishing `prqlc`.

max-sixty (commented):
This is great, thank you @vanillajonathan !

How confident are you that it works? Is it possible to try testing locally? Once we merge it, it'll only run every release, so it's not that easy to iterate on.

.github/workflows/publish-snap.yaml
vanillajonathan: Yes, but we should wait probably get this pipeline to run green once first.

Currently this pipeline is configured to publish into the `edge` channel and the 'snapcraft.yaml` configured with `grade` set to `devel`. We can move this later into `release.yml` when we set `grade` to `stable` we publish into the `stable` or `candidate` channels instead of the `edge` channel.

prql-compiler/snap/snapcraft.yaml
vanillajonathan: It seems that wouldn't be possible.
https://github.com/so-fancy/diff-so-fancy/pull/434

.github/workflows/publish-snap.yaml
max-sixty: Good point

.github/workflows/publish-snap.yaml
```diff
@@ -0,0 +1,26 @@
+name: publish-snap
+on:
+  release:
+    types: [published]
```
max-sixty: Let's do the same thing that we're doing on https://github.com/PRQL/prql/pull/1883 — have it build on this PR so we can test it, and then when we can see it works, we'll move to `release.yaml`

prql-compiler/snap/snapcraft.yaml
```diff
@@ -0,0 +1,25 @@
...
+base: core22 # the base snap is the execution environment for this snap
+version: "0.1" # just for humans, typically '1.2+git' or '1.3.2'
+summary: Transpile PRQL queries into SQL! # 79 char long summary
```
max-sixty:
```suggestion
summary: PRQL is a modern language for transforming data
```

prql-compiler/snap/snapcraft.yaml
max-sixty: Thanks for looking into it. Possibly we should move into `prql-compiler/prqlc`, but we can change that later if that's easier

.github/workflows/publish-snap.yaml
max-sixty: I've added this to the repo. (Though until this PR is merged, it won't get access to it, so we should expect auth errors while we test.) 

Let's test that this gets as far as the auth error before we merge

.github/workflows/publish-snap.yaml
vanillajonathan: Yeah, sounds good.

vanillajonathan:
> This is great, thank you @vanillajonathan !
> 
> How confident are you that it works? Is it possible to try testing locally? Once we merge it, it'll only run every release, so it's not that easy to iterate on.

Not so confident, I've never made a Snap before, but this Snap should be rather simple; it has no dependencies and only need permission to the file system.

We could duplicate the pipeline to have it run whenever we want to.

max-sixty:
OK, let's enable it to run on this PR and see if it gets as far as an auth error...

eitsupi:
I know prqlc is a single binary tool, so is there any advantage to distributing the Snap and deb packages in a GitHub release?
If we distribute it as a single binary, we don't have to choose the distribution, but in the case of Snap and deb, can it be installed only on some Linux machines?

max-sixty:
> I know prqlc is a single binary tool, so is there any advantage to distributing the Snap and deb packages in a GitHub release?

This PR would publish `prqlc` to snap, so it's installable from there.

The "GitHub release" is just the event that triggers it to be published. 

Or am I misunderstanding?

eitsupi:
> This PR would publish `prqlc` to snap, so it's installable from there.

Sorry, I was totally mistaken, I agree it is great to be able to install right away with the snap command.

max-sixty:
OK! Thanks for asking, it's helpful!

vanillajonathan:
I am having some problems with this:

> error: failed to parse manifest at `/root/parts/prqlc/build/Cargo.toml`
>
> Caused by:
  error inheriting `version` from workspace root manifest's `workspace.package.version`

prql-compiler/snap/snapcraft.yaml
```diff
@@ -0,0 +1,25 @@
+name: prqlc # you probably want to 'snapcraft register <name>'
+base: core20 # the base snap is the execution environment for this snap
+version: "0.1" # just for humans, typically '1.2+git' or '1.3.2'
```
max-sixty: Lower priority but before we merge — we should have this be correct version and have something like this in the `prqlc` Cargo.toml so it gets update on every release:

https://github.com/PRQL/prql/blob/b430f03e05171c898ece420674d63511d17f2d69/prql-js/Cargo.toml#L44-L48

max-sixty:
> > error: failed to parse manifest at `/root/parts/prqlc/build/Cargo.toml`
> > Caused by:
> > error inheriting `version` from workspace root manifest's `workspace.package.version`

Have you tried running it locally? I would really encourage having a local environment for testing, otherwise the iteration cycle is very very long if you have to wait for GH Actions to run on each attempt. 

(Though if you can't get it working locally and using this to show the error, then that's great!)

---

My guess is here is that snapcraft has copied some of the repo into a path, but not the root `Cargo.toml`, which it requires in order to find the version, so it raises an error on not being able to find that file.

Instead, we might want to place the `snap/snapcraft.yaml` at the root of the repo, and tell it to compile with `cargo build -p prqlc`. That way, it has the context of the root `Cargo.toml`, which it needs, and has the package `prqlc` specified.

prql-compiler/snap/snapcraft.yaml
vanillajonathan: An alternative would be to set the version property to `git` which would expand it using the `git describe` command to something like `0.2.5-690-g311ccd9` as per the documentation. https://snapcraft.io/docs/snapcraft-yaml-reference

max-sixty: Great, good idea re `git`; this will give us a good result when we're on a tag

vanillajonathan:
I got it building locally.

The pipeline fails, but maybe you need to setup a secret?

```
Run snapcore/action-publish@v1
  with:
    release: edge
  env:
    SNAPCRAFT_STORE_CREDENTIALS: 
Publishing snap ""...
Error: login_data is empty
```

The Snap can be built locally by running `snapcraft`.

The Snap can be tested locally by running:

    unsquashfs prqlc_0.2.5+git693.4a8b915_amd64.snap
    snap try squashfs-root
    prqlc
    snap remove prqlc

snap/snapcraft.yaml
```diff
@@ -0,0 +1,29 @@
+name: prqlc # you probably want to 'snapcraft register <name>'
+title: PRQL Compiler
+base: core20 # the base snap is the execution environment for this snap
+version: git
+summary: PRQL is a modern language for transforming data
+description: |
+  PRQL is a modern language for transforming data — a simple, powerful,
+  pipelined SQL replacement.
+  It lets you transpile PRQL queries into SQL queries.
```
max-sixty:
```suggestion
  PRQL is a modern language for transforming data — a simple, powerful,
  pipelined SQL replacement.
```

I think we should stick to the standard description across all the releases. If we think the standard description isn't sufficiently descriptive, we can discuss at that level

max-sixty:
> The pipeline fails, but maybe you need to setup a secret?

Yes, I put the secret in the repo, so once it's merged it'll work.

So now let's put this in `release.yaml` and then we'll test it on the next release in a few days?

snap/snapcraft.yaml
vanillajonathan: I think we should include some some more paragraphs to better explain what it is and what it does and what it can do. A "one line" thing is good, but these descripion fields allow multiple paragraphs that we can use to help convey information better than just a oneliner can do alone.

I also think it should be mentioned that it is a CLI app and that the binary is `prqlc`else you might install this application and you might not know how to run it because it does not show up in any menu.

vanillajonathan:
Alright, but I think we could better utilize the `description` field with multiple paragraphs to better convey what PRQL is and what the PRQL compiler does, and how to run the it (The command is `prqlc`).

max-sixty:
> Alright, but I think we could better utilize the `description` field with multiple paragraphs to better convey what PRQL is and what the PRQL compiler does, and how to run the it (The command is `prqlc`).

Good point, edited in #1919

vanillajonathan:
Could you add a reference to this in the CHANGELOG?
I don't know under which header in the CHANGELOG it would go under.

max-sixty:
> Could you add a reference to this in the CHANGELOG? I don't know under which header in the CHANGELOG it would go under.

Can you try copying the one I just pushed for the `.deb`?

PR #1883, opened 2023-02-18
build: Package prqlc as .deb package

vanillajonathan:
This uses the [build-deb-action](https://github.com/marketplace/actions/build-deb-action) GitHub Action to create a `.deb` package for Debian and Debian-based Linux distributions such as Ubuntu.

To install download and run:

    sudo dpkg -i prqlc-0.5.1.deb


max-sixty:
Great! Currently getting a failure at https://github.com/PRQL/prql/actions/runs/4213264607/jobs/7312869920, but up for merging this when we get it working.

What does it do with the `.deb` file? Ideally we want it to be published as a release artifact

vanillajonathan:
It should run when tagged with a release that is a version number, now it fails because:

>  'Version' field value 'refs/pull/1883/merge': version number does not start with digit

Because the config is:
```yaml
version: '${{ github.ref }}' # refs/tags/v*.*.*
```
If I were to change it to `version: 1.0.0` then I guess it would work.

I don't know what it does with the `.deb` file. Maybe nothing. I don't know how to make a release artifact. I followed the instructions at [build-deb-action](https://github.com/marketplace/actions/build-deb-action) which didn't mention anything about about artifacts.

max-sixty:
OK, for the moment let's upload it as a normal artifact like https://github.com/PRQL/prql/blob/27f2c59369ffa63b5bed481db5b3b76b828e434e/.github/actions/time-compilation/action.yaml#L29-L33

Then we can upload it as a release artifact when we do releases from GitHub in the future

max-sixty:
> If I were to change it to `version: 1.0.0` then I guess it would work.

And we're going to need to do something to change the `refs/tags/0.6.0` to `0.6.0`, is that right?

vanillajonathan:
> > If I were to change it to `version: 1.0.0` then I guess it would work.
> 
> And we're going to need to do something to change the `refs/tags/0.6.0` to `0.6.0`, is that right?

I don't know, I never made a .deb package before, but it requires a version to be set, so we need a higher version number every time we make a release, and we wouldn't want to hardcode a value in the pipeline, we want it to take the pipeline from GitHub when we make a new release. The documentation for the build-deb-action uses `${{ github.ref }}`, with the comment `# refs/tags/v*.*.*`.

max-sixty:
> The documentation for the build-deb-action uses `${{ github.ref }}`, with the comment `# refs/tags/v*.*.*`

Right, we need to transform `refs/tags/*.*.*` that is returned by `${{ github.ref }}` into the `*.*.*`. 

The current version would fail with a similar error to the one we're seeing now https://github.com/PRQL/prql/actions/runs/4213264607/jobs/7312869920 with `'refs/pull/1883/merge'`

I can look if you need, but I think you'll be able to figure it out. How do other repos which use this action do that?

eitsupi:
https://github.com/PRQL/prql/pull/1881#issuecomment-1435811938

As I mentioned in my comment above, I don't think it is common practice to distribute single binary tools as deb packages on GitHub releases.

max-sixty:
> I don't think it is common practice to distribute single binary tools as deb packages on GitHub releases.

At the moment this would create a `.deb` artifact, it doesn't actually create a release artifact.

But even so — I'm looking at very popular projects like https://github.com/BurntSushi/ripgrep/releases/tag/13.0.0 — they publish `.deb`s. I agree many folks prefer binaries, but there's some audience for traditional packages. Or am I missing something?

eitsupi:
> But even so — I'm looking at very popular projects like [BurntSushi/ripgrep@`13.0.0` (release)](https://github.com/BurntSushi/ripgrep/releases/tag/13.0.0) — they publish `.deb`s. I agree many folks prefer binaries, but there's some audience for traditional packages. Or am I missing something?

I simply thought it might be a good idea to start with a single binary first, considering the management hassle.
I believe that the effort to install a deb package from a GitHub release is much the same as the effort to install a single binary.

If you agree with this, no objection.

max-sixty:
> If you agree with this, no objection.

Strongly agree! 

(And this doesn't even create the release item yet. But I'm very keen that folks can make small contributions which then add up into larger ones, it doesn't have to be fully complete to be helpful)

vanillajonathan:
https://github.com/PRQL/prql/actions/runs/4213264607/jobs/7312869920

Here we see that `${{ github.ref }}` expands into "refs/pull/1883/merge" and that the script runs:

    echo refs/pull/1883/merge
    sed -E s,^refs/tags/,,

This removes "refs/tags/" from the start of the string, which when tagged would transform "refs/tags/v1.0.0" into "v1.0.0" or "refs/tags/v1.0.0" into "1.0.0". Then it runs:

    sed -E s/^v//
    
Which strips the leading character "v" from the string, which would transform "v1.0.0" into "1.0.0".

This script should when we tag a release. I noticed previous releases has been tagged without a v-prefix. It seems the convention in Git is to prefix the version number with a "v". https://git-scm.com/book/en/v2/Git-Basics-Tagging

I don't know if any artifact will be produced or where it will be placed or how to reach it.

max-sixty:
But at the moment I think we get this error:

```
 'Version' field value 'refs/pull/1883/merge': version number does not start with digit
```

I haven't looked through the `sed` — but is that working correctly?

---

I also found this on Google, which looks like a simpler way: https://github.com/orgs/community/discussions/26686

---

We just do the bare number; no `v`, both approaches are common and acceptable

vanillajonathan:
The sed works as explained in the post above, it removes "refs/tags/" from the start of the string. If we tagged a release, we would get "refs/tags/1.0.0" which would transform into "1.0.0" but since the CI runs on this PR we get "refs/pull/1883/merge" which no transform happens on because it doesn't start with "refs/tags/". So the version will be transform correctly when the CI is triggered when a release is tagged, but does not transform on this PR.

max-sixty:
> The sed works as explained in the post above, it removes "refs/tags/" from the start of the string. If we tagged a release, we would get "refs/tags/1.0.0" which would transform into "1.0.0" but since the CI runs on this PR we get "refs/pull/1883/merge" which no transform happens on because it doesn't start with "refs/tags/". So the version will be transform correctly when the CI is triggered when a release is tagged, but does not transform on this PR.

Ah OK, that makes sense, sorry for misunderstanding.

But
- I don't see `sed` in the currently proposed code?
- I think just using `${{ github.ref_name }}` would be simpler than using `sed`

vanillajonathan:
`sed` is not in the current proposed code but it is part of the `build-deb-action` GitHub Action. If you look at https://github.com/PRQL/prql/actions/runs/4220116103/jobs/7326108887 then you will see it in the output. We use `${{ github.ref }}` in the yaml file then `build-deb-action` will transform it using `sed` which is used internally by the GitHub Action.

.github/workflows/create-deb.yaml
```diff
@@ -0,0 +1,34 @@
...
+on:
+  release:
+    types: [released]
+  pull_request:
+    paths:
+      - .github/workflows/create-deb.yaml
+  workflow_call:
+
+jobs:
+  package:
```
max-sixty:
```suggestion
  package:
  # This doesn't currently publish anywhere. We can add publishing in a follow-up PR
```

.github/workflows/create-deb.yaml
```diff
@@ -0,0 +1,34 @@
+name: create-deb
+
+on:
+  release:
+    types: [released]
```
max-sixty: Now this works, let's move this to `release.yaml` and remove this file?

max-sixty:
OK great!

Then as soon as we move the workflow into `release.yaml`, we can merge.

.github/workflows/create-deb.yaml
vanillajonathan: Yeah, I don't really know what happens with the build, maybe it ends up somewhere where you can add it as an artifact on a release or something? Not sure how this works.

max-sixty: Yes, we can do something like https://github.com/PRQL/prql/blob/4a8b915eef940b5ef1231b915ceaa210b7a1e86b/.github/actions/time-compilation/action.yaml#L29 for normal artifacts.

Then there are also "release artifacts", but that requires running the whole release through GitHub, which we don't do yet https://github.com/softprops/action-gh-release

.github/workflows/create-deb.yaml
vanillajonathan: Yeah, unless you feel the `release.yaml` file would get messy, and there is a way for the `release.yaml` file to call upon this `create-deb.yaml` file as a step. I don't know how this works but in the future we might want to use [`build-rpm-action`](https://github.com/marketplace/actions/build-rpm-action) to build `.rpm` packages for Fedora and Fedora-based Linux distributions, and packages for macOS.

max-sixty: Yes, I think we should try and consolidate them unless there's a reason to have them be separate. 

We have different ones for `test-python` etc, because we want to run those on changes to the `prql-python` path. but that's not required here.

vanillajonathan: Alright

vanillajonathan:
Okay, I moved it into the `release.yaml` file.

.github/workflows/release.yaml
```diff
@@ -24,6 +24,30 @@ jobs:
...
+          cp target/release/prqlc .debpkg/usr/bin/prqlc
+          chmod +x .debpkg/usr/bin/prqlc
+      - name: 📦 Build .deb package
+        uses: jiro4989/build-deb-action@v2
+        with:
+          package: prqlc
+          package_root: .debpkg
+          maintainer: The PRQL Project
+          version: '${{ github.ref }}'
+          desc: PRQL compiler transpiles PRQL queries into SQL queries.
```
```suggestion
          desc:  prqlc is the CLI for PRQL, a modern language for transforming data
```

.github/workflows/release.yaml
```diff
@@ -24,6 +24,30 @@ jobs:
...
             })
 
+  build-deb-package:
+    runs-on: ubuntu-latest
+    steps:
+      - name: 📂 Checkout code
+        uses: actions/checkout@v3
+      - uses: actions-rs/toolchain@v1
+        with:
+          toolchain: stable
```
```suggestion
```

This is included in the runner

.github/workflows/release.yaml
```diff
@@ -24,6 +24,27 @@ jobs:
               }
             })
 
+  build-deb-package:
```
max-sixty, line 27:
```suggestion
  build-deb-package:
  # TODO: currently this doesn't publish it as an artifact; would be a welcome contribution
  # to add that (and eventually as a release asset)
```

max-sixty:
Made a couple of edits, hope that's OK.

Now merging!

vanillajonathan:
The description could contain a few more paragraph to better explain what it is and what it does.

Could you add this to CHANGELOG? I am not sure under where it should go or how it should be described.

max-sixty:
> Could you add this to CHANGELOG? I am not sure under where it should go or how it should be described.

Done!

PR #1893, opened 2023-02-19
build: add basic devcontainer.json (for Rust, JavaScript, Python)

eitsupi:
Adresses #1808


max-sixty:
I tried this — it works pretty well! Because it's a devcontainer, I lose shell customizations, but if I were first arriving at a project, then it would still be a good way of quickly starting on that project.

It does take a long time to run `task install-cargo-tools`, and yes I got an error too (though on restarting VSCode it went away). Given that the image is partly pre-built, is there any way of building those in CI, maybe as a Docker image, or as one of the "features"? They already build into our standard Docker image in CI.

eitsupi:
Thank you for trying this!
Yeah, the installation takes an incredibly long time, I think it took over 30 minutes when I tried it on 2 core Codespaces......

> It does take a long time to run `task install-cargo-tools`, and yes I got an error too (though on restarting VSCode it went away). Given that the image is partly pre-built, is there any way of building those in CI, maybe as a Docker image, or as one of the "features"? They already build into our standard Docker image in CI.

It is possible, but wouldn't it still take a long time to build in the end?
In the end, most of the time is spent compiling the Rust packages, so it will take the same amount of time no matter which layer we install as long as you install these packages.

I am wondering if it is possible to define another task that can install only minimal dependencies in a short time.

max-sixty:
> It is possible, but wouldn't it still take a long time to build in the end?
> In the end, most of the time is spent compiling the Rust packages, so it will take the same amount of time no matter which layer we install as long as you install these packages.

I was thinking it might be possible to compile them in CI and then publish them.

For example, with Docker, we could compile these tools in a `docker build` and `docker push` them to a repository. Then when someone `docker pull`s the image, they have the tools immediately, there's no need for them to re-compile them.

Is that possible with this model too? For example, a `"ghcr.io/prql/devcontainer-features/cargo-tools:1"`?

max-sixty:
From a quick browse, it seems like pre-building is not the focus of devcontainers: https://code.visualstudio.com/blogs/2022/09/15/dev-container-features and https://containers.dev/implementors/spec/ — instead the just package the `.sh` scripts. We could make our own base image, but maybe that's more maintenance.

So maybe this isn't viable.

Another alternative is `cargo binstall`, which would make it much faster — one compilation step, and then mostly just downloading binaries. We could move the main `task` onto this. 

(or any other ideas are very welcome)

> I am wondering if it is possible to define another task that can install only minimal dependencies in a short time.

If we really need to, we could do this. I would say that tools like this tend to decay, and there's a potentially bad cycle of them falling out of date -> people using them less -> them not being maintained -> falling further out of date. 

So one list which is tested is much better than two lists. But if we need to, that's totally fine with me.

eitsupi:
Thanks for your response.

> From a quick browse, it seems like pre-building is not the focus of devcontainers: [code.visualstudio.com/blogs/2022/09/15/dev-container-features](https://code.visualstudio.com/blogs/2022/09/15/dev-container-features) and [containers.dev/implementors/spec](https://containers.dev/implementors/spec/) — instead the just package the `.sh` scripts. We could make our own base image, but maybe that's more maintenance.

It is possible to pre-build using the Dev Container CLI.
In fact, the image `mcr.microsoft.com/devcontainers/rust:0-1-bullseye` used here as a base image was built that way.
https://github.com/devcontainers/images/blob/b361b069291b78b97396eda62ed6d6f62266a36f/src/rust/README.md

However, the problem here is the packages installed by cargo anyway, so I think we can greatly reduce the build time for each user by simply pushing an image built from a simple Dockerfile like the following to GHCR and using that as the base image.

```dockerfile
FROM mcr.microsoft.com/devcontainers/rust:0-1-bullseye
USER vscode
RUN cargo install cargo install cargo-audit cargo-insta cargo-release \
        default-target mdbook mdbook-admonish mdbook-toc wasm-bindgen-cli \
        wasm-pack
USER root
```

The cons of this method are that it takes probably 3 hours to do a Docker Build with a multi-platform build of amd64 and arm64, and we need to periodically push the built image to the Registry. (Perhaps it would make sense to use Dependabot to update the Dockerfile once a month or so when the base image is updated and trigger the build.)

Nevertheless, I would give this a try as it will definitely improve the development experience for users.

> If we really need to, we could do this. I would say that tools like this tend to decay, and there's a potentially bad cycle of them falling out of date -> people using them less -> them not being maintained -> falling further out of date.
>
> So one list which is tested is much better than two lists. But if we need to, that's totally fine with me.

My concern was installing all the packages, including the need to install `mdbook` even when doing work unrelated to the book.
I think sometimes it is better to install the necessary tools as needed for the subprojects that need them.
Of course, we can support the method of defining higher-level tasks and installing them all at once.

max-sixty:
Thanks for the response @eitsupi 

> However, the problem here is the packages installed by cargo anyway, so I think we can greatly reduce the build time for each user by simply pushing an image built from a simple Dockerfile like the following to GHCR and using that as the base image.

Yes, that could be great, that's somewhat what I was thinking re pre-building

> The cons of this method are that it takes probably 3 hours to do a Docker Build with a multi-platform build of amd64 and arm64, and we need to periodically push the built image to the Registry. (Perhaps it would make sense to use Dependabot to update the Dockerfile once a month or so when the base image is updated and trigger the build.)

I think this is fine — we can build in CI and push automatically. It's very fast when it builds from cache; e.g. https://github.com/PRQL/prql/actions/runs/4226792538/jobs/7340613823, even though I imagine it does take a long time to build from scratch. But worst case we can just build on every release. I think we can do this without any marginal manual work on each image update. 

Seeing whether dependabot can update the base image would be great too.

(I'm not sure whether we can use the same image for this and our existing docker setup — maybe we can _only_ use the `mcr.microsoft.com/devcontainers/rust` base and replace the current base?)

> My concern was installing all the packages, including the need to install `mdbook` even when doing work unrelated to the book.
> I think sometimes it is better to install the necessary tools as needed for the subprojects that need them.

I agree with the point, but it's also work to maintain the lists (and not easy to automatically test them), so it's a tradeoff. And assuming we can build first, then it's very low cost to have all the tools — something like 2MB per binary.

That said, if folks feel strongly and are OK maintaining them, we can split up the lists.

eitsupi:
> we can build in CI and push automatically. It's very fast when it builds from cache

I don't know how best to use cache here......perhaps this feature?
I have never used it and will have to give it a try.
https://github.com/moby/buildkit/blob/master/frontend/dockerfile/docs/reference.md#run---mounttypecache

> (I'm not sure whether we can use the same image for this and our existing docker setup — maybe we can _only_ use the `mcr.microsoft.com/devcontainers/rust` base and replace the current base?)

I see no particular reason why it cannot be replaced.
Eventually it would be great if the pre-built images could be used for other purposes (mdbook preview via docker compose) as well.

max-sixty:
> I don't know how best to use cache here......perhaps this feature?
> I have never used it and will have to give it a try.
> https://github.com/moby/buildkit/blob/master/frontend/dockerfile/docs/reference.md#run---mounttypecache

We already do this! https://github.com/PRQL/prql/blob/68752b7806cee40af10b22d44609230ed2a78dc1/.github/workflows/nightly.yaml#L92-L93

So I think the only things we need to change to get the Docker part of this working:
- Change the base of our existing image
- Push to an image repository in CI

> I see no particular reason why it cannot be replaced.

I agree

eitsupi:
> We already do this!
> 
> https://github.com/PRQL/prql/blob/68752b7806cee40af10b22d44609230ed2a78dc1/.github/workflows/nightly.yaml#L92-L93

My understanding is that this cache is disabled when the base image (base layer) is changed.

I don't think it's a big deal if the cache doesn't work, so anyway I'll prototype and push it to this branch.......

max-sixty:
> My understanding is that this cache is disabled when the base image (base layer) is changed.
> 
> I don't think it's a big deal if the cache doesn't work

Yes agree!

eitsupi:
> I don't know how best to use cache here......perhaps this feature?
> I have never used it and will have to give it a try.
> [moby/buildkit@`master`/frontend/dockerfile/docs/reference.md#run---mounttypecache](https://github.com/moby/buildkit/blob/master/frontend/dockerfile/docs/reference.md?rgh-link-date=2023-02-21T13%3A57%3A41Z#run---mounttypecache)

Note: It does not seem possible to run this on CI and share the cache between images now. docker/build-push-action#716

.github/workflows/devcontainer.yaml
```diff
@@ -0,0 +1,47 @@
...
+          if [ "${{ github.ref }}" = "refs/heads/main" ]; then
+            echo "platforms=linux/amd64,linux/arm64" >>$GITHUB_OUTPUT
+          else
+            echo "platforms=linux/amd64" >>$GITHUB_OUTPUT
+          fi
+
+      - name: Pre-build dev container image
+        uses: devcontainers/ci@v0.2
+        with:
+          subFolder: .devcontainer/base
```
eitsupi: It does not seem to be possible to configure this way; I have created an issue.
devcontainers/ci#212

.github/workflows/devcontainer.yaml
```diff
@@ -0,0 +1,53 @@
...
+          imageName: ghcr.io/prql/prql-devcontainer
+          cacheFrom: ghcr.io/prql/prql-devcontainer
```
eitsupi: Is this a good name for the image to push?
Or like `ghcr.io/prql/prql-base-devcontainer`?

max-sixty: Yes I think it's great

.github/workflows/devcontainer.yaml
max-sixty: Hmmm. I wonder whether this is substantially better than using a normal docker workflow like we have now. Because I _think_ the only thing that's pushed is the output of the Docker build. Is that right?

If it is, it might be that we can mostly swap out the existing Dockerfile for the new one, and use the existing workflow...

max-sixty:
> > I don't know how best to use cache here......perhaps this feature?
> > I have never used it and will have to give it a try.
> > [moby/buildkit@`master`/frontend/dockerfile/docs/reference.md#run---mounttypecache](https://github.com/moby/buildkit/blob/master/frontend/dockerfile/docs/reference.md?rgh-link-date=2023-02-21T13%3A57%3A41Z#run---mounttypecache)
> 
> Note: It does not seem possible to run this on CI and share the cache between images now. [docker/build-push-action#716](https://github.com/docker/build-push-action/issues/716)

@eitsupi it's much easier than this for the basic Docker caching — this does the basic build cache already, using the image repository: https://github.com/PRQL/prql/blob/de474182a23ccc7de70b417692228905e717a697/.github/workflows/nightly.yaml#L92-L93

The "cache mount" is for a more advanced form of caching where, for example, you build lots of _different_ images and want to share a cargo cache between them. But this isn't that helpful for us.

Does that make sense? Tell me if I'm being confusing! (or mistaken!)

.github/workflows/devcontainer.yaml
eitsupi: > it's much easier than this for the basic Docker caching — this does the basic build cache already, using the image repository:

Inline cache is used here.

.github/workflows/devcontainer.yaml
eitsupi: > I wonder whether this is substantially better than using a normal docker workflow like we have now. Because I _think_ the only thing that's pushed is the output of the Docker build. Is that right?

Since Dev Container Features are a fairly long shell script, I think it provides a better setup than simply installing Python, etc.
Dev Container CLI is required to build a Docker image with Dev Container Features installed.

But as you point out, it doesn't support all the features of `buildx build` (such as GitHub Actions cache support), so it's worth moving to a simple Docker file-only build.
Is that better for you?
(In that case, change the installation of Dev Container Features to be done at the user's end. This takes some time, but is far more minimal than `cargo install`.)

.github/workflows/devcontainer.yaml
max-sixty: Yes great! That's all the caching we need; no need for the build mount

.github/workflows/devcontainer.yaml
max-sixty: Oh I think I see now — this will cache running the steps in `.devcontainer/devcontainer.json` as well as the Dockerfile? That's excellent if so!

~In that case I would probably keep it as-is, because then folks don't need to do any building themselves, they get a fully baked image...~ Writing down my understanding below

max-sixty:
If my understanding is correct here: https://github.com/PRQL/prql/pull/1893/files#r1117593795, then I think we have two options:

**Build docker image to hold everything**

- We build a docker image based on the devcontainer using this `devcontainer.json` in CI: https://github.com/PRQL/prql/pull/1893/files#diff-c28c246b0792371d4cb69440d78cc47b4876ea34eb98618d7daafa41f96ce10bR3. We use the `devcontainer` GitHub Action
- We push that to the image repository
- Then we supply a _different_ `devcontainer.json` which is similar to that file, but the image value is a reference to the image we've built & pushed

That way, folks get a devcontainer with everything already built onto the base image. But we need two `devcontainer.json` files.

**Build base docker image only**

- Build a docker image which contains the cargo tools with a normal docker workflow. We put these in an image since they take a long time to build (and we could add anything else there that was expensive, but need to express them as Dockerfile commands, not as references to devcontainer features).
- Reference that image in a single `devcontainer.json`

This would be simpler, and IIUC is what you're suggesting in https://github.com/PRQL/prql/pull/1893#discussion_r1117103494. It requires a small amount of building on the client side — the devcontainer features that aren't built.

We then wouldn't use the devcontainer GitHub Action. Is that correct?

---

Both seem reasonable, the second seems simpler and easier to get something working; and could grow into the second in time if needed.

I think we could replace our existing docker build with this either way: https://github.com/PRQL/prql/blob/main/Dockerfile.

---

Thanks for keeping pushing on this. Feel free to proceed as you think best!

book/src/contributing/using-dev-container.md
```diff
@@ -0,0 +1,68 @@
...
+
+2. Edit files locally (say, using VS Code) to work on PRQL. The tools in the Dev
+   Container watch for changed files and rebuild the project as needed.
+
+3. When you are satisfied with the changes, you can commit them to the repo, and
+   push the changes as usual.
+
+## Starting and stopping the Dev Container
+
+_Instructions needed - both from VS Code and maybe from CLI?._
```
max-sixty: I would generally deemphasize documenting anything that isn't PRQL-specific, such as how to use a dev container, and instead just link to the official documentation. It's often not as good as the official version, difficult to maintain, and accrues debt that's difficult to assess.

That said, I think this is well-written, so we don't have to remove it, but I wouldn't invest more.

book/src/contributing/using-dev-container.md
```diff
@@ -0,0 +1,68 @@
...
+
+- [Docker](https://www.docker.com/) (follow any of the instructions on the
+  internet for your computer)
+- [VS Code](https://code.visualstudio.com/) Interactive development environment
+- [Dev Containers](https://marketplace.visualstudio.com/items?itemName=ms-vscode-remote.remote-containers)
+  extension for VS Code
+
+When you first start the Dev Container, the build process may take a long time
+(as much as 20-40 minutes) as the container collects all the packages of the
+full PRQL toolchain. Once it's running, though, startup is fairly fast.
```
max-sixty: This part is helpful PRQL-specific context though

book/src/contributing/using-dev-container.md
```diff
@@ -0,0 +1,68 @@
...
+  (Port 3000)
+
+- `task run-website` starts a `hugo` server. As you edit the files (in the
+  `website` directory), `hugo` rebuilds those pages. (Port 1313)
+
+- `task run-playground` starts a Node server to build the Playground. As you
+  edit the files (in the `playground` directory), the server rebuilds those
+  pages. (Port 3000)
+
+- `task WHAT ELSE?` _Provide explanation of other useful commands._
```
max-sixty: I would just link to our existing docs here

book/src/contributing/using-dev-container.md
```diff
@@ -0,0 +1,68 @@
...
+_Codespaces (from Github) provide equivalent facilities (and a web-based IDE).
+This document could also describe how to set this up._
+
+A Dev Container is a way to package a number of "developer tools" (compilers,
+bundlers, package managers, loaders, etc.) into a single object. This is helpful
+when many people want to contribute to a project: each person only has to
+install the (single) Dev Container on their own machine to start working. By
+definition, the Dev Container has a consistent set of tools that are known to
+work together. This avoids a fuss with finding the proper versions of each of
+the build tools.
```
max-sixty: This is still a reasonable summary to include, it's very well-written!

book/src/contributing/using-dev-container.md
```diff
@@ -0,0 +1,68 @@
...
+## How you use it
+
+1. Clone the git repo onto your hard drive as usual. Then you start the Dev
+   Container (say, using VS Code) that bundles all the developer tools.
+
+2. Edit files locally (say, using VS Code) to work on PRQL. The tools in the Dev
+   Container watch for changed files and rebuild the project as needed.
+
+3. When you are satisfied with the changes, you can commit them to the repo, and
+   push the changes as usual.
```
max-sixty: I probably wouldn't keep this, though. But fine to leave it in for the moment and see how we go

book/src/contributing/using-dev-container.md
eitsupi: For the record, this document is entirely the credit of @richb-hanover.
Sorry, but I have very less hand in the documentation yet.

I agree that it is not necessary to explain here how to use Dev Container in general.
And, this document does not sufficiently distinguish Dev Container from VS Code Dev Container (extension) and I need to minor correct this.......

For example, https://github.com/pandas-dev/pandas has included devcontainer.json for several years, and described as this.
https://pandas.pydata.org/docs/development/contributing_environment.html?highlight=devcontainer#id5

richb-hanover:
All good points. I wrote this up as an on-boarding document for newcomers to PRQL and especially those who are also new to Dev Containers. The hints here are a collection of the stuff I had to learn while casting about to get it going. 

Most importantly, I feel the need to describe "what a Dev Container is good for" (working on the Book, Playground, Website, not so much improving the compiler) so that newcomers can see the goal and decide whether the startup hassle is worth it.

That said, please edit away! Thanks.

book/src/contributing/using-dev-container.md
max-sixty: > For the record, this document is entirely the credit of @richb-hanover.

Ah congrats @richb-hanover ! It is very well-written!

> For example, https://github.com/pandas-dev/pandas has included devcontainer.json for several years, and described as this.
https://pandas.pydata.org/docs/development/contributing_environment.html?highlight=devcontainer#id5

Yes I think that's about the right amount of explanation...

book/src/contributing/using-dev-container.md
```diff
@@ -0,0 +1,68 @@
...
+pre-requisites:
+
+- [Docker](https://www.docker.com/) (follow any of the instructions on the
+  internet for your computer)
+- [VS Code](https://code.visualstudio.com/) Interactive development environment
+- [Dev Containers](https://marketplace.visualstudio.com/items?itemName=ms-vscode-remote.remote-containers)
+  extension for VS Code
+
+When you first start the Dev Container, the build process may take a long time
+(as much as 20-40 minutes) as the container collects all the packages of the
```
eitsupi: Note: This should not be the case when this is merged since we will have pre-built images.

eitsupi:
@max-sixty I have changed to build images to be pushed to GHCR from pure Dockerfile.

I would like to improve the documentation and the Taskfile after this (split only the part copied in the Dockerfile into a separate Taskfile to minimize cache invalidation due to Taskfile updates), but can we merge first to see if it works in its current state?

Also, I believe this base image can replace most of the Dockerfile roles currently in the project root, and we can confirm this after merging this PR.

book/src/contributing/using-dev-container.md
max-sixty: Agree, can we remove?

book/src/contributing/using-dev-container.md
```diff
@@ -0,0 +1,55 @@
...
+full PRQL toolchain. Once it's running, though, startup is fairly fast.
+
+## How you use it
+
+1. Clone the git repo onto your hard drive as usual. Then you start the Dev
+   Container (say, using VS Code) that bundles all the developer tools.
+2. Edit files locally (say, using VS Code) to work on PRQL. The tools in the Dev
+   Container watch for changed files and rebuild the project as needed.
```
max-sixty: Is there actually watching going on? I think the volumes are mapped in, but there's no "watch" task, as it would be commonly known as.

(btw this is an example where just linking to docs is easier — can we actually explain this better than VSCode itself can?)

book/src/contributing/using-dev-container.md
```diff
@@ -0,0 +1,55 @@
...
+3. When you are satisfied with the changes, you can commit them to the repo, and
+   push the changes as usual.
+
+## Starting and stopping the Dev Container
+
+_Instructions needed - both from VS Code and maybe from CLI?._
```
max-sixty: I would vote to just link to the standard docs

.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,26 @@
...
+    "ghcr.io/devcontainers/features/python:1": {},
+    "ghcr.io/devcontainers/features/node:1": {}
+  },
+  "customizations": {
+    "vscode": {
+      "extensions": [
+        "budparr.language-hugo-vscode",
+        "mitsuhiko.insta",
+        "prql-lang.prql-vscode",
+        "esbenp.prettier-vscode"
```
max-sixty: If this supports comments, let's link to https://github.com/PRQL/prql/blob/e132c3de9b1099f0fa0d7929c921a301eaccccc4/.vscode/extensions.json (and link back from there to here) so we keep them all in sync. We should add `rust-analyzer` here too

.devcontainer/base-image/Dockerfile
```diff
@@ -0,0 +1,15 @@
+FROM mcr.microsoft.com/devcontainers/rust:0.203.7-1
+
+RUN apt-get update \
+    && apt-get install -y --no-install-recommends \
+        cmake \
+    && rm -rf /var/lib/apt/lists/* \
+    && sh -c "$(curl --location https://taskfile.dev/install.sh)" -- -d -b /usr/local/bin
+
+# ========= Install cargo-tools for non-root user (vscode) =========
```
max-sixty: I'm guessing this is here for good reason — I added a link to the docs, though I couldn't get to a conclusive answer. 

If it still works without switching users, I would vote to not switch users — unless we have a good reason?

```suggestion
# ========= Install cargo-tools for non-root user (vscode) =========
# Based on https://code.visualstudio.com/docs/devcontainers/create-dev-container#_install-additional-software
```

.github/workflows/devcontainer.yaml
```diff
@@ -0,0 +1,63 @@
+name: devcontainer
+
+on:
+  push:
+    # ToDo: Add partial Taskfile (only include rust-tools)
```
max-sixty: We should indeed add `Taskfile.yml`. 

It's fine to run workflows too often, and we're never going to get it precise, so we should lean on over-building.

.github/workflows/devcontainer.yaml
```diff
@@ -0,0 +1,63 @@
+name: devcontainer
+
+on:
+  push:
```
max-sixty: I think the simplest approach would be to copy the existing workflows like `test-java.yaml` — run on `pull_request`, and then add a call from `test-all.yaml` so it runs on every merge. That removes the need for a `push` action.

Then if something changes that we don't track, we still find out about it. And assuming this is cached, then the build is very fast, so it's fine to run on every merge.

Taskfile.yml
```diff
@@ -52,6 +52,17 @@ tasks:
           🟢 Setup complete! ✅🚀"
         silent: true
 
+  setup-devcontainer:
```
max-sixty: 👍

book/src/contributing/using-dev-container.md
```diff
@@ -0,0 +1,55 @@
+# Development in Dev Containers
+
+```admonish note
+This document is a heavy work in progress and may not contain sufficient information.
+```
```
max-sixty: I applaud the approach of disclaiming when something isn't complete, and merging quickly rather than waiting until everything is perfect.

But is there a specific thing these docs lack? To me this looks fairly complete! 

Disclaimers like this should either a) be specific, so we know the next thing to do, or b) have someone's name attached, so they can own getting it to the next stage. Otherwise it's very likely to be there for a long time...

.github/workflows/devcontainer.yaml
```diff
@@ -0,0 +1,63 @@
...
+
+      - name: Set up QEMU
+        uses: docker/setup-qemu-action@v2
+
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v2
+
+      - id: set-platforms
+        run: |
+          if [ "${{ github.ref }}" = "refs/heads/main" ]; then
```
max-sixty: Can we try running this on the PR once, so we can test it locally? When I try testing locally, it can't find the image, I guess because it hasn't been pushed yet.

book/src/contributing/using-dev-container.md
```diff
@@ -0,0 +1,55 @@
...
+
+_Instructions needed - both from VS Code and maybe from CLI?._
+
+## Using PRQL in a Dev Container
+
+After the Dev Container starts up, the VS Code Terminal pane shows a Dev
+Container command line. Enter one of the following commands to work on a
+component of PRQL.
+
+VS Code offers a link to "Open in Browser" to the proper port to see the results
```
max-sixty: FYI I wasn't familiar with this, and couldn't test it yet — as long as someone is confident then great

book/src/contributing/using-dev-container.md
```diff
@@ -0,0 +1,55 @@
...
+- `task -l` lists all the available tasks.
+- `task run-book` starts an `mdbook` server. As you edit the files of the
+  Language Book (in the `book` directory), `mdbook` rebuilds those pages.
+  (Port 3000)
+- `task run-website` starts a `hugo` server. As you edit the files (in the
+  `website` directory), `hugo` rebuilds those pages. (Port 1313)
+- `task run-playground` starts a Node server to build the Playground. As you
+  edit the files (in the `playground` directory), the server rebuilds those
+  pages. (Port 3000)
+- `task WHAT ELSE?` _Provide explanation of other useful commands._
```
max-sixty: This is actually quite a good list — maybe move to the `developing.md`? Unless there's a specific reason to have this here?

Or put them in `Taskfile.yaml`'s tasks' summary so it's next to the code?

This section should cover just what's _different_ about a dev container.

book/src/contributing/using-dev-container.md
```diff
@@ -0,0 +1,55 @@
+# Development in Dev Containers
```
max-sixty: FWIW I would match the title of the file with the name of the file

We also need to add this file to the book index

max-sixty (approved):
Great, code looks v good. 

Some comments about firming up / slimming down the docs — they are well written but it's also important we set them up to improve through time rather than decay — which is more likely if they're concise & specific.

Definitely good to merge and iterate!

.devcontainer/base-image/Dockerfile
eitsupi: My understanding is that `cargo install` and `task` are usually assumed to be run with user privileges.
So while it may not be necessary to switch from root user here (I may have tried, but I forget), I think it would be more appropriate to do so.

.devcontainer/devcontainer.json
eitsupi: I think I commented on this before, but did not write it here because the `rust-analizer` is already included in the base image metadata.

However, adding it here is not a problem, so let's add it.

eitsupi:
@max-sixty Thank you for your review.
After reviewing your review and thinking again, it seems to me that it would be better to split the PR since this PR contains the following three things.

1. Definition and building workflow for devcontainer base image.
2. Definition for Dev Container.
3. Document about 2.

As you pointed out, 2 will not work until after 1 is merged and the image is built.
So first, 1 is separated from this PR and merged, and the base image is ready to be pulled.
Then further improve and merge this PR with 2 and 3 (and also push to `web` branch).

.github/workflows/devcontainer.yaml
eitsupi: As already commented https://github.com/PRQL/prql/pull/1893#issuecomment-1455409575, I will separate this part into a separate PR and merge it first so that we can try it out.

max-sixty:
> it seems to me that it would be better to split the PR since this PR contains the following three things.

That's reasonable; it probably would be a better split.

But it's also fine to just merge! Whatever gets us to the final destination fastest :)

.devcontainer/base-image/Dockerfile
max-sixty: You're correct — they use the `vscode` user upstream... https://github.com/microsoft/vscode-dev-containers/blob/5a084a93b0736ea86395ac99019a5b72a00b6341/containers/rust/.devcontainer/base.Dockerfile

eitsupi: FYI, that is the old definition.
Current definition is this. (The name `vscode` is retained for compatibility)
https://github.com/devcontainers/images/blob/c087e361571e2200c3b40ed86d14a8ced84c388a/src/rust/.devcontainer/devcontainer.json#L9

eitsupi:
The base image is now available by #2028
https://github.com/PRQL/prql/pkgs/container/prql-devcontainer-base

Try using this and revise the documentation.

book/src/contributing/using-dev-container.md
eitsupi: I believe VS Code detects when some program uses the port in the container and issues a notification and forwards it to the host machine port. (I don't understand the details of how it works!)

In any case, this will be obvious if anyone use it, so there is little need to specify it here.

@richb-hanover Any thoughts?

book/src/contributing/using-dev-container.md
eitsupi: If the explanation about Taskfile is not elsewhere, it may indeed be better to move it elsewhere.

It should be noted, however, that not all tasks can be executed in all environments.
I was imagining something like a list of tasks available in the container in this page.

.github/workflows/devcontainer.yaml
eitsupi: I was simply concerned about the frequent releases of containers.

book/src/contributing/using-dev-container.md
eitsupi: I believe that when the Rust source code changes, rust-analizer automatically builds it.

It is not devcontainer specific, so I agree that it just doesn't need to be explained here.

book/src/contributing/using-dev-container.md
eitsupi: Yes, done.

eitsupi:
I tried this with Codespace, which has the lowest specs, and had the following problems:

1. It took a few dozen minutes to build by rustanalyzer. It seemed to build all workspace members, so perhaps you may need to add a setting to limit them.
2. rusanalyzer failed to build duckdb (libduckdb-sys). It may be a lack of machine specs.
3. I could not see the preview when I ran `mdbook serve`. I do not know what the problem is.

As far as running on VS Code Dev Containers on my local computer (Ryzen 2700X, 32GB RAM), the above problem did not occur.

book/src/contributing/using-dev-container.md
richb-hanover: I say, leave out that "open in browser" comment, in the spirit of retaining only what's always going to be true, and not better explained elsewhere.

eitsupi:
I think we can merge this for now and deploy to the web branch as well.

Any further documentation improvements would include a wider range of issues (Taskfile improvements, workflow improvements using Docker, etc.)

.github/workflows/devcontainer.yaml
eitsupi: I recall that frequent builds can be a problem.
That is, even if all the layers that make up this image are the same, the digest of the image itself will be new, making any downstream image cache unavailable.
So we should not build and push the image for every push.

max-sixty:
FYI I tried testing it, but I can't seem force it to use the `amd64` image. I tried adding `  "runArgs": ["--platform=linux/amd64"]` to the `devcontainer.json` but it doesn't supply this to `docker pull`. When I pull it manually prior, I then get this error — looks like it's still relying on the platform for part of the process:

<details>
<summary>log</summary>

```log
[33 ms] Dev Containers 0.282.0 in VS Code 1.76.0 (92da9481c0904c6adfe372c12da3b7748d74bdcb).
[32 ms] Start: Resolving Remote
[42 ms] Setting up container for folder or workspace: /Users/maximilian/workspace/prql
[43 ms] Start: Check Docker is running
[43 ms] Start: Run: docker version --format {{.Server.APIVersion}}
[222 ms] Server API version: 1.41
[222 ms] Start: Run: docker volume ls -q
[321 ms] Start: Run: docker ps -q -a --filter label=vsch.local.folder=/Users/maximilian/workspace/prql --filter label=vsch.quality=stable
[460 ms] Start: Run: docker ps -q -a --filter label=devcontainer.local_folder=/Users/maximilian/workspace/prql --filter label=devcontainer.config_file=/Users/maximilian/workspace/prql/.devcontainer/devcontainer.json
[544 ms] Start: Run: docker ps -q -a --filter label=devcontainer.local_folder=/Users/maximilian/workspace/prql
[667 ms] Start: Run: docker ps -q -a --filter label=devcontainer.local_folder=/Users/maximilian/workspace/prql
[785 ms] Start: Run: /Applications/Visual Studio Code.app/Contents/Frameworks/Code Helper (Plugin).app/Contents/MacOS/Code Helper (Plugin) --ms-enable-electron-run-as-node /Users/maximilian/.vscode/extensions/ms-vscode-remote.remote-containers-0.282.0/dist/spec-node/devContainersSpecCLI.js up --user-data-folder /Users/maximilian/Library/Application Support/Code/User/globalStorage/ms-vscode-remote.remote-containers/data --container-session-data-folder /tmp/devcontainers-ed28eb90-9dda-4332-ac16-7f985020db651678145467696 --workspace-folder /Users/maximilian/workspace/prql --workspace-mount-consistency cached --id-label devcontainer.local_folder=/Users/maximilian/workspace/prql --id-label devcontainer.config_file=/Users/maximilian/workspace/prql/.devcontainer/devcontainer.json --log-level debug --log-format json --config /Users/maximilian/workspace/prql/.devcontainer/devcontainer.json --default-user-env-probe loginInteractiveShell --mount type=volume,source=vscode,target=/vscode,external=true --skip-post-create --update-remote-user-uid-default on --mount-workspace-git-root true
[965 ms] (node:57320) [DEP0005] DeprecationWarning: Buffer() is deprecated due to security and usability issues. Please use the Buffer.alloc(), Buffer.allocUnsafe(), or Buffer.from() methods instead.
[965 ms] (Use `Code Helper (Plugin) --trace-deprecation ...` to show where the warning was created)
[965 ms] @devcontainers/cli 0.31.0. Node.js v16.14.2. darwin 22.3.0 arm64.
[965 ms] Start: Run: docker buildx version
[1242 ms] github.com/docker/buildx v0.10.0 876462897612d36679153c3414f7689626251501
[1242 ms] 
[1242 ms] Start: Resolving Remote
[1257 ms] Start: Run: git rev-parse --show-cdup
[1277 ms] Start: Run: docker ps -q -a --filter label=devcontainer.local_folder=/Users/maximilian/workspace/prql --filter label=devcontainer.config_file=/Users/maximilian/workspace/prql/.devcontainer/devcontainer.json
[1383 ms] Start: Run: docker inspect --type image ghcr.io/prql/prql-devcontainer-base:latest
[1584 ms] local container features stored at: /Users/maximilian/.vscode/extensions/ms-vscode-remote.remote-containers-0.282.0/dist/node_modules/vscode-dev-containers/container-features
[1587 ms] Start: Run: tar --no-same-owner -x -f -
[1607 ms] * Processing feature: ghcr.io/devcontainers/features/hugo:1
[2422 ms] * Processing feature: ghcr.io/devcontainers/features/python:1
[2806 ms] * Processing feature: ghcr.io/devcontainers/features/node:1
[3207 ms] * Processing feature: ghcr.io/eitsupi/devcontainer-features/go-task:1
[3638 ms] * Fetching feature: hugo_1_oci
[4005 ms] * Fetching feature: python_2_oci
[4400 ms] * Fetching feature: node_3_oci
[4770 ms] * Fetching feature: go-task_4_oci
[5105 ms] Start: Run: docker buildx build --load --build-context dev_containers_feature_content_source=/var/folders/vk/btr3xc6x6f1_2cbv4h0xzp500000gn/T/devcontainercli/container-features/0.31.0-1678145469826 --build-arg _DEV_CONTAINERS_BASE_IMAGE=ghcr.io/prql/prql-devcontainer-base:latest --build-arg _DEV_CONTAINERS_IMAGE_USER=root --build-arg _DEV_CONTAINERS_FEATURE_CONTENT_SOURCE=dev_container_feature_content_temp --target dev_containers_target_stage -t vsc-prql-793278263691d95561acba5161b934dc-features -f /var/folders/vk/btr3xc6x6f1_2cbv4h0xzp500000gn/T/devcontainercli/container-features/0.31.0-1678145469826/Dockerfile.extended /Users/maximilian/Library/Application Support/Code/User/globalStorage/ms-vscode-remote.remote-containers/data/empty-folder
[+] Building 15.5s (8/8) FINISHED                                               
 => [internal] booting buildkit                                            9.2s
 => => pulling image moby/buildkit:buildx-stable-1                         6.5s
 => => creating container buildx_buildkit_sharp_raman0                     2.7s
 => [internal] load build definition from Dockerfile.extended              0.0s
 => => transferring dockerfile: 4.77kB                                     0.0s
 => [internal] load .dockerignore                                          0.0s
 => => transferring context: 2B                                            0.0s
 => resolve image config for docker.io/docker/dockerfile:1.4               2.7s
 => [auth] docker/dockerfile:pull token for registry-1.docker.io           0.0s
 => CACHED docker-image://docker.io/docker/dockerfile:1.4@sha256:9ba7531b  0.0s
 => => resolve docker.io/docker/dockerfile:1.4@sha256:9ba7531bd80fb0a8586  0.0s
 => ERROR [internal] load metadata for ghcr.io/prql/prql-devcontainer-bas  2.3s
 => [context dev_containers_feature_content_source] load .dockerignore     0.0s
 => => transferring dev_containers_feature_content_source: 2B              0.0s
------
 > [internal] load metadata for ghcr.io/prql/prql-devcontainer-base:latest:
------
Dockerfile.extended:13
--------------------
  11 |     RUN chmod -R 0700 /tmp/build-features
  12 |     
  13 | >>> FROM $_DEV_CONTAINERS_BASE_IMAGE AS dev_containers_target_stage
  14 |     
  15 |     USER root
--------------------
ERROR: failed to solve: no match for platform in manifest sha256:57593976654df7c02f8e8bd06d08f621cbc0196a6a8d19d41f5f15c5fc61ab31: not found
[20850 ms] Error: Command failed: docker buildx build --load --build-context dev_containers_feature_content_source=/var/folders/vk/btr3xc6x6f1_2cbv4h0xzp500000gn/T/devcontainercli/container-features/0.31.0-1678145469826 --build-arg _DEV_CONTAINERS_BASE_IMAGE=ghcr.io/prql/prql-devcontainer-base:latest --build-arg _DEV_CONTAINERS_IMAGE_USER=root --build-arg _DEV_CONTAINERS_FEATURE_CONTENT_SOURCE=dev_container_feature_content_temp --target dev_containers_target_stage -t vsc-prql-793278263691d95561acba5161b934dc-features -f /var/folders/vk/btr3xc6x6f1_2cbv4h0xzp500000gn/T/devcontainercli/container-features/0.31.0-1678145469826/Dockerfile.extended /Users/maximilian/Library/Application Support/Code/User/globalStorage/ms-vscode-remote.remote-containers/data/empty-folder
[20850 ms]     at cie (/Users/maximilian/.vscode/extensions/ms-vscode-remote.remote-containers-0.282.0/dist/spec-node/devContainersSpecCLI.js:1930:3293)
[20851 ms]     at S7 (/Users/maximilian/.vscode/extensions/ms-vscode-remote.remote-containers-0.282.0/dist/spec-node/devContainersSpecCLI.js:1930:3229)
[20851 ms]     at process.processTicksAndRejections (node:internal/process/task_queues:96:5)
[20851 ms]     at async xie (/Users/maximilian/.vscode/extensions/ms-vscode-remote.remote-containers-0.282.0/dist/spec-node/devContainersSpecCLI.js:1945:2625)
[20851 ms]     at async Wf (/Users/maximilian/.vscode/extensions/ms-vscode-remote.remote-containers-0.282.0/dist/spec-node/devContainersSpecCLI.js:1945:3750)
[20851 ms]     at async noe (/Users/maximilian/.vscode/extensions/ms-vscode-remote.remote-containers-0.282.0/dist/spec-node/devContainersSpecCLI.js:2069:17325)
[20851 ms]     at async roe (/Users/maximilian/.vscode/extensions/ms-vscode-remote.remote-containers-0.282.0/dist/spec-node/devContainersSpecCLI.js:2069:17066)
[20853 ms] Exit code 1
[20855 ms] Command failed: /Applications/Visual Studio Code.app/Contents/Frameworks/Code Helper (Plugin).app/Contents/MacOS/Code Helper (Plugin) --ms-enable-electron-run-as-node /Users/maximilian/.vscode/extensions/ms-vscode-remote.remote-containers-0.282.0/dist/spec-node/devContainersSpecCLI.js up --user-data-folder /Users/maximilian/Library/Application Support/Code/User/globalStorage/ms-vscode-remote.remote-containers/data --container-session-data-folder /tmp/devcontainers-ed28eb90-9dda-4332-ac16-7f985020db651678145467696 --workspace-folder /Users/maximilian/workspace/prql --workspace-mount-consistency cached --id-label devcontainer.local_folder=/Users/maximilian/workspace/prql --id-label devcontainer.config_file=/Users/maximilian/workspace/prql/.devcontainer/devcontainer.json --log-level debug --log-format json --config /Users/maximilian/workspace/prql/.devcontainer/devcontainer.json --default-user-env-probe loginInteractiveShell --mount type=volume,source=vscode,target=/vscode,external=true --skip-post-create --update-remote-user-uid-default on --mount-workspace-git-root true
[20856 ms] Exit code 1
```

</details>

One concern is that it's a) VSCode encourages you to use a Devcontainer with a pop-up and b) it's a very confusing error message, which suggests that the image doesn't exist, which makes the library look poorly maintained.

Is there any way of making that a more friendly failure? Can we disable the pop-up? Or run something before attempting to launch that checks for the arch and gives a better message?

I'm surprised this isn't more of an issue with devcontainers — is it just us? Or maybe they're not that widely adopted yet?

max-sixty:
> * It took a few dozen minutes to build by rustanalyzer. It seemed to build all workspace members, so perhaps you may need to add a setting to limit them.
> 
> * rusanalyzer failed to build duckdb (libduckdb-sys). It may be a lack of machine specs.

Yes, I think this just it running `cargo build` which does take a lot of resources, and will break on insufficient RAM because of DuckDB. I'm not sure there's much we can do there — maybe checking RAM and warning? But that seems difficult to maintain...

eitsupi:
@max-sixty Thanks for testing.

Unfortunately I don't use macOS so I don't know about that.
I have looked and could not even find a similar issue. Could you please report about it here?
https://github.com/microsoft/vscode-remote-release

The only solution I have found that may be the only solution is to set the environment variable `DOCKER_DEFAULT_PLATFORM`.
https://docs.docker.com/engine/reference/commandline/cli/#environment-variables

max-sixty:
I can report it, though my guess is that the solution is "provide an image with the correct platform".

How close do you think providing both architectures is? Is this something that's around the corner or months away? If it's the former then maybe we prioritize that?

eitsupi:
> How close do you think providing both architectures is? Is this something that's around the corner or months away? If it's the former then maybe we prioritize that?

I do not see this as a problem that can be solved quickly.

`DOCKER_DEFAULT_PLATFORM` did not work?

eitsupi:
https://medium.com/@tonistiigi/faster-multi-platform-builds-dockerfile-cross-compilation-guide-part-1-ec087c719eaf

I tried to cross-compile (e.g. `cargo install --target aarch64-unknown-linux-gnu mdbook`) in an attempt to perform the steps in this article without success.
Do you know how to cross-compile with `cargo install`?

If cross-compilation cannot be performed, building images for arm64 does not seem realistic until GHA's arm64 runner comes along.

This is the half-finished Dockerfile

```dockerfile
FROM mcr.microsoft.com/devcontainers/rust:0.203.8-1-bullseye AS base

FROM --platform=$BUILDPLATFORM base AS build

RUN apt-get update \
    && apt-get install -y \
        cmake \
        gcc-aarch64-linux-gnu \
        binutils-aarch64-linux-gnu \
    && rm -rf /var/lib/apt/lists/* \
    && sh -c "$(curl --location https://taskfile.dev/install.sh)" -- -d -b /usr/local/bin

# ========= Install cargo-tools for non-root user (vscode) =========
COPY Taskfile.cargo-tools.yml /tmp/Taskfile.cargo-tools.yml
WORKDIR /tmp
USER vscode
ARG TARGETPLATFORM
RUN if [ $TARGETPLATFORM = "linux/arm64" ]; then \
        export TARGET="aarch64-unknown-linux-gnu" \
        export CARGO_TARGET_AARCH64_UNKNOWN_LINUX_GNU_LINKER="aarch64-linux-gnu-gcc" \
    && rustup target add $TARGET \
    else \
        export TARGET="default" \
    fi \
    && cargo install --locked --target $TARGET mdbook
```

max-sixty:
> `DOCKER_DEFAULT_PLATFORM` did not work?

It does work! 

Though the selling point of this was that it was a super-easy setup "If you just want to contribute in 20 minutes, here's a setup that will get you productive in 2 minutes"; and that's less the case if it requires setting env vars before launching. It also is much slower because of the emulation.

But it's OK — there's still a case for folks who want everything packaged, and setting the env var is the one thing they need to do.

---

When I launch, I'm getting an error `/usr/local/cargo/bin/cargo: No such file or directory` — is that just me? This means that rust-analyzer returns `[ERROR rust_analyzer::lsp_utils] rust-analyzer failed to load workspace: "cargo" "--version" failed, exit status: 127`.

Can you see better output here? 
image

---

It's not that fast to start, one of the features takes 207 seconds. It's unfortunate that it's not easy to cache this. (I don't think we should try and do this ourselves; possibly VSCode will allow better caching in future..., rather than us having two devcontainer.json files.)

```
 => [dev_containers_target_stage 2/7] COPY --from=dev_containers_feature_  0.0s
 => [dev_containers_target_stage 3/7] RUN echo "_CONTAINER_USER_HOME=$(ge  0.2s
 => [dev_containers_target_stage 4/7] RUN --mount=type=bind,from=dev_cont  4.2s
 => [dev_containers_target_stage 5/7] RUN --mount=type=bind,from=dev_con  47.0s
 => [dev_containers_target_stage 6/7] RUN --mount=type=bind,from=dev_co  207.9s
 => [dev_containers_target_stage 7/7] RUN --mount=type=bind,from=dev_cont  2.4s
 ```

**Comment by max-sixty** [comment]
Re the architectures — my understanding is that enabling QEMU would allow for emulation, which will be slow, but will work. (and that's fine, there's no urgency for the building...) — https://blog.thesparktree.com/docker-multi-arch-github-actions

...but presumably this relies on VSCode having arm64 images available. Am I correct in thinking they don't have them available?

**Comment by eitsupi** [comment]
Thanks for trying this!

> Though the selling point of this was that it was a super-easy setup "If you just want to contribute in 20 minutes, here's a setup that will get you productive in 2 minutes"; and that's less the case if it requires setting env vars before launching. It also is much slower because of the emulation.

Unfortunately, Docker (Linux containers) run on Linux, so I think running them on macOS is generally regarded as not a very good experience (due to file system differences, etc.)
Also, since it only supports amd64, running it on an arm64 mac will further degrade performance.

In short, the main target users here are amd64 Linux (including WSL2) users.

> my understanding is that enabling QEMU would allow for emulation, which will be slow, but will work.

`docker buildx build` already uses QEMU.
https://github.com/PRQL/prql/blob/f04baf22a69e02bfa858f23b77926ad702d56ee8/.github/workflows/devcontainer.yaml#L36-L37

The problem here is that `cargo install` fails under too high a load for emulation by QEMU.

**Comment by eitsupi** [comment]
> When I launch, I'm getting an error `/usr/local/cargo/bin/cargo: No such file or directory` — is that just me? This means that rust-analyzer returns `[ERROR rust_analyzer::lsp_utils] rust-analyzer failed to load workspace: "cargo" "--version" failed, exit status: 127`.
> 
> Can you see better output here?

I haven't seen any such error.
Maybe it's an emulation problem...

**Comment by max-sixty** [comment]
OK great, thanks for the clear answers.

So maybe we just say it's only supported for amd64... That's a lot of WSL2 users.

File: book/src/contributing/dev-container.md [hunk 21]
```diff
@@ -0,0 +1,46 @@
+# Develop in Dev Containers
+
+```admonish note
+Currently the Dev Container included in this repository only supports the `amd64` platform.
```
max-sixty, line 4: Perfect :)

book/src/contributing/dev-container.md
```diff
@@ -0,0 +1,46 @@
...
+[VS Code Dev Containers extension](https://marketplace.visualstudio.com/items?itemName=ms-vscode-remote.remote-containers).
+
+Please refer to the documents for general instructions on how to use these
+tools.
+
+## Using PRQL in a Dev Container
+
+[Task](https://taskfile.dev/) is installed in the container for quick access to
+tasks defined on the `Taskfiles.yml`. Autocompression works when using the `zsh`
+shell installed in the container.
```
max-sixty: What is autocompression?

book/src/contributing/dev-container.md
```diff
@@ -0,0 +1,46 @@
...
+- `task -l` lists all the available tasks.
+- `task run-book` starts an `mdbook` server. As you edit the files of the
+  Language Book (in the `book` directory), `mdbook` rebuilds those pages.
+  (Port 3000)
+- `task run-website` starts a `hugo` server. As you edit the files (in the
+  `website` directory), `hugo` rebuilds those pages. (Port 1313)
+- `task run-playground` starts a Node server to build the Playground. As you
+  edit the files (in the `playground` directory), the server rebuilds those
+  pages. (Port 3000)
+- `task WHAT ELSE?` _Provide explanation of other useful commands._
```
max-sixty: I would move somewhere else, maybe in a separate PR if @richb-hanover wants to do that. These would be excellent summaries in the Taskfile, and we could just point folks to `task --list` & `task --summary`. 

I was going to suggest https://prql-lang.org/book/contributing/development.html#components-of-prql — but this is out of date now. (which further increases my resolve to ensure we only take docs that we can keep up to date. I'll open an issue for this.)

max-sixty (approved):

book/src/contributing/dev-container.md
eitsupi: Sorry, it's a typo...... "autocompletion"

The available task suggestions are displayed when the Tab key is pressed.
https://taskfile.dev/installation/#setup-completions

max-sixty: Oh OK, yes that's great. But I wouldn't include that in the PRQL docs...

eitsupi:
Perhaps another devcontainer.json could be added to build from the Dockerfile locally to support arm64 machines.
(VS Code seems to have recently added support for multiple devcontainer.json)

I can do that in another PR in the next time.

book/src/contributing/dev-container.md
eitsupi:
```suggestion
tasks defined on the `Taskfiles.yml`.
```

eitsupi:
Thank you for taking the time to review this.

max-sixty:
Thank you very much @eitsupi !

Maybe we could also add to the Changelog so folks learn about it?

eitsupi:
> Maybe we could also add to the Changelog so folks learn about it?

Thanks, I forgot about that. I will update Changelog too with a follow-up PR.

max-sixty:
> Perhaps another devcontainer.json could be added to build from the Dockerfile locally to support arm64 machines.

I think possibly we could see how people find the amd64 version, and we could see whether MS make it easier to build arm64 images — my guess is that it'll become much easier...

eitsupi:
I believe the GHA arm64 runner preview was set for the 2023Q3 milestone.

Conversely, we will not be able to provide pre-built images for arm64 for at least half a year.

max-sixty:
@eitsupi FYI I swapped the labels

.github/workflows/devcontainer.yaml
max-sixty: Sorry, I missed this.

If we're concerned about that, we could only push a new image in `release.yaml`.
Or even in `publish-web.yaml`, which would give us the ability to change it faster (we had planned to call that branch `stable`, which would be a more suitable name now...)

But if the files are the same the hash will be the same. But you're correct that the hash can change for a very small reason.

I'll add a TODO to the code — no rush to do this, but if folks reference the workflows it's more cohesive if they follow a similar pattern or have a note where they don't.

eitsupi:
> @eitsupi FYI I swapped the labels

Thank you for correcting the label. (But seems to have failed to deploy? No problem to carry over until next release)

max-sixty:
> (But seems to have failed to deploy? No problem to carry over until next release)

Ah, sorry, I got the label wrong too! We should have done backport.

Yeah it'll go out later anyway!

.github/workflows/devcontainer.yaml
eitsupi: That makes sense.
It seems reasonable for me to trigger it when releasing, so I will make that change.

PR #1894, opened 2023-02-19
fix: website tests

aljazerzen:


max-sixty:
FYI not backporting this since it's relevant for the latest code only

PR #1896, opened 2023-02-19
docs: Fix "LSP server"

richb-hanover:


max-sixty:
Thanks!

vanillajonathan:
This _could_ link to https://microsoft.github.io/language-server-protocol/

This _could_ use the HTML abbr element.
```html
an <abbr title="Language Server Protocol">LSP</abbr> server could greatly
```

max-sixty:
Feel free to adjust! (I'm not sure whether Hugo allows for inline HTML, we'd have to try it, I think it probably does)

PR #1642, opened 2023-01-29
feat!: loop

aljazerzen:
Work in progress.

Ref #407

PR #1899, opened 2023-02-20
revert: "feat!: loop"

max-sixty:
Reverts PRQL/prql#1642; book is failing, links are failing.

I can try and fix later. And also assess why the book test doesn't fail


max-sixty:
This is why the book tests didn't fail, unfortunately: https://github.com/rust-lang/mdBook/issues/1094

PR #1905, opened 2023-02-20
refactor: Options as &Options

max-sixty:
Not sure if this is worthwhile — I started in one function and then kept on replacing. Fine to close if not an improvement.


aljazerzen (approved):
It's a marginal improvement I think.

PR #1913, opened 2023-02-20
internal: Enable auto-conversion of anyhow's error to `ErrorMessages`

max-sixty:


aljazerzen (approved):
If you search for .map_err, there is a lot of other places where this can be cleaned up.

PR #1915, opened 2023-02-21
refactor: Attempt to replace our `IntoOnly` with `ExactlyOne`

max-sixty:
Since I originally wrote this (and others have iterated on it), Itertools released `ExactlyOne`, which has better errors and reduces our custom code.

Unfortunately, I couldn't fix a rust type error, and spent too long on it already. So pushing what I have in case anyone wants to take a look.

There's also a decent chance that we replace the `parser.rs` code, in which case this type error becomes moot, and we can merge this anyway.

### DIFF
--- a/prql-python/src/lib.rs
+++ b/prql-python/src/lib.rs
@@ -1,7 +1,7 @@
 #![cfg(not(target_family = "wasm"))]
 use std::str::FromStr;
 
-use prql_compiler::{self, IntoOnly, Target};
+use prql_compiler::{self, Target};
 use pyo3::{exceptions, prelude::*};
 
 #[pyfunction]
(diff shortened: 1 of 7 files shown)


prql-python/src/lib.rs
```diff
@@ -11,7 +11,7 @@ pub fn compile(prql_query: &str, options: Option<CompileOptions>) -> PyResult<St
         .and_then(prql_compiler::pl_to_rq)
         .and_then(|rq| prql_compiler::rq_to_sql(rq, &options.map(|o| o.into()).unwrap_or_default()))
         .map_err(|e| e.composed("", prql_query, false))
-        .map_err(|e| (PyErr::new::<exceptions::PySyntaxError, _>(e.into_only().unwrap().reason)))
+        .map_err(|e| (PyErr::new::<exceptions::PySyntaxError, _>(e.to_string())))
```
max-sixty, line 14: @aljazerzen does this look reasonable to you?

With the existing code, I think we would have panicked if there had been more than one `ErrorMessage`. Is just putting the whole thing into a string fine?

aljazerzen: Yes, looks good.

This is a change in API - no formally since we still return a string, but it's much larger formatted string now.

In any case, we should return the whole error here, maybe serialized to JSON as we do for js.

max-sixty:
The error message is actually much better between `0.5.2` and now:

Existing:

```python
❯ python x.py
Traceback (most recent call last):
  File "/Users/maximilian/workspace/prql/x.py", line 4, in <module>
    compile('AOU')
SyntaxError: Unknown name `AOU`
```

Main:

```python
❯ python x.py
Traceback (most recent call last):
  File "/Users/maximilian/workspace/prql/x.py", line 4, in <module>
    compile('AOU')
SyntaxError: Error:
   ╭─[:1:1]
   │
 1 │ AOU
   · ─┬─
   ·  ╰─── Unknown name `AOU`
───╯
```

I'm actually not sure what we'd add — I'd started an issue re the JSON, but potentially this is as good an output as we'd hope for?

PR #1917, opened 2023-02-21
feat: Add .NET bindings

### LINKED ISSUE #1845: .NET bindings
PRQL should have bindings for .NET. I found some docs at: https://learn.microsoft.com/en-us/dotnet/standard/native-interop/cross-platform

vanillajonathan:
Closes #1845.


.github/workflows/test-dotnet.yaml
```diff
@@ -0,0 +1,36 @@
...
+
+jobs:
+  test:
+    runs-on: ubuntu-latest
+    steps:
+      - name: 📂 Checkout code
+        uses: actions/checkout@v3
+      - uses: actions-rs/toolchain@v1
+        with:
+          toolchain: stable
```
max-sixty:
```suggestion
```

prql-net/PrqlCompiler/PrqlCompiler.csproj
```diff
@@ -0,0 +1,28 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>netstandard2.0</TargetFramework>
+  </PropertyGroup>
+
+  <PropertyGroup>
+    <PackageId>Prql.Compiler</PackageId>
+    <PackageVersion>0.1.0</PackageVersion>
```
max-sixty: Should we align this with our standard version?

We can use the replacements in `Cargo.toml` so it updates on every release.

prql-net/PrqlCompiler/PrqlCompiler.csproj
```diff
@@ -0,0 +1,28 @@
...
+    <Title>PRQL Query Compiler</Title>
+    <Authors>PRQL</Authors>
+    <Copyright>Copyright © The PRQL Project 2023</Copyright>
+    <Description>PRQL compiler bindings for .NET which lets you transpile PRQL queries to SQL queries.</Description>
```
max-sixty:
```suggestion
    <Description>.NET bindings for the PRQL compiler. PRQL is a modern language for transforming data</Description>
```

prql-net/PrqlCompiler/PrqlCompiler.csproj
```diff
@@ -0,0 +1,28 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>netstandard2.0</TargetFramework>
+  </PropertyGroup>
+
+  <PropertyGroup>
+    <PackageId>Prql.Compiler</PackageId>
+    <PackageVersion>0.1.0</PackageVersion>
+    <Title>PRQL Query Compiler</Title>
```
max-sixty:
```suggestion
    <Title>PRQL Compiler</Title>
```

prql-net/prql-net.sln
```diff
@@ -0,0 +1,28 @@
+﻿
+Microsoft Visual Studio Solution File, Format Version 12.00
+# Visual Studio Version 17
+VisualStudioVersion = 17.0.31903.59
+MinimumVisualStudioVersion = 10.0.40219.1
+Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "PrqlCompiler", "PrqlCompiler\PrqlCompiler.csproj", "{339EA2A6-23D2-4938-884F-052431AC0674}"
```
max-sixty: Can I ask where this comes from?

Does it need to be updated in the future? Would it be possible to write a couple of sentences on this in a comment or the Readme?

prql-net/README.md
```diff
@@ -0,0 +1,27 @@
+# prql-net
```
max-sixty: My sense is we should rename this to its full `dotnet` name. I've often seen `dotnet`, and I haven't seen `-net` used as much — is that accurate?

```suggestion
# prql-dotnet
```

Hopefully this will be a single find/replace and won't be too much work...

max-sixty (commented):
Another good one! Thanks @vanillajonathan !

prql-net/PrqlCompiler/PrqlCompiler.csproj
vanillajonathan: I don't know. Maybe. We could that. This is a early pre-release though, I would really like the prql-lib to be updated to expose the `compile` function if possible so I don't have to use the `to_sql` function which does not accept PRQL compiler options.

prql-net/README.md
vanillajonathan: I don't know, the framework is called ".NET", but using a dot isn't always possible, so sometimes "dotnet" is used, like in the `dotnet` command.

prql-net/prql-net.sln
vanillajonathan: No, it doesn't need to be updated. That gets auto-generated when you create a new solution (`.sln`) file using `dotnet new sln` and add the projects to the solution using `dotnet add reference PrqlCompiler\PrqlCompiler.csproj`.

prql-net/README.md
max-sixty: Yes, that's why I'm suggesting we rename this everywhere to `prql-dotnet` — including the folder name. Is that OK?

prql-net/PrqlCompiler/PrqlCompiler.csproj
max-sixty: OK, if we don't do that now, can we add a TODO saying that in the code?

prql-net/README.md
vanillajonathan: Yeah, if you prefer it that way, sure. Maybe it is more clear what it is, so it doesn't get confused with something network related.

prql-net/PrqlCompiler/PrqlCompiler.csproj
vanillajonathan: We could do that now if you want. To say what in the code?

But users would downloaded the .NET project wouldn't be able to build it without also having cargo installed, because `dotnet` refuses to build it if its not a valid string.

> /usr/lib/dotnet/sdk/7.0.102/NuGet.targets(132,5): error : 'abc' is not a valid version string. (Parameter 'value')

prql-net/prql-net.sln
vanillajonathan: Technically the solution file isn't really needed, its just a convenient thing so that you if you use Visual Studio then you can open the solution file which holds both the compiler project and the unit test project. It is possible for `dotnet` to compile the '.csproj` project files, so we could remove the solution file if you don't like it.

prql-net/PrqlCompiler/PrqlCompiler.csproj
max-sixty: To say something like `# TODO: this is currently at 0.1.0 because we're waiting to update prql-lib for the latest API. When we've done that, we can match the version here with the broader PRQL version`.

Generally we do this so others know what state the project is in, and what the next steps are to make progress. Does that make sense? (Not trying to slow you down, tell me if this is a burden and we can skip it...)

max-sixty:
Looks great! I think this is ready to go! Would be nice to add a TODO about what we'll do next, but either way we can merge.

@vanillajonathan if you'd like to add an entry to the Changelog, we can do that in this PR if you want.

Thank you!

prql-net/PrqlCompiler/PrqlCompiler.csproj
vanillajonathan: Alright, I added a TODO to the README.

Yeah, that is fine.

max-sixty:
Thanks again @vanillajonathan !

PR #1918, opened 2023-02-21
build: Package prqlc as .rpm package

vanillajonathan:
This uses the [build-rpm-action](https://github.com/marketplace/actions/build-rpm-action) GitHub Action to create a `.rpm` package for Fedora and Fedora-based Linux distributions such as AlmaLinux, CentOS and RHEL.

To install download and run:

    sudo rpm -i  prqlc-0.5.3.rpm


.github/workflows/release.yaml
```diff
@@ -24,6 +24,32 @@ jobs:
...
+          chmod +x .rpmpkg/usr/bin/prqlc
+      - name: 📦 Build .rpm package
+        uses: jiro4989/build-rpm-action@v2
+        with:
+          summary: PRQL Compiler
+          package: prqlc
+          package_root: .rpmpkg
+          maintainer: The PRQL Project
+          version: ${{ github.ref }}
+          desc: PRQL is a modern language for transforming data — a simple, powerful, pipelined SQL replacement
```
max-sixty: We could use the latest one I just pushed...

.github/workflows/release.yaml
```diff
@@ -24,6 +24,32 @@ jobs:
...
             })
 
+  build-rom-package:
+    runs-on: ubuntu-latest
+    steps:
+      - name: 📂 Checkout code
+        uses: actions/checkout@v3
+      - uses: actions-rs/toolchain@v1
+        with:
+          toolchain: stable
```
```suggestion
```

No need for these

.github/workflows/release.yaml
```diff
@@ -24,6 +24,32 @@ jobs:
...
+        run: cargo build --release
+      - name: Copy files into .rpm package
+        run: |
+          mkdir -p .rpmpkg/usr/bin
+          cp target/release/prqlc .rpmpkg/usr/bin/prqlc
+          chmod +x .rpmpkg/usr/bin/prqlc
+      - name: 📦 Build .rpm package
+        uses: jiro4989/build-rpm-action@v2
+        with:
+          summary: PRQL Compiler
```
```suggestion
          summary: CLI for PRQL, a modern language for transforming data
```

.github/workflows/release.yaml
```diff
@@ -24,6 +24,32 @@ jobs:
...
+          package: prqlc
+          package_root: .rpmpkg
+          maintainer: The PRQL Project
+          version: ${{ github.ref }}
```
max-sixty, line 71:
```suggestion
          version: ${{ github.ref_name }}
```

Unless I'm misunderstanding something (I might be) — this gives the actual version number

.github/workflows/release.yaml
```diff
@@ -24,6 +24,32 @@ jobs:
...
+  build-rom-package:
+    runs-on: ubuntu-latest
+    steps:
+      - name: 📂 Checkout code
+        uses: actions/checkout@v3
+      - uses: actions-rs/toolchain@v1
+        with:
+          toolchain: stable
+      - name: 👷 Build prql-compiler
+        run: cargo build --release
```
```suggestion
      - name: 👷 Build prqlc
        run: cargo build --bin prqlc --release
```

Slightly more efficient I think

.github/workflows/release.yaml
vanillajonathan: Yeah, but `build-rpm-action` will strip the `refs/tags/` from it anyway, so it doesn't matter. The example at https://github.com/marketplace/actions/build-rpm-action uses `github.ref`.

max-sixty: OK. I think the reason why they use `github.ref` is because that is older than `refname`, but you're right that they use it

vanillajonathan: Do you want me to use `github.ref` or `github.refname`? Feel free to edit as you wish.

.github/workflows/release.yaml
```diff
@@ -44,10 +44,36 @@ jobs:
           package: prqlc
           package_root: .debpkg
           maintainer: The PRQL Project
-          version: "${{ github.ref }}"
+          version: ${{ github.ref }}
           desc:
             prqlc is the CLI for PRQL, a modern language for transforming data
 
+  build-rpm-package:
```
max-sixty, line 51:
```suggestion
  build-rpm-package:
  # TODO: This doesn't publish the rpm yet, that would be a welcome follow-up (even as a CI artifact)
```

.github/workflows/release.yaml
max-sixty: My guess is that `refname` is more direct, but I'm not 100% sure. If `ref` works then it's OK!

max-sixty:
Added a comment and merging! Thank you @vanillajonathan !

PR #1920, opened 2023-02-21
chore: Add changelog for #1883

max-sixty:
As requested by @vanillajonathan.

@vanillajonathan one thing we could add for all these is to upload them as artifacts, as an easy way to have them published, using something like https://github.com/PRQL/prql/pull/1883#discussion_r1112342801.

It's less permanent than publishing them as release assets, but we can do the artifact upload now without changing our release process.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,8 @@
 
 - [prql-php] Added PHP bindings (@vanillajonathan, #1860)
 - [prql-lib] Added C header file (@vanillajonathan, #1879)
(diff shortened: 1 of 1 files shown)


vanillajonathan:
Yeah, but I have never used any artifacts on GitHub, so I don't know how that work.

max-sixty:
OK! You seem to be good at doing varied things! But no stress — your contributions so far are greatly appreciated, and it would be easy for someone to build on them, for example with the artifacts.

Issue #1923, opened 2023-02-22
Non-zero exit value is preferred when prqlc failed to compile query


zhicwu:
Hi, below case should return `failed` IMO. Is there any specific reason of returning zero?

```bash
echo "fr0m t" | ./target/debug/prqlc compile && echo succeeded || echo failed
Error:
   ╭─[:1:1]
   │
 1 │ fr0m t
   · ──┬─
   ·   ╰─── Unknown name fr0m
───╯
succeeded
```

max-sixty:
Definitely a bug!

A fix to this would be welcome...

max-sixty:
I'm taking a look at this

PR #1924, opened 2023-02-22
fix: Return a non-zero exit code for `prqlc compile` errors

max-sixty:


max-sixty:
Closes https://github.com/PRQL/prql/issues/1923

PR #1928, opened 2023-02-22
fix: Panic with multiple terms after a `from` transform

### LINKED ISSUE #1785: put the second word after `from` cause panics at bad transform cast
I noticed that compiling the following queries would cause panics.

```elm
from a select b
```

```elm
from a b
```

max-sixty:
Closes #1785


aljazerzen:
This error should be caught earlier. Calling any function with too much arguments should be an error.

max-sixty:
@aljazerzen I think this is better, and simpler...

aljazerzen (approved):
Nice, exactly how I'd do it

Issue #1779, opened 2023-02-11
Playground default SQL vs Arrow


max-sixty:
We discussed whether the default output should be SQL or Arrow, and making the RQ / PL views nicer.

One thing we should change now (happy to work it out how unless anyone knows and is up for making the change) is that currently if you go to the playground on a new browser and enter `from x`, you get an error, and it's quite unclear what's going on.

Clicking `output.sql` makes it work again, but that's over on the right.

I guess none of us realized because the playground retains state, and we had a different default page.

<img width="287" alt="image" src="https://user-images.githubusercontent.com/5635139/218231948-d221d846-9fd9-4bc4-8512-1ffd474cb46e.png">

vanillajonathan:
I think the Arrow view is nice, but I don't know what Arrow, RQ and PL means.

Issue #1728, opened 2023-02-08
Enhancements to Playground for Newcomers


richb-hanover:
As I try new queries in the Playground, I stumble on things... A few thoughts:

- I wish there was better feedback for the "no pipeline" state than the red squiggly underline. It feels disconcerting that an empty page already has an "error". As a newcomer, I have the mental state, 'Uh oh. What am I supposed to type...?" And this state of anxiety remains if I type a comment (starting with `#`) or either of the two most common legal keywords (next point).

   Perhaps the left-hand pane should report the syntactic state of the PRQL query being built, while the right-hand pane indicates the "no pipeline" state. That way, a blank page (or several lines of comments) would indicate: "No SQL generated" on the right.

- I wish there was better feedback when starting a valid query. Starting with the empty query and typing the following gives these errors:
	- `let` - "Unknown name let"
	- `from` - "expected pipeline that resolves to a table..."

- Names of output tabs: Maybe the first tab could be "PRQL Output" (or "SQL Output") and the second could be "Chinook Output"? And the third and fourth be could be "PL Output" and "RQ Output"?

- I like how Playground defaults to showing a real PRQL query and the results from the chinook database. And how typing additional PRQL commands affects the results. I also like how clearing out the PRQL query blanks the right-hand pane. 

    BUT...  when I type a valid PRQL (say, `from foo`), the right-hand pane stays blank. I think the gesture of clearing out the PRQL pane should switch to the PRQL Output tab.

Thanks!

richb-hanover:
Further incentive to implement the fourth item above (clearing query pane should switch to SQL output tab)... 

I was trying out one of the new Showcase queries. In the Playground, I cleared the query pane and pasted in the Showcase query. I was mystified that it caused an an error message. It "looked broken" to me until I remembered that I had to click the SQL Output tab.

max-sixty:
Yes, that's https://github.com/PRQL/prql/issues/1779

PR #1939, opened 2023-02-23
web: Attempt to default to the SQL view in playground

max-sixty:
At attempt to solve https://github.com/PRQL/prql/issues/1779

This doesn't seem to work though — it still defaults to the arrow tab. I tried opening in Incognito but that doesn't help clear the state. 

@aljazerzen I know it's not fun if you build something cool, and then people nag you for support — but do you have any ideas of where the default gets set?

### DIFF
--- a/playground/src/examples.js
+++ b/playground/src/examples.js
@@ -1,6 +1,6 @@
 const examples = {
   "introduction.prql": [
-    "arrow",
+    "sql",
     `from invoices
 filter invoice_date >= @1970-01-16
 derive [                        # This adds columns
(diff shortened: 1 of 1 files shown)


aljazerzen:
They are set per-file:

https://github.com/PRQL/prql/blob/48b16e82a737d344ae929c1a15aa2b6ed0725df3/playground/src/workbench/Workbench.js#L33-L38

So for examples, you should change their definitions:

https://github.com/PRQL/prql/blob/48b16e82a737d344ae929c1a15aa2b6ed0725df3/playground/src/examples.js#L2-L4

max-sixty:
Great — now switched!

richb-hanover:
This looks great, and solves the final bullet point of https://github.com/PRQL/prql/issues/1728 

That is, I clear out the PRQL pane and start typing a new query, and the SQL side updates. Thanks!

PR #1941, opened 2023-02-23
build: update prql-lib API

aljazerzen:


vanillajonathan:
Could you add some information about what PL and RQ are?

PR #1942, opened 2023-02-23
feat: Add compile function to library

### LINKED ISSUE #1859: Update prql-lib, dotnet & php libraries
The `prql-lib` library does not expose the [`compile`](https://docs.rs/prql-compiler/latest/prql_compiler/fn.compile.html) method.

vanillajonathan:
I need someone to review this.

Should the struct be named `CompileOptions` or `CompilerOptions`?
Should the struct be passed as `options` or `&options`?

I named it `CompileOptions` because that is what it is named in the Python binding.
The `compile` C library function returns a string (well a `*const c_char`), just like the in the Python binding, however contrary to the Rust code itself which returns a  `Result<T>` enum containing the success result or an [ErrorMessage](https://docs.rs/prql-compiler/latest/prql_compiler/struct.ErrorMessage.html) struct. On error it returns a empty string.

You can test it with the following program:
```c
#include <stdio.h>

#include <libprql_lib.h>

int main() {
    char *prql_query;
    prql_query = "from albums | select [album_id, title] | take 3";

    struct CompileOptions options;
    options.format = true;
    options.signature_comment = true;
    options.target = "haha";

    const char *sql_query = compile(prql_query, options);

    printf("%s", sql_query);
    return 0;
}
```

Closes #1859

### DIFF
--- a/prql-lib/libprql_lib.h
+++ b/prql-lib/libprql_lib.h
@@ -3,6 +3,33 @@
 #include <stdint.h>
 #include <stdlib.h>
 
+typedef struct CompileOptions {
+  /**
+   * Pass generated SQL string trough a formatter that splits it
+   * into multiple lines and prettifies indentation and spacing.
+   *
+   * Defaults to true.
+   */
+  bool format;
+  /**
+   * Target and dialect to compile to.
+   */
+  const char *target;
+  /**
+   * Emits the compiler signature as a comment after generated SQL
+   *
+   * Defaults to true.
+   */
+  bool signature_comment;
+} CompileOptions;
+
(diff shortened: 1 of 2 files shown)


aljazerzen:
This looks good. Can I incorporate this PR in my update to the prql-lib?

vanillajonathan:
Yes, please do!

max-sixty:
Great, thanks @vanillajonathan !

max-sixty:
> This looks good. Can I incorporate this PR in my update to the prql-lib?

FYI one way of doing that is to change the base of this PR to your branch

PR #1946, opened 2023-02-23
build: Fix .NET bindings

vanillajonathan:
This fixes the .NET bindings that #1941 broke.


max-sixty:
@vanillajonathan thank you very much for such a speedy fix!

PR #1949, opened 2023-02-23
refactor: Improve PHP bindings

vanillajonathan:
This PR restructures the PHP code to avoid duplicate code, adds a new unit test, adds support for finding the libprql library on macOS, adds support for specifying the path to the library.

Unfortunately, I haven't gotten the `compile` method to work, so that one is still broken.


prql-php/src/Compiler.php
```diff
@@ -53,108 +92,94 @@ function compile(string $prql_query, ?Options $options = null): string
...
+     * @param string $pl_json PL in JSON format.
      *
-     * @return string SQL query.
+     * @return string RQ string.
      * @throws \InvalidArgumentException If no query is given or the query cannot
      * be compiled.
      * @api
      */
-    function toSql(string $prql_query): string
+    function pLToRQ(string $pl_json): string
```
max-sixty: What's the convention in Java for upper case words?

In rust, the convention is to only ever have the first character of each word be upper case. So `trait Sql`, not `trait SQL`. That would be `plToRq` I guess?

(only matters on the external API, no need to change everything tbc)

max-sixty (approved):
Super @vanillajonathan !

I left one question.

prql-php/src/Compiler.php
vanillajonathan: PHP uses camel case for method names and camel case is the recommendation by [PHP-FIG](https://www.php-fig.org/) and the configuration for [PHP_CodeSniffer](https://github.com/squizlabs/PHP_CodeSniffer) linter.

I don't know about acronyms, there aren't so many in PHP, but acronyms seem to be all uppercase in class names such as `SQLite3` and lowercase in method names such as [`Swoole\Async::dnsLookup`](https://www.php.net/manual/en/swoole-async.dnslookup) but that name starts with a lower case letter. PHP_CodeSniffer doesn't complain on method names that repeat capital letters.

I could leave `toSql` as is, or rename it to `toSQL`.
As for `pLToRQ` that one is trickier, but I should probably rename it to `plToRQ`maybe?

max-sixty: Thanks for the links!

Yes, it should either be `plToRQ` or `plToRq`. My slight preference from Googling would be the latter — e.g. [here](https://stackoverflow.com/a/15526526/3064736), but totally fine with the former.

vanillajonathan: Microsoft's recommendation for .NET is that acronyms with two letters (such as IO and UI) should have both letters uppercase, while acronyms longer two letters (such as TCP or HTTP) should be lowercase. So in the .NET world Microsoft would favor the former not the latter.

max-sixty: I had thought that's for `foo.IO`, but not for `foo.IOTransport`, but you're completely right it's ambiguous, honestly either is fine, as long as we make the `pL` -> `pl`

vanillajonathan: Done!

max-sixty:
Thank you @vanillajonathan !

PR #1952, opened 2023-02-24
docs: Add a note re reverting

max-sixty:


max-sixty:
As discussed with @aljazerzen

PR #1962, opened 2023-02-25
feat: Add a `--format=yaml` option to `prqlc parse`

### LINKED ISSUE #1912: Add a `--format` option to prqlc to produce yaml
Currently `prqlc` only returns JSON:

```
❯ echo "from x" | prqlc resolve
{
  "def": {
    "version": null,
    "other": {}
  },
  "tables": [
    {
...
```

In the playground, we produce YAML, which is arguably more friendly for people (no object-level discussion of whether this claim is accurate please!).

We could allow a `--format` option to `prqlc` to produce this output. Would be a nice contribution, reasonably well modularized while also introducing the main rust code.

max-sixty:
Would close #1912 

Unfortunately it doesn't work on producing RQ, because `serde_yaml` doesn't seem to support its AST. I added a link to the relevant `serde_yaml` code inline.

RQ serialization works in the playground because we serialize to YAML in JS.

---

I think this might need some refactoring before merging — it currently adds a `--format` option to every command, but actually only works on one. This might mean we need to make an Args struct for each command (the Args struct is currently named `CommandIO`).

### DIFF
--- a/prql-compiler/prqlc/src/cli.rs
+++ b/prql-compiler/prqlc/src/cli.rs
@@ -14,10 +14,11 @@ use prql_compiler::{downcast, Options};
 
 use crate::watch;
 
+/// Entrypoint called by [main::main]
 pub fn main() -> color_eyre::eyre::Result<()> {
     env_logger::builder().format_timestamp(None).init();
     color_eyre::install()?;
-    let mut cli = Cli::parse();
+    let mut cli = dbg!(Cli::parse());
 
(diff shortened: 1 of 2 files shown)


max-sixty:
I based #1963 off this. Ideally it would have been the other way — refactor and then add the feature — but inevitably it's only possible to see what should be refactored after the feature :)

aljazerzen (approved):

PR #1963, opened 2023-02-26
refactor: Refactor CLI arg handling

max-sixty:
Merge #1962 first

### DIFF
--- a/prql-compiler/prqlc/src/cli.rs
+++ b/prql-compiler/prqlc/src/cli.rs
@@ -158,8 +158,9 @@ impl Command {
                 combine_prql_and_frames(source, frames).as_bytes().to_vec()
             }
             Command::Resolve(_) => {
-                // Currently failing based on serde_yaml not being able to serialize an
-                // Enum of an Enum; from https://github.com/dtolnay/serde-yaml/blob/68a9e95c9fd639498c85f55b5485f446b3f8465c/tests/test_error.rs#L175
+                // We can't currently have `--format=yaml` here, because
+                //  serde_yaml is unable to serialize an Enum of an Enum; from
+                // https://github.com/dtolnay/serde-yaml/blob/68a9e95c9fd639498c85f55b5485f446b3f8465c/tests/test_error.rs#L175
                 let ast = prql_to_pl(source)?;
                 let ir = semantic::resolve(ast)?;
                 serde_json::to_string_pretty(&ir)?.into_bytes()
@@ -172,6 +173,9 @@ impl Command {
     }
 
     fn read_input(&mut self) -> Result<(String, String)> {
+        // TODO: possibly this should be called by the relevant subcommands
+        // passing in `input`, rather than matching on them and grabbing `input`
+        // from `self`.
         use Command::*;
         let mut input = match self {
             Parse { input, .. } => input.clone(),
--- a/prql-compiler/prqlc/src/cli.rs
+++ b/prql-compiler/prqlc/src/cli.rs
@@ -31,12 +31,12 @@ pub fn main() -> color_eyre::eyre::Result<()> {
 #[derive(Parser, Debug, Clone)]
 struct Cli {
     #[command(subcommand)]
-    command: Commands,
+    command: Command,
 }
 
 #[derive(Subcommand, Debug, Clone)]
 #[clap(name = env!("CARGO_PKG_NAME"), about, version)]
-enum Commands {
+enum Command {
     /// Parse into PL AST
     Parse {
         #[clap(value_parser, default_value = "-")]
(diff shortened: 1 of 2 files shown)


prql-compiler/prqlc/src/cli.rs
```diff
@@ -14,61 +14,84 @@ use prql_compiler::{downcast, Options};
...
 
     /// Watch a directory and compile .prql files to .sql files
-    Watch(watch::WatchCommand),
+    Watch(watch::WatchArgs),
 }
 
-#[derive(clap::Args, Default)]
-pub struct CommandIO {
+// TODO: Should this be named `IoArgs`? IIUC it's just the args; its parent
+// represents the Command. I struggled mapping this to clap docs for a while.
```
aljazerzen:
```suggestion
```

prql-compiler/prqlc/src/cli.rs
```diff
@@ -14,61 +14,84 @@ use prql_compiler::{downcast, Options};
...
+    #[command(subcommand)]
+    command: Commands,
+}
+
+#[derive(Subcommand, Debug, Clone)]
 #[clap(name = env!("CARGO_PKG_NAME"), about, version)]
-pub enum Cli {
-    /// Parse PL AST
-    Parse(CommandIO),
+enum Commands {
```
aljazerzen:
```suggestion
enum Command {
```

prql-compiler/prqlc/src/cli.rs
```diff
@@ -128,46 +154,55 @@ impl Cli {
...
-
-                let mut source = String::new();
-                io.input.read_to_string(&mut source)?;
-                let source_id = (*io.input.path()).to_str().unwrap().to_string();
-                Ok((source, source_id))
-            }
-            Cli::Watch(_) => unreachable!(),
+        use Commands::*;
+        let mut input = match self {
+            // TODO: are these clones sound?
```
aljazerzen: That depends on what the Input is. But they are probably not necessarily, you should be able to do an exclusive borrow here.

prql-compiler/prqlc/src/cli.rs
```diff
@@ -128,46 +154,55 @@ impl Cli {
...
     }
 
     fn write_output(&mut self, data: &[u8]) -> std::io::Result<()> {
-        use Cli::*;
-        match self {
-            Parse(io) | Format(io) | Debug(io) | Annotate(io) | Resolve(io) | Compile(io) => {
-                io.output.write_all(data)
+        use Commands::*;
+        let mut output = match self {
+            Parse { output, .. } => output.to_owned(),
```
aljazerzen, line 202: Again, this clone is not needed, I think.

prql-compiler/prqlc/src/cli.rs
```diff
@@ -205,12 +240,28 @@ fn combine_prql_and_frames(source: &str, frames: Vec<(Span, Frame)>) -> String {
...
+    // Usage: prqlc parse [OPTIONS] [INPUT] [OUTPUT]
+    //
+    // Arguments:
+    //   [INPUT]   [default: -]
+    //   [OUTPUT]  [default: -]
+    //
+    // Options:
+    //       --format <FORMAT>  [possible values: json, yaml]
+    //   -h, --help             Print help
+
```
aljazerzen, line 261: Yes it would - we'd have this output in source code,  so it could be compared between versions (there was a change when clap was updated and we didn't even notice I think).

aljazerzen (approved):
Yup, way better.

prql-compiler/prqlc/src/cli.rs
aljazerzen: https://docs.rs/clio/0.2.7/clio/enum.Input.html#impl-Clone-for-Input

It looks like it is sound. It should not have any performance implications.

prql-compiler/prqlc/src/cli.rs
max-sixty: I re-up-ed https://github.com/PRQL/prql/issues/177

prql-compiler/prqlc/src/cli.rs
max-sixty: Yeah, in the narrow "does it compile without the clones" — it doesn't. But I was a bit confused by this — in the old approach they weren't needed. 

I think this is one of those times that the compiler was being helpful, we adjust the flow of the code, the compiler can no longer see that it can still be helpful, and so the result is somewhat confusing...

max-sixty: I left a TODO about inverting how this function is called; I think that is likely to make the code better and avoid this problem...

PR #1964, opened 2023-02-26
docs: RFC for type system

aljazerzen:


book/src/internals/type-system.md
```diff
@@ -0,0 +1,179 @@
...
+Using Python, JavaScript, C++ or Rust, one could define many different a data
+structure that would correspond to our idea of "relation". Most of them would be
+an object/struct that has column names and types and then a generic array of
+arrays for rows.
+
+PRQL's type system should also be able to express relation as composed from
+primitive types, but have only one idiomatic way of doing so.
+
+In practice this means that builtin types include only primitives (int, text,
+bool), struct (for product), enum (for sum) and list (for repeating).
```
max-sixty:
```suggestion
bool, float), struct (for product), enum (for sum) and list (for repeating).
```

book/src/internals/type-system.md
```diff
@@ -0,0 +1,179 @@
...
+
+An SQL row would translate to struct, and a relation would translate to a list
+of structs.
+
+I would also strive for the type system to be minimal - don't differentiate
+between tuples, objects and structs. Choose one and stick to it.
```
max-sixty: I like this very much

book/src/internals/type-system.md
```diff
@@ -0,0 +1,179 @@
...
+
+# users-defined types
+type my_int = int
+
+# sum type (union or enum)
+type number = int | float
+type scalar = number | bool | str | char
+
+# by default types are not nullable
+type my_nullable_int = int | null
```
max-sixty: Minor point but given that most SQL columns are default nullable, we could have a concise way of communicating this — e.g. `Option<int>`

book/src/internals/type-system.md
```diff
@@ -0,0 +1,179 @@
...
+	title = text,
+	age = int
+]}
+```
+
+## Built-in primitives
+
+I've mentioned having `int32` and `int64` as distinct types, but initial
+implementation I don't see a need to do that. Instead we can start with a
+built-in `int`, which can later be replaced by `type int = int32 | int64`.
```
max-sixty: Perfect

book/src/internals/type-system.md
```diff
@@ -0,0 +1,179 @@
...
+Say, for example, that we have a pipeline like this:
+
+```
+derive color = switch [x => 'red', true => 'green']
+derive is_red = switch [color == 'red' => true, color == 'green' => false]
+```
+
+It should be possible to infer that `color` is of type `text`, but only when
+equal to `'red'` or `'green'`. This means that the second switch covers all
+possible cases and `is_red` cannot be `null`.
```
max-sixty: I'm no type theory expert, but FWIW my understanding is that this is quite difficult... Not out of reach forever ofc

book/src/internals/type-system.md
```diff
@@ -0,0 +1,179 @@
...
+## Built-in primitives
+
+I've mentioned having `int32` and `int64` as distinct types, but initial
+implementation I don't see a need to do that. Instead we can start with a
+built-in `int`, which can later be replaced by `type int = int32 | int64`.
+
+The general rule for "when to make a distinction between types" would be "as
+soon as we find an operation that would be expressed differently". In this
+example, that would require some operation on `int32` to have different syntax
+than same operation over `int64`.
```
max-sixty: Excellent rule!

max-sixty (approved):
Overall great! And thanks for writing it up!

Very much agree with the approach of starting simple — e.g. just `int` rather than `int64` — and seeing where we get to. At least in SQL, that's going to give us almost all the benefits, like an error on trying to add an int & a string.

---

One thing we could consider — though maybe you've considered it and rejected it, is to model the type system on something like [DuckDB](https://duckdb.org/docs/sql/data_types/overview.html). This is what @ghalimi suggested. To the extent that the industry coalesces on that standard, it'd be good. To the extent it's just another standard, it's less good, but could still be a helpful template.

That said, I think the approaches aren't _that_ different — both would start with some basic types and attempt to map in other type systems. And there isn't an easy way of using DuckDB's types in code (it's not in rust), so we're just using the spec.

---

My guess is that a core challenge here will be adding in types to the library gradually — e.g. can we get an error on `5 + "foo"` with reasonable work? There's so much we could do here — and likely we'll get lots of utility from the first 20% of work.

ghalimi:
> Overall great! And thanks for writing it up!
> 
> Very much agree with the approach of starting simple — e.g. just `int` rather than `int64` — and seeing where we get to. At least in SQL, that's going to give us almost all the benefits, like an error on trying to add an int & a string.
> 
> One thing we could consider — though maybe you've considered it and rejected it, is to model the type system on something like [DuckDB](https://duckdb.org/docs/sql/data_types/overview.html). This is what @ghalimi suggested. To the extent that the industry coalesces on that standard, it'd be good. To the extent it's just another standard, it's less good, but could still be a helpful template.
> 
> That said, I think the approaches aren't _that_ different — both would start with some basic types and attempt to map in other type systems. And there isn't an easy way of using DuckDB's types in code (it's not in rust), so we're just using the spec.
> 
> My guess is that a core challenge here will be adding in types to the library gradually — e.g. can we get an error on `5 + "foo"` with reasonable work? There's so much we could do here — and likely we'll get lots of utility from the first 20% of work.

DuckDB or Iceberg. They're highly compatible with each other, but we like DuckDB better because it offers more control regarding encoding sizes. Here is a mapping table between the two: https://github.com/sutoiku/puffin/blob/main/TYPES.md

adriangb:
I like the comparisons with Postgres, SQLite and DuckDB. I think it would also be interesting to compare to [Apache Arrow](https://arrow.apache.org/docs/python/api/datatypes.html)

ghalimi:
> I like the comparisons with Postgres, SQLite and DuckDB. I think it would also be interesting to compare to [Apache Arrow](https://arrow.apache.org/docs/python/api/datatypes.html)

Apache Arrow is a great candidate for that as well, and is really close to DuckDB. You can't go wrong with Arrow.

ghalimi:
> I like the comparisons with Postgres, SQLite and DuckDB. I think it would also be interesting to compare to [Apache Arrow](https://arrow.apache.org/docs/python/api/datatypes.html)

Here is a mapping table between DuckDB, Arrow, and Iceberg types:

https://github.com/sutoiku/puffin/blob/main/TYPES.md

Apache Arrow is probably more "universal", but DuckDB's types look a bit cleaner to me. It's purely cosmetic though.

max-sixty:
To what extent should we have a single type system for:
- "PRQL types", such as relation / column / function, and 
- "Data types", such as int / string?

My impulse is that they should be common, at least at the "front-end". If so, possibly we should add those to the doc. (Am happy to do it if you'd like me to)

book/src/internals/type-system.md
aljazerzen: Do you think that `Option<int>` would be more concise than `int | null`?

book/src/internals/type-system.md
aljazerzen: Yes, this is very difficult to do perfectly. But it's much easier in PRQL, where we consider all functions to be pure. We can also focus on simpler cases like the one described.

aljazerzen:
@max-sixty 
> To what extent should we have a single type system for:

I've added a section on physical layout that talks about this (if I understood correctly what you meant).

In summary, we don't have to define "data types" right now.

aljazerzen:
Regarding DuckDB and Apache Iceberg: we will surely use them as a base for defining primitives, but also steer away from their handling of null and a few other quirks.

ghalimi:
> Regarding DuckDB and Apache Iceberg: we will surely use them as a base for defining primitives, but also steer away from their handling of null and a few other quirks.

Nulls are painful...

RandomFractals:
Why not just use standard Arrow data types?

I believe those are more common and with more databases and systems like [Data Fusion](https://arrow.apache.org/blog/2023/01/19/datafusion-16.0.0/) and [Polars](https://www.pola.rs/)  providing native arrow and parquet data support, plus Arrow Flight and [ADBC](https://arrow.apache.org/blog/2023/01/05/introducing-arrow-adbc/), PRQL target system submodules could just convert arrow data types to the target system defined/supported data types.

See this api for example or arrow data types defined for other languages on their site: 

https://arrow.apache.org/docs/python/api/datatypes.html

aljazerzen:
@RandomFractals as I've written, this is not talking about physical layout of types, but about logical type system in PRQL. For physical layout we will use a standard, probably Apache Arrow, but this is not needed right now.

I am strongly against just adopting type system that is vaguely defined for Apache Arrow Python client or the type system of DuckDB, because it does not cohere principles described in the proposal. Again, for example, handling of null values.

book/src/internals/type-system.md
max-sixty: `int | null` is def fine.

I guess the example of `type my_nullable_int` wouldn't actually happen, folks would just use `int | null`

(or something like `int?` if we really wanted to shorten it)

aljazerzen: Yes, the `type my_nullable_int =` was meant as an example of "you can declare your types".

vanillajonathan:
I _really_ like the data types `int32` and `int64`. I've been thinking about database type systems and those two data types are exactly what I had in mind.

But what about `int8` and `int16`?
We would need something to map to `tinyint` and `smallint`.

Note: Interestingly enough Rust have a [`i128`](https://doc.rust-lang.org/std/primitive.i128.html) and a [`u128`](https://doc.rust-lang.org/std/primitive.u128.html) data type.

## `datetime` instead of `timestamp`?
As for `timestamp` maybe this should be named `datetime`?
It is `datetime` in Microsoft SQL Server ([documentation](https://learn.microsoft.com/en-us/sql/t-sql/data-types/datetime-transact-sql?view=sql-server-ver16)) which also offers the `date` and `time`data types, so if we were to offer a `date` and `time` data type (which I think we should) then `datetime` would be much more suitable, coherent and predictable than `timestamp`.

## A boolean data type?
Maybe there could be a boolean data type?
But keep in mind that MySQL have a built-in boolean data type, while Microsoft SQL Server does not and instead uses the `bit` data type.

## A string data type?
Most programming languages have a `string` datatype so that is what developers, especially front-end developers and NoSQL developers are accustomed to, but SQL uses `varchar` (and `char`). Oh, and Microsoft SQL Server also have `nvarchar` and `nchar` for Unicode.

## Vendor-specific data types?
Some databases have non-standard vendor-specific data types, such as `jsonb`in PostgreSQL. One thing that bothers me about this is that just by looking at a database schema it is difficult to discern whether a data type is a standard data type or a non-standard data type, and to which DBMS it belongs to.

CSS deals with this with vendor prefixes, example Chrome might have their vendor-specific CSS property called `-webkit-user-drag` and Firefox might have a vendor-specific CSS property named `-moz-radial-gradient`, then these are denoted as such by the `-webkit-` and `-moz-` prefix respectively.

Media types addresses this with "vendor trees", example `application/vnd.microsoft.portable-executable`.

book/src/internals/type-system.md
```diff
@@ -0,0 +1,224 @@
...
+Apache Arrow.
+
+## Syntax
+
+```
+# built-in types
+type int
+type float
+type bool
+type text
```
aljazerzen: > @vanillajonathan 
> A string data type?
>
> Most programming languages have a string datatype so that is what developers, especially front-end developers and NoSQL developers are accustomed to, but SQL uses varchar (and char). Oh, and Microsoft SQL Server also have nvarchar and nchar for Unicode.

I've used "text" here, because we found term "string" too programmer-like, which may be confusing to non-tech-savy people (see https://github.com/PRQL/prql/issues/286#issuecomment-1332656023). Apart from being consistent in the language, I don't have a strong preference here.

Whatever we choose, this type does not have a specific physical layout, by which I mean that I could be implemented by VARCHAR in one DBMS and by NVARCHAR in another.

book/src/internals/type-system.md
```diff
@@ -0,0 +1,224 @@
...
+In the future, PRQL may define a common physical layout of types, probably using
+Apache Arrow.
+
+## Syntax
+
+```
+# built-in types
+type int
+type float
+type bool
```
aljazerzen: > @vanillajonathan
> A boolean data type?
>
> Maybe there could be a boolean data type?
> But keep in mind that MySQL have a built-in boolean data type, while Microsoft SQL Server does not and instead uses the bit data type.

There is one proposed. And we really need it, because it's the type of all comparison expressions.

max-sixty:
> I've added a section on physical layout that talks about this (if I understood correctly what you meant).

Yes I think so. My main point was that we already have a type system for "PRQL types" such as functions and relations — attempting to compile `select (f a)` will fail if `f` isn't a function`. 

I'm guessing that will be part of this type system, even though our `function` type will never map to a DB type, because it's evaluated before it gets to the DB...

book/src/internals/type-system.md
max-sixty: (minor point — I was the person who thought `from_string` was too programmer-y. I'm less against it here, since types will predominantly be handled by the programmer-y folks rather than those more used to Excel)

book/src/internals/type-system.md
```diff
@@ -0,0 +1,224 @@
...
+would attach `#[repr(INTEGER)]` or `#[repr(TEXT)]` to the type. This would
+effect how `timestamp - timestamp` is translated into SQL. INTEGER can use
+normal int subtraction, but TEXT must apply `unixepoch` first.
+
+A similar example is a "string array type" in PTS that could be represented by
+an `text[]` (if DBMS supports arrays) or `json` or it's variant `jsonb` in
+Postgres. Again, the representation would affect operators: in Postgres arrays
+can be access with `my_array[1]` and json uses `my_json_array -> 1`. This
+example may not be applicable, if we decide that we want a separate JSON type in
+PST.
```
aljazerzen: > @vanillajonathan 
> Vendor-specific data types?
> 
> Some databases have non-standard vendor-specific data types, such as jsonbin PostgreSQL. One thing that bothers me about this is that just by looking at a database schema it is difficult to discern whether a data type is a standard data type or a non-standard data type, and to which DBMS it belongs to.
>
> CSS deals with this with vendor prefixes, example Chrome might have their vendor-specific CSS property called -webkit-user-drag and Firefox might have a vendor-specific CSS property named -moz-radial-gradient, then these are denoted as such by the -webkit- and -moz- prefix respectively.
>
> Media types addresses this with "vendor trees", example application/vnd.microsoft.portable-executable.

What I meant here, is that we may be able to support vendor-specific data types jsonb in Postgres by defining them as normal structs and lists in PRQL and mark them as having a "Postgres JSON" representation.

This would tell the compiler to use postgres's operators (i.e. `->`) for implementing standard PRQL operations (in this case field indirection - accessing a field of a struct).

book/src/internals/type-system.md
```diff
@@ -0,0 +1,224 @@
...
+# relations are list of structs
+type my_relation = {[
+	id = int,
+	title = text,
+	age = int
+]}
+
+type invoices = {[
+    invoice_id = int64,
+    issued_at = timestamp,
```
aljazerzen: > @vanillajonathan 
> datetime instead of timestamp?
>
> As for timestamp maybe this should be named datetime?
> It is datetime in Microsoft SQL Server ([documentation](https://learn.microsoft.com/en-us/sql/t-sql/data-types/datetime-transact-sql?view=sql-server-ver16)) which also offers the date and timedata types, so if we were to offer a date and time data type (which I think we should) then datetime would be much more suitable, coherent and predictable than timestamp.

I'm torn on this question.

datetime is more user-friendly, but also a bit misleading in that it refers to a date and a time including a timezone, when in fact it is refering to a moment, which is timezone independent.

My feeling here is that we should find a good (modern) API that deals with time and build on top of their findings.

aljazerzen: A solid API that hasn't disappointed me yet is [chrono crate](https://docs.rs/chrono/0.4.23/chrono/index.html)

Also Java 8 [java.time package](https://docs.oracle.com/javase/8/docs/api/java/time/package-summary.html)

book/src/internals/type-system.md
```diff
@@ -0,0 +1,224 @@
...
+DBMSs, since the physical layout of the result will vary.
+
+In the future, PRQL may define a common physical layout of types, probably using
+Apache Arrow.
+
+## Syntax
+
+```
+# built-in types
+type int
```
aljazerzen: > @vanillajonathan 
> I really like the data types int32 and int64. I've been thinking about database type systems and those two data types are exactly what I had in mind.
>
> But what about int8 and int16?
> We would need something to map to tinyint and smallint.
> 
> Note: Interestingly enough Rust have a [i128](https://doc.rust-lang.org/std/primitive.i128.html) and a [u128](https://doc.rust-lang.org/std/primitive.u128.html) data type.

Do you mean naming `int32, int64` as opposed to `integer, bigint` or `i32 and i64` or `int, long, long long`? That's a clear benefit of defining a new language and defining things to be consistent instead of just adding a `long` before your type name.

But as noted in section "Built-in primitives", we will not be adding these types soon, as there is no need to do so.

aljazerzen:
> I'm guessing that will be part of this type system, even though our function type will never map to a DB type, because it's evaluated before it gets to the DB...

Yes, this is meant to replace or extend the type system we have now. It was built without much design work, so I lacks the big picture view.

And yes, I completely forgot about the types of functions :D

book/src/internals/type-system.md
RandomFractals: I would argue you'd need int32, int64, float64, etc. and even though you are only proposing logical PRQL data type primitives here, having them would allow to define operations on large numbers.

aljazerzen: We can define operations on all kinds of integers by saying: use operator "+" of our target (SQL). Because all SQL dialects support "+" for all int types, this works.

Am I missing something? Is there an operation on ints that has different syntax in SQL depending on the int size?

RandomFractals: maybe operations was not a good example. What you are missing is that all the modern data stack tools I've seen in the past year or so now use arrow data types and arrow data, and databases are migrating to vectorized data with typed arrays/datasets too. If PRQL is the tool to use with OLAP systems mostly, or at least that's how I see it, I'd like it more if you just adopted arrow data types like everyone else in that space, even if what you are defining is abstract data types for PRQL.

book/src/internals/type-system.md
```diff
@@ -0,0 +1,224 @@
...
+type my_timestamp_with_timezone = [int, text]
+
+# list type (I'm not sure about this syntax)
+type list = {T}
+
+# relations are list of structs
+type my_relation = {[
+	id = int,
+	title = text,
+	age = int
```
eitsupi: The indentation seems to be tab.

aljazerzen: That's another thing we should talk about :D But not here.

book/src/internals/type-system.md
```diff
@@ -0,0 +1,224 @@
...
+type bool
+type text
+type char
+type null
+
+# users-defined types
+type my_int = int
+
+# sum type (union or enum)
+type number = int | float
```
eitsupi: It's a naive question, but I think it's hard to understand if pipe and OR use the same symbol `|`.
Couldn't we use some other symbol here?
(I don't know of any other symbol that would be better suited for OR except `||`, so it might be better to leave it as is.)

aljazerzen: Yes, I agree - using same syntax as pipe it not ideal.

I think having `||` is better.

book/src/internals/type-system.md
aljazerzen: If we decide on the proposed abstract types now, we can still adopt Arrow as the standard physical layout and expand the abstract types with primitives that correspond to Arrow logical types.

This was actually my intention, but I don't want to "just adopt" anything for the sake of standardization.

vanillajonathan:
> Very much agree with the approach of starting simple — e.g. just `int` rather than `int64`.

I disagree.

What exactly really is `int`?

It is ambiguous.
- Is it an alias to `int32`?
- Is it an alias to `int64`?
- Is it an union type that means either `int32` or `int64`?

I believe introducing `int` would be a mistake. Ambiguity is bad.

aljazerzen:
> What exactly really is int?

It is an integer number. Without specified size, but with many arithmetic operations defined. That is enough for now, as the size does not affect any of our translations. If needed, we can define it as a union of all `intXX` types in the future.

I wouldn't say this is ambiguous - type definition and all of its operations are well defined. But it may be too "dynamic"...

book/src/internals/type-system.md
RandomFractals: Perfect! in that case, I am good with keeping it simple and refine/add more types as needed based on use cases we might encounter when adding other system outputs like query plans from Data Fusion or even generating Polars Py code. I think you got the base types defined well. The rest we can sort out in a way you propose we expand on what int means later: `type int = int32 | int64`.

book/src/internals/type-system.md
```diff
@@ -0,0 +1,224 @@
...
+> For variables, the type system determines the allowed values of that term.
+>
+> -- Wikipedia
+
+## Purpose
+
+Each of the SQL DBMSs has their own type system. Thanks to SQL standard, they
+are very similar, but have key differences. For example, SQLite does not have a
+type for date or time or timestamps, but it has functions for handling date and
+time that take ISO8603 strings or ints representing Unix timestamps. So it does
```
snth: ISO 8603 is about Solid Fertilizers: https://www.iso.org/standard/15908.html

I think you meant ISO 8601 😆

book/src/internals/type-system.md
```diff
@@ -0,0 +1,224 @@
...
+
+PRQL's task is to define common description of _data formats_, just as how it
+already defines _data transformations_.
+
+I believe this should best be done in two steps:
+
+1. Define PRQL's Type System (PTS), following principles we think a relational
+   language should have (and not focus on what existing SQL DBMSs have).
+
+2. Define a mapping from the SQL's Type System (STS) into PTS, for each of the
```
snth: Since it's not going to be a bijection, shouldn't the mapping be from PTS to STS? The image of this mapping may be a strict subset of the co-domain (i.e. STS) but that doesn't matter since we only care about mapping PRQL to SQL at the moment.

(We discussed that it would be nice to map SQL back to PRQL but is that in scope at the moment?)

book/src/internals/type-system.md
```diff
@@ -0,0 +1,224 @@
...
+flexibility, all while being conceptually simple.
+
+**Composable** - as with transformation, we'd want types to compose together.
+
+Using Python, JavaScript, C++ or Rust, one could define many different a data
+structure that would correspond to our idea of "relation". Most of them would be
+an object/struct that has column names and types and then a generic array of
+arrays for rows.
+
+PRQL's type system should also be able to express relation as composed from
```
snth: relation^s^

book/src/internals/type-system.md
```diff
@@ -0,0 +1,224 @@
...
+In many programming languages, physical layout of a logical type is dependent on
+the target platform. Similarly, physical layout of a PRQL logical type is
+dependent on representation of that type in target STS.
+
+```
+PTS logical type  --->  STS logical type  ---> STS physical layout
+```
+
+Note that STS types do not have a single physical layout. Postgres has a logical
+type `anyelement`, which is a super type of any data type. It can be used a
```
snth: It can be used ^as^ a ...

book/src/internals/type-system.md
```diff
@@ -0,0 +1,224 @@
...
+function parameter type, but does not have a single physical layout so it cannot
+be used a in column declarations.
```
snth: be used in<->a column declaration~s~.

book/src/internals/type-system.md
```diff
@@ -0,0 +1,224 @@
...
+### Non-bijection cases between PTS and STS
+
+There are cases where a PTS construct has multiple possible and valid
+representations in some STSs.
+
+For such cases, we'd want to have something similar to Rust's `#[repr(X)]` which
+says "data in this type is represented as X" (we'd probably want a different
+syntax).
+
+This is needed because translation from PRQL operation to SQL may depend on the
```
snth: operation^s^

book/src/internals/type-system.md
```diff
@@ -0,0 +1,224 @@
...
+representation.
+
+Using SQLite as an example again, users may have some data stored as INTEGER and
+some as TEXT, but would want to define both of them as PTS `timestamp`. They
+would attach `#[repr(INTEGER)]` or `#[repr(TEXT)]` to the type. This would
+effect how `timestamp - timestamp` is translated into SQL. INTEGER can use
```
snth: affect

book/src/internals/type-system.md
```diff
@@ -0,0 +1,224 @@
...
+normal int subtraction, but TEXT must apply `unixepoch` first.
+
+A similar example is a "string array type" in PTS that could be represented by
+an `text[]` (if DBMS supports arrays) or `json` or it's variant `jsonb` in
```
snth: a~n~ `text[]`

snth (commented):
Looks great! Mostly just pointing out minor typos in this review.

I have some thoughts on the type syntax which I'll put in a conversation as that's less broken up than this view.

book/src/internals/type-system.md
max-sixty: @snth easy to add suggestions and then the author can just click

```suggestion
PRQL's type system should also be able to express relations as composed from
```

book/src/internals/type-system.md
max-sixty: FYI this was me that suggested `from_text` rather than `from_string`.

I'm actually +1 on using `string` here, since this part is much more engineer-y, and the "Excel poweruser archetype user" won't be interacting with these.

eitsupi:
Perhaps a large number of types like Apache Arrow data types will be needed when implementing something like `cast` function?

book/src/internals/type-system.md
snth: Thanks @max-sixty . I didn't know that. Will try and do that from now on.

snth: Couldn't figure this out on my own but got it now thanks to this: https://egghead.io/lessons/github-add-suggestions-in-a-github-pr-review

book/src/internals/type-system.md
snth:
```suggestion
type `anyelement`, which is a super type of any data type. It can be used as a
```

book/src/internals/type-system.md
snth:
```suggestion
be used in a column declaration.
```

book/src/internals/type-system.md
snth:
```suggestion
This is needed because translation from PRQL operations to SQL may depend on the
```

book/src/internals/type-system.md
snth:
```suggestion
affect how `timestamp - timestamp` is translated into SQL. INTEGER can use
```

book/src/internals/type-system.md
snth:
```suggestion
a `text[]` (if DBMS supports arrays) or `json` or it's variant `jsonb` in
```

snth (commented):
Added suggestions

snth:
As hinted at before, I have some suggestions for the syntax of the type system.

The principles guiding my suggestions are:
1. Try to reuse whatever syntax we have already.
2. Where we need to introduce new syntax, try to keep this consistent with other popular conventions out there.

We currently use angle brackets `<>` for type annotations. I have mostly seen this used in discussions in Github issues and in the [PRQL std lib](https://prql-lang.org/book/language-features/standard-library.html). To me it seems like `<>` acts as a "type constructor" operator (I'm not sure if that's completely correct in the most technical sense of the word) in the sense that it takes an expression and turns it into a type. I think we should try to keep that and ensure that whatever syntax we decide on stays consistent with that. What I mean by that is the following:

* While for function defintions we write
    ```elm
    func interp lower:0 higher x -> (x - lower) / (higher - lower)
    ```
   this is equivalent to
    ```elm
    let interp = lower:0 higher x -> (x - lower) / (higher - lower)
    ```
    but the first form is preferred for readability.
* Similarly for types we could(/should) have that
    ```elm
    type number = int | float
    ```
    is equivalent to
    ```elm
    let number = <int | float>
    ```
    where again the first is preferred for readability but they are equivalent.

Other syntax that we have already firmly established is (square) brackets `[]` for lists. I think we should carry this over into the type system rather than introducing `{}` for this. Moreover, in terms of point 2. above, `{}` is quite commonly used for struct types in other languages such as Javascript, JSON, Python, ... .

Moreover, we have already established `:` as syntax for named parameters, as for example in the `interp` function definition above. So then if we view `{}` as a "struct constructor", then it would make sense that it takes named parameters with `:` and purely positional ones (for tuples) simply separated by `,`. This has the nice consequence that the resulting syntax is very similar, if not identical to, JSON.

Therefore I propose rewriting the Syntax section as follows:

## Syntax

```elm
# built-in types
type int
type float
type bool
type text
type char
type null

# users-defined types
type my_int = int

# sum type (union or enum)
type number = int || float
type scalar = number || bool || str || char

# by default types are not nullable
type my_nullable_int = int || null

# user-defined enum
type open
type pending
type closed
type status = open || pending || closed

# product type
type my_struct = {id: int_nullable, name: str}
type my_timestamp_with_timezone = {int, text}

# list type
type list = [T]

# relations are list of structs
type my_relation = [{
	id: int,
	title: text,
	age: int
}]

type invoices = [{
    invoice_id: int64,
    issued_at: timestamp,
    labels: [text]

    #[repr(json)]
    items = [{
        article_id: int64,
        count: int16 where x -> x >= 1,
    }],
    paid_by_user_id: int64 || null,
    status: status,
}]
```

WDYT?

EDIT:  I didn't remember about the `#[repr(json)]` bit and will have to reread that and comment on that another time.

book/src/internals/type-system.md
aljazerzen:
```suggestion
time that take ISO 8601 strings or ints representing Unix timestamps. So it does
```

book/src/internals/type-system.md
aljazerzen: Thinking about it, I realize we actually need two mappings:

- STS to PTS, for when people (or LSP) want to describe their database in PRQL,
- PTS to STS, for compiler to generate valid SQL for the defined PRQL types.

As you said, image of the second mapping must be a subset of the target STS, so we can translate the queries to SQL. (it will take quite a lot of effort to achieve this, we will start with a working impl for basic types and a few dialects).

book/src/internals/type-system.md
```diff
@@ -0,0 +1,224 @@
...
+type my_nullable_int = int | null
+
+# user-defined enum
+type open
+type pending
+type closed
+type status = open | pending | closed
+
+# product type
+type my_struct = [id = int_nullable, name = str]
```
aljazerzen: > @snth 
> Other syntax that we have already firmly established is (square) brackets [] for lists. I think we should carry this over into the type system rather than introducing {} for this. Moreover, in terms of point 2. above, {} is quite commonly used for struct types in other languages such as Javascript, JSON, Python, ... .
> 
> Moreover, we have already established : as syntax for named parameters, as for example in the interp function definition above. So then if we view {} as a "struct constructor", then it would make sense that it takes named parameters with : and purely positional ones (for tuples) simply separated by ,. This has the nice consequence that the resulting syntax is very similar, if not identical to, JSON.

There is a clash between what we now call "lists" and what list is in this PR.

Our `select` and `derive` take something that has a fixed number elements, they can be of different types and they can be named. This definition collides with what this PR proposes under the name "struct". So effectively, this PR renames "lists" to "structs".

E.g.:
```
select [1.4, false, "foo"] # this is now named "struct"
```

It also proposes a new construct named "list" which acts as "a repeated type" - number of repetitions is unknown at compile time, all elements are of the same type and they cannot be named. Think of a column in a relation.

Now, there are two options for the syntax:
- (what this PR currently proposes) try to preserve current syntax: use `[]` for structs and `{}` for lists,
- (what you described) mimic JSON and basically all other languages: use `{}` for structs and `[]` for lists.

book/src/internals/type-system.md
```diff
@@ -0,0 +1,224 @@
...
+# built-in types
+type int
+type float
+type bool
+type text
+type char
+type null
+
+# users-defined types
+type my_int = int
```
aljazerzen: > @snth 
> We currently use angle brackets <> for type annotations. I have mostly seen this used in discussions in Github issues and in the [PRQL std lib](https://prql-lang.org/book/language-features/standard-library.html). To me it seems like <> acts as a "type constructor" operator (I'm not sure if that's completely correct in the most technical sense of the word) in the sense that it takes an expression and turns it into a type. I think we should try to keep that and ensure that whatever syntax we decide on stays consistent with that. What I mean by that is the following:
> 
>  -  While for function defintions we write
>     ```
>     func interp lower:0 higher x -> (x - lower) / (higher - lower)
>     ```
>     this is equivalent to
>     ```
>     let interp = lower:0 higher x -> (x - lower) / (higher - lower)
>     ```
>     but the first form is preferred for readability.
>  -  Similarly for types we could(/should) have that
>     ```
>     type number = int | float
>     ```
>     is equivalent to
>     ```
>     let number = <int | float>
>     ```
>     where again the first is preferred for readability but they are equivalent.

This is a great point: I haven't really defined how types interact with values and what `< >` means.

I think we should define a type of a variable to be "a set of all possible values of that variable". And if types are just sets, any type expression are just set expressions (e.g. `int || null` is a union of two sets).

`param <X>` would then mean "param can accept any value from set X".

I'd say that the syntax for your last example should then be:

```
type number = int || float
let number = int || float
```

vanillajonathan:
@snth For types declared over multiple lines, I think it would be nice have to use squrae brackets, so that curly braces would be enough, just like in the majority of other languages.

book/src/internals/type-system.md
max-sixty: > This definition collides with what this PR proposes under the name "struct". So effectively, this PR renames "lists" to "structs".
> 
> E.g.:
> 
> ```
> select [1.4, false, "foo"] # this is now named "struct"
> ```

I hadn't thought about it like that, but I think that's quite accurate.

IIUC most DBs name their List type as Array, so we could have 

I'm keen to keep the PRQL construct of `select [a, b, c]` be called something simple, like List (and not Struct), since it's a very simple type that everyone would encounter very early. 

Whereas the "repeated type" is more advanced and OK to call something technical.

aljazerzen: We'd then have lists and arrays. And lists would represent relation rows.

What about tuples and arrays? Is the name "tuple" approachable for non-programmer people? Would it be confusing that `[]` is used for tuples?

I think that tuple is very close to relational algebra and would be a good fit here.

max-sixty: > I think that tuple is very close to relational algebra and would be a good fit here.

It's a bit foreign but I agree it's very close to relational algebra. Definitely fine in the typing docs; I'm less keen on having the opening example explaining PRQL be "Then we pass a _tuple_ of `[1.4, false, "foo"]` to `select`"...

It's also unfortunate that our `[]` syntax is not standard tuple syntax. Though this could give us an opportunity to make our parentheses syntax even more complicated! (no)

> And lists would represent relation rows.

I didn't have this in my mind — if anything the List would be closer to columns, or values. (Or maybe that's what you mean? Values in one row?)

aljazerzen: This naming gets confusing fast :D Let me clarify:

- lists (currently expressed with `[]`) or tuples (proposed) correspond to rows.
- arrays (proposed) correspond to columns.

book/src/internals/type-system.md
snth: > I think we should define a type of a variable to be "a set of all possible values of that variable".

Nice, I like this!

> And if types are just sets, any type expression are just set expressions (e.g. int || null is a union of two sets).

Minor nit: do we distinguish between singleton sets and the value they contain or are values automatically upgraded to singleton sets? For example technically it should probably be

    nullable_int = int || {null}

but I think it might make sense to treat values the same as singleton sets containing them as long as that doesn't cause any problems elsewhere.

vanillajonathan:
I think there should be a `uuid` type. Example `123e4567-e89b-12d3-a456-426614174000`.

book/src/internals/type-system.md
snth: Very interesting! I had thought about the `select` case briefly before posting but I had only considered the case of a list of column names so I thought it was still covered by `lists / arrays`. I had forgotten that you can also include literal values and do renaming like you mentioned which brings you to structs as you pointed out, e.g.

    select [num = 1.4, truthy = false, bar = "foo", double_col = col * 2]

Having `[ ]` for structs / records is uncommon but not unprecedented. This is actually taking us pretty close to [Power Query M Formula Language](https://learn.microsoft.com/en-us/powerquery-m/) because M also uses `[]` for records and `{}` for lists. See for example this example from their [Quick Tour](https://learn.microsoft.com/en-us/powerquery-m/quick-tour-of-the-power-query-m-formula-language):

    let Orders = Table.FromRecords({  
        [OrderID = 1, CustomerID = 1, Item = "fishing rod", Price = 100.0],  
        [OrderID = 2, CustomerID = 1, Item = "1 lb. worms", Price = 5.0],  
        [OrderID = 3, CustomerID = 2, Item = "fishing net", Price = 25.0]}),  
        #"Capitalized Each Word" = Table.TransformColumns(Orders, {"Item", Text.Proper})  
    in  
        #"Capitalized Each Word"

M is actually underappreciated IMHO, probably because few programmers bother with Power Query and the people that use Power Query just want to get the data out and use the UI and aren't too interested in the underlying M code. 

M is actually pretty similar to PRQL in it's aims and use cases, namely it's a **pipelined, functional language for data munging**. I quite like it in its structure and approach. 

What I don't like about it is:
* it tends to be quite verbose
* the syntax for structs and lists is unconventional (and the wrong way around IMHO)

Given these (at least to me) new insights that we've got now, I think we should take a careful look at PRQL syntax again and perhaps try out some variations just to see what they look like.

book/src/internals/type-system.md
aljazerzen: Wait, I just realized that we already use `or` for logical disjunction. Can we use it here?

Let's say we have two sets:
```
let a = {1, 2, 3}
let b = bool

# does this make sense?
let union_of_a_and_b = a or b

# or is the better?
let union_of_a_and_b = a + b
```

P.S. Python uses `|` (binary or) for unioning sets.

book/src/internals/type-system.md
aljazerzen: > do we distinguish between singleton sets and the value they contain

Technically we should distinguish, but "automatic coercion to singletons" you propose would work.

The thing is that apart from this, we don't need syntax for sets so we don't need to use `{}` and it remains available for other applications.

book/src/internals/type-system.md
eitsupi: I agree that it is better to unify either "`&&` and `||`" or "`and` and `or`".
I prefer the former, but given that it is easier to distinguish from `|`, maybe the latter is better.

book/src/internals/type-system.md
aljazerzen: Woah, M is really similar to PRQL in some aspects, nice find.

Looking at the example you've posted, I see the unconventional aspect of `[]` for structs and `{}` for lists.

I'm now leaning to the "mimic JSON" syntax. An example:

```elm
from my_table
select {num = 1.4, truthy = false, bar = "foo", double_col = col * 2}
group {col_2} (derive a = sum double_col)
sort {num, truthy}
```

And yes, we should try out changing the syntax on more examples before switching. Maybe translate all examples on the website?

max-sixty: > This naming gets confusing fast :D Let me clarify:
> 
>     * lists (currently expressed with `[]`) or tuples (proposed) correspond to rows.
> 
>     * arrays (proposed) correspond to columns.

Would it be possible to further clarify what we mean by "correspond to {rows,columns}"?

I would have thought that lists correspond to columns, because `col_1` is a column in `select [col_1, col_2]`?

max-sixty: (edit: possibly we should push the syntax changes off to a different thread, unless it's required for typing decisions, sorry for sidetracking us)


> ```elm
> from my_table
> select {num = 1.4, truthy = false, bar = "foo", double_col = col * 2}
> group {col_2} (derive a = sum double_col)
> sort {num, truthy}
> ```

I agree with the internal elegance of this. And we can say "`select {a}` is equivalent to `select {a=a}`.

But `sort {num, truthy}` is still more alien than `sort [num, truthy]`, so I think we should pause before making this change. For most people who aren't programmers, the squiggly brackets are barely used!

RandomFractals:
Frankly, if we have this many comments in PR, maybe we should start/use design/feature ticket or discussions next time to get consensus and iron out the details. PRs this long should never happen mainly b/c not many github users actually read PR comments, but they do read issue comments and discussions. And if you want to involve your dev/users community in these discussions, maybe move them out to discussions next time. :)

book/src/internals/type-system.md
aljazerzen: > Would it be possible to further clarify what we mean by "correspond to {rows,columns}"?

Right now, this is the way I'd model our relations:
```
relation = array of tuples
tuple = a list of named types
```
From this I hope it's obvious what I mean when I say that a relation row is expressed as a tuple.

----

> For most people who aren't programmers, the squiggly brackets are barely used!

I'd say the same thing for square brackets.

snth: I agree with @max-sixty that we should pause for a while before a change like this (if we went ahead) because it's quite fundamental. I'm actually pro it and would also favour bringing things around to be more consistent with JSON type syntax but IMHO we should take care to try and think through as many of the repercussions as we can because we don't want to make changes like this to often.

So, in terms of current release numbers I would propose something like the following (numbers are just indicative and could be incremented appropriately depending on when the work gets done):
* 0.6.0: release imminent, pretty much in "feature freeze" mode
* 0.7.0: more QOL improvements, (settle syntax deliberations)
* 0.8.0: last release with old syntax, advertise upcoming syntax change
* 0.9.0 or 0.10.0: release new syntax

My main point is that given that it might be a fair amount of work to rewrite existing queries, people should be given adequate warning and they should still be given the opportunity to benefit from the next round or two of QOL improvements before having to rewrite their queries for the new syntax.

Maybe I'm too conservative but that's how I would go about it.

---
I agree that we should probably move this to the Github Discussions section since we seem to have stumbled onto a bigger topic than we anticipated.

max-sixty: > tuple = a list of named types

OK, makes sense.

---

(Let's move the syntax conversation elsewhere, I'm muddying the discussion...)

max-sixty:
> if we have this many comments in PR, maybe we should start/use design/feature ticket or discussions next time to get consensus and iron out the details.

At the risk of getting meta / having another thread of discussion here — I do think anchoring in a specific proposal is very helpful; rather than a discussion on "what does anyone want to see in our type system".

I find this a very difficult problem. There are tradeoffs between gathering more ideas vs. having a focused discussion; between collecting a wider range of viewpoints vs. having a more informed discussion.

A couple of things I've found helpful:
1. Anchor in an actual proposal
2. Move other threads elsewhere, to keep the discussion focused
3. Produce a summary, so folks don't need to read the whole thing to catch up on the conclusions. Can do midway through to outline the current state of the the discussion and then restart.

We get (1) & (3) by working on this doc and then merging it, and then having another discussion if we need one. I'm fine to have that discussion in GH Discussion if folks prefer...

book/src/internals/type-system.md
rslabbert: Fully agreed on using a good modern API for datetime logic. A major source of logic bugs in data transformations comes from datetime handling.

Chrono is definitely a great reference and personally I think there should definitely be at least two native datetime types for sql usage: datetime (with TZ) and naive_datetime (without TZ).

book/src/internals/type-system.md
```diff
@@ -0,0 +1,323 @@
...
+    invoice_id = int64,
+    issued_at = timestamp,
+    labels = {text}
+
+    #[repr(json)]
+    items = [{
+        article_id = int64,
+        count = int16 where x -> x >= 1,
+    }],
+    paid_by_user_id = int64 | null,
```
eitsupi:
```suggestion
    paid_by_user_id = int64 or null,
```

book/src/internals/type-system.md
```diff
@@ -0,0 +1,323 @@
...
+
+## Appendix
+
+### Built-in primitives
+
+This document mentions `int32` and `int64` as distinct types, but there is no
+need for that in the initial implementation. The built-in `int` can associate
+with all operations on integers and translate PRQL to valid SQL regardless of
+the size of the integer. Later, `int` be replaced by
+`type int = int8 | int16 | int32 | int64`.
```
eitsupi:
```suggestion
`type int = int8 or int16 or int32 or int64`.
```

book/src/internals/type-system.md
```diff
@@ -0,0 +1,323 @@
...
+Apache Arrow.
+
+<!-- ## Enums
+
+```
+# user-defined enum
+type open
+type pending
+type closed
+type status = open | pending | closed
```
eitsupi:
```suggestion
type status = open or pending or closed
```

eitsupi (commented):
Perhaps missed replacements?

aljazerzen:
I've updated the RFC with discussed changes. Mainly these are under sections Theory, Type annotations, Type definitions and Container types. I still have to expand definitions to include functions.

book/src/internals/type-system.md
aljazerzen: Let's continue the discussion here: https://github.com/PRQL/prql/issues/2124

book/src/internals/type-system.md
aljazerzen: Filed here: https://github.com/PRQL/prql/issues/2125

book/src/internals/type-system.md
```diff
@@ -0,0 +1,323 @@
...
+It should be possible to infer that `color` is of type `text`, but only when
+equal to `'red'` or `'green'`. This means that the second switch covers all
+possible cases and `is_red` cannot be `null`.
+
+## Theory
+
+> For any undefined terms used in this section, refer to set theory and
+> mathematical definitions in general.
+
+A "type of a variable" is a "set of all possible values of that variable". This
```
max-sixty: No strong opinion, but isn't the word `type` more familiar than the word `set`? I get the theory here. For the sake of our docs, I'd favor going back to `type` when these become docs....

book/src/internals/type-system.md
```diff
@@ -0,0 +1,323 @@
...
+
+But similar to how both `func` and `let` can be used to define functions (when
+we introduce lambda function syntax), let's also define syntactic sugar for type
+definitions:
+
+```
+# these two are equivalent
+let my_type <set> = set_expr
+type my_type = set_expr
+```
```
max-sixty: FWIW I'm increasingly in favor of this approach — i.e. functions can be `let add = a b -> a + b`

OTOH having two ways of defining these is not ideal. So we could only do the first, and lint towards including the type annotation for set/types & functions...

book/src/internals/type-system.md
```diff
@@ -0,0 +1,323 @@
...
+is known at compile-time. Each field has a type itself and an optional name.
+Fields are not necessarily of the same type.
+
+In other languages, similar constructs are named struct, tuple, named tuple or
+(data)class.
+
+**Array** is a container type that contains n ordered fields, where n is not
+known at compile-time. All fields are of the same type and cannot be named.
+
+**Relation** is an array of tuples.
```
max-sixty: Very clear, thanks!

book/src/internals/type-system.md
```diff
@@ -0,0 +1,323 @@
...
+
+So let's introduce a "set" as a PRQL expression construct (alongside existing
+idents, literals, ranges and so on). For now, it does not need any special
+syntax. Because sets are normal expressions, existing syntax can be repurposed
+to define operations on sets:
+
+- Binary operation `or` of two sets represents a union of those two sets:
+
+  ```
+  let number = int or float
```
max-sixty: Separate issue filed at https://github.com/PRQL/prql/issues/2170

max-sixty (approved):

book/src/internals/type-system.md
aljazerzen: > going back to type

Changing to "or of two types represents a union of those two types"?

max-sixty: My focus here iz the discussion below where we start to refer to types exclusively as sets. I support making the link between the concepts, I just think when we start documenting this, we should use the term "type" rather than "set".

aljazerzen: I see. I'm using term "set", because it's more suitable here in theory section and because it will be implemented this way - type declarations will be just variable declarations of sets.

Let's hold on with this until we convert this to documentation.

aljazerzen:
I'm merging this now, as it is generally finished. That does not mean that all details described here are set in stone, but rather that the outline of the type system is ready for further discussion in separate threads.

Thank you all for comments!

PR #1967, opened 2023-02-26
fix: [draft] Expand keywords

### LINKED ISSUE #1966: Expand set of quoted keywords to include sqlite keywords
I have a tables named "Assessment" and "Appraisal", so I entered the statement `from as=Assessment`. It generates (obviously) the following SQL:

```
SELECT
  *
FROM
  "Assessment" AS as
```
and SQLite throws an error: `Result: near "as": syntax error`

It's easy enough to work around (`from ass=Assessment`) but could this be solved in PRQL, perhaps by quoting the variable name? (I don't know enough about SQL...)

max-sixty:
Would close #1966, but with lots of changes. Let's confirm the SQLite requirements before merging this, and prepare to change lots of the examples so we don't get so many columns quoted in our examples, which adds syntactic noise to them.


max-sixty:
I think the path forward in https://github.com/PRQL/prql/issues/1966 of copying the keywords from that list into a list in the repo makes sense. We can reference that list in the `SqlLite` dialect, otherwise referencing the default list.

So closing this.

Issue #873, opened 2022-07-27
Design of `translator.rs`


max-sixty:
We're starting to write more dialect-specific code in `translator.rs`; e.g. https://github.com/prql/prql/pull/858

What's the best design for this?
- Currently we have a mix of methods on `Dialect` structs https://github.com/prql/prql/blob/a1919f5504974878a1f8c763cb43cb5cac4f7d7f/prql-compiler/src/ast/dialect.rs#L90-L97 and `match` expressions in `translate_*` functions https://github.com/prql/prql/blob/a1919f5504974878a1f8c763cb43cb5cac4f7d7f/prql-compiler/src/sql/translator.rs#L801-L802. If we continue with this, what should go in each?
- We're sometimes using bare functions like https://github.com/prql/prql/blob/a1919f5504974878a1f8c763cb43cb5cac4f7d7f/prql-compiler/src/sql/translator.rs#L699 and sometimes using `From` impls https://github.com/prql/prql/blob/a1919f5504974878a1f8c763cb43cb5cac4f7d7f/prql-compiler/src/sql/translator.rs#L901
  - `rust-analyzer` has upcoming improved support for `From`, so it'll be less of an issue to track down the specific impl, which I've found inconvenient historically
- Should we do something like [`ast_fold.rs`](https://github.com/prql/prql/blob/main/prql-compiler/src/ast/ast_fold.rs), where each Dialect "inherits" from a generic dialect, using this "Fold" pattern to generate a sql AST from constituent parts? This kinda emulates inheritance in Rust.
  - Note that these functions have different input and output types, so it might be a slightly different design
- I added a question https://github.com/prql/prql/blob/a1919f5504974878a1f8c763cb43cb5cac4f7d7f/prql-compiler/src/ast/dialect.rs#L29-L31 on whether we had the right design for Dialects. We're swapping between them at the moment; e.g. https://github.com/prql/prql/blob/a1919f5504974878a1f8c763cb43cb5cac4f7d7f/prql-compiler/src/sql/translator.rs#L798

Anything else?

aljazerzen:
I very much agree that dialect specifics should not be part of translator.rs, but reside in `dialects.rs` or even their own module.

> Should we do something like [ast_fold.rs](https://github.com/prql/prql/blob/main/prql-compiler/src/ast/ast_fold.rs), where each Dialect "inherits" from a generic dialect

[...]

PR #1970, opened 2023-02-27
fix: Fix sqlite datetime output

### LINKED ISSUE #1969: Incorrect datetime value outputs in sqlite
For this query:

```
from test_table
select [date = @2022-12-31, time = @08:30, timestamp = @2020-01-01T13:19:55-0800]
```

current implementation emits following query for sqlite dialect:

```
SELECT
  DATE '2022-12-31' AS date,
  TIME '08:30' AS time,
  TIMESTAMP '2020-01-01T13:19:55-0800' AS timestamp
FROM
  test_table
```

However, according to sqlite's doc: https://www.sqlite.org/lang_datefunc.html , these typed field outputs are incorrect as sqlite uses function to convert string to time values.

bcho:
fix #1969 

This pull request attempts to fix the datetime values output in sqlite by using a different translate functions for sqlite dialect. I added a `is` helper to the `DialectHandler` to allow detecting what's the dialect on runtime.


max-sixty:
Thanks a lot @bcho !

Interesting approach re getting the type at runtime. Maybe we should be doing that for all these...

I see a false negative test failure on my phone, not sure whether you see the same. I'll look again when I'm at a computer.

Overall looks great!

prql-compiler/src/sql/gen_query.rs
```diff
@@ -846,4 +846,46 @@ mod test {
           _expr_0 > 3
         "###);
     }
+
+    #[test]
```
max-sixty: Generally we've put the standard "does this PRQL create this SQL" in `test.rs`, though looks like we have a few others here. Unless you have a preference, would it be OK to move to `test.rs`?

prql-compiler/src/sql/gen_expr.rs
```diff
@@ -184,6 +183,60 @@ pub(super) fn translate_literal(l: Literal, ctx: &Context) -> Result<sql_ast::Ex
...
+    ctx: &Context,
+) -> sql_ast::Expr {
+    if ctx.dialect.is::<crate::sql::dialect::SQLiteDialect>() {
+        translate_datetime_literal_with_sqlite_function(data_type, value)
+    } else {
+        translate_dateatime_literal_with_typed_string(data_type, value)
+    }
+}
+
+fn translate_dateatime_literal_with_typed_string(
```
max-sixty:
```suggestion
fn translate_datetime_literal_with_typed_string(
```

(unless this was intentional)

prql-compiler/src/sql/gen_expr.rs
```diff
@@ -184,6 +183,60 @@ pub(super) fn translate_literal(l: Literal, ctx: &Context) -> Result<sql_ast::Ex
...
 
+fn translate_datetime_literal(
+    data_type: sql_ast::DataType,
+    value: String,
+    ctx: &Context,
+) -> sql_ast::Expr {
+    if ctx.dialect.is::<crate::sql::dialect::SQLiteDialect>() {
+        translate_datetime_literal_with_sqlite_function(data_type, value)
+    } else {
+        translate_dateatime_literal_with_typed_string(data_type, value)
```
max-sixty:
```suggestion
        translate_datetime_literal_with_typed_string(data_type, value)
```

prql-compiler/src/sql/dialect.rs
```diff
@@ -109,7 +110,7 @@ pub(super) enum ColumnExclude {
     Except,
 }
 
-pub(super) trait DialectHandler {
+pub(super) trait DialectHandler: Any {
```
max-sixty, line 113: Could I ask why this is needed? (You probably know more than me about the complexities of rust's type system...)

max-sixty (approved):
@bcho great code! Thank you!

(looks like the test error I mention above only shows on the mobile app, so all good)

prql-compiler/src/sql/gen_expr.rs
```diff
@@ -184,6 +183,60 @@ pub(super) fn translate_literal(l: Literal, ctx: &Context) -> Result<sql_ast::Ex
...
+    value: String,
+) -> sql_ast::Expr {
+    sql_ast::Expr::TypedString { data_type, value }
+}
+
+fn translate_datetime_literal_with_sqlite_function(
+    data_type: sql_ast::DataType,
+    value: String,
+) -> sql_ast::Expr {
+    let timezone_indicator_regex = Regex::new(r"([+-]\d{2}):?(\d{2})$").unwrap();
```
max-sixty, line 212:
```suggestion
    // TODO: promote parsing timezone handling to the parser; we should be storing
    // structured data rather than strings in the AST
    let timezone_indicator_regex = Regex::new(r"([+-]\d{2}):?(\d{2})$").unwrap();
```

(no need to do as part of this tbc)

prql-compiler/src/sql/gen_expr.rs
```diff
@@ -184,6 +183,60 @@ pub(super) fn translate_literal(l: Literal, ctx: &Context) -> Result<sql_ast::Ex
     })
 }
 
+fn translate_datetime_literal(
+    data_type: sql_ast::DataType,
+    value: String,
+    ctx: &Context,
+) -> sql_ast::Expr {
+    if ctx.dialect.is::<crate::sql::dialect::SQLiteDialect>() {
```
max-sixty, line 192: @aljazerzen let us know if you have thoughts on this vs methods on `DialectHandler`. This approach seems good...

prql-compiler/src/sql/gen_expr.rs
```diff
@@ -184,6 +183,60 @@ pub(super) fn translate_literal(l: Literal, ctx: &Context) -> Result<sql_ast::Ex
     })
 }
 
+fn translate_datetime_literal(
```
max-sixty, line 187: @aljazerzen this is an example of the question in https://github.com/PRQL/prql/issues/873 — should this be a method on `DialectHandler`, rather than a function which branches based on the value of `ctx.dialect`?

aljazerzen: Yes, that would be object-oriented-approach best practice. That does not apply in Rust - I dont't think it matters where we implement this complexity, so I can remain here for now.

prql-compiler/src/sql/gen_query.rs
bcho: yeah I couldn't find the test.rs at the first place. Will move these cases to test.rs.

prql-compiler/src/sql/gen_expr.rs
bcho: yeah I agree we should try parsing the timezone information from user input to provide more flexible transpiling.

prql-compiler/src/sql/gen_query.rs
bcho: updated

prql-compiler/src/sql/dialect.rs
bcho: the type_id thing is granted by std::any (https://doc.rust-lang.org/std/any/index.html). I learnt this trick from https://github.com/sqlparser-rs/sqlparser-rs/blob/0c0d088ec21d7f3883165d36c34300745816c5c7/src/dialect/mod.rs#L110-L113

prql-compiler/src/sql/gen_expr.rs
bcho: IMO, more or less we will need a similar way to tell which target we are generating the query to. This would be more helpful if we are going to generate query for languages beyond SQL-like.

Or maybe another approach is we push down the full generation into each dialect, and for most of them, we can share the same implementation as we currently doing. It's like the folder pattern we are using in the parse phase.

But I do feel like this needs more discussions / designs. I used this trick as 1) I couldn't come up a good function name for this datetime settings; 2) this datetime usage seems limited to sqlite only, therefore, I feel it's better to explicitly to call it out.

prql-compiler/src/sql/dialect.rs
max-sixty: Great, thanks for the reference! TIL

max-sixty:
Thanks for the excellent PR! And for your patience @bcho 

The project would benefit a lot from more PRs like this...

PR #1971, opened 2023-02-27
feat: adjust panic to Java Exception

linux-china:
Now if PRQL compile failed, and Java app will exit from JNI call.   Adjust panic to Java Exception and add `throws Exception` to JNI methods.


prql-java/src/lib.rs
```diff
@@ -14,11 +14,23 @@ pub extern "system" fn Java_org_prql_prql4j_PrqlCompiler_toSql(
         .get_string(query)
         .expect("Couldn't get java string!")
         .into();
-    let rs_sql_str: String = prql_compiler::compile(&prql_query, &Options::default())
-        .expect("Couldn't compile query to prql!");
-    env.new_string(rs_sql_str)
-        .expect("Couldn't create java string!")
-        .into_raw()
+    let result = prql_compiler::compile(&prql_query, &Options::default());
+    return java_string_with_exception(result, &env);
```
max-sixty:
```suggestion
   java_string_with_exception(result, &env)
```

In rust we automatically return a final statement with no semicolon

max-sixty (approved):
This looks good!

A couple of linting issues, let me know if you have any questions on these.

Would be great to have a test that it throws an Exception (e.g. here: https://github.com/PRQL/prql/pull/1971/files#diff-03cf862c0ad084867ea3a0adce5550d1cc83d428a9f8c47101be139677df93eaL7). But no great stress, there are some bigger changes at https://github.com/PRQL/prql/pull/1929

prql-java/src/lib.rs
linux-china: Yes, fixed.   Java test method for compile error added too.

```
    @Test(expected = Exception.class)
    public void compileWithError() throws Exception {
       PrqlCompiler.toSql("from table | filter id >> 1");
    }
```

prql-java/java/src/test/java/org/prql/prql4j/PrqlCompilerTest.java
```diff
@@ -16,4 +16,9 @@ public void compile() {
                 "  table";
         assert expected.equalsIgnoreCase(found);
     }
+
+    @Test(expected = Exception.class)
+    public void compileWithError() throws Exception {
+       PrqlCompiler.toSql("from table | filter id >> 1");
+    }
```
max-sixty: Great test!

prql-java/src/lib.rs
max-sixty: Great! There are a couple more that the linter found. When we fix those let's merge

max-sixty:
@linux-china i see some commits, let me know if you need a hand with the clippy errors at all? More than happy to help.

linux-china:
> @linux-china i see some commits, let me know if you need a hand with the clippy errors at all? More than happy to help.

Fixed some problems reported by clippy, and you can have a check.

max-sixty:
Great re clippy issues. `cargo clippy --fix` will fix some of them in future

Now run `cargo fmt --all` for the formatting

prql-java/src/lib.rs
```diff
@@ -1,24 +1,64 @@
...
+use prql_compiler::sql::Dialect;
+use prql_compiler::{json, pl_to_prql, prql_to_pl, ErrorMessages, Options, Target};
 
 #[no_mangle]
 #[allow(non_snake_case)]
 pub extern "system" fn Java_org_prql_prql4j_PrqlCompiler_toSql(
     env: JNIEnv,
     _class: JClass,
     query: JString,
+    dialect: JString,
```
max-sixty: We should merge because this is an improvement, but this really should be `target`, and it should be matching against strings like `sql.bigquery`, rather than `bigquery`, from https://github.com/PRQL/prql/blob/main/book/src/language-features/target.md#L10-L11

linux-china: Fixed.

max-sixty:
Is this change OK? It uses the existing code in the compiler to evaluate the target https://github.com/PRQL/prql/pull/1971/commits/1b3f46fbb9f9affc86cd54cef5046bbb37f1bf26

linux-china:
Add Javadoc for toSql() to clear parameters.

max-sixty:
Thanks a lot @linux-china !

PR #1974, opened 2023-02-27
build: add very experimental MegaLinter config files

### LINKED ISSUE #1944: Use MegaLinter?
Since this repository contains files in quite a few different languages, setting up MegaLinter could provide many suggestions.
https://megalinter.io/latest/

There is some overlap in role with the pre-commit hook, though.

eitsupi:
Close #1944

- Automatic fixes are disabled to avoid conflicts with existing pre-commits.
- The linters for languages (Rust, JS, and Python) which are already configured (?) are disabled.
- stylelint is disabled because of long execution time.
- Configured to ignore Linter errors, which currently generate errors for all codes in this repo.


eitsupi:
@max-sixty I have configured it to ignore all linters errors that were causing errors.
Revert a665b2dd19bf6b16ba437d76b106a4e2e1582a43 and we can merge without generating errors.

Is that good enough for you?
Or is there a linters you would like to disable altogether or a setting you would like to add?
(Personally, I think it would be better to add them as needed after the merge than to do that in this PR. Unless this is merged, the number of codes that linters complain about will continue to grow.)

max-sixty:
It does actually find a couple of good results, so yes, let's try it. We can evaluate in a few weeks whether it's adding value.

I find the PR comment very long and noisy — I would vote to have that turned off, at least initially. I'm also not that keen on a sponsor message on every single PRQL PR without a really good reason.

I also think we should:
- Update the `development.md` with a sentence to say we're experimenting with this
- Add a comment somewhere in the config file saying this is experimental and that we're exploring using this more / open to changes / etc.
- Resolve it adding annotations — possibly these are correct and we should fix them, but we don't want these added to every PR until they're fixed — maybe we disable, open an issue to fix?
image


And then at some point:
- Look into turning on auto-fixes for formatters
- Move it to the `pull-request` workflow
- Decide whether it's adding value — are we using it to make changes to the code or is it just producing perma-warnings? My experience is that in order to be helpful, linters either need to fix the errors they find (e.g. formatters) or discriminate between important and unimportant errors. 
  - So my guess is that we might benefit instead from adding more pre-commit tasks, which have strict pass/fails, driven by the folks who are working on that language binding. But let's see...

eitsupi:
@max-sixty Thanks for your advice.

> I find the PR comment very long and noisy — I would vote to have that turned off

Done. (The comment left above is previous and will no longer be updated.)

> Update the development.md with a sentence to say we're experimenting with this

Done.

> Add a comment somewhere in the config file saying this is experimental and that we're exploring using this more / open to changes / etc.

Done.

> Resolve it adding annotations — possibly these are correct and we should fix them, but we don't want these added to every PR until they're fixed — maybe we disable, open an issue to fix?

I don't think this appears on all PRs.
I temporarily ran the workflow for all code in the project (a665b2dd19bf6b16ba437d76b106a4e2e1582a43), but usually only for files that have changed.
In other words, this will not show up in PRs that do not edit php files.

Does this answer your concern?

max-sixty:
Perfect! Looking forward to trying this.

Some things we could do from here (no urgency / obligations!):
- Turn on autofixes for formatters
- Turn on failures for the checks that we've fixed and are reliable (e.g. thanks for fixing #1988, we could try enforcing that if we trust it to have minimal false positives)

max-sixty (approved):

eitsupi:
Thanks, I merged this and created a issue about future follow up #1994

PR #1978, opened 2023-02-27
fix: Attempt to fix the chumsky break on MacOS

max-sixty:
Ref https://github.com/PRQL/prql/pull/1818

### DIFF
--- a/.github/workflows/test-all.yaml
+++ b/.github/workflows/test-all.yaml
@@ -28,10 +28,6 @@ jobs:
         os: [ubuntu-latest, windows-latest, macos-latest]
         target_option: ["", --target=wasm32-unknown-unknown]
 
-        # Combination of macos and wasm32 has a problem with crate psm, a dependency of chumsky
-        exclude:
-          - os: macos-latest
-            target_option: --target=wasm32-unknown-unknown
     uses: ./.github/workflows/test-rust.yaml
     with:
       os: ${{ matrix.os }}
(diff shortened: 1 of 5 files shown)


max-sixty:
OK, so I think this is the problem, and it's not just on `prql-js` — this sits locally on 100% CPU:

```sh
echo "Mississippi has four S’s and four I’s." | cargo run -p prqlc -- compile
```

max-sixty:
One very confusing thing:

`resolver = "2"` is required for this to pass, but is also supposed to be the default? https://doc.rust-lang.org/cargo/reference/resolver.html#resolver-versions:

> The version "1" resolver is the original resolver that shipped with Cargo up to version 1.50. The default is "2" if the root package specifies edition = "2021" or a newer edition. Otherwise the default is "1".

max-sixty:
We can try and merge #1982, and then hopefully this will pass (without the commented out test)

max-sixty:
Given that this is actually quite isolated — I just added a three character repro at https://github.com/PRQL/prql/pull/1982#issuecomment-1447476157 — I think we're in a place where:
- We can likely fix this in the short term
- The aggregate view is that we shouldn't revert
- We shouldn't release until it's fixed

So I'll change the standard JS test to exclude the unicode characters, merge this, and then aim to fix this issue soon.

---

Zooming out, as discussed on Discord, I think we should really try and avoid being in a place where the current `main` is broken / isn't possible to release / isn't where people should start contributing from. Instead, we should be treating it like a shared resource whose breaks impact lots of people and are difficult to oversee.

There are some differences of opinion though! And I understand that my view is dependent on the number of contributors, and I have ambitions for more contributors than we currently have — so hence maybe the different conclusions.

I'm happy to find a compromise, and I'm sure over time we'll find a synthesis of our views...

PR #1980, opened 2023-02-28
test: repro timestamp parsing issue

bcho:
image

This is the output I got from 0.5.2

### DIFF
--- a/prql-compiler/src/test.rs
+++ b/prql-compiler/src/test.rs
@@ -3320,3 +3320,20 @@ fn test_params() {
     "###
     );
 }
+
+#[test]
+fn test_datetime_parsing() {
+    assert_display_snapshot!(compile(r#"
+    from test_tables
+    select [date = @2022-12-31, time = @08:30, timestamp = @2020-01-01T13:19:55-0800]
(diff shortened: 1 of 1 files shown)


max-sixty:
Great repro, thanks.

Depending on what we do with https://github.com/PRQL/prql/pull/1978, I'll look at fixing this

prql-compiler/src/test.rs
```diff
@@ -3320,3 +3320,20 @@ fn test_params() {
     "###
     );
 }
+
+#[test]
```
max-sixty, line 3325:
```suggestion
// TODO: fix this based on https://github.com/PRQL/prql/pull/1818
#[test]
#[should_panic]
```

Then we can merge

prql-compiler/src/test.rs
```diff
@@ -3320,3 +3320,21 @@ fn test_params() {
...
```
max-sixty, line 3325: Are we OK to include a reference to the issue?

```suggestion
// TODO: fix this based on https://github.com/PRQL/prql/pull/1818
#[test]
```

PR #1982, opened 2023-02-28
test: Add test for query which blocks on chumsky

max-sixty:
Based on #1978.

It will block tests, probably until timeout

### DIFF
--- a/prql-compiler/src/parser/mod.rs
+++ b/prql-compiler/src/parser/mod.rs
@@ -54,7 +54,13 @@ pub fn parse(source: &str) -> Result<Vec<Stmt>> {
 }
 
 fn convert_lexer_error(source: &str, e: Cheap<char>) -> Error {
-    let found = source[e.span()].to_string();
+    // TODO: is there a neater way of taking a span? We want to take it based on
+    // the chars, not the bytes, so can't just index into the str.
+    let found = source
+        .chars()
+        .skip(e.span().start)
+        .take(e.span().end() - e.span().start)
+        .collect();
     let span = common::into_span(e.span());
 
     Error::new(Reason::Unexpected { found }).with_span(span)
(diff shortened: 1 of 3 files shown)


max-sixty:
I have to pause on this now, but I isolated this down to lexing a very specific string: `’s `. 

Note that that's not a normal apostrophe — copy it from the text or the minimal test in the PR if you're copying it.

max-sixty:
OK, so I think this was caused by https://github.com/zesterer/chumsky/issues/301, and is now fixed in PRQL and tested.

We lose some error recovery, but I'm not sure there's any alternative.

max-sixty:
I'll merge, but hit me with feedback and I can look tomorrow.

prql-compiler/src/parser/mod.rs
```diff
@@ -54,7 +54,13 @@ pub fn parse(source: &str) -> Result<Vec<Stmt>> {
...
 
 fn convert_lexer_error(source: &str, e: Cheap<char>) -> Error {
-    let found = source[e.span()].to_string();
+    // TODO: is there a neater way of taking a span? We want to take it based on
+    // the chars, not the bytes, so can't just index into the str.
+    let found = source
+        .chars()
+        .skip(e.span().start)
+        .take(e.span().end() - e.span().start)
+        .collect();
```
aljazerzen, line 63: I believed that indexing str was indexing chars, not bytes! Huh, this is unexpected.

```
echo 'fn main() { let a: &str = &"😙"[1..2]; println!("{a}"); }' > a.rs && rustc a.rs && ./a
thread 'main' panicked at 'byte index 1 is not a char boundary; it is inside '😙' (bytes 0..4) of `😙`', a.rs:1:28
```

Well, after a quick look around, what you wrote seems like the best approach. Sadly.

zesterer: Using char indices instead of bytes for the default impl of `From<&str> for Stream` was an API misstep that has been corrected in `zero-copy`.

zesterer:
> OK, so I think this was caused by [zesterer/chumsky#301](https://github.com/zesterer/chumsky/issues/301), and is now fixed in PRQL and tested.
> 
> We lose some error recovery, but I'm not sure there's any alternative.

FWIW I've just fixed this and released `0.9.2` with the fix. Apologies for this, a very subtle edge case!

PR #1991, opened 2023-03-01
fix: Allow no `:` in timezones

max-sixty:
Fixes issue in #1818


max-sixty:
This reverts to the previous behavior, but explicitly, rather than just passing any colons and digits through. 

Both with a `:` and without are supported by https://en.wikipedia.org/wiki/ISO_8601

Thanks to @bcho for finding the issue. We can now merge #1970

PR #1992, opened 2023-03-01
internal: Use `Display` for `Error::Display`

max-sixty:


max-sixty:
I did this while trying to figure out why the errors were difficult to read while writing tests. 

But actually does this make any sense? I'm almost surprised it doesn't recurse on itself?

prql-compiler/src/error.rs
```diff
@@ -115,14 +115,14 @@ impl StdError for Errors {}
 // Needed for StdError
 impl Display for Error {
     fn fmt(&self, f: &mut Formatter<'_>) -> fmt::Result {
-        Debug::fmt(&self, f)
+        Display::fmt(&self, f)
```
aljazerzen, line 118: This will recurse on itself. But it was probably never called, because it is converted to ErrorMessage, before it is displayed.

PR #1993, opened 2023-03-01
feat: error code

aljazerzen:
Ref https://github.com/PRQL/prql-vscode/issues/128

### DIFF
--- a/prql-compiler/src/error.rs
+++ b/prql-compiler/src/error.rs
@@ -17,6 +17,7 @@ pub struct Error {
     pub span: Option<Span>,
     pub reason: Reason,
     pub help: Option<String>,
+    pub code: Option<&'static str>,
 }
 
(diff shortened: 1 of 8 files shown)


max-sixty:
Looked great!

Would it be more tractable to have the codes as short strings, like `"no-main-pipeline"` rather than `E0034`?

That the approach of clippy / mypy; I find it both easy to understand as a person, and simple of parse for tools.

(We could even define them in an Enum, but no need to, especially now)

aljazerzen:
I don't think readability is the concern here. Code is something short, easy to copy and compare. Something that should **never** change.

And this format was plagiarized from Rust:
```
error[E0308]: `?` operator has incompatible types
  --> prql-compiler/src/semantic/lowering.rs:48:19
   |
48 |
```

max-sixty:
Great!

PR #1995, opened 2023-03-01
feat: validate Options.target and allow sql.any

### LINKED ISSUE #1907: Special string target expression to fall back to the target in query header?
Related to #1855 and PRQL/prql-vscode#98

Wondering if it is worth implementing a special string expression for not specifying a target in the compiler options.
For example, I assume a name like `"sql.none"` or simply `"none"`.

This is converted to `Target::Sql(None)` or `Target::default()` (now `Target::Sql(None)`) at the following location.

https://github.com/PRQL/prql/blob/c6c73ef3631b414730cc14aa1fa4130183f6519f/prql-compiler/src/lib.rs#L153-L172

aljazerzen:
Closes #1907

### DIFF
--- a/prql-compiler/src/lib.rs
+++ b/prql-compiler/src/lib.rs
@@ -142,10 +142,12 @@ impl Default for Target {
 
 impl Target {
     pub fn names() -> Vec<String> {
-        sql::Dialect::names()
-            .into_iter()
-            .map(|d| format!("sql.{d}"))
-            .collect()
+        let mut names = vec!["sql.any".to_string()];
+
+        let dialects = sql::Dialect::names();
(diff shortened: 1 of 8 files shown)


eitsupi (approved):
Thank you for implementing this!

What do you think about including `sql.any` in `Target::names()` so that we can easily verify that `sql.any` exists as an available target?
https://github.com/PRQL/prql/blob/42112a5a760e839d93a5fc2c605c221b9b898d74/prql-compiler/src/lib.rs#L143-L150

aljazerzen:
Yup, good catch.

max-sixty:
Thanks for chasing down all the bindings!

PR #2002, opened 2023-03-02
feat: Improve lexer error recovery

max-sixty:
Upgrade to Chumsky 0.9.2 and add back recovery

### DIFF
--- a/prql-compiler/src/test.rs
+++ b/prql-compiler/src/test.rs
@@ -2954,6 +2954,20 @@ fn test_errors() {
        ·                       ┬
        ·                       ╰── unexpected ’
     ───╯
+    Error:
+       ╭─[:1:36]
+       │
+     1 │ Mississippi has four S’s and four I’s.
+       ·                                    ┬
+       ·                                    ╰── unexpected ’
+    ───╯
+    Error:
+       ╭─[:1:39]
+       │
+     1 │ Mississippi has four S’s and four I’s.
+       ·                                       ┬
+       ·                                       ╰── identifier expected an identifier or *, but found
+    ───╯
     "###);
 
     let err = compile(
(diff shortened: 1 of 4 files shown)


prql-compiler/src/test.rs
```diff
@@ -2954,6 +2954,20 @@ fn test_errors() {
...
+     1 │ Mississippi has four S’s and four I’s.
+       ·                                    ┬
+       ·                                    ╰── unexpected ’
+    ───╯
+    Error:
+       ╭─[:1:39]
+       │
+     1 │ Mississippi has four S’s and four I’s.
+       ·                                       ┬
+       ·                                       ╰── identifier expected an identifier or *, but found
```
max-sixty: It's not great prose, but we can iterate...

max-sixty:
Unfortunately it has volatile outputs...

Is this expected? Can we make error recovery consistent?

aljazerzen:
What do you mean "volatile"? Non deterministic, so different every run?

I'm guessing you are talking about order of expected tokens. We can sort that before displaying.

prql-compiler/src/test.rs
aljazerzen: I did not know that at the time, but expected token `None` means end of file. In the convert_parser_error I'm converting that `None` into empty string, which causes this output.

max-sixty:
Yes exactly, non-deterministic is a better term.

Ok, great if we construct the message ourselves, easy for us to sort the items, I can have a look

aljazerzen:
https://github.com/PRQL/prql/blob/main/prql-compiler/src/parser/mod.rs#L69-L116

prql-compiler/src/test.rs
max-sixty: Great, fixed in #2008, thanks

PR #2008, opened 2023-03-03
feat: Improve error messages on EOI

max-sixty:
Upgrade to Chumsky 0.9.2 and add back recovery


prql-compiler/src/test.rs
```diff
@@ -2926,7 +2926,7 @@ fn test_errors() {
        │
      1 │ Mississippi has four S’s and four I’s.
        ·                                       ┬
-       ·                                       ╰── identifier expected * or an identifier, but found
+       ·                                       ╰── identifier expected * or an identifier, but found end of input
```
max-sixty, line 2929: I don't love "..but found end of input" — this is not how people talk — but I'm not sure what's better. 

Maybe "...but didn't find them before reaching the end of the query" (though we'd have to change the way the message is constructed)

We can still merge this, there's lots of lower-hanging fruit on error messages

aljazerzen: If you want that, you can add a special case for when `found == None` and construct `Reason::simple("")`.

max-sixty: Great, done in https://github.com/PRQL/prql/pull/2012

PR #2015, opened 2023-03-05
docs: Add error messages to the book

max-sixty:
This allows us to show and test error messages in the book, and adds an initial example.

Unfortunately the initial example doesn't have a great error message! So that's something we could work on.

It also updates the docs to add late binding — now functions support that!


max-sixty:
The size of the diff is unfortunately because of our enumerating of snapshots. Now we delete unreferenced snapshots, we could change this to be a hash, so adding one example didn't cause a big diff.

PR #2017, opened 2023-03-05
docs: update changelog about sql.any

eitsupi:
The target priority (option of the compile function v.s. query header) and `"sql.any"` behavior is not fully explained in the book and prql-compiler documentation, so I think the documentation needs to be improved, but for now, only the changelog update is included in this PR.


max-sixty:
Thanks @eitsupi !

max-sixty:
#2023 for a typo fyi

PR #2019, opened 2023-03-05
feat: upper and lower function

### LINKED ISSUE #1793: Implement uppercase and lowercase string transform
Implement a way to transform a string to lowercase and a string to uppercase.

This can be done by calling the `upper()` and `lower()` functions in SQL.

Jelenkee:
Fixes #1793


aljazerzen:
This is great! Even the `func interp` now uses better naming (in my opinion)!

max-sixty:
Agree! Thanks @Jelenkee !

PR #2025, opened 2023-03-06
build: add definitions for devcontainer base image

eitsupi:
Split from #1893
Push a Docker image with the time-consuming cargo packages installed to GHCR.

Changes are also included to separate only the `install-cargo-tools` task in the Taskfile referenced in the Dockerfile into a separate Taskfile, from a Dokcer cache perspective.

### DIFF
--- a/.github/workflows/devcontainer.yaml
+++ b/.github/workflows/devcontainer.yaml
@@ -1,0 +1,63 @@
+name: devcontainer
+
+on:
+  push:
+    # ToDo: Add partial Taskfile (only include rust-tools)
+    paths:
+      - .devcontainer/base-image/Dockerfile
+  pull_request:
+    paths:
+      - .devcontainer/base-image/Dockerfile
+  workflow_dispatch:
+
+concurrency:
+  group: ${{ github.workflow }}
+  cancel-in-progress: true
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: docker/metadata-action@v4
+        id: meta
+        with:
+          images: ghcr.io/${{ github.repository_owner }}/prql-devcontainer-base
+          tags: |
+            type=raw,latest
+
+      - name: Login to GitHub Container Registry
+        uses: docker/login-action@v2
+        with:
+          registry: ghcr.io
+          username: ${{ github.repository_owner }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Set up QEMU
+        uses: docker/setup-qemu-action@v2
+
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v2
+
+      - id: set-platforms
+        run: |
+          if [ "${{ github.ref }}" = "refs/heads/main" ]; then
+            echo "platforms=linux/amd64,linux/arm64" >>$GITHUB_OUTPUT
+            echo "push=true" >>$GITHUB_OUTPUT
+          else
+            echo "platforms=linux/amd64" >>$GITHUB_OUTPUT
+            echo "push=false" >>$GITHUB_OUTPUT
+          fi
+
+      - name: Build
+        uses: docker/build-push-action@v4
+        with:
+          file: "./.devcontainer/base-image/Dockerfile"
+          tags: ${{ steps.meta.outputs.tags }}
+          platforms: ${{ steps.set-platforms.outputs.platforms }}
+          push: ${{ steps.set-platforms.outputs.push }}
+          cache-from: |
+            ${{ env.IMAGE_NAME }}
+            type=gha
+          cache-to: |
+            type=inline
+            type=gha,mode=max
--- a/.github/workflows/devcontainer.yaml
+++ b/.github/workflows/devcontainer.yaml
@@ -2,12 +2,13 @@ name: devcontainer
 
 on:
   push:
-    # ToDo: Add partial Taskfile (only include rust-tools)
     paths:
       - .devcontainer/base-image/Dockerfile
+      - Taskfile.cargo-tools.yml
   pull_request:
     paths:
       - .devcontainer/base-image/Dockerfile
+      - Taskfile.cargo-tools.yml
   workflow_dispatch:
 
 concurrency:
--- a/Taskfile.cargo-tools.yml
+++ b/Taskfile.cargo-tools.yml
@@ -1,0 +1,19 @@
+# yaml-language-server: $schema=https://json.schemastore.org/taskfile.json
+
+version: 3
+
+tasks:
+  install:
+    # factored this out because it takes a long time to build
+    desc: Install cargo tools for PRQL development.
+    cmds:
+      # `--locked` installs from the underlying lock files (which is not the
+      # default?!)
+      - cargo install --locked bacon cargo-audit cargo-insta cargo-release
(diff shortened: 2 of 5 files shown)


eitsupi:
@max-sixty Since the initial (multi-platform) build will take several hours, sorry but I will merge this now without your review.
As always, please revert this if we have problems.

max-sixty:
> I will merge this now without your review.

Always feel free!

Taskfile.cargo-tools.yml
```diff
@@ -0,0 +1,19 @@
+# yaml-language-server: $schema=https://json.schemastore.org/taskfile.json
+
+version: 3
+
+tasks:
```
max-sixty, line 5: No stress at all, but out of interest — any reason to split this out into a different file?

max-sixty:
ICYMI: https://github.com/PRQL/prql/actions/runs/4340298602

Taskfile.cargo-tools.yml
eitsupi: Each time the `Taskfile.yml` (Files copied into the container at build time) is modified (even a mere modification of a comment, for example), the later layers' cache is invalidated.
By copying files that are less likely to be modified, I believe we can make better use of the cache.

eitsupi:
> ICYMI: [PRQL/prql/actions/runs/4340298602](https://github.com/PRQL/prql/actions/runs/4340298602)

Sorry for failing the workflow. I will investigate and fix it......

max-sixty:
> Sorry for failing the workflow. I will investigate and fix it......

Zero stress, keep merging until it works. It ran on this PR, so possibly it was just flaky on that run?

eitsupi:
> Zero stress, keep merging until it works. It ran on this PR, so possibly it was just flaky on that run?

Sorry, I was completely mistaken about the base image. To my surprise, the base image does not support the arm64 platform!

```sh
$ docker buildx imagetools inspect mcr.microsoft.com/devcontainers/rust:0.203.7-1
Name:      mcr.microsoft.com/devcontainers/rust:0.203.7-1
MediaType: application/vnd.oci.image.index.v1+json
Digest:    sha256:427f31e3c81cf8189e4a4253c332dac381fe2e0646569634e57d1781d65122c3
           
Manifests: 
  Name:        mcr.microsoft.com/devcontainers/rust:0.203.7-1@sha256:f8a481ea7a30621a1ba07c16fac67fd330f6a25d6d064f8086de70f1f1553264
  MediaType:   application/vnd.oci.image.manifest.v1+json
  Platform:    linux/amd64
               
  Name:        mcr.microsoft.com/devcontainers/rust:0.203.7-1@sha256:62645b8a13342c442cf8c0cdf877f6dc699edb3338ab80e8236ae81674bbcb5f
  MediaType:   application/vnd.oci.image.manifest.v1+json
  Platform:    unknown/unknown
  Annotations: 
    vnd.docker.reference.digest: sha256:f8a481ea7a30621a1ba07c16fac67fd330f6a25d6d064f8086de70f1f1553264
    vnd.docker.reference.type:   attestation-manifest
```

Taskfile.cargo-tools.yml
max-sixty: OK. But the Taskfile doesn't change often — I would lean towards having the repo simpler even if it means running too many workflows. If we did this for everything we'd have many many tiny files and it would be difficult to work with the repo.

If we had a full build system like Bazel, then we could do this, but it's very approximate at best, and so somewhat fruitless to optimize it...

max-sixty:
Oh, weird! Do none of the standard Microsoft VSCode images?

eitsupi:
I was completely mistaken, but it seems that the default variant is still `buster` and the `bullseye` needs to be explicitly stated.
https://github.com/devcontainers/images/blob/c087e361571e2200c3b40ed86d14a8ced84c388a/src/rust/README.md?plain=1#L13

```sh
$ docker buildx imagetools inspect mcr.microsoft.com/devcontainers/rust:0.203.7-1-bullseye
Name:      mcr.microsoft.com/devcontainers/rust:0.203.7-1-bullseye
MediaType: application/vnd.oci.image.index.v1+json
Digest:    sha256:7eb80e1054ba35dd4865e1fb9d9c27f7401b6ba6e406d7fc986e9f7ff9507741
           
Manifests: 
  Name:        mcr.microsoft.com/devcontainers/rust:0.203.7-1-bullseye@sha256:8e0fc0744f644ea9f64c0cebcc3ee87a77035b3af99ddb874bd910541eb5f9e9
  MediaType:   application/vnd.oci.image.manifest.v1+json
  Platform:    linux/amd64
               
  Name:        mcr.microsoft.com/devcontainers/rust:0.203.7-1-bullseye@sha256:21b4d0c618a3dd24b1a37a38ef2cbc150c6dc038b72edbed5227d95780de5661
  MediaType:   application/vnd.oci.image.manifest.v1+json
  Platform:    linux/arm64
               
  Name:        mcr.microsoft.com/devcontainers/rust:0.203.7-1-bullseye@sha256:5b2ef13f21aff24c6999c5d9a2d82c8785ef658b15271bcdf659a95a049deebf
  MediaType:   application/vnd.oci.image.manifest.v1+json
  Platform:    unknown/unknown
  Annotations: 
    vnd.docker.reference.digest: sha256:8e0fc0744f644ea9f64c0cebcc3ee87a77035b3af99ddb874bd910541eb5f9e9
    vnd.docker.reference.type:   attestation-manifest
               
  Name:        mcr.microsoft.com/devcontainers/rust:0.203.7-1-bullseye@sha256:feffae93572a62643ede3aac2130304e3193a50ae7c5f28cd07bfd0277518375
  MediaType:   application/vnd.oci.image.manifest.v1+json
  Platform:    unknown/unknown
  Annotations: 
    vnd.docker.reference.digest: sha256:21b4d0c618a3dd24b1a37a38ef2cbc150c6dc038b72edbed5227d95780de5661
    vnd.docker.reference.type:   attestation-manifest
```

eitsupi:
> Oh, weird! Do none of the standard Microsoft VSCode images?

If I remember correctly, this was due to a bug in ~~OpenSSH~~ OpenSSL, and the buster and `ubuntu:focal` arm64 images were not built.
https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=989604

I mistakenly thought that the default was already moved to `bullseye` since it has been so long since `bullseye` was released...

.github/workflows/devcontainer.yaml
```diff
@@ -0,0 +1,64 @@
...
+
+      - id: set-platforms
+        run: |
+          if [ "${{ github.ref }}" = "refs/heads/main" ]; then
+            echo "platforms=linux/amd64,linux/arm64" >>$GITHUB_OUTPUT
+            echo "push=true" >>$GITHUB_OUTPUT
+          else
+            echo "platforms=linux/amd64" >>$GITHUB_OUTPUT
+            echo "push=false" >>$GITHUB_OUTPUT
+          fi
```
max-sixty, line 50: @eitsupi On reflection it's probably simpler to run both architectures on tests, not just amd64. Again — we should be liberal with workflows...

eitsupi: Ok, I will fix this.
There is a disadvantage that multi-platform builds can't load images, so we can't run tests on containers, etc., but since we are only doing builds at this point, it shouldn't be a problem.

PR #2040, opened 2023-03-07
feat!: Re-revert switch syntax change

max-sixty:
Re-reverts https://github.com/PRQL/prql/pull/1587 after a lengthy discussion

This is WIP and tests will fail


max-sixty:
Did we already change the other grammars (e.g. playground, vscode) or was that still a TODO? I couldn't see it in the commits. No stress if not; want to ensure I'm not double-doing things though.

max-sixty:
I changed the grammars in this repo. Is there anything else?

We need to update the textmate & tree-sitter grammars, but they're in separate repos.

I'll merge now but lmk anything else.

PR #2043, opened 2023-03-07
build: Publish prqlc to Microsoft Store?

vanillajonathan:
This uses the [Microsoft Store Publish GitHub Action](https://github.com/marketplace/actions/microsoft-store-publish). A Windows MSI installer is created created using [another GitHub Action](https://github.com/marketplace/actions/build-windows-installer-msi-from-exe-file).

This requires a Microsoft store developer account:
https://developer.microsoft.com/en-us/microsoft-store/register/
> Individual accounts cost approximately $19 USD, and company accounts cost approximately $99 USD.

> The Microsoft Store on Windows is providing free accounts to female identifying developers who want to publish an app.

### DIFF
--- a/.github/workflows/publish-ms-store.yaml
+++ b/.github/workflows/publish-ms-store.yaml
@@ -5,7 +5,7 @@ on:
     types: [released]
   pull_request:
     paths:
-      - .github/workflows/publish-microsoft-store.yaml
+      - .github/workflows/publish-ms-store.yaml
   workflow_call:
 
 jobs:
--- a/.github/workflows/publish-ms-store.yaml
+++ b/.github/workflows/publish-ms-store.yaml
@@ -11,7 +11,7 @@ on:
 jobs:
   microsoft_store:
     name: Publish Microsoft Store
-    runs-on: ubuntu-latest
+    runs-on: windows-latest
     steps:
       - name: 📂 Checkout code
         uses: actions/checkout@v3
--- a/.github/workflows/publish-ms-store.yaml
+++ b/.github/workflows/publish-ms-store.yaml
@@ -1,0 +1,60 @@
+name: publish-microsoft-store
+
+on:
+  release:
+    types: [released]
+  pull_request:
+    paths:
+      - .github/workflows/publish-microsoft-store.yaml
+  workflow_call:
+
+jobs:
+  microsoft_store:
+    name: Publish Microsoft Store
+    runs-on: ubuntu-latest
+    steps:
+      - name: 📂 Checkout code
+        uses: actions/checkout@v3
+      - uses: actions-rs/toolchain@v1
+        with:
+          toolchain: stable
+      - name: 👷 Build prql-compiler
+        run: cargo build --release
+      - name: 📁 Copy files into .msi package
+        run: |
+          mkdir application
+          cp target/release/prqlc application/prqlc
+          mv prql-compile/prqlc/wix.json .
+      - name: 👷 Build Windows Installer MSI from exe file
+        uses: AliceOh/CreateWindowsInstaller@1.0.0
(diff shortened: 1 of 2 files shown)


max-sixty:
Is this where folks are likely to install CLI apps from?

How about https://learn.microsoft.com/en-us/windows/package-manager/winget/? That seems like the Windows version of brew?

vanillajonathan:
Yes, there is winget and [chocolatey](https://chocolatey.org/) ([create packages](https://docs.chocolatey.org/en-us/create/create-packages)). We should look into that. for chocolatey we need to create a `.nuspec` file.
For that I think we need have `prqlc.exe` available over HTTP, either hosted on the PRQL website or as a release artifact on GitHub like:

    https://github.com/PRQL/prql/releases/download/v1.0.0/prqlc.exe

And also have prql-query (pq) on winget and/or chocolatey.

max-sixty:
> For that I think we need have `prqlc.exe` available over HTTP, either hosted on the PRQL website or as a release artifact on GitHub like:

We can do this as a release artifact. I think that would be quite useful more generally.

I think winget (or chocolatey, I'm less familiar with) would be better than Windows Store — which seems much less relevant for CLI apps. So I would vote to probably not proceed with this, unless you strongly disagree.

eitsupi:
winget can also install MS Store apps.
For example, a search for DBeaver will show two sources.

```powershell
❯ winget search dbeaver
名前       ID              バージョン ソース
----------------------------------------------
DBeaver CE 9PNKDR50694P    Unknown    msstore
DBeaver    dbeaver.dbeaver 23.0.0     winget
```

That said, since winget will soon support single binary installations[^2], I recommend simply attaching the exe file to the GitHub release and registering that in the repository[^1].

[^1]: https://github.com/microsoft/winget-pkgs
[^2]: https://github.com/microsoft/winget-cli/releases/tag/v1.4.10173

max-sixty:
@vanillajonathan I see some changes to the PR — just to confirm my message before because I don't want you to lose any time. I don't think we could pursue Winget or Chocolatey rather than Microsoft Store — the former seems much more used by developers / CLI apps.

Do you agree / does that make sense?

vanillajonathan:
The benefit of Microsoft store is that it is installed and readily available out-of-the-box on Windows. But I think it would be better to pursue Winget or Chocolatey than Microsoft Store because it cost a fee to put software on Microsoft Store and none of us want to pay that.

max-sixty:
OK great. Feel free to repurpose this PR, or close and open a new one

vanillajonathan:
Is there any interest in eventually having prqlc on the Microsoft Store longer down the road after we get it on Chocolatey and/or winget?

eitsupi:
Releasing in the MS Store has the advantage of being updated automatically without the user having to do anything.

Nevertheless, isn't it too late to release prqlc in the MS Store only after it has become popular enough and users have requested it?

vanillajonathan:
Yes, that is an advantage. Yes can wait with and do it later if there is any interest in having prqlc on Microsoft Store. Should I close this PR or leave it open?

max-sixty:
> Is there any interest in eventually having prqlc on the Microsoft Store longer down the road after we get it on Chocolatey and/or winget?

If there are folks who look for CLI apps on the Microsoft Store, then yes, eventually. (But I'm not sure whether there are...)

> Should I close this PR or leave it open?

I'll add a label to mark it as postponed but not canceled, and then we can do Winget on a different one?

max-sixty:
> Is there any interest in eventually having prqlc on the Microsoft Store longer down the road after we get it on Chocolatey and/or winget?

vanillajonathan:
Sounds good!

Issue #2044, opened 2023-03-07
`Components of PRQL` out of date


max-sixty:
[Components of PRQL](https://prql-lang.org/book/contributing/development.html#components-of-prql) is out of date now. I think probably we slim this down to the few categories in https://github.com/PRQL/prql/issues/2029, and individual items such as `prql-X` become implicit.

More broadly — docs are excellent, but we should only add docs that we can keep up-to-date; where the costs of keeping them up-to-date and worth the tax on development velocity (i.e. we can't merge the code until the docs are updated).

I'm also *very* open to ideas for synthesizing for a better tradeoff. Concepts like doctests are both great documentation and very maintainable — if there are more, then that's great.

PR #2046, opened 2023-03-07
docs: Move task docs out of devcontainers

max-sixty:
These are good docs, but there's no reason for them to be here. They'd be good in the Taskfile, or very open to other suggestions

While I _really_ appreciate docs that we write (CC @richb-hanover ), it's important that they're focused, concise, and close to the code that they document -- because they also need to be maintained, and the project takes on that responsibility.

I really don't want to lose folks' generosity and ideas, but I'm going to start being a bit firmer on these sorts of things, because we're already starting to see some stale docs (#2044). There are other ways of writing things that have fewer guarantees of continued support -- blog posts, gists, etc.


richb-hanover:
> There are other ways of writing things that have fewer guarantees of continued support -- blog posts, gists, etc.

Dang. You beat me. :-) As part of https://github.com/PRQL/prql/pull/1893#discussion_r1128635211, I was about to suggest that we need a new "newcomer's description" and that it should be a blog posting since that doesn't carry the "full weight" of needing to be true for all time. I'll look at reviving my original document as a blog post

PR #2048, opened 2023-03-08
chore: Redirect case.html

max-sixty:
This will fix the tests in https://github.com/PRQL/prql/pull/2040


max-sixty:
Sorry, this was a terrible git mistake. I'm reverting the `web` branch and fixing.

max-sixty:
OK I applied only the correct diff to `web` and pushed manually. It's correct that the redirect won't be on `main`, and so will get blasted away when we force push `main` when we release

aljazerzen:
I git pinged from the commit message. Is using the backport bot really easier than a push to web branch?

max-sixty:
I think you get pinged because it has every commit in it, because I messed up the merge.

> Is using the backport bot really easier than a push to web branch?

Yes — much!

PR #2050, opened 2023-03-08
docs: Add some guidance on writing docs

max-sixty:


max-sixty:
I'll leave this since we're releasing soon

PR #2062, opened 2023-03-09
docs: Add a new policy for bindings support

max-sixty:
As discussed on Discord:
- I / we are very grateful for the contributions of folks such as @vanillajonathan & @linux-china & @kasvith for adding bindings to PRQL
- We're currently in a place where upgrading `prql-lib` is quite difficult, and @aljazerzen has spent lots of time on this in the last few days, because some of the bindings are initial early efforts
- It's great to have these initial early efforts — everything starts out small before it becomes big
- But we shouldn't gate the changes to PRQL on the compatibility of these yet, because it slows down PRQL development

So this establishes three tiers of support, and proposes to only gate PRQL changes on bindings that have sufficient support and tooling.

When we have more support & tooling, then we can promote these to "Supported"

Does this make sense to folks?

### DIFF
--- a/book/src/bindings/README.md
+++ b/book/src/bindings/README.md
@@ -36,6 +36,7 @@ nascent.
 
 - [Java](./java.md)
 - [Elixir](./elixir.md)
+- `prql-lib`, the C bindings
 
 ## Nascent
 
--- a/book/src/bindings/README.md
+++ b/book/src/bindings/README.md
@@ -2,11 +2,44 @@
 
 PRQL has bindings for many languages. These include:
 
-- [Java](./java.md)
+We have three tiers of bindings:
+
+- Supported
+- Unsupported
+- Nascent
+
+## Supported
+
+Supported bindings require:
+
+- A maintainer.
+- Implementations of the
+  [core compile functions](https://docs.rs/prql-compiler/latest/prql_compiler/#functions).
+- Test coverage for these functions.
+- A published package to the language's standard package repository.
+- A script in `Taskfile.yml` to bootstrap a development environment.
+- Any dev tools, such as a linter & formatter, in pre-commit or MegaLinter.
+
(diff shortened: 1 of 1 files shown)


vanillajonathan:
Sure, as long as there is al least one working binding so that we can can know that it is possible to write a binding against `prql-lib`. So with the Python and JavaScript bindings listed as supported, then as long as those bindings get updated upon any changes to `prql-lib` then that is good because then hopefully it should be possible to get the other bindings to work too.

aljazerzen (approved):
Nice writeup

max-sixty:
> So with the Python and JavaScript bindings listed as supported, then as long as those bindings get updated upon any changes to `prql-lib`

Good point. These don't rely on `prql-lib` — they use rust libraries to facilitate the bindings.

Probably we should have a support level for `prql-lib` there too. It's probably "Unsupported" but fairly close to "Supported"; do you agree @aljazerzen ?

vanillajonathan:
I kind of would like one binding that uses FFI to be supported. Getting FFI done can be tricky, so if I see it has been done for at least one binding then it gives me faith that with enough trial-and-error I might somehow get it to work.

max-sixty:
> I kind of would like one binding that uses FFI to be supported. Getting FFI done can be tricky, so if I see it has been done for at least one binding then it gives me faith that with enough trial-and-error I might somehow get it to work.

I 100% empathize with this. 

I think `prql-java` is the furthest along. If @zhicwu were up for merging his code, that would get us all the code we needed. Then I'm not sure whether @zhicwu or @linux-china would be up for maintaining it...

aljazerzen:
`prql-lib` does not have a build process established - I don't even know how it should be packaged, so I'd leave it at "unsupported".

For FFI example impl, look at php bindings, which are now at the level where all the functionality is implemented. But it also does not have a build process.

Issue #2060, opened 2023-03-08
Remove redundant code snippets in book


Jelenkee:
Currently the PRQL-snippets exist in the markdown files and in  `book/tests/prql`.  That is not good for maintenance of these snippets.

A solution could be to use a "macro" in the markdown files to import them at build time.
Or you parse the markdown files and extract the snippets when running the tests.

I would prefer the first solution

max-sixty:
Hi @Jelenkee — these are auto-generated by https://github.com/PRQL/prql/blob/60b7b0750c1c7095c2c21cdb55cdc8e23724526a/book/tests/snapshot.rs#L44, and get run with `cargo test`

Does that make sense? It sounds a bit like your suggested solution.

Possibly we should document this better? There's a bullet "Examples" at https://github.com/PRQL/prql/blob/60b7b0750c1c7095c2c21cdb55cdc8e23724526a/book/src/contributing/development.md#how-we-test, but it's kinda far from the code.

Jelenkee:
I see. And you didn't add those files to gitignore because the tests run faster, did you?

max-sixty:
Good question...

We currently split up the extractions from the testing (which is at https://github.com/PRQL/prql/blob/9d0e6fdc37c3075c6623f0b21d8f1e4ce79ae15e/book/tests/snapshot.rs#L188), and the testing reads the files.

Though this isn't strictly necessary, we could avoid writing the PRQL files and just extract them each time we wanted them. There's no performance benefit.

Two points:
- We want to test our autoformatting when that's further along, and writing the autoformatted examples is a good way. We could try and write these _into_ the markdown, but that's more complicated.
- One caveat is that the extraction has slightly different behavior on Windows, and we rely on the files for running the Windows tests, https://github.com/PRQL/prql/issues/356. So we'd need to write a function to do that, which requires someone with a Windows machine and an hour.

Jelenkee:
* Autoformatting sounds good. But is it really that complicated to inject it into markdown? You already add the transpiled SQL to the markdown. Seems not had to me, provided that a function to format prql already exists.
* I will look into that Windows problem later

max-sixty:
> Autoformatting sounds good. But is it really that complicated to inject it into markdown? You already add the transpiled SQL to the markdown. Seems not had to me, provided that a function to format prql already exists.

[...]

Issue #1895, opened 2023-02-19
Add actual results in the book


max-sixty:
From https://github.com/PRQL/prql/pull/1642#discussion_r1111315974

Would showing results of queries be helpful in the book? Currently we compile to SQL. For some queries, showing the actual results, executed against an `artists` table, might be helpful.

We would need to:
- Producing the HTML that displays a small table of results as well as the SQL. Possibly with "tabs"? Likely there are existing approaches to this so we wouldn't have to write much HTML ourselves (though I'm not sure if there are native mdbook approaches, so we may have to embed some HTML...)
- Run queries in the book's `snapshot.rs`, probably with duck-db, to generate the results. We could use `prql_results` as the language to hint to this file that it should produce results as well as the SQL, in the same way we use `prql_no_test` to avoid attempting to produce SQL.
- Adjust some queries so they can run on the `artists` db we have in the playground. We don't need to adjust them all, we can do this very gradually.

Any of these would make a welcome PR. While they'd all be required in order to have this functional, we can merge individual PRs to split up the work.

Jelenkee:
I guess the next step is making the PRQL-snippet editable and executable like in the offical Rust book :wink:

max-sixty:
> I guess the next step is making the PRQL-snippet editable and executable like in the offical Rust book 😉

Yes that would be very cool!

A button for "open in playground" might be a tractable thing?

To have it work _live_ in the book would be possible, but would require bundling the wasm in the book, and writing the logic to do the live updating — not trivial at all. (Though actually much easier than doing it for rust, since there's a wasm module which will compile it quickly enough to have it on every keystroke)

eitsupi:
An easy way might be to pass a query to the DuckDB CLI, receive a Markdown table, and embed it, as I wrote here. https://github.com/PRQL/prql/pull/3634#issuecomment-1751975845

- DuckDB CLI binary is required, but tedious DuckDB compilation is not required.
- We don't need to format the results into a Markdown table.

eitsupi:
> An easy way might be to pass a query to the DuckDB CLI, receive a Markdown table, and embed it, as I wrote here. [#3634 (comment)](https://github.com/PRQL/prql/pull/3634#issuecomment-1751975845)

Here is an example of this.
https://github.com/eitsupi/prql-book-test/blob/f962f1be627da025c85f6bd8b96575e565920524/src/main.rs#L132-L138

max-sixty:
[...]

PR #2070, opened 2023-03-09
feat: Code snippets from book in Playground

Jelenkee:
I added the prql snippets from the book to the playground.
The naming of the files may not be ideal.

image

If you have some ideas for better design, tell me.


max-sixty:
I think this could be great. I think _at the moment_ there are too many queries, and they don't target the Chinook files we have in the playground (nor do some of the existing queries, but that's the direction we'd like to move towards)

A couple of approaches I could see:
- Put this section at the bottom and don't expand it by default
- Add an indicator for queries we want to include in the book, like `prql,playground` (see how we use `prql_no_test` for queries that are expected to error), write out a comment like `# include in playground` when we write the query, and then only grab those queries for the playground
- Rewrite some of the queries to use the Chinook tables. Mostly this is just changing some of the column names. Then all our example queries also generate data.

The latter two would be awesome when combined — indeed they could replace all of the current queries, but are also much more work.

Jelenkee:
* The folders are not expanded by default. Only files and folders in the root folder are shown (which are a lot due to these syntax files)
* An indicator sounds good. I thought about using the markdown directly instead of the generated files (at the latests, when #2060 is done). Then it is also possible to use the name of the page and headline to name the example (instead of syntax-1.prql, syntax-2.prql, ...)
* Using chinook tables is a good idea. It is mention in #1895 as well

max-sixty:
> * Only files and folders in the root folder are shown (which are a lot due to these syntax files)

OK great, so maybe just putting this below Chinook and reducing the visual emphasis on the chapter headings would be sufficient to merge this, and then we can iterate more as we go?

max-sixty:
> * Then it is also possible to use the name of the page and headline to name the example (instead of syntax-1.prql, syntax-2.prql, ...)

This is a great idea. I guess not everything has a precise heading, but would still be much more descriptive.

If you're up for some futzing around with cmark, would be an interesting piece of work... (but probably not trivial, would involve some trial & error...)

max-sixty:
I see changes here — but surely we can't do both this and #2105 , since this relies on the files that #2105 removes?

Jelenkee:
It uses the markdown files now for fetching the prql snippets.


https://user-images.githubusercontent.com/59470612/224540781-640cd99f-b46b-48a1-89b9-c0e651b15c76.mp4




Does it look good?

aljazerzen:
Much better!

What about the arrow before the name?

image

Jelenkee:
> What about the arrow before the name?

Good idea

image

web/playground/src/app/App.js
```diff
@@ -26,6 +27,7 @@ class App extends React.Component {
     library: {
       examples,
       chinook,
+      book,
```
aljazerzen, line 30: Nice, this is very clean.

web/playground/src/generateBook.js
```diff
@@ -0,0 +1,138 @@
+const { readdir, stat, readFile, writeFile } = require("fs/promises");
```
aljazerzen: Wait, `generateBook.js` should not be part of `src/`, this would go into `tools/` or top-level.

web/playground/src/sidebar/Sidebar.js
```diff
@@ -1,24 +1,58 @@
...
+  const [v, sv] = useState(1);
 
   for (const [section, files] of Object.entries(library)) {
     const fileRows = [];
     for (const [index, filename] of Object.keys(files).entries()) {
+      const array = files[filename];
+      const depth = array[2];
+      const parent = array[3];
+      const id = array[4];
+      const name = array[5];
```
aljazerzen, line 39: At some point, this will have to be refactored from arrays into objects. I knew it when I changed `source` into `[output, source]`.

aljazerzen (approved):
Looks nice, I like it.

web/playground/src/generateBook.js
Jelenkee: Moved it

web/playground/src/sidebar/Sidebar.js
Jelenkee: Sure. This will need changes for the examples and chinook as well

Jelenkee: And it may not be compatble with currently saved queries in local storage

aljazerzen: Oh yeah, I haven't though about that. Well, we'll do it some time later then.

max-sixty:
@Jelenkee this is really excellent, thank you!

Grabbing the title works very well indeed. I might see whether I can do this quickly in the rust version too, and avoid the large churn anytime we add an example at the top of the page.

I'm still not sure whether we should have kept the old approach of storing the `.prql` files — now we have two chunks of code in two languages to extract them! But the output is great and that's the most important thing.

I'll merge. If you want to make an additional PR to add something to the Changelog, and possible a short docstring saying what the function does, that would be nice.

max-sixty:
@Jelenkee do you want to retry this? I've added https://github.com/PRQL/prql/pull/2181 so assuming it passes those tests, it'll work. I'm guessing it's not much code or comments to get it to pass.

FYI `git revert c1d05f76` gets you your old PR back on top of the current main.

Again sorry for not having this test, appreciate your patience.

PR #2075, opened 2023-03-10
ci: Fix GHPages deploment

max-sixty:
Not sure what changed, but it wasn't working. This more closely replicates the example GH pages example workflow. It works on `web` branch.


max-sixty:
FYI I just force pushed because the last release is so recent

PR #2080, opened 2023-03-10
internal: Derive `Debug` for `Context`

max-sixty:


max-sixty:
@aljazerzen merging, but let me know if there's any reason not to do this without much consideration...

PR #2093, opened 2023-03-10
web: Remove underline from logo

max-sixty:
Ref https://github.com/PRQL/prql/pull/1667. Even if we want it for the headings, it's not the logo. We could replace the logo with an image of the text (which we have available) if we want to be strict about text links having underline?


aljazerzen (approved):
Sure

PR #2094, opened 2023-03-10
devops: Add a GitHub issue template for bug reports

max-sixty:


max-sixty:
Merging so we can see it — post any feedback at all here

PR #2096, opened 2023-03-10
chore: PHP - fix lints and docs

vanillajonathan:


max-sixty:
Looking good! 

I see the number of errors in MegaLinter has gone down. (Which I don't place _that_ much weight on — I trust your judgement above MegaLinter's...)

I'll merge and let folks comment for any follow-ups

PR #2105, opened 2023-03-11
test: Remove writing of test files to disk

### LINKED ISSUE #356: Allow examples to be extracted on Windows
I just fixed the GHA spec so it actually ran on Windows.

Unfortunately this is now failing! I'm going to comment out the test, which runs on main, and open a PR to add it back.

Is anyone familiar with what could be going on? Or have easy access to a Windows machine to debug? I haven't used Windows in a long time...

Jelenkee:
Fixes #356

See #2060


max-sixty:
OK, I'm quite sad to lose my beautiful code, but this is both simpler and better. 

Thanks a lot @Jelenkee; this is a high quality PR.

max-sixty:
(and if we ever need these again — for example we want to include them in the playground — then it's easy to revert some of this)

PR #2069, opened 2023-03-09
build: update MegaLinter configs for PHP

eitsupi:
Part of #1994

Reflects the Lint setting from #2039

PR #2106, opened 2023-03-11
chore: prql-php in PSR-12 coding standard

vanillajonathan:
This fixes the inconsistency due to conflicting rules in #2069.


eitsupi (approved):
Thanks!

PR #2107, opened 2023-03-11
chore(prql-php): Add comment and define to header file

vanillajonathan:
This `#define` can later be used by the PHP binding for caching.

As per the PHP documentation: https://www.php.net/manual/en/ffi.load.php
> The header file _should_ contain a #define statement for the FFI_SCOPE variable, e.g.: #define FFI_SCOPE "MYLIB". Refer to the [class introduction](https://www.php.net/manual/en/class.ffi.php#ffi.intro) for details.


max-sixty:
Looks great! 

(anyone who knows PHP well, feel free to comment if you have thoughts)

PR #2110, opened 2023-03-11
fix: ensure use chumsky 0.9.2 or later

### LINKED ISSUE #2103: Multiple error messages do not work in prqlr (R)
### What's up?

I know this is not strictly a prql-compiler issue, but any help in developing prqlr would be appreciated.

After adopting prql-compiler 0.6.0, I noticed that the R session running prqlr in the following examples froze. (eitsupi/prqlr#100)

```r
"from a | select {b}" |> prqlr:::compile(target = NULL, format = FALSE, signature_comment = FALSE)
```

```r
"from a | select {b}" |> prqlr:::prql_to_pl()
```

It works fine with other clients. For example, prqlc.

```sh
$ echo 'from a | select {b}' | prqlc compile
Error:
   ╭─[-:1:17]
   │
 1 │ from a | select {b}
   ·                 ┬
   ·                 ╰── unexpected {
───╯
Error:
   ╭─[-:1:19]
   │
 1 │ from a | select {b}
   ·                   ┬
   ·                   ╰── unexpected }
───╯
```

After some testing, I found that `prql_compiler::prql_to_pl` was the cause.

https://github.com/PRQL/prql/blob/ed1e896ecbb1c5e18f0f1286639118e3a77b0100/prql-compiler/src/lib.rs#L229-L233

I suspect that this is probably because the following part does not work in R (the process does not end).
Could you please give me some tips for debugging?

https://github.com/PRQL/prql/blob/ed1e896ecbb1c5e18f0f1286639118e3a77b0100/prql-compiler/src/error.rs#L229-L276

eitsupi:
Fix #2103 (Follow up #2002)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@
 
 **Fixes**:
 
+- The version specification of the dependent Chumsky was bumped from `0.9.0` to `0.9.2`. `0.9.0` has a bug that causes an infinite loop. (@eitsupi, #2110)
+
 **Documentation**:
 
 **Web**:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,8 @@
 
 **Fixes**:
 
-- The version specification of the dependent Chumsky was bumped from `0.9.0` to `0.9.2`. `0.9.0` has a bug that causes an infinite loop. (@eitsupi, #2110)
+- The version specification of the dependent Chumsky was bumped from `0.9.0` to
+  `0.9.2`. `0.9.0` has a bug that causes an infinite loop. (@eitsupi, #2110)
 
 **Documentation**:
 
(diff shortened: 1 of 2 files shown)


max-sixty:
Ah ha! Sorry, so this was on `prql-compiler`.

---

Possibly we should add https://crates.io/crates/cargo-minimal-versions to CI?

I just tried running it, it failed, but we hit https://github.com/wangfenjin/duckdb-rs/issues/134 before finding this problem, so we'll have to wait for that before turning it on.

eitsupi:
> Possibly we should add [crates.io/crates/cargo-minimal-versions](https://crates.io/crates/cargo-minimal-versions) to CI?
> 
> I just tried running it, it failed, but we hit [wangfenjin/duckdb-rs#134](https://github.com/wangfenjin/duckdb-rs/issues/134) before finding this problem, so we'll have to wait for that before turning it on.

I didn't know about it, but looks great!
Thank you for working on that.

PR #2111, opened 2023-03-11
chore(prql-dotnet): Fix .NET bindings

vanillajonathan:
This fixes the .NET bindings to work with the new compiler library update with the refined API.

The unit test passes, but I ran into some problems marshaling the `CompileResult` struct so the source locations are bogus.


prql-dotnet/PrqlCompiler/PrqlCompiler.cs
```diff
@@ -38,53 +36,95 @@ public static string Compile(string prqlQuery)
...
+        }
+
+        /// <summary>
+        /// Finds variable references, validates functions calls, determines frames and converts PL to RQ.
+        /// </summary>
+        /// <param name="plJson">A PRQL query.</param>
+        /// <returns>JSON.</returns>
+        /// <exception cref="ArgumentException"><paramref name="plJson"/> is null or empty.</exception>
+        /// <exception cref="FormatException"><paramref name="plJson"/> cannot be compiled.</exception>
+        /// <remarks>https://docs.rs/prql-compiler/latest/prql_compiler/ast</remarks>
```
max-sixty: +1 on these links, great for reducing docs duplication

prql-dotnet/PrqlCompiler/PrqlCompiler.cs
```diff
@@ -38,53 +36,95 @@ public static string Compile(string prqlQuery)
         /// <exception cref="ArgumentException"><paramref name="prqlQuery"/> is null or empty.</exception>
         /// <exception cref="ArgumentNullException"><paramref name="options"/> is <c>null</c>.</exception>
         /// <exception cref="FormatException"><paramref name="prqlQuery"/> cannot be compiled.</exception>
-        public static string Compile(string prqlQuery, PrqlCompilerOptions options)
+        public static Result Compile(string prqlQuery, PrqlCompilerOptions options)
         {
             if (string.IsNullOrEmpty(prqlQuery))
             {
                 throw new ArgumentException(nameof(prqlQuery));
```
max-sixty: Is this name supposed to match the name at https://github.com/PRQL/prql/pull/2111/files#diff-628f74a4e1fbc63c5979f552afd14a7043936043a54700da168b3186fa36dc22L39? One has `Null` in...

max-sixty (approved):
Looking great @vanillajonathan 

(usual caveat about me not knowing C# well...)

prql-dotnet/PrqlCompiler/PrqlCompiler.cs
vanillajonathan: It documents the guard clause which throws if the input parameter `options` is `null`. If any property inside the argument is null that is a different matter.

max-sixty: Ah yes I see, `ArgumentException` is also on L36. Thanks!

vanillajonathan:
@max-sixty Can you help me move this to /bindings/prql-dotnet/ or /bindings/dotnet/ and merge?

max-sixty:
Great — merging! Thanks @vanillajonathan !

PR #2112, opened 2023-03-12
ci: Test minimum versions of dependencies

max-sixty:
This will fail because of https://github.com/wangfenjin/duckdb-rs/issues/134, so we'll wait for that to resolve before merging


eitsupi:
The test seems passing now?

eitsupi:
Currently this seems to fail because it cannot successfully download `tiberius`'s dependency `rust-encoding` (which abandoned years ago...).

```log
error: process didn't exit successfully: `/home/runner/.rustup/toolchains/1.65.0-x86_64-unknown-linux-gnu/bin/cargo metadata --format-version=1` (exit status: 101)
--- stderr
 Downloading crates ...
error: failed to download `encoding-index-tradchinese v1.0.20140915`

Caused by:
  unable to get packages from source

Caused by:
  failed to parse manifest at `/home/runner/.cargo/registry/src/github.com-1ecc6299db9ec823/encoding-index-tradchinese-1.0.20140915/Cargo.toml`

Caused by:
  library target names cannot contain hyphens: encoding-index-tradchinese

error: process didn't exit successfully: `/home/runner/.rustup/toolchains/1.65.0-x86_64-unknown-linux-gnu/bin/cargo hack test` (exit status: 1)
Error: Process completed with exit code 1.
```

This is obviously a development dependency and would have a chance of passing the test if it could be excluded from the test, but is there a way to exclude it?

max-sixty:
Ah, interesting @eitsupi. I think the best way is to figure out where it's coming from — it's not that easy but I've used `cargo tree -i` successfully to figure out the original issue. Then we can raise upstream / remove a dependency from prql...

eitsupi:
@max-sixty `tiberius` is this.

https://github.com/PRQL/prql/blob/65b7fefbf3cb1c2a8a86924b2f3f44fdc8809d0e/prql-compiler/Cargo.toml#L56

This is a test dependency, so it is not worth doing a min-version check here. But I don't know how to exclude this.

(BTW, I completely forgot to mention that this crate issue was pointed out earlier on this issue #2351)

max-sixty:
I see you put in https://github.com/prisma/tiberius/issues/292! Thanks.

If we can exclude it, great. Otherwise I guess we wait for tiberius to move to a new crate...

.github/workflows/test-all.yaml
```diff
@@ -150,3 +150,22 @@ jobs:
...
         working-directory: prql-compiler
         run: cargo msrv verify
+
+  test-package-min-versions:
+    runs-on: ubuntu-latest
+    steps:
+      - name: 📂 Checkout code
+        uses: actions/checkout@v3
+      - run: rustup toolchain add nightly
+      - uses: baptiste0928/cargo-install@next
```
```suggestion
      - uses: baptiste0928/cargo-install@v2
```

.github/workflows/test-all.yaml
```diff
@@ -150,3 +150,22 @@ jobs:
...
+        with:
+          crate: cargo-hack
+      - uses: baptiste0928/cargo-install@next
```
```suggestion
      - uses: baptiste0928/cargo-install@v2
```

max-sixty:
The docs don't say anything about exclusions, so I think we wait for https://github.com/prisma/tiberius/issues/292

max-sixty:
I think we need to wait for the next release of `clio`, to upgrade minimal clap to clap 4 (CC @aj-bagwell , without wanting to disturb!)

aj-bagwell:
`clio` will work with [clap 3 and clap 4](https://github.com/aj-bagwell/clio/blob/v0.3.3/Cargo.toml#L28), you should be fine to bump you minumum `clap` version. Cargo sometimes gets confused and adds in two versions of `clap` and uses one for clio and a different one for prql which means you have to fix the lock file by hand.

max-sixty:
> `clio` will work with [clap 3 and clap 4](https://github.com/aj-bagwell/clio/blob/v0.3.3/Cargo.toml#L28), you should be fine to bump you minumum `clap` version. Cargo sometimes gets confused and adds in two versions of `clap` and uses one for clio and a different one for prql which means you have to fix the lock file by hand.

Yes, I'm not sure how the resolver does this, but running the min version tests downgrades clap to 3, and then the code breaks.

But sounds like this is a cargo problem (or at least a PRQL problem) rather than a Clio problem.

max-sixty:
FYI this seems to come from when clap refers to a `clap_lex` struct `RawArgs`:

```
   Compiling clap v3.2.0
error[E0599]: no method named `is_end` found for mutable reference `&mut RawArgs` in the current scope
   --> /home/runner/.cargo/registry/src/github.com-1ecc6299db9ec823/clap-3.2.0/src/parser/parser.rs:434:63
    |
434 |                 if cfg!(feature = "unstable-v4") || !raw_args.is_end(&args_cursor) {
    |                                                               ^^^^^^ method not found in `&mut RawArgs`

For more information about this error, try `rustc --explain E0599`.
```

I guess we may need to wait until clap 3 is unsupported by any dependency

max-sixty:
This works now! Merging...

eitsupi:
Awesome! Thanks.

Issue #2029, opened 2023-03-06
Rationalizing the repo


max-sixty:
The repo is getting very busy in its root. If someone arrived and thought "Where can I find the core code", they might have no idea. (this happens to me occasionally in other repos)

Could we:
- Move al the bindings into a `bindings` path?
- Move `playground` & `website` & `book` into `web` path?
- I tried to see whether some of our config files could go into `.config`. But only `insta` seems to support that at the moment; we have a lot of them. I don't think there are any we don't use at all. 
- Any others? `snap` could maybe go somewhere? Maybe into a `build` path? Docker might be subsumed by the devcontainer work.

<img width="983" alt="image" src="https://user-images.githubusercontent.com/5635139/223037983-7e7b4960-53ae-4fc4-bb74-35f1586f5df5.png">

aljazerzen:
> Move al the bindings into a bindings path?

Yes, +1

> Move playground & website & book into web path?

Yes, +1

> Any others? snap could maybe go somewhere? 

Into `packages/`?

vanillajonathan:
Maybe move the playground out to its own repository, `/PRQL/playground`?
Maybe move the playground to its own subdomain, `play.prql-lang.org`?

eitsupi:
I don't think prql-js and playground should be moved from this repository given the current deployment process.

max-sixty:
> Maybe move the playground out to its own repository, `/PRQL/playground`?
> Maybe move the playground to its own subdomain, `play.prql-lang.org`?

We actually wanted the playground to be _more_ integrated with the website — e.g. we added a banner at the top of the playground to mirror the website, so it feels like the same property.

Possibly we'd get more contributions from folks who are comfortable with JS if it were a separate repo; there's some tradeoff there. But yes, the current deployment would be much more difficult, I wouldn't think it were worth the extra work there without more confidence it would lead to more contributions.

max-sixty:
Mostly done — just need to do a `packages` and `grammars` one, each with exactly one item...

max-sixty:
Closed by #2135, #2117, #2121

PR #2117, opened 2023-03-12
refactor: Place all web into `web` path

max-sixty:
Part of #2029


max-sixty:
I'm sure something will break somewhere, but I'm really not sure what, and there's only one way to find out...

I'll be on tomorrow to fix anything that does break

PR #2121, opened 2023-03-12
refactor: Move all bindings code to `bindings` path

max-sixty:


max-sixty:
Same comment as #2117 

> I'm sure something will break somewhere, but I'm really not sure what, and there's only one way to find out...

> I'll be on tomorrow to fix anything that does break

PR #2126, opened 2023-03-12
feat(prql-lib): Generate C++ header file

vanillajonathan:
The C++ header file `libprql_lib.hpp` was generated using cbindgen.
To generate a new one run:

    cbindgen --crate prql-lib --lang C++ --output libprql_lib.hpp


max-sixty:
Looks good! As ever, anyone who knows this better feel free to comment...

bindings/prql-lib/README.md
```diff
@@ -97,3 +97,10 @@ The C header file `libprql_lib.h` was generated using
 
     cargo install --force cbindgen
     cbindgen --crate prql-lib --output libprql_lib.h
+
+### C++ header file
+
+The C++ header file `libprql_lib.hpp` was generated using cbindgen. To generate
+a new one run:
+
+    cbindgen --crate prql-lib --lang C++ --output libprql_lib.hpp
```
max-sixty, line 106: If you want to make this more integrated, add a task in the Taskfile of `build-lib` and have these two commands there. Then we can also test them on every merge...

(but no obligation)

max-sixty: Actually even better than these would be to run them in `build.rs` — https://github.com/eqrion/cbindgen/blob/master/docs.md#buildrs — then it runs on every build, no need to have a separate script / Readme section

That would be a nice enhancement & simplification

PR #2127, opened 2023-03-12
chore(prql-php): Add PHPStan configuration file

vanillajonathan:
Since [PHPStan](https://phpstan.org/) doesn't know about the existence of the functions in libprql it complains since they get dynamically invoked on the FFI object instance without being statically defined so they're unknown.

https://phpstan.org/config-reference

### DIFF
--- a/bindings/prql-php/phpstan.neon
+++ b/bindings/prql-php/phpstan.neon
@@ -1,0 +1,9 @@
+parameters:
+    ignoreErrors:
+        - '#^Call to an undefined method FFI::compile\(\)\.$#'
+        - '#^Call to an undefined method FFI::prql_to_pl\(\)\.$#'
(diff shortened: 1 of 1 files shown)


bindings/prql-php/phpstan.neon
```diff
@@ -0,0 +1,13 @@
+parameters:
+    level: 9
+    paths:
+        - src
+        - tests
+    ignoreErrors:
```
max-sixty, line 6:
```suggestion
    ignoreErrors:
    # Since [PHPStan](https://phpstan.org/) doesn't know about the existence of the functions in libprql, it complains since they get dynamically invoked on the FFI object instance without being statically defined, so they're unknown.
```

max-sixty:
Thanks! Even better to add comments to the code than the PR, so the next person who comes knows why they're there...

Issue #1535, opened 2023-01-16
Resolve opaqueness of idents


max-sixty:
Edit: Renamed from "Resolve idents based on type"; new proposal below. Original issue follows:

---

Continuing on from #1516

Currently backticks are required in:

```elm
from `schema.table`
```

...which compiles to 

```sql
SELECT * FROM 'schema'.'table' -- (or just `schema.table`, no backticks, but *not* 'schema.table' with quotes)
```

That's because the resolver will try to resolve...

```elm
from schema.table
```

...as a namespace `schema` and a column name `table`, which then breaks.

---

I think ideally — we would instead resolve the argument based on its type:
- `from` & `join` would resolve to namespaces
- `select` & `derive` etc would resolve to columns

Then:
- `from schema.table` would be parsed as a single namespace `schema.table`, not a namespace `schema` and column `table`.
- `select table.column` would still be parsed as a namespace `table` & column `column`, because that's typed as a column

And then:
- any ident in backticks could be opaque — we wouldn't need to split by `.`, clearing up #1516
- we'd take `from schema.table`, not ```from `schema.table` ``` and produce the correct thing

max-sixty:
@aljazerzen as discussed

max-sixty:
We discussed this on the dev call and I said I'd think more about the tradeoffs. I think this issue demonstrates some weaknesses is our current name handling. We have a tri-lemma; we want:
- The ability to specify an identifier as opaque.
  - This:
    ```elm
    from `https://example.com/foo.parquet` 
    ```
    should definitely not compile to 
    ```sql
    FROM "https://example"."com/foo".parquet
    ```
- To not need to manage database schema hierarchy in PRQL
  - Ideally we could avoid managing more than a) Table and b) Column. 
    -   Specifically, all identifiers are `$namespace.$column`, where `$namespace` can be opaque
  - i.e. `from foo.bar` can be handled as the table `foo.bar`; `select baz.foo.bar` can just be the table `baz.foo` & the column `bar`. 
  - NB: Databases require breaking up the idents when quoted — i.e. `SELECT * FROM "foo.bar"` will try and query a Table named `foo.bar`, not a schema named `foo` and table named `bar`
    - So currently we split up the namespace by periods when compiling to SQL
- Names work similarly wherever they're used
[...]

PR #2129, opened 2023-03-12
docs: RFC modules

aljazerzen:
A crude draft on module and multiple queries per file. It's much more concrete than the type system one.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,9 +6,9 @@ target*/
 
 lcov.info
 
-.vscode/settings.json
-.vscode/launch.json
-.vscode/tasks.json
+**/.vscode/settings.json
+**/.vscode/launch.json
+**/.vscode/tasks.json
 .idea
 
 /*.prql
--- a/web/book/src/internals/modules.md
+++ b/web/book/src/internals/modules.md
@@ -15,19 +15,77 @@ Module is a namespace that contains declarations.
 Modules can be defined with module keyword and a code block encased in curly
 braces:
 
-```
-module my_module {
+```prql_no_test
+module my_playlists {
     let bangers = (from tracks | take 10)
 }
 ```
 
+Because module declaration is a declaration itself, modules can contain nested
+child modules:
+
+```prql_no_test
+module my_playlists {
+    let bangers = (from tracks | take 10)
+
+    module soundtracks {
+        let movie_albums = (from albums | filter id == 3)
+    }
+}
+```
+
+## Usage
+
+Any declarations within a module can be referenced from the outside of the
+module:
+
+```prql_no_test
+# using module structure declared above
+
+let great_tracks = my_playlists.bangers
+
+let movie_scores = my_playlists.soundtracks.movie_albums
+```
+
+All identifiers are resolved relative to current module.
+
+```prql_no_test
+module my_playlists {
+    module soundtracks {
+        let movie_albums = (from albums | filter id == 3)
+    }
+
+    from soundtracks.movie_albums
+}
+from my_playlists.soundtracks.movie_albums
+```
+
+## Main var declaration
+
+If last declaration in a module is a variable declaration that is named `main`,
+then the leading `let main = ` can be omitted and expressed only by the
+expression itself.
+
+```
+module my_playlists {
+    let bangers = (from tracks | take 10)
+
+    from playlists | take 10
+}
+
+let album_titles = my_playlists.main
+```
+
+When a file `my_file` is compiled to a query, variable `my_file.main` will be
+compiled to RQ.
+
 ## File loading
 
-If a module definition of `my_module` does not have a code block, it's contents
-are loaded from `./my_module.prql`.
+If a module definition of `my_playlists` does not have a code block, it's
+contents are loaded from `./my_playlists.prql`.
 
 If the file does not exist, contents are loaded from file
-`./my_module/mod.prql`.
+`./my_playlists/mod.prql`.
 
 When compilation is invoked compiler will search for the root of the module
 structure of the project:
@@ -42,36 +100,6 @@ prql-compiler will provide an interface that has:
 - a function that takes a "file loader", which load files on-demand,
 - a simple function, that always report "file does not exist" to the compiler,
   and can be used for compiling a single file to a single query.
-
-## Behavior
-
-Any declarations within a module can be referenced from the outside of the
-module:
-
-```
-module my_module {
-    let bangers = (from tracks | take 10)
-}
-
-let great_tracks = my_module.bangers
-```
-
-If last declaration in a module is a variable declaration that is named `main`,
-then the leading `let main = ` can be omitted and expressed only by the
-expression itself.
-
-```
-module my_module {
-    let bangers = (from tracks | take 10)
-
-    from albums | select [title]
-}
-
-let album_titles = my_module.main
-```
-
-When a file `my_file` is compiled to a query, variable `my_file.main` will be
-compiled to RQ.
 
 ## Built-in module structure
 
--- a/web/book/src/internals/modules.md
+++ b/web/book/src/internals/modules.md
@@ -18,7 +18,7 @@ Design goals:
 ## Definition
 
 A module is a namespace that contains declarations. A module is itself a
-declaration , which means that it can contain nested child modules.
+declaration, which means that it can contain nested child modules.
 
 This means that modules form a
 [tree graph](<https://en.wikipedia.org/wiki/Tree_(graph_theory)>), which we call
@@ -49,6 +49,7 @@ module:
 
 ```prql no-eval
 # using module structure declared above
+module my_playlists
 
 let great_tracks = my_playlists.bangers
 
@@ -115,15 +116,15 @@ let album_titles = my_playlists
 
 ## File importing
 
-To include PRQL source code from other files, one can use the following syntax:
+To include PRQL source code from other files, we can use the following syntax:
 
 ```
 module my_playlists
 ```
 
-Such declaration loads either `./my_playlists.prql` (a leaf module) or
+This loads either `./my_playlists.prql` (a leaf module) or
 `./my_playlists/_my_playlists.prql` (a directory module) and uses its contents
-as the declarations in module `my_playlists`. If none or both of the files are
+as module `my_playlists`. If none or both of the files are
 present, a compilation error is raised.
 
 Only directory modules can contain module declarations. If a leaf module
--- a/web/book/src/internals/modules.md
+++ b/web/book/src/internals/modules.md
@@ -47,7 +47,7 @@ module my_playlists {
 Any declarations within a module can be referenced from the outside of the
 module:
 
-```prql_no_test
+```prql no-eval
 # using module structure declared above
 
 let great_tracks = my_playlists.bangers
@@ -57,7 +57,7 @@ let movie_scores = my_playlists.soundtracks.movie_albums
 
 Identifiers are resolved relative to current module.
 
-```prql_no_test
+```prql no-eval
 module my_playlists {
     module soundtracks {
         let movie_albums = (from albums | filter id == 3)
@@ -72,7 +72,7 @@ If an identifier cannot be resolved relative to the current module, it tries to
 resolve relative to the parent module. This is repeated, stepping up the module
 hierarchy until a match is found or root of the module structure is reached.
 
-```prql_no_test
+```prql no-eval
 module my_playlists {
     let decl_1 = ...
 
@@ -138,7 +138,7 @@ adding this restriction.
 For example, the following module structure is annotated with files names in
 which the modules would reside:
 
-```prql_no_test
+```prql no-eval
 
 module my_project {
     # _my_project.prql
--- a/web/book/src/internals/modules.md
+++ b/web/book/src/internals/modules.md
@@ -91,9 +91,8 @@ module my_playlists {
 
 ## Main var declaration
 
-If the final declaration in a module is a variable declaration that is named
-`main`, then the leading `let main = ` can be omitted and expressed only by the
-expression itself.
+The final variable declaration in a module can omit the leading `let main =` and
+acquire an implicit name main.
 
 ```
 module my_playlists {
--- a/web/book/src/internals/modules.md
+++ b/web/book/src/internals/modules.md
@@ -8,15 +8,15 @@ PRQL needs modules that would:
 - have unambiguous module structure within a project,
 - be able to compile individual files that are a part of a project.
 
-## Basic definition
+## Declaration
 
 Module is a namespace that contains declarations.
 
-Modules can be defined with mod keyword and a code block encased in curly
+Modules can be defined with module keyword and a code block encased in curly
 braces:
 
 ```
-mod my_module {
+module my_module {
     let bangers = (from tracks | take 10)
 }
 ```
@@ -49,7 +49,7 @@ Any declarations within a module can be referenced from the outside of the
 module:
 
 ```
-mod my_module {
+module my_module {
     let bangers = (from tracks | take 10)
 }
 
@@ -61,7 +61,7 @@ then the leading `let main = ` can be omitted and expressed only by the
 expression itself.
 
 ```
-mod my_module {
+module my_module {
     let bangers = (from tracks | take 10)
 
     from albums | select [title]
@@ -77,13 +77,13 @@ compiled to RQ.
 
 ```
 # root module of every project
-mod project {
-	mod std {
+module project {
+	module std {
 		func sum a -> ...
 		func mean a -> ...
 	}
 
-	mod default_db {
+	module default_db {
 		# all inferred tables and defined CTEs
 	}
 
--- a/web/book/src/internals/modules.md
+++ b/web/book/src/internals/modules.md
@@ -124,8 +124,8 @@ module my_playlists
 
 This loads either `./my_playlists.prql` (a leaf module) or
 `./my_playlists/_my_playlists.prql` (a directory module) and uses its contents
-as module `my_playlists`. If none or both of the files are
-present, a compilation error is raised.
+as module `my_playlists`. If none or both of the files are present, a
+compilation error is raised.
 
 Only directory modules can contain module declarations. If a leaf module
 contains a module declaration, a compilation error is raised, suggesting the
--- a/web/book/src/internals/modules.md
+++ b/web/book/src/internals/modules.md
@@ -1,0 +1,95 @@
+# Modules
+
+PRQL needs modules that would:
+
+- allow importing declarations from other files,
+- have namespaces for things like `std`,
+- have hierarchical structure so we can represent files in directories,
+- have unambiguous module structure within a project,
+- be able to compile individual files that are a part of a project.
+
+## Basic definition
+
+Module is a namespace that contains declarations.
+
+Modules can be defined with mod keyword and a code block encased in curly
+braces:
+
+```
+mod my_module {
+    let bangers = (from tracks | take 10)
+}
+```
+
+## File loading
+
+If a module definition of `my_module` does not have a code block, it's contents
+are loaded from `./my_module.prql`.
+
+If the file does not exist, contents are loaded from file
+`./my_module/mod.prql`.
+
+When compilation is invoked compiler will search for the root of the module
+structure of the project:
+
+1. attempt to load `./project.prql`. If successful use it as the root,
+2. attempt to load `./mod.prql`. If successful visit parent directory and go
+   back to step 1.
+3. use compiled file as the root for the module structure.
+
+prql-compiler will provide an interface that has:
+
+- a function that takes a "file loader", which load files on-demand,
+- a simple function, that always report "file does not exist" to the compiler,
+  and can be used for compiling a single file to a single query.
+
+## Behavior
+
+Any declarations within a module can be referenced from the outside of the
+module:
+
+```
+mod my_module {
+    let bangers = (from tracks | take 10)
+}
+
+let great_tracks = my_module.bangers
+```
+
+If last declaration in a module is a variable declaration that is named `main`,
+then the leading `let main = ` can be omitted and expressed only by the
+expression itself.
+
+```
+mod my_module {
+    let bangers = (from tracks | take 10)
+
+    from albums | select [title]
+}
+
+let album_titles = my_module.main
+```
+
+When a file `my_file` is compiled to a query, variable `my_file.main` will be
+compiled to RQ.
+
+## Built-in module structure
+
+```
+# root module of every project
+mod project {
+	mod std {
+		func sum a -> ...
+		func mean a -> ...
+	}
+
+	mod default_db {
+		# all inferred tables and defined CTEs
+	}
+
+	let main = (
+		from t = tracks
+		select [track_id, title]
+	)
+}
+```
--- a/web/book/src/internals/modules.md
+++ b/web/book/src/internals/modules.md
@@ -1,40 +1,48 @@
 # Modules
 
-PRQL needs modules that would:
+> Status: proposal
 
-- allow importing declarations from other files,
-- have namespaces for things like `std`,
-- have hierarchical structure so we can represent files in directories,
-- have unambiguous module structure within a project,
-- be able to compile individual files that are a part of a project.
+> This is a technical document. For a "how to use" or a TLDR; skip to the
+> [Example](#Example) section.
 
-## Declaration
+Design goals:
 
-Module is a namespace that contains declarations.
+1. Allow importing declarations from other files.
 
-Modules can be defined with module keyword and a code block encased in curly
-braces:
+2. Have namespaces for things like `std`.
 
-```prql_no_test
-module my_playlists {
-    let bangers = (from tracks | take 10)
-}
+3. Have hierarchical structure so we can represent files in directories.
+
+4. Have unambiguous module structure within a project.
+
+## Definition
+
+Module is a namespace that contains declarations. A module declaration itself,
+which means that it can contain nested child modules.
+
+This means that modules form a
+[tree graph](<https://en.wikipedia.org/wiki/Tree_(graph_theory)>), which we call
+"the module structure".
+
+For the sake of this document, we will express the module structure with
+`module` keyword and a code block encased in curly braces:
+
 ```
-
-Because module declaration is a declaration itself, modules can contain nested
-child modules:
-
-```prql_no_test
 module my_playlists {
-    let bangers = (from tracks | take 10)
+    let bangers = ... # a declaration
 
     module soundtracks {
-        let movie_albums = (from albums | filter id == 3)
+        let movie_albums = ... # another declaration
     }
 }
 ```
 
-## Usage
+> The syntax `module name { ...decls... }` is not part of PRQL language, with
+> the objection that it is unnecessary as it only adds more ways of defining
+> modules. If a significant upside of this syntax is found, it may be added in
+> the future.
+
+## Name resolution
 
 Any declarations within a module can be referenced from the outside of the
 module:
@@ -47,7 +55,7 @@ let great_tracks = my_playlists.bangers
 let movie_scores = my_playlists.soundtracks.movie_albums
 ```
 
-All identifiers are resolved relative to current module.
+Identifiers are resolved relative to current module.
 
 ```prql_no_test
 module my_playlists {
@@ -58,6 +66,26 @@ module my_playlists {
     from soundtracks.movie_albums
 }
 from my_playlists.soundtracks.movie_albums
+```
+
+If an identifier cannot be resolved relative to the current module, it tries to
+resolve relative to the parent module. This is repeated, stepping up the module
+hierarchy until a match is found or root of the module structure is reached.
+
+```prql_no_test
+module my_playlists {
+    let decl_1 = ...
+
+    module soundtracks {
+        let decl_2 = ...
+    }
+
+    module upbeat_rock {
+        let decl_3 = ...
+
+        from decl_1 | join soundtracks.decl2 | join decl_3
+    }
+}
 ```
 
 ## Main var declaration
@@ -76,32 +104,137 @@ module my_playlists {
 let album_titles = my_playlists.main
 ```
 
-When a file `my_file` is compiled to a query, variable `my_file.main` will be
-compiled to RQ.
+When a module is referenced as a value, the `main` variable is used instead.
+This is especially useful when referring to a module which is to be compiled to
+RQ (and later SQL).
 
-## File loading
+## File importing
 
-If a module definition of `my_playlists` does not have a code block, it's
-contents are loaded from `./my_playlists.prql`.
+To include PRQL source code from other files, one can use the following syntax:
 
-If the file does not exist, contents are loaded from file
-`./my_playlists/mod.prql`.
+```
+module my_playlists
+```
 
-When compilation is invoked compiler will search for the root of the module
-structure of the project:
+Such declaration loads either `./my_playlists.prql` (a leaf module) or
+`./my_playlists/_my_playlists.prql` (a directory module) and uses its contents
+as the declarations in module `my_playlists`. If none or both of the files are
+present, a compilation error is raised.
 
-1. attempt to load `./project.prql`. If successful use it as the root,
-2. attempt to load `./mod.prql`. If successful visit parent directory and go
-   back to step 1.
-3. use compiled file as the root for the module structure.
+Only directory modules can contain module declarations. If a leaf module
+contains a module declaration, a compilation error is raised, suggesting the
+leaf module to be converted into a directory module. This is a step toward any
+module structure having a single "normalized" representation in the file system.
+Such normalization is desired because it restrains the possible file system
+layouts to a comprehensible and predictable layout, while not sacrificing any
+functionality.
 
-prql-compiler will provide an interface that has:
+Described importing rules don't achieve this "single normalized representation"
+in full, since any leaf modules could be replaced by a directory module with
+zero submodules, without any semantic changes. Restricting directory modules to
+have at least one sub-module would not improve approachability enough to justify
+adding this restriction.
 
-- a function that takes a "file loader", which load files on-demand,
-- a simple function, that always report "file does not exist" to the compiler,
-  and can be used for compiling a single file to a single query.
+For example, the following module structure is annotated with files names in
+which the modules would reside:
+
+```prql_no_test
+
+module my_project {
+    # _my_project.prql
+
+    module sales {
+        # sales.prql
+    }
+
+    module projections {
+        # projections/_projections.prql
+
+        module year_2023 {
+            # projections/year_2023.prql
+        }
+
+        module year_2024 {
+            # projections/year_2024.prql
+        }
+    }
+}
+```
+
+If module `my_project.sales` wants to add a submodule `util`, it has to be
+converted to a directory modules. This means that it has to be moved to
+`sales/_sales.prql`. The new module would reside in `sales/util.prql`.
+
+The annotated layout is not the only possible layout for this module structure,
+since any of the modules `sales`, `year_2023` or `year_2024` could be converted
+into a directory module with zero sub-modules.
+
+Point 4 of design goals means that each declaration within a project has a
+single fully-qualified name within this project. This is insured by strict rules
+regarding importing files and the fact that the module structure is a tree.
+
+## Declaration order
+
+The order of declarations in a module holds no semantic value, except the "last
+`main` variable".
+
+References between modules can be cyclic.
+
+```
+module mod_a {
+    let decl_a_1 = ...
+    let decl_a_2 = (from mod_b.decl_b | take 10)
+}
+module mod_b {
+    let decl_b = (from mod_a.decl_a | take 10)
+}
+```
+
+References between variable declarations cannot be cyclic.
+
+```
+let decl_a = (from decl_b)
+let decl_b = (from decl_a) # error: cyclic reference
+```
+
+```
+module mod_a {
+    let decl_a = (from mod_b.decl_b)
+}
+module mod_b {
+    let decl_b = (from mod_a.decl_a) # error: cyclic reference
+}
+```
+
+## Compiler interface
+
+`prql-compiler` provides two interfaces for compiling files.
+
+**Multi-file interface** requires three arguments:
+
+- path to the file containing the module which is the root of the module
+  structure,
+- identifier of the pipeline that should be compiled to RQ and,
+- a "file loader", which can load files on-demand.
+
+The path to the root module can be automatically detected by searching for
+`.prql` files starting with `_` in the current working directory.
+
+Example prqlc usage:
+
+```
+$ prqlc compile _project.prql sales.projections.orders_2024
+$ prqlc compile sales.projections.orders_2024
+```
+
+**Single-file interface** requires a single argument; the PRQL source. Any
+attempts to load modules in this mode result in compilation errors. This
+interface is needed, for example, when integrating the compiler with a database
+connector (i.e. JDBC) where no other files can be loaded.
 
 ## Built-in module structure
+
+> Work In Progress
 
 ```
 # root module of every project
@@ -120,4 +253,59 @@ module project {
 		select [track_id, title]
 	)
 }
+```
+
+## Example
+
+This is an example project, where each of code block is a separate file.
+
+```
+# _project.prql
+
+module employees
+module sales
+module util
+```
+
+```
+# employees.prql
+
+let employees = (...)
+
+let salaries = (...)
+
+let departments = (...)
+```
+
+```
+# sales/_sales.prql
+
+module orders
+module projections
+
+let revenue_by_source = (...)
+```
+
+```
+# sales/orders.prql
+
+let current_year = (...)
+
+let archived = (...)
+
+let by_employee = (from orders | join employees.employees ...)
+```
+
+```
+# sales/projections.prql
+
+let orders_2023 = (from orders.current_year | append orders.archived)
+
+let orders_2024 = (...)
+```
+
+```
+# util.prql
+
+func pretty_print_num col -> (...)
 ```
--- a/web/book/src/internals/modules.md
+++ b/web/book/src/internals/modules.md
@@ -13,7 +13,7 @@ Design goals:
 
 3. Have a hierarchical structure so we can represent files in directories.
 
-4. Have unambiguous module structure within a project.
+4. Have an unambiguous module structure within a project.
 
 ## Definition
 
@@ -175,7 +175,7 @@ since any of the modules `sales`, `year_2023` or `year_2024` could be converted
 into a directory module with zero sub-modules.
 
 Point 4 of design goals means that each declaration within a project has a
-single fully-qualified name within this project. This is insured by strict rules
+single fully-qualified name within this project. This is ensured by strict rules
 regarding importing files and the fact that the module structure is a tree.
 
 ## Declaration order
--- a/web/book/src/internals/modules.md
+++ b/web/book/src/internals/modules.md
@@ -11,14 +11,14 @@ Design goals:
 
 2. Have namespaces for things like `std`.
 
-3. Have hierarchical structure so we can represent files in directories.
+3. Have a hierarchical structure so we can represent files in directories.
 
 4. Have unambiguous module structure within a project.
 
 ## Definition
 
-Module is a namespace that contains declarations. A module declaration itself,
-which means that it can contain nested child modules.
+A module is a namespace that contains declarations. A module is itself a
+declaration , which means that it can contain nested child modules.
 
 This means that modules form a
 [tree graph](<https://en.wikipedia.org/wiki/Tree_(graph_theory)>), which we call
@@ -90,8 +90,8 @@ module my_playlists {
 
 ## Main var declaration
 
-If last declaration in a module is a variable declaration that is named `main`,
-then the leading `let main = ` can be omitted and expressed only by the
+If the final declaration in a module is a variable declaration that is named
+`main`, then the leading `let main = ` can be omitted and expressed only by the
 expression itself.
 
 ```
@@ -309,3 +309,10 @@ let orders_2024 = (...)
 
 func pretty_print_num col -> (...)
 ```
+
+---
+
+Sources:
+
+- [Notes On Module System](https://matklad.github.io/2021/11/27/notes-on-module-system.html),
+  by @matklad.
--- a/web/book/src/internals/modules.md
+++ b/web/book/src/internals/modules.md
@@ -108,6 +108,11 @@ When a module is referenced as a value, the `main` variable is used instead.
 This is especially useful when referring to a module which is to be compiled to
 RQ (and later SQL).
 
+```
+# last line from previous example could thus be shortened to:
+let album_titles = my_playlists
+```
+
 ## File importing
 
 To include PRQL source code from other files, one can use the following syntax:
@@ -214,7 +219,8 @@ module mod_b {
 
 - path to the file containing the module which is the root of the module
   structure,
-- identifier of the pipeline that should be compiled to RQ and,
+- identifier of the pipeline that should be compiled to RQ (this can also be an
+  identifier of a module that has a `main` pipeline) and,
 - a "file loader", which can load files on-demand.
 
 The path to the root module can be automatically detected by searching for
--- a/web/book/src/internals/modules.md
+++ b/web/book/src/internals/modules.md
@@ -115,6 +115,9 @@ let album_titles = my_playlists
 
 ## File importing
 
+> This section is under discussion. Current implementation plans do not include
+> `module` declarations, but loading of all files under the compilation path.
+
 To include PRQL source code from other files, we can use the following syntax:
 
 ```


web/book/src/internals/modules.md
```diff
@@ -0,0 +1,95 @@
...
+
+Module is a namespace that contains declarations.
+
+Modules can be defined with mod keyword and a code block encased in curly
+braces:
+
+```
+mod my_module {
+    let bangers = (from tracks | take 10)
+}
```
aljazerzen, line 37: Syntax is similar to current discussion on tuples/arrays, but I don't think it will clash on the parser-level. That's because this is on the "level" of the statements and those two are both on the "level" of expressions.

web/book/src/internals/modules.md
```diff
@@ -0,0 +1,95 @@
...
+```
+
+## File loading
```
aljazerzen: This design is heavily plagiarized from how Rust modules work. Thank god for opensource!

web/book/src/internals/modules.md
```diff
@@ -0,0 +1,95 @@
...
+# root module of every project
+mod project {
+	mod std {
+		func sum a -> ...
+		func mean a -> ...
+	}
+
+	mod default_db {
+		# all inferred tables and defined CTEs
+	}
```
aljazerzen: To be honest I don't remember why this module is needed. Maybe we could get away by having the tables on the top-level.

It would simplify the compiler a bit.

.gitignore
```diff
@@ -6,9 +6,9 @@ target*/
 
 lcov.info
 
-.vscode/settings.json
-.vscode/launch.json
-.vscode/tasks.json
+**/.vscode/settings.json
```
max-sixty, line 9: (sorry, this was my misunderstanding — `foo` is the same as `**/foo` but `foo/bar` is not the same as `**/foo/bar`...

web/book/src/internals/modules.md
```diff
@@ -0,0 +1,95 @@
...
+If a module definition of `my_module` does not have a code block, it's contents
+are loaded from `./my_module.prql`.
+
+If the file does not exist, contents are loaded from file
+`./my_module/mod.prql`.
+
+When compilation is invoked compiler will search for the root of the module
+structure of the project:
+
+1. attempt to load `./project.prql`. If successful use it as the root,
```
max-sixty: What do you see as the difference between a project and a module? If we don't need a distinction, we can avoid having one at the moment, and just use `mod` for both...

max-sixty (commented):
Overall looks great!

---

I found this quite insightful: https://matklad.github.io/2021/11/27/notes-on-module-system.html. It recommends some simplifications to the module system:
- No `mod {}` — is it worth the complication, as opposed to just files?
- No `mod.rs` — instead `foo/foo.rs` (it actually has `_foo.rs`, but why the leading underscore?) — I like this, it reduces the number of `mod.rs` files

---

Are you thinking that imports would be implicit? Or would that be a separate discussion?

web/book/src/internals/modules.md
aljazerzen: My intention was for project to represent the root of the module tree.

This may not be needed if we would also have `super` (a reference to parent module), which would be nice to have in any case.

aljazerzen:
>    No mod {} — is it worth the complication, as opposed to just files?

Good idea, it does not bring much to the table. We can also add it later without breaking changes.

>    No mod.rs — instead foo/foo.rs (it actually has _foo.rs, but why the leading underscore?) — I like this, it reduces the number of mod.rs files

I also like this. The pain with `mod.rs` is searching because you have to look at folder names instead of just file name.

The underscore is here so it's first alphabetically.

> Are you thinking that imports would be implicit? Or would that be a separate discussion?

(We can discuss here)

I'm not exactly sure what do you mean:

Implicit `use foo.*`. That is problematic because it will cause broken code just by adding a module. We do it now it `std` module and it's already causing bugs.

Implicit `mod foo`. I'd like to avoid automatic file discovery, because would allow attacks that place malicious `mod.prql` in common locations like `Downloads/` and hope you try to compile a `.prql` file in same directory. It's not a huge risk now, but it would be backward incompatible to remove later.

aljazerzen:
Thinking about it I realize that similar attacks could be possible by injecting `project.prql` with current proposal.

Huh. I'll do another pass on the proposal.

max-sixty:
> Implicit `use foo.*`. That is problematic because it will cause broken code just by adding a module. We do it now it `std` module and it's already causing bugs.

So are you thinking that we'd have `use foo` as a way of importing? (Currently I don't see this in the RFC — ofc no prob if we do it later)

> The underscore is here so it's first alphabetically.

Hmmm, I don't love the underscore aesthetically, given it's the "main" file, but only aesthetically, low weight.

aljazerzen:
> > Implicit `use foo.*`. That is problematic because it will cause broken code just by adding a module. We do it now it `std` module and it's already causing bugs.
> 
> So are you thinking that we'd have `use foo` as a way of importing? (Currently I don't see this in the RFC — ofc no prob if we do it later)

We don't have to add it now, because you can use longer idents: `foo.bar.baz`. But yes, ultimately, `use` will be needed.
 
> > The underscore is here so it's first alphabetically.
> 
> Hmmm, I don't love the underscore aesthetically, given it's the "main" file, but only aesthetically, low weight.

I agree, it doesn't look nice. We could do without it, but that:
- makes it hard to find the "parent" file in a directory,
- has a conflict when you want to have modules `foo` and `foo.foo` (because both want to have a file at `foo/foo.prql`). I'd be fine with forbidding that all together.

max-sixty:
> has a conflict when you want to have modules `foo` and `foo.foo` (because both want to have a file at `foo/foo.prql`). I'd be fine with forbidding that all together.

lol, this is technically possible, it just requires doing `foo/foo/foo.prql`!

---


> We don't have to add it now, because you can use longer idents: `foo.bar.baz`. But yes, ultimately, `use` will be needed.

OK, so we do have implicit imports — i.e. someone can call `foo.bar` without having a `use foo` statement?

web/book/src/internals/modules.md
```diff
@@ -0,0 +1,95 @@
...
+
+## Basic definition
+
+Module is a namespace that contains declarations.
+
+Modules can be defined with mod keyword and a code block encased in curly
+braces:
+
+```
+mod my_module {
```
snth: How come `mod` rather than `module`?

The reason I ask is because elsewhere we've gone for longer names, e.g. `func` rather than `fn`, `aggregate` rather than `agg`, `derive` rather than `drv`, ...

I find those long names a bit of a pain to type sometimes and would probably like all of them shorter. However if we want to be more beginner friendly and explicit then we should probably go for `module` here too in order to be consistent? WDYT?

web/book/src/internals/modules.md
```diff
@@ -0,0 +1,95 @@
...
+    let bangers = (from tracks | take 10)
+}
+```
+
+## File loading
+
+If a module definition of `my_module` does not have a code block, it's contents
```
snth:
```suggestion
If a module definition of `my_module` does not have a code block, its contents
```

web/book/src/internals/modules.md
snth: I was going to ask about that. Are modules not also structs where the keys are names and the values are either (sub)modules, relations or functions?

Shouldn't we therefore be consistent in the use of symbols and use `[]` possibly rather than `{}`? Your current proposal reads a lot more naturally ofc but curious about the consistency angle.

snth (commented):
Overall I found this a bit less clear than the other RFC and feel that it could be fleshed out a bit more. I had to read a few sections two or three times to try and ensure that I understood it correctly.

web/book/src/internals/modules.md
max-sixty: (fwiw this part is likely going away anyway https://github.com/PRQL/prql/pull/2129#issuecomment-1468861960)

web/book/src/internals/modules.md
aljazerzen: No, semantically modules are not equivalent to structs. They don't have a type, cannot be referred to as a value and have different parsing rules.

So I think that here it's ok to choose the syntax that looks best, instead of seeking consistency with structs.

max-sixty: (also moot if we use just files / https://github.com/PRQL/prql/pull/2129#issuecomment-1468861960)

web/book/src/internals/modules.md
```diff
@@ -0,0 +1,95 @@
...
+are loaded from `./my_module.prql`.
+
+If the file does not exist, contents are loaded from file
```
vanillajonathan: I dislike this this, I think it make the module system more complicated, more to learn, more ways to do one thing, it increases the needs for best practices and to reach consensus.
I think it would better to have just one way, that it should always load `./my_module.prql` without any fallback.

I think having several files with the same filename (`mod.prql`) is a nightmare, it makes it difficult to search, to find files, to determine which file something is in.

web/book/src/internals/modules.md
vanillajonathan: I think the `mod { ... }` code block should not exist, I think modules should always be one self-contained file.

web/book/src/internals/modules.md
vanillajonathan: I think Rust is poor choice to draw inspiration from for a module system. In my opinion it is very confusing and one of the worst parts of Rust.
I think the module system should aim to be as simple and predictable as possible.
I think it would be better to look at PHP and it's [`include`](https://www.php.net/manual/en/function.include.php) statement which is dead simple.

web/book/src/internals/modules.md
vanillajonathan: Why have a code block at all?
It complicated things for the compiler, for users, more to learn, more ways to do one thing, more decide and figure our the preferred way to do things.
I would much prefer there was no `mod` code block and that modules were **always** declared in a self-contained file.

web/book/src/internals/modules.md
```diff
@@ -0,0 +1,95 @@
...
+1. attempt to load `./project.prql`. If successful use it as the root,
+2. attempt to load `./mod.prql`. If successful visit parent directory and go
+   back to step 1.
+3. use compiled file as the root for the module structure.
+
+prql-compiler will provide an interface that has:
+
+- a function that takes a "file loader", which load files on-demand,
+- a simple function, that always report "file does not exist" to the compiler,
+  and can be used for compiling a single file to a single query.
```
vanillajonathan: I would prefer that file loading was not customizable, that it always works in the same predetermined predictable way.

web/book/src/internals/modules.md
vanillajonathan: Does there need to be a "roof of the module"?
Shouldn't **one** module **always** be just **one** file?

web/book/src/internals/modules.md
vanillajonathan: I don't like the nesting of `mod` statements. It is confusing and creates indention. I would prefer just have `std.prql` and `default_db.prql` as files.

vanillajonathan:
The documention doesn't mention how to load a module. I expected to find a `include "foo.prql"`, `import Foo`, `using AcmeCorp.Foo`, or `use Foo` statement.

web/book/src/internals/modules.md
vanillajonathan: I agree with @max-sixty about not having `mod {}` and `mod.rs`, but I think just `foo.rs` might be enough and not have it's own directory such as `foo/foo.rs`.
**One** module would **always** be just **one** file.

vanillajonathan:
> I agree, it doesn't look nice. We could do without it, but that:
>
> *    makes it hard to find the "parent" file in a directory,
> *    has a conflict when you want to have modules `foo` and `foo.foo` (because both want to have a file at `foo/foo.prql`). I'd be fine with forbidding that all together.

Do we even really needed modules in paths like that?
Maybe only `foo` should be allowed, not `foo.foo` with dot.

web/book/src/internals/modules.md
max-sixty: Please read the previous comments @vanillajonathan — this has already been discussed

web/book/src/internals/modules.md
aljazerzen: This is talking about interfaces prql-compiler will expose to different programs. For example, if you use it in some database connector (i.e. JDBC) there is no other files to load. Because only input a single source string, we need an interface that will report "file does not exist" when that source string tries to load some other file.

Will expand this section.

aljazerzen:
> The documention doesn't mention how to load a module. I expected to find a `include "foo.prql"`, `import Foo`, `using AcmeCorp.Foo`, or `use Foo` statement.

It does, but perhaps not clearly enough. You just use `module foo` and that will import `foo.prql`.

> Do we even really needed modules in paths like that?
> Maybe only foo should be allowed, not foo.foo with dot.

I don't think we can afford that. It would be a ruff arbitrary prohibition that may lead to IDE's overwriting your source files when you rename some module. And we (almost) run into such case in prql-compiler: we have a module with path `tests::test`. https://github.com/PRQL/prql/tree/main/prql-compiler/src/tests

web/book/src/internals/modules.md
```diff
@@ -0,0 +1,311 @@
...
+        let decl_2 = ...
+    }
+
+    module upbeat_rock {
+        let decl_3 = ...
+
+        from decl_1 | join soundtracks.decl2 | join decl_3
+    }
+}
+```
```
aljazerzen: I'm not sure about this.

Upsides:

- the std module can be defined just in the root module and be accessible
  everywhere under name `std`,

Downsides:

- we need to find the root of the module structure,

web/book/src/internals/modules.md
aljazerzen: I've changed the keyword to `module`, as I agree it is more approachable.

web/book/src/internals/modules.md
aljazerzen: I tried to redesign it to come closer to PHP's `include()` or Common.js's `require()` and I've realized that this is a very close already.

The main distinctions are:
- we have `module hello` declaration instead of a *function* `include("hello.prql")`,
- we have more restrictions on importing that make the module structure more "normalized" (see goal 4),
- these restrictions also allow cyclic references between modules,
- child modules don't have to re-import modules that are already imported into parent modules:
  ```
  # parent.prql
  module child1
  module child2
  ```
  ```
  # child1.prql

  # no need to re-import child2, as it is already in the module structure
  let _ = (from child2.some_decl | ...)
  ```

aljazerzen:
I've updated the RFC with all the suggestions. I hope I've also made it more readable @snth :D

web/book/src/internals/modules.md
max-sixty: (edited after reading more)

When does this know when to stop going up the file tree?

max-sixty: There's some possible modules-as-inheritance approaches from the implicit shadowing.

But also possible some quite confusing error messages when a parent changes or adds a definition!

web/book/src/internals/modules.md
```diff
@@ -0,0 +1,311 @@
...
+
+## Main var declaration
+
+If last declaration in a module is a variable declaration that is named `main`,
+then the leading `let main = ` can be omitted and expressed only by the
+expression itself.
```
max-sixty:
```suggestion
If the final declaration in a module is a variable declaration that is named
`main`, then the leading `let main = ` can be omitted and expressed only by the
expression itself.
```

web/book/src/internals/modules.md
```diff
@@ -0,0 +1,311 @@
...
+
+```
+module my_playlists {
+    let bangers = (from tracks | take 10)
+
+    from playlists | take 10
+}
+
+let album_titles = my_playlists.main
```
max-sixty, line 104: Is this overall feature helpful though? Why not just have a final unassigned expression take main.

So could this just be:

```suggestion
my_playlists
```

?

web/book/src/internals/modules.md
```diff
@@ -0,0 +1,311 @@
...
+When a module is referenced as a value, the `main` variable is used instead.
+This is especially useful when referring to a module which is to be compiled to
+RQ (and later SQL).
+
+## File importing
+
+To include PRQL source code from other files, one can use the following syntax:
+
+```
+module my_playlists
```
max-sixty, line 116: Nitty, but to me this look a bit like a definition rather than an import; relative to `import` or `use`. No strong view tho

web/book/src/internals/modules.md
```diff
@@ -0,0 +1,311 @@
...
+> This is a technical document. For a "how to use" or a TLDR; skip to the
+> [Example](#Example) section.
+
+Design goals:
+
+1. Allow importing declarations from other files.
+
+2. Have namespaces for things like `std`.
+
+3. Have hierarchical structure so we can represent files in directories.
```
max-sixty:
```suggestion
3. Have a hierarchical structure so we can represent files in directories.
```
(tell me if these tweaks on idioms are helpful / annoying)

web/book/src/internals/modules.md
```diff
@@ -0,0 +1,311 @@
...
+}
+
+let album_titles = my_playlists.main
+```
+
+When a module is referenced as a value, the `main` variable is used instead.
+This is especially useful when referring to a module which is to be compiled to
+RQ (and later SQL).
+
+## File importing
```
max-sixty, line 116: What's the advantage of this relative to just including all PRQL files as modules automatically, like python does?

web/book/src/internals/modules.md
```diff
@@ -0,0 +1,307 @@
...
+## Compiler interface
+
+`prql-compiler` provides two interfaces for compiling files.
+
+**Multi-file interface** requires three arguments:
+
+- path to the file containing the module which is the root of the module
+  structure,
+- identifier of the pipeline that should be compiled to RQ and,
+- a "file loader", which can load files on-demand.
```
max-sixty, line 227: What's this third arg? Presumably `prqlc` is supplying its own? But what would alternatives be?

(Everything else seems perfect.)

web/book/src/internals/modules.md
```diff
@@ -0,0 +1,307 @@
...
+
+## Compiler interface
+
+`prql-compiler` provides two interfaces for compiling files.
+
+**Multi-file interface** requires three arguments:
+
+- path to the file containing the module which is the root of the module
+  structure,
+- identifier of the pipeline that should be compiled to RQ and,
```
max-sixty:
```suggestion
- identifier of the pipeline that should be compiled, defaulting to `main` or the final expression, and,
```

web/book/src/internals/modules.md
```diff
@@ -0,0 +1,307 @@
...
+`.prql` files starting with `_` in the current working directory.
+
+Example prqlc usage:
+
+```
+$ prqlc compile _project.prql sales.projections.orders_2024
+$ prqlc compile sales.projections.orders_2024
+```
+
+**Single-file interface** requires a single argument; the PRQL source. Any
```
max-sixty, line 239: Another way of conceptualizing this is that there's a new stage in the compiler before the current parser, where files are collected. It's possible to skip that stage by passing PRQL directly.

max-sixty:
Overall I'm very excited about this. I added some comments, and changed a few lines myself where it wasn't material (feedback welcome)

web/book/src/internals/modules.md
aljazerzen: The `let` statement is here just to highlight how `my_playlist.main` can be referenced.

web/book/src/internals/modules.md
aljazerzen: Yes, `prqlc` would supply the file loader itself and just implement it by reading from the file system.

web/book/src/internals/modules.md
aljazerzen: You could omit the `.main` (as described by the next paragraph), but if you omit `let album_titles = ` that would have a different semantics.

aljazerzen: I'm not sure if I understand your question here, though.

web/book/src/internals/modules.md
aljazerzen: What do you mean include automatically? In Python you still have to have an "import x from y" for each file. 

Even more so, if file a.py imports b.py file, some other file c.py has to re-import b.py. With our spec, c.py could just refer to `a.b` to get to the contents of b.py.

web/book/src/internals/modules.md
aljazerzen: Yes. That will fit better with the architecture of the compiler.

web/book/src/internals/modules.md
```diff
@@ -0,0 +1,324 @@
...
+        from decl_1 | join soundtracks.decl2 | join decl_3
+    }
+}
+```
+
+## Main var declaration
+
+If the final declaration in a module is a variable declaration that is named
+`main`, then the leading `let main = ` can be omitted and expressed only by the
+expression itself.
```
aljazerzen: `main` could also be `default` (as in CommonJS) or `self` (as in Rust, but in methods) or `this` (as in Java, but in classes).

WDYT?

web/book/src/internals/modules.md
max-sixty: OIC, sorry, this is clear in retrospect

web/book/src/internals/modules.md
max-sixty: OK, that makes sense — but in that case I'm confused by something else; I'll move the comment there

web/book/src/internals/modules.md
```diff
@@ -0,0 +1,324 @@
...
+```
+module my_playlists
+```
+
+Such declaration loads either `./my_playlists.prql` (a leaf module) or
+`./my_playlists/_my_playlists.prql` (a directory module) and uses its contents
+as the declarations in module `my_playlists`. If none or both of the files are
+present, a compilation error is raised.
+
+Only directory modules can contain module declarations. If a leaf module
```
max-sixty, line 132: I don't get this, then (from https://github.com/PRQL/prql/pull/2129/files#r1161422818) — and [here](https://github.com/PRQL/prql/pull/2129/files#diff-4923b398dd490faaaf22d6d981064e8fda7d70906cfb1c3af2144666a03cc86aR289) there's an import (declaration?) in a leaf module? (I'm sure I'm misunderstanding...)

web/book/src/internals/modules.md
```diff
@@ -0,0 +1,324 @@
...
+have at least one sub-module would not improve approachability enough to justify
+adding this restriction.
+
+For example, the following module structure is annotated with files names in
+which the modules would reside:
+
+```prql no-eval
+
+module my_project {
+    # _my_project.prql
```
max-sixty, line 152: How do we decide what the root module is called? What if there were a file here named `_my_project2.prql`?

web/book/src/internals/modules.md
```diff
@@ -0,0 +1,324 @@
...
+If module `my_project.sales` wants to add a submodule `util`, it has to be
+converted to a directory modules. This means that it has to be moved to
+`sales/_sales.prql`. The new module would reside in `sales/util.prql`.
+
+The annotated layout is not the only possible layout for this module structure,
+since any of the modules `sales`, `year_2023` or `year_2024` could be converted
+into a directory module with zero sub-modules.
+
+Point 4 of design goals means that each declaration within a project has a
+single fully-qualified name within this project. This is insured by strict rules
```
max-sixty:
```suggestion
single fully-qualified name within this project. This is ensured by strict rules
```

web/book/src/internals/modules.md
```diff
@@ -0,0 +1,324 @@
...
+	}
+
+	let main = (
+		from t = tracks
+		select [track_id, title]
+	)
+}
+```
+
+## Example
```
max-sixty, line 267: This is v clear, thanks

web/book/src/internals/modules.md
```diff
@@ -0,0 +1,324 @@
...
+
+    from playlists | take 10
+}
+
+let album_titles = my_playlists.main
+```
+
+When a module is referenced as a value, the `main` variable is used instead.
+This is especially useful when referring to a module which is to be compiled to
+RQ (and later SQL).
```
max-sixty, line 109: One thought — how do we know when it's referenced as a value? If we have:

```
let foo_mod = bar_mod
```

Are we basically saying we can never assign a module itself?

Maybe that's OK? Or maybe we need some way to assign different names if there are name conflicts?

web/book/src/internals/modules.md
```diff
@@ -0,0 +1,324 @@
...
+
+Design goals:
+
+1. Allow importing declarations from other files.
+
+2. Have namespaces for things like `std`.
+
+3. Have a hierarchical structure so we can represent files in directories.
+
+4. Have unambiguous module structure within a project.
```
max-sixty:
```suggestion
4. Have an unambiguous module structure within a project.
```

web/book/src/internals/modules.md
vanillajonathan: I think `main` or `default` are fitting, but I don't like `self` or `this` in this context.

web/book/src/internals/modules.md
aljazerzen: > When does this know when to stop going up the file tree?

Well, at the root? So `std` can be defined in the root only.

> But also possible some quite confusing error messages when a parent changes or adds a definition!

If we give declarations in child priority over declarations in parent, no change in parent will be able to overshadow the child declaration.

But it could happen that a module would be referring to a grand-parent module and a change in parent would break the child.

```
 grandparent -> decl
      parent -> we add decl here
      child  -> ref to decl now resolves to parent instead of grandparent
```

web/book/src/internals/modules.md
aljazerzen: From the second link, there is `sales/_sales.prql`, which is a directory module, so it can contain module declarations.

If we wanted `sales.orders` (which is a leaf module, because it is in `sales/orders.prql`) to have submodules, we'd have to convert it to directory module; move it to `sales/orders/_order.prql`.

Basically all Rust code I've seen already adheres to this rule, but it is not enforced, so there may be some code that does not. I feel like this is something that a beginner might do and it might be confusing, so I feel it's something we should forbid.

web/book/src/internals/modules.md
aljazerzen: Good question. I see two options:
- hardcode this to always be `_project.prql` (or something similar),
- allow any name, which would mean it will probably be the name of the project/workspace/gibberish. This would also mean it would be possible to have multiple entry-points into your project, which is something I'm -0.5 on. If there is a good use case for it then sure, but otherwise I'd avoid it.

web/book/src/internals/modules.md
aljazerzen: I think it's okay to not support assigning a module to a variable. It's not something that would have a value and type anyway, it's just a namespace.

In this case, we'd need another keyword for creating an alias to a module. Rust has the `use module::submodule::Declaration;`.

web/book/src/internals/modules.md
max-sixty: OK, I _think_ I get the root of my confusion. What is a "module declaration"? 

Is it `module foo` or the module file? (I had thought this line was saying we could only put `module foo` in directory modules, since that's how it's referred to in the paragraph above)

I'll add a couple of suggestions to clear this up

web/book/src/internals/modules.md
```diff
@@ -0,0 +1,324 @@
...
+RQ (and later SQL).
+
+```
+# last line from previous example could thus be shortened to:
+let album_titles = my_playlists
+```
+
+## File importing
+
+To include PRQL source code from other files, one can use the following syntax:
```
max-sixty:
```suggestion
To include PRQL source code from other files, we can use the following syntax:
```

web/book/src/internals/modules.md
```diff
@@ -0,0 +1,324 @@
...
+
+```
+module my_playlists
```
max-sixty, line 124:
```suggestion
use my_playlists
```

...no strong view on the exact term, feel free to close. 

My reasoning is that `module` might be used for an actual declaration later, like we have in the examples. And possibly it more clearly distinguishes the language away from "declaration" give my confusion below

web/book/src/internals/modules.md
```diff
@@ -0,0 +1,324 @@
...
+```
+
+Such declaration loads either `./my_playlists.prql` (a leaf module) or
```
max-sixty:
```suggestion
This loads either `./my_playlists.prql` (a leaf module) or
```

web/book/src/internals/modules.md
```diff
@@ -0,0 +1,324 @@
...
+`./my_playlists/_my_playlists.prql` (a directory module) and uses its contents
+as the declarations in module `my_playlists`. If none or both of the files are
```
max-sixty:
```suggestion
as module `my_playlists`. If none or both of the files are
```

web/book/src/internals/modules.md
```diff
@@ -0,0 +1,324 @@
...
+present, a compilation error is raised.
+
+Only directory modules can contain module declarations. If a leaf module
+contains a module declaration, a compilation error is raised, suggesting the
```
max-sixty, line 133: How would a leaf module contain a module declaration? Is this what `foo/bar.prql` would be?

web/book/src/internals/modules.md
max-sixty: I think starting with `_project.prql` would be very reasonable.

If we ever need a config file, such as `Cargo.toml` or `pyproject.toml`, then we could widen the requirements later...

web/book/src/internals/modules.md
```diff
@@ -0,0 +1,324 @@
...
+2. Have namespaces for things like `std`.
+
+3. Have a hierarchical structure so we can represent files in directories.
+
+4. Have an unambiguous module structure within a project.
+
+## Definition
+
+A module is a namespace that contains declarations. A module is itself a
+declaration , which means that it can contain nested child modules.
```
max-sixty:
```suggestion
declaration, which means that it can contain nested child modules.
```

web/book/src/internals/modules.md
```diff
@@ -0,0 +1,324 @@
...
+> the future.
+
+## Name resolution
+
+Any declarations within a module can be referenced from the outside of the
+module:
+
+```prql no-eval
+# using module structure declared above
+
```
max-sixty:
```suggestion
# using module structure declared above
module my_playlists

```
(or `use`)

web/book/src/internals/modules.md
```diff
@@ -0,0 +1,324 @@
...
+## File importing
+
+To include PRQL source code from other files, one can use the following syntax:
+
+```
+module my_playlists
+```
+
+Such declaration loads either `./my_playlists.prql` (a leaf module) or
+`./my_playlists/_my_playlists.prql` (a directory module) and uses its contents
```
max-sixty, line 128: How set are you on having the leading underscore?

One (mostly aesthetic) issue is that `_foo` is frequently considered to mean "private" — e.g. throughout python, rust vars.

Does having it first in a file listing justify this downside? Personally I'm rarely looking through an alphabetical list of files...

web/book/src/internals/modules.md
max-sixty: > > When does this know when to stop going up the file tree?
> 
> Well, at the root? So `std` can be defined in the root only.

This is moot if we have a file that can only be at the root, such as `_project.prql` or `project.prql` — we can stop there

But otherwise we need some way of knowing when to stop searching upwards, so we don't keep going to `/`. We could do that when we no longer find a match of `$dir` containing `_$dir.prql`...

I also think this "keep searching up" would be fine to push into a later version if it's not trivial to implement — very reasonable to adjust at implementation time. Though we'd need a way to reference a parent module, such as `module ..foo`.

aljazerzen: Oh, this is a tad different thing I was talking about here.

Here I meant name resolution only, which happens after the module structure has been read from files, parsed and resolved.

When deciding what files to load, the compiler should get the path to the root module and then find references to other files which need to be loaded from there.

web/book/src/internals/modules.md
aljazerzen: module declaration is `module foo`. And yes, that's what I meant: you can only put `module foo` into directory modules such as the one that is loaded from `sales/_sales.prql`.

If you'd put `module foo` into `sales/orders.prql`, you'd get this error:
```
module `sales.orders` is not a directory module, but it declares sub modules.

Hint: convert it to a directory module by moving file `sales/order.prql` to `sales/orders/_orders.prql`
```

web/book/src/internals/modules.md
aljazerzen: ... or even declare configs in the `_project.prql` itself?

web/book/src/internals/modules.md
aljazerzen: I like the synergy between "actual declaration" `module foo { ...decls ... }` and "file import declaration" `module foo`. They are doing the same thing after all, it's just that the latter it loading it from a file.

web/book/src/internals/modules.md
aljazerzen: Using the same annotation syntax:
```
module foo {
    # foo/_foo.prql

    module bar {
        # foo/bar.prql

        module hello {
            # foo/hello.prql  <- right next to _foo.prql!
        }

        module world {
            # foo/world/_world.prql  <- foo/world/ is next to foo/_foo.prql
            #                          even though world is a grandchild of foo!
        }
    }

}
```

web/book/src/internals/modules.md
aljazerzen: My preference lies with distinction from files that contain submodules. It should be easy to spot the parent file.

`mod.rs` is a poor attempt at that and there is nothing special about `_`.

Looking at available symbols, there is no better option. Maybe `foo/*foo.prql`?

web/book/src/internals/modules.md
max-sixty: > ... or even declare configs in the `_project.prql` itself?

Though PRQL isn't a great format for structured data — if we want to enable external dependencies / etc...

web/book/src/internals/modules.md
max-sixty: OK, I'm looking at the Example now — I should have started there, sorry for this noise — that's the clearest way of anchoring the discussion. 

I see the intention now, I'll comment there.

web/book/src/internals/modules.md
```diff
@@ -0,0 +1,324 @@
...
+	}
+
+	let main = (
+		from t = tracks
+		select [track_id, title]
+	)
+}
+```
+
+## Example
```
max-sixty, line 267: Re all the confusion above — sorry — the example is actually the most anchored thing to discuss from; I should have started here.

My main question is:

Why place the `module foo` declarations in the directory modules, rather than have import statements near the code that uses them?

- It doesn't supply _that_ much new information — just that we should use the file that's already there. We're replicating the file structure in declarations that are arguably duplicative
- Instead, the import statement can just look up the file — same as python.
- How do we discriminate between `from foo.bar` where `foo` is a DB schema vs. a module (assuming I eventually solve #1535)? If we have `use foo` in the file itself, it's explicit.

So by moving from "declarations in directory modules" to "import statements at the call site", we'd have something like:

```
# _project.prql

# We can have code here, but there's no need for `module` declarations
```

```
# This is identical to below
# employees.prql
let employees = (...)
let salaries = (...)
let departments = (...)
```

```
# sales/orders.prql
use ../employees  # or some variant of `../` that shows it's a level above

let current_year = (...)
let archived = (...)
let by_employee = (from orders | join employees.employees ...)
```

WDYT? I hope I'm making more sense now...

web/book/src/internals/modules.md
max-sixty: Given file naming constraints, we're quite limited. 

Having `foo/*foo.prql` would be quite the troll given shell semantics.

Fine to proceed with the underscore; we can easily change later if we learn new things.

web/book/src/internals/modules.md
max-sixty: Yeah, this was based on my giant misunderstanding about how this worked, sorry. Resolving

web/book/src/internals/modules.md
vanillajonathan: Do we even need submodules inside directories?
Can't we just import single files instead? So that `import foo` imports `foo.prql` not `foo/_foo.prql`.

web/book/src/internals/modules.md
vanillajonathan: I agree with that it should just directly load a file with the same name instead of look inside a subdirectory.

I am a bit skeptical against having imports from parent directories though (as you mention with `../`). I lean towards that files should only be able to use other files in the same directory and child directories but not in parent directories.

max-sixty: > I lean towards that files should only be able to use other files in the same directory and child directories but not in parent directories.

Surely this would be super restrictive — any sort of utils package wouldn't be possible, unless every file was in the same path? Are there any other languages that do this? (Or am I misunderstanding the point?)

web/book/src/internals/modules.md
aljazerzen: Not now, but if we settle on syntax for arrays and tuples, I think we are pretty close to typed JSON.

But this is a digression in this PR.

web/book/src/internals/modules.md
aljazerzen: @vanillajonathan Because we want to place .prql files into directories? Having all project files in a single directory is not scalable.

web/book/src/internals/modules.md
aljazerzen: > Why place the module foo declarations in the directory modules, rather than have import statements near the code that uses them?

Because with this proposal:
- you only "import"/"declare" a module once and then it's in the module structure where you can use it simply by referencing that name, no `use` or `module` statements required,
- you don't need a `import ../foo` (i.e. import parent) statement, because modules are imported/declared from root down (but you can still reference parent modules after they are imported/declared),
- refactoring is simpler, since there is a single import/declaration of a module,

> How do we discriminate between from foo.bar where foo is a DB schema vs. a module (assuming I eventually solve 
Resolve opaqueness of idents #1535)? If we have use foo in the file itself, it's explicit.

Let's discuss that after this RFC is done. I have an idea how to do it on top of modules and declarations.

web/book/src/internals/modules.md
eitsupi: I don't understand the need to repeat `foo` here (like `foo/_foo.prql`).
Is there any reason not to fix it with something like the special name `_main.prql`? (like `foo/_main.prql`)

web/book/src/internals/modules.md
vanillajonathan: > Surely this would be super restrictive — any sort of utils package wouldn't be possible, unless every file was in the same path? Are there any other languages that do this? (Or am I misunderstanding the point?)

C# and Java imports using namespaces so nothing can be from a parent directory. PHP lets you [`include`](https://www.php.net/manual/en/function.include.php) from any file path including relative paths which may be parent directories. JavaScript (Node.js) lets you import from parent directories. I am not sure about Python, but I don't think it is common. I also worry about the security impact of importing from parent directories.

In the PHP world when using Composer it is common that imports from your own code is in your `your_project/src/` directory, and third-party code from a `your_project/vendor/` directory.

web/book/src/internals/modules.md
aljazerzen: `foo/mod.prql` was in the initial proposal, see this comment and the following thread: https://github.com/PRQL/prql/pull/2129#pullrequestreview-1336188529

(maybe I should add "refused alternatives" section)

eitsupi: Thanks for the reply. Sorry I missed that one.

max-sixty: (marking as resolved given I started the thread)

vanillajonathan: > @vanillajonathan Because we want to place .prql files into directories? Having all project files in a single directory is not scalable.

Sure, but I am thinking `import foo` imports `foo.prql` and `import foo/bar` imports `foo/bar.prql`, instead of having `import foo` import `foo/_foo.prql`. This way you could still structure files in many directories, so it is scalable.

web/book/src/internals/modules.md
max-sixty: > Because with this proposal:
> 
>     * you only "import"/"declare" a module once and then it's in the module structure where you can use it simply by referencing that name, no `use` or `module` statements required,
> 
>     * you don't need a `import ../foo` (i.e. import parent) statement, because modules are imported/declared from root down (but you can still reference parent modules after they are imported/declared),

The main issue I see is that the action is quite far from the consequence. When we're missing an item in python, it's fairly obvious where the item should be:

```
from ..utils import read_file
```

"Couldn't find read_file in utils" — OK great

But in this design, what do we say if something isn't found? There are so many places we might be missing a declaration. The user needs to understand the module system well enough to then go and check each parent in the hierarchy for the missing item.

>     * refactoring is simpler, since there is a single import/declaration of a module,

Indeed — I can imagine refactorings being quite complicated, because the compiler can no longer tell you "we can't find this item here". I'm fine pushing the conversation re "DB schema vs. a module", but this will further confuse things, since `from foo.bar` will also be legitimate SQL even when `foo` isn't available as a module.

Or are my fears overblown? I can definitely see that it's an elegant system when everything is ideal — but I can also imagine it being quite confusing with newer users or messier projects.

---

My secondary — much more minor — concern is that the `module` declarations aren't doing _that_ much. IIUC they can only refer to a single file location. Is requiring them (relative to just finding them in the file tree) going to make things clearer?

---

To put this into the example, I would do something more like python, with a touch of rust, and do:

```
# _project.prql
```

```
# employees.prql
let employees = (...)
let salaries = (...)
let departments = (...)
```

```
# sales/_sales.prql
let revenue_by_source = (...)
```

```
# sales/orders.prql
use ..employees

let current_year = (...)
let archived = (...)
let by_employee = (from orders | join employees.employees ...)
```

```
# sales/projections.prql
use ..orders

let orders_2023 = (from orders.current_year | append orders.archived)
let orders_2024 = (...)
```

```
# util.prql
func pretty_print_num col -> (...)
```

web/book/src/internals/modules.md
```diff
@@ -0,0 +1,325 @@
...
```
max-sixty, line 267: @vanillajonathan starting a new thread for the "import parent" discussion:

> > Surely this would be super restrictive — any sort of utils package wouldn't be possible, unless every file was in the same path? Are there any other languages that do this? (Or am I misunderstanding the point?)

> C# and Java imports using namespaces so nothing can be from a parent directory. PHP lets you [include](https://www.php.net/manual/en/function.include.php) from any file path including relative paths which may be parent directories. JavaScript (Node.js) lets you import from parent directories. I am not sure about Python, but I don't think it is common. I also worry about the security impact of importing from parent directories.

> In the PHP world when using Composer it is common that imports from your own code is in your your_project/src/ directory, and third-party code from a your_project/vendor/ directory.

Ah, so I think that's where the misunderstanding is — I agree it might be dangerous to allow importing from a parent outside of the package.

Python allows for importing parents within the package — e.g. `from ... import utils`, because the path is relative to the current file. IIUC Java allows for `import utils` in a file `src/foo/bar/baz.java` from a file `src/utils.java`, because the path is relative to the package root.

So given this, I agree possibly we should downweight being able to import from a parent outside the package, which might require a definition of what a package is. (Though I'm not sure, let's see how the other discussions shake out; possibly it's a nice feature to be less tethered to the concept of a package)

web/book/src/internals/modules.md
aljazerzen: That's a fair point, declarations are far from where they are then used. In essence, the problem with my proposal is that "the module hierarchy" is a whole concept you have to know about as opposed to just knowing how to pull a declaration from one file into another.

I've purposefully made the module system repr in filesystem normalized, so we can have very good hints about where to put your module declarations:

```
select [sales.utils.my_func my_column]
              -----
                 \___ name utils does not exist
Hint: add `module utils` to `sales/_sales.prql`
```

```
module orders
       ------
          \__ cannot load sales/orders.prql or sales/orders/_orders.prql
```

My point is that knowing about "the module hierarchy" is not a requirement for anyone, because we can provide hints to guide users until it makes sense to them.

----

My fear is that it will end up having way-too much `use`s with the extreme case being:

```
# _project.prql
```

```
# employees.prql
use ..sales
use ..util

let employees = (...)
let salaries = (...)
let departments = (...)
```

```
use projections
use orders
use ..employees
use ..utils

# sales/_sales.prql
let revenue_by_source = (...)
```

```
# sales/orders.prql
use ..
use ..projections
use ...employees
use ...utils


let current_year = (...)
let archived = (...)
let by_employee = (from orders | join employees.employees ...)
```

```
# sales/projections.prql
use ..
use ..orders
use ...employees
use ...utils

let orders_2023 = (from orders.current_year | append orders.archived)
let orders_2024 = (...)
```

```
# util.prql
func pretty_print_num col -> (...)
```

web/book/src/internals/modules.md
```diff
@@ -0,0 +1,325 @@
...
+
```
aljazerzen, line 268: @max-sixty Another thread for "Opaqueness of idents".

----

(Bear with me, I need to explain all this to so I can explain how I'd change it)

Right now, when we infer that a table exists, we add its declaration to module `default_db` (this declaration is updated when we discover new columns).

Using a made-up syntax:
```
module default_db {
    table foo = ...
}
```

If a table name is segmented (i.e. `my_db.my_schema.foo`), we infer this:
```
module default_db {
    table `my_db.my_schema.foo` = ...
}
```
... essentially, the table name contains the dots, which is then problematic when actual dots are inserted into the name segments (i.e. ``my_db.`my.schema`.foo``).

----

Instead, I propose we infer this:
```
module default_db {
    module my_db {
        module my_schema {
            table foo = ...
        }
    }
}
```

max-sixty: Yes, very much agree. that's what I'd been trying to do in my branch for #1535 — when it finds a table type of `my_schema.foo`, it would create a module `my_schema` in the root (though should actually be in `default_db`, good point).

My branch is super-messy, but maybe I push it and you can criticize / guide it towards something helpful. I haven't worked on it in a couple of weeks now.

web/book/src/internals/modules.md
max-sixty: > I've purposefully made the module system repr in filesystem normalized, so we can have very good hints about where to put your module declarations:
> 
> ```
> select [sales.utils.my_func my_column]
>               -----
>                  \___ name utils does not exist
> Hint: add `module utils` to `sales/_sales.prql`
> ```

I agree this is a very good hint. But given how we would search upwards for a module, if `sales` isn't found — i.e. a "root" item, there could be lots of places that it could be.

One option would be to do completely automatic imports — no need for `module`, no need for `use`. I might even prefer that to only having `module` — which seems to ask for more than gives, at least until we had non-file `module`s.

My synthesis after thinking about this is:
- Just using a file structure is simpler than offering non-file modules (we agree)
- `use` statements are explicit, reduce confusion, not _that_ costly
- Given the language goals & users, we should favor simplicity. And imports are often pits of confusion.
- I do think your example of the hint is quite good, but I'm still worried about the ambiguity because of the hierarchy search, and whether the hint is so good that the language should infer it itself

WDYT?

Any thoughts from others? @eitsupi & @snth ? (Note — please read the comparative examples carefully, don't make the same mistake I did...)

web/book/src/internals/modules.md
vanillajonathan: That's a lot of indention. Maybe it could be:

    module default_db.my_db.my_schema {
        table foo = ... # just one level of indention
    }
    
Or just remove the indention all together and make one module per file:

    module default_db.my_db.my_schema;
    
    table foo = ... # no indention

But why does the module even need to be explicitly declared? The module could be the file, then the filename is the module name.

max-sixty: @vanillajonathan tbc, that's not real syntax, it's just a representation of the AST — there would be no way of declaring modules like that. Does that make sense?

vanillajonathan: Yes, that makes sense. My bad, I had misunderstood it.

web/book/src/internals/modules.md
```diff
@@ -0,0 +1,324 @@
...
+        from decl_1 | join soundtracks.decl2 | join decl_3
+    }
+}
+```
+
+## Main var declaration
+
+If the final declaration in a module is a variable declaration that is named
+`main`, then the leading `let main = ` can be omitted and expressed only by the
```
max-sixty: Looking at the example — should this instead be "The final statement in a module is implicitly named `main`"? In the example there is no variable declaration named `main`...

aljazerzen: That may be written poorly, I wanted to say that:
- if the last declaration in a module is a variable declaration,
- and it is named `main`,
- it may omit the leading `let main = `.

It can also be named something else, but it needs to have the `let x =` in this case.

Another way to say it, would be that:

```
Last unnamed pipeline in a module is implicitly named `main` 
```
... but this implies that we can have unnamed pipelines elsewhere in a module.

(ps: I've been using term "declaration" as a cover term for variable, function and type declarations. We can switch that to "statement" if you feel it is more appropriate or future-proof)

max-sixty: > * if the last declaration in a module is a variable declaration,
> 
>     * and it is named `main`,
> 
>     * it may omit the leading `let main = `.

I think this is the sticking point... it seems like the conclusion is required for the premise, since it can't be named `main` if it omits the leading `let main =`...
(sorry if I'm misunderstanding — my guess is that this is actually a small point of writing rather than some fundamental disagreement)


> ```
> Last unnamed pipeline in a module is implicitly named `main` 
> ```
> 
> ... but this implies that we can have unnamed pipelines elsewhere in a module.

How about a "An unnamed pipeline as the module's final declaration is implicitly named `main`" ?

aljazerzen: Oh, I see the problem! In my mind, the last pipeline remains named `main` even if you remove the leading `let main =`. That may be confusing, yes.

> How about a "An unnamed pipeline as the module's final declaration is implicitly named main" ?

That is still implying there are unnamed pipelines.

Last variable declaration in a module can omit the leading `let main =` and acquire an implicit name `main`.

(yes this appears to be a minor point in writing)

web/book/src/internals/modules.md
aljazerzen: Nobody wants to weigh-in here :(

> I do think your example of the hint is quite good, but I'm still worried about the ambiguity because of the hierarchy search, and whether the hint is so good that the language should infer it itself

Good point, it may be more confusing than helpful.

> One option would be to do completely automatic imports — no need for module, no need for use. I might even prefer that to only having module — which seems to ask for more than gives, at least until we had non-file modules.

That's probably the safest way to go here. It's also the easiest to implement, so would't be wasting much work if we decide to change later.

max-sixty: > Nobody wants to weigh-in here :(

Yeah, one issue with these big discussions is that it's more difficult for folks to weigh in on some smaller but important issues — there are 114 comments — and it's difficult to know how much context is required for weighing in. I don't know a great way around it...

> That's probably the safest way to go here. It's also the easiest to implement, so would't be wasting much work if we decide to change later.

Great, very much agree. 

I also think that approach is the best way of getting more feedback — merge this with a path forward, and once we start building, folks will raise anything sufficient important...

web/book/src/internals/modules.md
max-sixty: > Last variable declaration in a module can omit the leading `let main =` and acquire an implicit name `main`.

Perfect

(nitty: I'd use "final" rather than "last" as it's less ambiguous, though here it's fairly clear given the "in a module" modifier)

aljazerzen:
I think that we have a consensus on the outline for modules. There are things left to discuss, but we can get to them in separate issues.

max-sixty:
Great work!

Issue #2065, opened 2023-03-09
`deb`, `rpm` & `snap` tasks failing


max-sixty:
I think this is probably a small issue, around quoting things: https://github.com/PRQL/prql/actions/runs/4371758464/jobs/7647949921

I had thought we'd tested them, possibly something small changed after we merged

Would be great to fix them, and then to use them to start publishing release assets — then they would be quite useful. If we can't fix them, we should temporarily disable them until then work.

CC @vanillajonathan (I know I copied you on another issue today — hope that's OK — we're very grateful for your contributions and hope they continue!)

vanillajonathan:
Okay I investigated this.

As for the rpm and deb packages the package description (the `desc` field in the YAML file) spans multiple lines. We used the `|` syntax and that didn't work good, I tried to use the `>` syntax which worked better for the rpm package, but still fails on the deb package because they handle string replacement in a slight different way.

See https://yaml-multiline.info/ for a description of various ways for doing multi-line strings in YAML.
I filed a issue with `build-deb-action` at jiro4989/build-deb-action/issues/43.

So maybe we have to shorten the description or have just one really long line without line breaks.

As for the Snap I believe it fails because you have not declared a secret with the name `STORE_LOGIN`. By the way, since we might want to publish this to some other software store, we might want to rename this to `SNAP_STORE_LOGIN` or `SNAPCRAFT_STORE_CREDENTIALS` or something.

max-sixty:
> As for the Snap I believe it fails because you have not declared a secret with the name `STORE_LOGIN`. By the way, since we might want to publish this to some other software store, we might want to rename this to `SNAP_STORE_LOGIN` or `SNAPCRAFT_STORE_CREDENTIALS` or something.

Apologies; I had called it a different name. I've renamed it in #2072

max-sixty:
These failed again, I think the `deb` & `rpm` for the same reasons.

The `snap` failed because it can't find the icon: https://github.com/PRQL/prql/actions/runs/4402023644/jobs/7708766674

Do you know what the correct relative path should be there, now that it's in the `packages` path?

vanillajonathan:
Yes, the `deb` and `rpm` packages failed for the same reason as last time. I described why in the post above. I contacted the developer of the build-deb-action but he hasn't replied yet. We could fix it by changing the description for the `.deb` package from a multi-line description to a single-line description. For the `.rpm` package we can keep the multiline description by changing the YAML syntax from `|` to `>`.

[...]

PR #2135, opened 2023-03-13
refactor: Move `snap` to `packages` path

max-sixty:
...in line with the broader refactor.

I'm not sure of the best way of testing this; our current `release.yaml` isn't really set up to test things. Possibly we wait until the next release and fix if it's broken -- not ideal but maybe the lowest cost approach?

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -88,6 +88,8 @@ jobs:
         uses: actions/checkout@v3
       - name: 📦 Build Snap
         uses: snapcore/action-build@v1
+        with:
+          path: packages/
       - name: 🆙 Publish Snap
         uses: snapcore/action-publish@v1
         env:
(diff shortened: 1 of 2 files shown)


max-sixty:
@vanillajonathan lmk if you have any thoughts...

vanillajonathan:
You could create a new `snap-experiment.yaml` file and move the Snap stuff from `release.yaml` into that file, and open a PR, then GitHub Actions will run on the PR if you declared in the YAML file that it should trigger on PR.

max-sixty:
Yes, I agree that would work, but it's some overhead.

Possibly we change `release.yaml` to be able to run on PRs without doing an actual release. But unclear whether we'd get a good enough test.

For the moment, if you want to fix #2065, that would be great — my guess is that once they're working, they won't require much maintenance.

Issue #2122, opened 2023-03-12
Add `build-web` test


max-sixty:
### What's up?

It failed on `main` after my recent refactor PR, but it's not in CI

max-sixty:
Closed by #2147

PR #2144, opened 2023-03-13
revert: "feat: Code snippets from book in Playground (#2070)"

max-sixty:
This is causing the build to break and so we can't deploy 0.6.1 to the web. It's completely my fault; we should have had a test for this; ironically I added an issue for this yesterday https://github.com/PRQL/prql/issues/2122.

I just spent 20 mins trying to fix the issue, but couldn't manage, and need to log off for the night, so reverting this as per our standard approach. I'm hoping it won't be a difficult fix and we can re-revert this and add it back easily.

The error is:

```
src/sidebar/Sidebar.js
  Line 30:10:  'v' is assigned a value but never used  no-unused-vars
```


max-sixty:
CC @Jelenkee — as I say above, this is our [standard approach — merge quickly and revert quickly](https://prql-lang.org/book/contributing/development.html#merges) — and I hope we can get this back tomorrow

PR #2153, opened 2023-03-13
docs: Fix book schema presentation

max-sixty:


max-sixty:
Surprising that such a small mistake has such a large impact:

image

PR #2159, opened 2023-03-13
build: Fix `deb`, `rpm` & `snap` tasks failing

### LINKED ISSUE #2065: `deb`, `rpm` & `snap` tasks failing

vanillajonathan:
Closes #2065

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -45,7 +45,7 @@ jobs:
           package_root: .debpkg
           maintainer: The PRQL Project
           version: ${{ github.ref }}
-          desc: |
+          desc: >
             prqlc is the CLI for the PRQL compiler. It compiles PRQL to SQL, and offers various diagnostics.
 
             PRQL is a modern language for transforming data — a simple, powerful, pipelined SQL replacement.
@@ -71,7 +71,7 @@ jobs:
           package_root: .rpmpkg
           maintainer: The PRQL Project
           version: ${{ github.ref }}
-          desc: |
+          desc: >
             prqlc is the CLI for the PRQL compiler. It compiles PRQL to SQL, and offers various diagnostics.
 
             PRQL is a modern language for transforming data — a simple, powerful, pipelined SQL replacement.
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -75,6 +75,7 @@ jobs:
           package: prqlc
           package_root: .rpmpkg
           maintainer: The PRQL Project
+          vendor: The PRQL Project
           version: ${{ github.ref_type == 'tag' && github.refname || 0 }}
           desc: >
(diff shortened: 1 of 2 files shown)


vanillajonathan:
@max-sixty `pre-commit.ci` is being mean.

max-sixty:
> @max-sixty `pre-commit.ci` is being mean.

Yeah, it can't fix workflow files, unfortunately ("GitHub prevented pre-commit.ci from autofixing this pr due to autofixes to a workflow file")

Do you have pre-commit locally? Otherwise I can fix...

max-sixty:
One quick way of testing this is to add a `pull_request` trigger for this workflow. We must remember to delete it before we merge though!

As long as it's on a fork (like this PR), it can't do any actual releasing, so it's "safe"

vanillajonathan:
No, I don't have it locally. Please feel free to fix it.

max-sixty:
OK, do you want to try adding the `pull_request` trigger so we can test this? 

Then I can run pre-commit before merging (but also try and get it locally, it should be very tractable to install and will let you be more independent!)

max-sixty:
`deb` & `rpm` look good, as far as they get! 

`snap` still having issues finding the correct path...

.github/workflows/release.yaml
```diff
@@ -71,7 +76,7 @@ jobs:
           package_root: .rpmpkg
           maintainer: The PRQL Project
           version: ${{ github.ref }}
```
max-sixty: Try this so we can test on non-tags. I'm not confident it'll work.

```suggestion
          version: ${{ github.ref_type == 'tag' && github.refname || 0 }}
```

max-sixty:
OK great, so that version approach works — worth also copying that to the others. And then we're back to the error around the description https://github.com/PRQL/prql/actions/runs/4410531443/jobs/7728094837#step:6:39

max-sixty:
`rpm` works! Can we get `deb` to pass too?

vanillajonathan:
The multi-line descriptions now works since I reported the issue upstreams and they've since been fixed.

Both the `.deb` and `.rpm` now also builds on PR with the version fallback you provided.

I had some problems with the Snap so now I copy it out from `packages/` to the project root in the GitHub pipeline and run it from the project root and then it works. The job fails fails but this is because secrets are not available in PR so this is not a real problem.

max-sixty:
@vanillajonathan super! Merging, thank you

PR #2167, opened 2023-03-13
ci: Limit docker caching in devcontainer to `main`

max-sixty:
Applies the logic in #2166. 

This doesn't run that often so is less important to exclude, adding it regardless.

I also simplified the workflow a bit.

### DIFF
--- a/.github/workflows/devcontainer.yaml
+++ b/.github/workflows/devcontainer.yaml
@@ -5,6 +5,9 @@ on:
   # https://github.com/PRQL/prql/pull/1893#discussion_r1125736478, possibly only
   # publishing on releases or on the `web` branch (which possibly should be
   # renamed back to `stable` if we use it more for this)
+  #
+  # TODO: integrate the Taskfiles — the extra complication of multiple files is
+  # not worth reducing the number of workflows we run
   push:
     paths:
       - .devcontainer/base-image/Dockerfile
--- a/.github/workflows/devcontainer.yaml
+++ b/.github/workflows/devcontainer.yaml
@@ -53,7 +53,7 @@ jobs:
           tags: ${{ steps.meta.outputs.tags }}
           # TODO: add linux/arm64
           platforms: linux/amd64
-          push: ${{ github.ref_name == 'main' && 'true' || 'false' }}
+          push: ${{ github.ref == 'refs/heads/main'}}
           cache-from: |
             ${{ env.IMAGE_NAME }}
             type=gha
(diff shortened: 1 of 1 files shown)


.github/workflows/devcontainer.yaml
```diff
@@ -43,25 +43,17 @@ jobs:
...
-            echo "push=true" >>$GITHUB_OUTPUT
-          else
-            echo "push=false" >>$GITHUB_OUTPUT
-          fi
-
       - name: Build
         uses: docker/build-push-action@v4
         with:
           file: "./.devcontainer/base-image/Dockerfile"
           tags: ${{ steps.meta.outputs.tags }}
```
max-sixty, line 53: Does this push to all tags? That seems unusual (but no real downside)?

eitsupi: I'm sorry, but I don't understand what your concern is.
I think it is common to specify multiple tags in `docker buildx build` command and push them at once.

Note that this is copied from the README of the docker metadata action.
https://github.com/docker/metadata-action
For now, only the `latest` raw tag is set, so the output is always `ghcr.io/prql/prql-devcontainer-base:latest`.
(No processing is done other than converting uppercase to lowercase!)
https://github.com/PRQL/prql/blob/e0d90d3a45fa5ca217c1e35a604d9ba9af0d7fbf/.github/workflows/devcontainer.yaml#L26-L31

.github/workflows/devcontainer.yaml
```diff
@@ -43,25 +46,17 @@ jobs:
...
           # TODO: add linux/arm64
           platforms: linux/amd64
-          push: ${{ steps.set-options.outputs.push }}
+          push: ${{ github.ref_name == 'main' && 'true' || 'false' }}
```
eitsupi: I often see the following. How about that?

```suggestion
          push: ${{ github.ref == 'refs/heads/main'}}
```

.github/workflows/devcontainer.yaml
```diff
@@ -43,25 +46,17 @@ jobs:
...
           cache-from: |
             ${{ env.IMAGE_NAME }}
             type=gha
           cache-to: |
+            ${{ github.ref_name == 'main' && 'type=gha,mode=max' || '' }}
```
eitsupi, line 61: Sorry, but I cannot be certain this will work.

eitsupi (approved):
Thank you for taking on the job I should have done!

max-sixty:
> Thank you for taking on the job I should have done!

Not at all, each of us makes incremental progress!

.github/workflows/devcontainer.yaml
max-sixty: We can try; I'm not sure whether it will interpret 0 and 1 correctly...

.github/workflows/devcontainer.yaml
max-sixty: Ok, let's try and we can revert...

PR #2172, opened 2023-03-14
internal: Remove nested enums, enabling RQ yaml

max-sixty:
This seems like the tail wagging the dog, but I wanted to try and produce more RQ and:
- The JSON representation is very verbose
- We can't produce YAML because it doesn't support nested Enums. I've spent some time researching this and can't find a solution.

So this is an attempt to remove them. If we don't think it has much impact, we can merge a version of this.


aljazerzen (approved):

max-sixty:
OK, simplified a bit and added comments, now planning to merge if you agree.

I left the old `Enum` there with a note. That said, it won't be trivial to undo — we'll have to undo the yaml serialization, which I'm hoping to start using more — are we sure we're OK with that?

One reason to think it's OK — it's not _that_ bad to just flatten nested Enums. Not great — sometimes we'll want similar behavior for one group, which will make `match` expressions more complicated.

aljazerzen (approved):

aljazerzen:
This is not a big change the AST and it does allow serializing to yaml, so this is an improvement.

max-sixty:
Unfortunately this doesn't always work, because of https://github.com/PRQL/prql/blob/79e5529413fecc66b824d1936e89ac3889f248fb/prql-compiler/src/ast/pl/expr.rs#L51 — and that's much harder to flatten. I'll put an issue in upstream.

PR #2225, opened 2023-03-18
docs: Substitute nbsp ("\00a0") for "\1F892"

richb-hanover:
Fixes indentation for all sections of the Playground's sidebar

PR #2227, opened 2023-03-18
revert: "docs: Substitute nbsp ("\00a0") for "\1F892""

max-sixty:
Reverts PRQL/prql#2225 give tree behavior, we can continue refining


richb-hanover:
@max-sixty I have the disclosure triangles working, but will create a PR in the morning...

PR #2233, opened 2023-03-18
chore: Fix GHA reference in `snap` release

max-sixty:


max-sixty:
@eitsupi i'll fix tomorrow!

.github/workflows/release.yaml
```diff
@@ -93,7 +93,7 @@ jobs:
         uses: actions/checkout@v3
       - name: Move Snap to project root directory
         run: cp -r packages/snap/ .
-      - name: 📦 Build Snap
+      - name: build
```
eitsupi: The name does not need to be modified, perhaps the `id: build ` needs to be set.

```suggestion
      - name: 📦 Build Snap
        id: build
```

PR #2237, opened 2023-03-18
docs(book): fix typo of dbt integrations page

eitsupi:


eitsupi:
The changes 1aead27d6f5f81b1677527d330051a4a4b670747 made by pre-commit CI is clearly wrong changes.
But I don't know how to stop this...

Before

image

After

image

@max-sixty Do you know where these changes are coming from?
I have looked at the pre-commit settings and cannot tell which formatter is making this change.
~~The formatter clearly does not understand gfm etc.'s footnote syntax.......~~
-> Sorry, I was wrong. It renders as intended on GitHub. However, footnotes is not a GFM or CommonMark spec! footnotes syntax seems come from pandoc... https://pandoc.org/MANUAL.html#footnotes

eitsupi:
@max-sixty Could you please take a look at the pre-commit settings?

max-sixty:
> @max-sixty Could you please take a look at the pre-commit settings?

Weirdly, I just reverted the lines and it seemed to stop complaining. Not sure what was happening. Unless I'm missing something. Maybe something tweaky like indentation...

max-sixty:
> -> Sorry, I was wrong. It renders as intended on GitHub. However, footnotes is not a GFM or CommonMark spec! footnotes syntax seems come from pandoc... https://pandoc.org/MANUAL.html#footnotes

Yes, this is unfortunate.

I think for this doc it's OK that it's imperfect, unless there are better approaches, because the plugin hasn't been a success, and requires a deeper integration with dbt to be useful. If you think it's better to be aesthetically correct, it's also OK for us to remove it.

eitsupi:
Thanks, I agree. We can merge this for now.

I have opened follow-up issue #2248 about pre-commit formatting.

PR #2244, opened 2023-03-18
feat: Use `DISTINCT ON` for dialects which support it

max-sixty:
Unfortunately this doesn't work yet — it would be easier with upstream support.

I'm putting the PR out there regardless; I can close if it gets stale.


prql-compiler/src/ast/rq/transform.rs
```diff
@@ -12,6 +12,7 @@ pub enum Transform {
     From(TableRef),
     Compute(Compute),
     Select(Vec<CId>),
+    DistinctOn(Vec<CId>),
```
aljazerzen: This should probably go into `sql::preprocess::SqlTransform` (where Distinct already is).

prql-compiler/src/sql/preprocess.rs
```diff
@@ -135,18 +135,18 @@ pub(super) fn distinct(
...
 
+                dbg!(&ctx.dialect);
                 if take_only_first && sort.is_empty() && matching_columns {
                     res.push(Distinct);
-                    continue;
+                } else if dbg!(ctx.dialect.supports_distinct_on()) {
+                    res.push(dbg!(SqlTransform::Super(DistinctOn(partition))));
+                } else {
+                    res.extend(create_filter_by_row_number(range, sort, partition, ctx));
                 }
```
aljazerzen: This is really concise, nice.

max-sixty:
This somewhat works, but I need to come back and work out how to do it well:
- Should `DistinctOn` be in RQ?
- Should it be a sibling to `Distinct` in SRQ? Or an Enum within that? Or is that building a whole extra AST analogous to `sqlparser-rs`'s?
- What's the best way of doing the translation for the items?

If @aljazerzen has thoughts on these lmk, otherwise I'll come back to it soon solo

prql-compiler/src/sql/gen_query.rs
```diff
@@ -107,8 +107,29 @@ fn translate_select_pipeline(
     let order_by = pipeline.pluck(|t| t.into_sort());
     let takes = pipeline.pluck(|t| t.into_take());
     let is_distinct = pipeline.iter().any(|t| matches!(t, SqlTransform::Distinct));
+    let distinct_owns = pipeline.pluck(|t| t.into_distinct_on());
```
aljazerzen: `distinct_ons`?

prql-compiler/src/ast/rq/transform.rs
max-sixty: Yeah, this wasn't needed at all! Thanks for pointing that out...

prql-compiler/src/sql/gen_query.rs
```diff
@@ -107,8 +107,29 @@ fn translate_select_pipeline(
     let order_by = pipeline.pluck(|t| t.into_sort());
     let takes = pipeline.pluck(|t| t.into_take());
     let is_distinct = pipeline.iter().any(|t| matches!(t, SqlTransform::Distinct));
+    let distinct_ons = pipeline.pluck(|t| t.into_distinct_on());
     let distinct = if is_distinct {
         Some(sql_ast::Distinct::Distinct)
+    } else if !distinct_ons.is_empty() {
+        // FIXME: this "works" but is very hacky — it's using the
+        // `translate_select_item` to translate each id into an Expr — is that
+        // correct?
```
aljazerzen, line 116: There is `translate_cid` that gives `Expr` directly.

prql-compiler/src/sql/gen_query.rs
```diff
@@ -107,8 +107,29 @@ fn translate_select_pipeline(
...
+        //
+        // Do we know we won't have more than one DistinctOn? We have that for
+        // `Distinct`. But this will panic if we have more than one.
+        // And if there the `SelectItem` is not an `UnnamedExpr`, it'll panic.
```
aljazerzen, line 120: Yes, we know there won't be multiple `Distinct`s, but you also have to inforce that for `DistinctOn`. You can do that here: https://github.com/prql/prql/blob/main/prql-compiler/src/sql/srq/anchor.rs#L342-L353

After that, it is safe to do this:
```
let distinct_on = pipeline.pluck(|t| t.into_distinct_on()).into_iter().next();
```

aljazerzen (approved):

PR #2245, opened 2023-03-19
fix: Raise an error on an invalid header arg

### LINKED ISSUE #2230: Additional named args in header don't error
### What happened?

We don't raise an error on args in the header which aren't supported

### PRQL input

```elm
prql dialect:sql.duckdb  # This should be `target`, not `dialect`

from x
```

### SQL output

```SQL
SELECT
  *
FROM
  x

```

### Expected SQL output

```SQL
Error, unknown arg `dialect`
```

### MVCE confirmation

- [X] Minimal example
- [X] New issue

### Anything else?

_No response_

max-sixty:


max-sixty:
Would close https://github.com/PRQL/prql/issues/2230

But possibly there was a reason for this flexibility — I could imagine it being helpful? I think we should error _somewhere_ in the compilation — maybe the parser should be more flexible than the proposed code; but then let's error later.

eitsupi:
Is there a test for when more than one header is set?
Something like the following

```elm
prql target:sql.mssql version:"0.6"

from a
take 5
```

max-sixty:
Good idea @eitsupi — pushed

aljazerzen (approved):

aljazerzen:
Nice, great idea to restrict headers. If we would not restrict them, adding any new header would be a breaking change, because before someone may have relied on the fact that they are ignored.

PR #2234, opened 2023-03-18
build: enable actionlint's warning from MegaLinter

eitsupi:
Part of #1994
actionlint's warnings from MegaLinter have been fixed by #2232 and #2233, so the warnings can be activated.

(#2233 should be merged first)

PR #2247, opened 2023-03-19
chore: revert temporary megalinter settings from #2234

eitsupi:
Revert settings changes from #2234.
It would be easier to see if only the Linter associated with the file changed in PR is displayed in the results.

### DIFF
--- a/.github/workflows/mega-linter.yaml
+++ b/.github/workflows/mega-linter.yaml
@@ -41,9 +41,8 @@ jobs:
         env:
           # All available variables are described in documentation
           # https://megalinter.io/configuration/
-          # VALIDATE_ALL_CODEBASE: ${{ github.event_name == 'push' && github.ref
-          #   == 'refs/heads/main' }} # Validates all source when push on main, else just the git diff with main. Override with true if you always want to lint all sources
-          VALIDATE_ALL_CODEBASE: true
+          VALIDATE_ALL_CODEBASE: ${{ github.event_name == 'push' && github.ref
+            == 'refs/heads/main' }} # Validates all source when push on main, else just the git diff with main. Override with true if you always want to lint all sources
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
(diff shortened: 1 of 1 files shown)


max-sixty:
Ah, what was wrong with them?

eitsupi:
Not a big problem, but I am concerned that linter errors in files unrelated to the PR will show up and confuse the PR author.
For example, the possibility that an update to linter could cause an error that was not an error before.

(I intended to set this up in #2234 temporarily for testing purposes.)

max-sixty:
If we can lock the version of the linter and dependabot upgrades it, then this shouldn't be a problem — that should be the only source of change assuming that every file is run through this check to make it to main 

(but no strong view my end, if the diff logic is perfect then that's strictly equivalent. And if it's not then the overall outcome is still fine)

eitsupi:
> If we can lock the version of the linter and dependabot upgrades it, then this shouldn't be a problem — that should be the only source of change assuming that every file is run through this check to make it to main

You are right. However, I think we pinned the major version here, so changes may occur depending on the minor version.
But I don't have a strong opinion either.

As for the difference detection, it has been in use for years since Super-Linter, so it is reliable.
https://github.com/github/super-linter

If you think it is always better to inspect all codes, I am fine with that.
(Generally the fewer Linters we run, the less time it takes to run, but in my experience it seems to make a small difference here)

max-sixty:
> As for the difference detection, it has been in use for years since Super-Linter, so it is reliable. https://github.com/github/super-linter
> 
> If you think it is always better to inspect all codes, I am fine with that. (Generally the fewer Linters we run, the less time it takes to run, but in my experience it seems to make a small difference here)

Great! Whatever you think for going forward! 

I think it was good to do one pass with everything — I'm happy that `actionlint` passes now!

eitsupi:
> I think it was good to do one pass with everything

Just to be clear, at the moment all code is inspected when pushed to the main branch, so every time we merge PRs, all code is inspected.

max-sixty:
Ok great, thanks for the clarification.

Issue #1840, opened 2023-02-14
Feedback wanted: contributing to the complier


max-sixty:
As discussed with @aljazerzen:

**How could we better enable folks to make their first few PRs to the compiler?**

PRQL has had many successes over the past year — there's much we can be happy and be grateful for. But something we've done a bad job at is enabling folks to contribute to the compiler.

### We're not doing well / the trend is negative

For context, over the past six months the pattern of project's contributions have diverged:
- Overall, code contributions to PRQL have increased a lot and the number of contributors has broadened  🙌. But the broadening is focused on the parts _around_ the compiler — such as bindings, docs, integrations such as `prql-vscode`
- The concentration of code contributions _to the compiler_ has narrowed — I have done less, and @aljazerzen has done much much more.

My mental mode of contributions is there's a funnel — folks see a project, think it's good, maybe try using it, maybe try contributing to docs, maybe try adding some code. Then as they contribute more, they build the context to make increasingly more significant contributions. But if we can't enable them to start contributing to the compiler, they can't build the context to contribute more there. So our biggest challenge is to enable making the 1st / 2nd / 3rd compiler PR.

### This matters

The current state is bad for the project for a few reasons:
- *Code progress* — the project doesn't make the progress it needs to. We've got a lot to do — [Our Roadmap](https://prql-lang.org/roadmap/) — and even before the big feature development, we still have some bugs / bad error messages / etc.
- *Context building* — beyond losing the code of those who can't contribute, we're losing their ideas. There's a dramatic difference in the speed — and to some extent the quality — of discussion between those who have the deep context of the language and those who don't. We still welcome ideas from all, and there are other ways to build that context, but having written the actual code is likely primary.
- *Bus factor* — It puts stress on the existing contributor. While I hope we're each taking joy from the project, it's less joyful if lots of folks are waiting for a bug fix and you're the only person who can fix something.
- *Ratcheting* — There's a self-reinforcing effect. When one person writes all the code, the code becomes fit to their style, the incentives to make the code engageable are less strong — e.g. comments explaining the code, modularity, explicit interfaces. Multiple contributors brings inherent robustness.
- *Diversity* — It reduces the diversity of the team, and the project — which has externalities beyond just the project
- *Sharing* — We're not sharing the joy of open-source (at least for me, if other projects were less accessible, I wouldn't have started contributing to them eight years ago, and I'd have missed out on the tremendous opportunities that this world has brought me)

### Examples

[...]

Issue #1355, opened 2022-12-28
Add color option to CLI


max-sixty:
Currently the CLI can't take options such as `--color` / `--no-color`. Ideally we'd be able pass these, and they would flow through with an `Options` struct to the compiler.

A future version could check whether we're in a TTY and default to `--no-color` if so (it's also possible that clap offers something like this?).

Labeling as Good first issue (which also make good nth issues). As ever, lots of folks will be happy to answer questions!

Shuozeli:
Hi Max, 

I'd like to work on this. My thought of the steps for adding the option are:

1. Add one `enable_color` boolean field in the [Option](https://github.dev/PRQL/prql/blob/467d508a45ea1472f027d1da5ea7700cb0caae4c/prql-compiler/src/sql/mod.rs#L60) struct under the `sql/mod.rs`.

2. Add the cli argument `enable_color`, `color`, or `no-color` to the compile command.

3. Pass the cli argument as an `Option` field to the compile function

Please correct me if my understanding is wrong.

max-sixty:
That would be great @Shuozeli !

I think that sounds very reasonable.

I would pass a full `Options` struct to the `compile` function (I think that's what you were suggesting). And yes re the `--color` / `--no-color` flags.

There's a "smaller" version where we strip out the color at the CLI level, I think that would be easier but less good (but OK!)

@aljazerzen lmk if you have any thoughts!

aljazerzen:
Hmm, sql does not do any highlighting ATM, so it doesn't need color option (at least for now). But the cli argument would be use to control [this param](https://github.com/prql/prql/blob/c7df3d4cc1d9830cd6c96cfd1ddbc843139eff59/prql-compiler/src/error.rs#L186) in error formatting. This function is called 

The best way to do this would be:
- add a new Options struct to `lib.rs`:
  ```rust
  struct Options {
      sql: sql::Options,
      color: bool
  }
  ```
- change `compile(prql: &str, options: Option<sql::Options>)`
  to this `compile(prql: &str, options: Option<Options>)`
- add a `impl Default for Options` and `impl From<sql::Options> for Options` for convenience
- make sure that all crates compile and pass `cargo test`

max-sixty:
OK, so we'd have a nested `Options` struct. That seems very reasonable too. 

@Shuozeli does that make sense to you?

Shuozeli:
Yes, thank you for the clarification. I will be working on this one.

vanillajonathan:
[...]

PR #2251, opened 2023-03-19
feat: Add option to output colors

max-sixty:
This moves #1355 forward, and enables a PR to add colors to the book. The main thing to review is whether the proposed structure and naming of the options is reasonable.

~On the names — it's not that consistent — we have `signature_comment` — should this just be `colors` rather than `use_colors`? Or does `signature_comment` suggest it contains an actual comment? I've generally tried to make bools self-describing, but possible an options struct can be more implicit.~ 

Edit: this is now just a `color` property and `.with_color(self, color)` method, consistent with something like https://github.com/zesterer/ariadne/blob/ccd465160129d2546d67f7ee78727a4098a64330/src/lib.rs#L81. Possibly we should standardize around something like that

---

It's slightly different from the proposal in #1355  — it doesn't introduce another layer of options.
- I think possibly the code has changed slightly since then, since there was a discussion on whether to pass an `Options` struct to the `compile` function, which we now already do. 
- And I would vote to lean towards having a flatter, simpler, structure until we have lots of confidence in the eventual structure. 
  - (Separate point, ref #1840 — possibly that would have also made the initial contribution easier, which was well-described in #1355 but was a decent amount of work? Possibly we should be trying to make the initial contributions as tiny as possible, even at the cost of them requiring later refactorings?)

---

I'm not completely clear how we support colors in `prqlc` now...
- We're currently [passing `false` to the `composed` method](https://github.com/PRQL/prql/compare/main...max-sixty:prql:colors?expand=1#diff-d8cf470395314450935c363c0f82ef23888873d109e7df81e131728c66265e84L128-R128), which is passed through to ariadne [here](https://github.com/PRQL/prql/blob/107f3d71a56805c59fa217090d3d213a50711f8b/prql-compiler/src/error.rs#L246)
- But we are getting colors! 
  image
- Possibly some combination of [Ariadne's features](https://github.com/zesterer/ariadne#cargo-features) and our [`color_eyre::install` usage](https://github.com/PRQL/prql/blob/79e5529413fecc66b824d1936e89ac3889f248fb/prql-compiler/prqlc/src/cli.rs#L20) auto configures it?
  - Though then how would we implement #1355 if passing `false` doesn't pass through to ariadne?

---

This doesn't implement support in bindings yet.

### DIFF
--- a/bindings/prql-elixir/native/prql/src/lib.rs
+++ b/bindings/prql-elixir/native/prql/src/lib.rs
@@ -78,7 +78,7 @@ impl From<CompileOptions> for prql_compiler::Options {
             target: target_from_atom(o.target),
             signature_comment: o.signature_comment,
             // TODO: add support for this
-            use_colors: false,
+            color: false,
         }
     }
 }
--- a/bindings/prql-elixir/native/prql/src/lib.rs
+++ b/bindings/prql-elixir/native/prql/src/lib.rs
@@ -77,6 +77,8 @@ impl From<CompileOptions> for prql_compiler::Options {
             format: o.format,
             target: target_from_atom(o.target),
             signature_comment: o.signature_comment,
+            // TODO: add support for this
+            use_colors: false,
         }
     }
 }
--- a/bindings/prql-java/src/lib.rs
+++ b/bindings/prql-java/src/lib.rs
@@ -27,6 +27,8 @@ pub extern "system" fn Java_org_prql_prql4j_PrqlCompiler_toSql(
         format: format != 0,
         target: prql_dialect,
         signature_comment: signature != 0,
+        // TODO: add support for this
+        use_colors: false,
     };
     let result = prql_compiler::compile(&prql_query, &opt);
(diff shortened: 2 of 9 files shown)


max-sixty:
@aljazerzen given you're quasi-out for two weeks, give this a thumbs-down if you want me to wait — e.g. the options aren't named well — otherwise I'll merge. (better to wait than revert)

eitsupi:
Could it be listed in the Changelog? (Downstream packages also need to add the new field to Options)

max-sixty:
> Could it be listed in the Changelog? (Downstream packages also need to add the new field to Options)

Definitely, will do

More generally, if there's some check we could add — e.g. anything with a `fix` or `feat` requires an edit to the Changelog file, then I'd support that. (But I'm less keen on a separate tool like changie, I find that adds overhead to tiny PRs)

aljazerzen (approved):

max-sixty:
> I'm not completely clear how we support colors in `prqlc` now...

FYI this was resolved in #2267

PR #2286, opened 2023-03-22
test: more RDBMS integration tests in CI

### LINKED ISSUE #2211: Integration test for different DBs in CI
### What's up?

Hi,
I noticed that there is an integration test with SQLite, DuckDB and Postgres that does not run in CI. Is there a special reason for that?

I think it would be cool to run tests for different vendors like MySQL (or MariaDB) in CI as well. Even IBM DB2 and MS SQL Server have public docker images that can be used.
The tests should cover the syntactic/semanctic differences between theses vendors (like SUBSTR vs SUBSTRING) so that issues like #1854 can be discovered earlier.

We can adapt the workflow from objection.js which looks pretty simple
[workflow](https://github.com/Vincit/objection.js/blob/master/.github/workflows/test.yml)
[docker-compose](https://github.com/Vincit/objection.js/blob/master/docker-compose.yml)

It may not be necessary to fill the DBs with chinook before testing. Just a few records in a table may be enough.

Jelenkee:
I added integration test for different DBs (currently DuckDB, SQLite, MySQL, Postgres and SQL Server).
The testcases are in a separate text file which is read at build time. I created a setup sql file for all DBs. There are some hacks because it's not compatible with every dialect, but I guess it's ok for tests.

Closes #2211


max-sixty:
Overall this looks extremely impressive @Jelenkee !

I'll look more later but a couple of quick questions:
- One thing that's important is a bare `cargo test` still working without someone having docker installed. Does that work now, or is the integration test run as part of normal tests? We can think of a way of handling that; maybe with a feature.
- This seems to eschew the [existing integration tests](https://github.com/PRQL/prql/blob/ad2f9adeb02a1b89b952ffa292fd93359afe7e65/prql-compiler/tests/integration/README.md), is that correct? While those currently only run the in-process DBs in CI, there is a postgres version that should run once set up. The config isn't nearly as elegant as yours (e.g. it requires copying the build artifacts), it does have an existing set of queries and snapshots. Ideally we would share the test code & queries & snapshots between databases, both to ensure they work similarly and to improve maintainability.

Jelenkee:
* Docker is not required because the test is ignored per default
* The new tests do not cover everything from the old chinook test. That's why I did not delete them. Currently `insta` is not used to run the tests because I'm not familiar with it. So no snapshots are generated

max-sixty:
> Docker is not required because the test is ignored per default

Ah yes, I see, thanks!

> The new tests do not cover everything from the old chinook test. That's why I did not delete them. Currently `insta` is not used to run the tests because I'm not familiar with it. So no snapshots are generated

I'm quite hesitant for the project to maintain a different set of integration tests, with different queries & data. What do you think about working to integrate your database connection code, which seems like the big & difficult piece of this, with the existing tests' queries & data?

It doesn't need to be refactored to exactly the same pattern (i.e. current each DB is a `mod`: https://github.com/prql/prql/blob/da173c39bb1c3af90ac8681bec9bd5295c379432/prql-compiler/tests/integration/main.rs#L208) — instead it could just be `mod external_dbs` or something similar.

I'm happy to help if needed. FWIW I think you'll find libraries like insta really nice, it significantly reduces boilerplate, as well as making the whole PRQL library more maintainable.

---

Otherwise things look very good. I can have another look over later and confirm if that's helpful, so you don't feel like there's always another change request.

Jelenkee:
I added all of the old integration test to the new test file except for [invoice_totals](https://github.com/PRQL/prql/blob/main/prql-compiler/tests/integration/queries/invoice_totals.prql). I'm not sure if all of the DBs work with these s-string.

The problem is that the tests don't run successfully because MsSQL has problem with `limit` and MySQL has problems with `loop` (maybe others too). The behaviour should be fixed before this PR is merged.

You have to help me with the snapshotting stuff. I don't get, how and why those snapshots are generated and so on

prql-compiler/tests/integration-vendors/main.rs
```diff
@@ -0,0 +1,151 @@
...
+                con.run_query(sql.as_str(), runtime);
+            });
+
+        for (prql, expected_rows) in get_test_cases() {
+            let options = Options::default().with_target(Sql(Some(con.get_dialect())));
+            let sql = prql_compiler::compile(prql.as_str(), &options).unwrap();
+            let mut actual_rows = con.run_query(sql.as_str(), runtime);
+            replace_booleans(&mut actual_rows);
+            println!("{} {:?}", &con.get_dialect(), &actual_rows);
+            assert_eq!(
```
max-sixty: Here — can we run `assert_display_snapshot!(actual_rows, con.get_dialect().to_string)`. Running that with `cargo insta test --accept` will write the snapshots, and remove the need for the current `testcases.txt` and associated code

max-sixty (commented):
For sure, very happy to help.

FWIW I'm out this weekend so won't really be online for a few days, but back at the start of next week.

I think we have two main choices:
- Use the connection code from this PR in the existing https://github.com/prql/prql/blob/da173c39bb1c3af90ac8681bec9bd5295c379432/prql-compiler/tests/integration/main.rs. The existing file has the canonical snapshots and the project's canonical data from Chinook.
- Use the data & snapshot approach from the existing file in this PR's tests. As well as having more DBs, this PR have an arguably better design for the connections, with each implementing a common Trait.

I don't have a strong preference of whether which we move into which (though I do definitely think we should consolidate them, as discussed above).

> Small nit: if we use the PR as a base and merge the existing code in, let's rename from "-vendors" to something like "-dbs" or "-rdmbs"; they're not really vendors per se

I left one comment for how to use insta — check that out — I don't think that part is much work at all. (Again, very happy to help there, though)

The main work will be loading the data into the DBs — the existing code loads CSVs directly, rather than `INSERT INTO` statements. I do think that gives us a bigger corpus of data to work with. And we are generally trying to move all our examples and tests in the direction of this specific dataset. Do all the DBs support loading from CSVs?

---

Does this make sense? I'm very appreciative of the contribution, I hope me asking for the change comes across as reasonable. I think this will be a great set of tests and I'm keen for the project keep using them and growing with them, hence the focus on ensuring these are easy to maintain. Very open to feedback.

.github/workflows/integration-test.yaml
```diff
@@ -0,0 +1,28 @@
...
+  cancel-in-progress: true
+
+jobs:
+  test:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v3
+      - name: Run docker compose
+        run: docker-compose up -d
```
eitsupi:
```suggestion
        run: docker compose up -d
```

prql-compiler/tests/integration-vendors/docker-compose.yml
```diff
@@ -0,0 +1,39 @@
+version: "3"
+
+services:
```
eitsupi: Note that `version` is not required in Compose Spec.

```suggestion
services:
```

prql-compiler/tests/integration-vendors/docker-compose.yml
```diff
@@ -0,0 +1,39 @@
...
+  postgres:
+    image: "postgres:15-alpine"
```
eitsupi: Is there any reason to specify version 15?

```suggestion
    image: "postgres:alpine"
```

prql-compiler/tests/integration-vendors/docker-compose.yml
```diff
@@ -0,0 +1,39 @@
...
+    ports:
+      - "5432:5432"
+    environment:
+      POSTGRES_DB: dummy
+      POSTGRES_USER: root
+      POSTGRES_PASSWORD: root
+  mysql:
+    image: "mysql:8"
```
eitsupi: Is there any reason to specify version 8?

```suggestion
    image: "mysql:debian"
```

prql-compiler/tests/integration-vendors/docker-compose.yml
```diff
@@ -0,0 +1,39 @@
...
+  #      LICENSE: accept
+  #      DBNAME: dummy
+  #      DB2INSTANCE: db2
+  #      DB2INST1_PASSWORD: root
+  #      BLU: false
+  #      TO_CREATE_SAMPLEDB: false
+  #      REPODB: false
+  #      IS_OSXFS: false
+  mssql:
+    image: "mcr.microsoft.com/mssql/server:2022-latest"
```
eitsupi:
```suggestion
    image: "mcr.microsoft.com/mssql/server:latest"
```

It does not seem necessary to specify the version.

```sh
$ docker buildx imagetools inspect mcr.microsoft.com/mssql/server:2022-latest
Name:      mcr.microsoft.com/mssql/server:2022-latest
MediaType: application/vnd.docker.distribution.manifest.v2+json
Digest:    sha256:7dfe6bf43d18d0cf929e1dc40d0c1fccce02e7cca1ed9e277e35b6815108d17b

$ docker buildx imagetools inspect mcr.microsoft.com/mssql/server:latest
Name:      mcr.microsoft.com/mssql/server:latest
MediaType: application/vnd.docker.distribution.manifest.v2+json
Digest:    sha256:7dfe6bf43d18d0cf929e1dc40d0c1fccce02e7cca1ed9e277e35b6815108d17b
```

eitsupi (commented):
It would be great to see integration testing!
I added some comments around Docker.

prql-compiler/tests/integration-vendors/docker-compose.yml
Jelenkee: No specific reason. I just took the latest version

prql-compiler/tests/integration-vendors/main.rs
Jelenkee: but the `testcases.txt` will be necessary when new tests are added, right?

max-sixty: > Here — can we run `assert_display_snapshot!(actual_rows, con.get_dialect().to_string)`

(Sorry, I got the order of the args wrong, now swapped)

> but the `testcases.txt` will be necessary when new tests are added, right?

We should be able to completely remove the machinery around them. We'll need to keep queries somewhere, but https://github.com/PRQL/prql/tree/90ec09b2f2f8bc0e3fc81fd49a888e85f4e96e58/prql-compiler/tests/integration/queries are good, and use the existing Chinook data. So I expect we won't keep `textcases.txt` at all... (unless you have a strong view...)

prql-compiler/tests/integration-vendors/docker-compose.yml
max-sixty: Does dependabot upgrade these? If it does, that's ideal — then we lock versions but they don't fall out of date.

(BTW fine to refine these sorts of things in another PR)

max-sixty:
> The problem is that the tests don't run successfully because MsSQL has problem with `limit` and MySQL has problems with `loop` (maybe others too). The behaviour should be fixed before this PR is merged.

That's totally fine, let's just keep a list of exclusions, either in rust or with some comment in the test (e.g. `-- mssql-exclude`) — whatever is closer (probably the first is easier?)

I think we should try and merge this and then later figure out how to fix them & remove the exclusions — otherwise I worry this won't get merged for a while, and it's good code!

---

> I added all of the old integration test to the new test file except for [invoice_totals](https://github.com/PRQL/prql/blob/main/prql-compiler/tests/integration/queries/invoice_totals.prql). I'm not sure if all of the DBs work with these s-string.

Sorry if I'm missing them — where are the old integration tests being run? I can't see them in `testcases.txt` or that the rust code in https://github.com/PRQL/prql/blob/b35f92eebaf5973f9f16c6098fde456a60e17912/prql-compiler/tests/integration-rdbms/main.rs is grabbing the old files.

Or maybe we haven't added them yet because of the problem above?

---

How can I run the tests locally? I've run `docker-compose -d up` and then `cargo test --test integration-rdbms` — they're passing but not running (I tried adding a `panic!()` to confirm.). 

Possibly I have an issue running the docker containers, I get some odd debug logs from mssql.

Could it at least run the in-process tests such as duckdb without an external service? The existing integration tests do this.

prql-compiler/tests/integration-rdbms/connection.rs
```diff
@@ -0,0 +1,227 @@
...
+            let mut columns = vec![];
+            // row.len() always gives 1. hence this workaround
+            for i in 0.. {
+                let v_ref = match row.get_ref(i) {
+                    Ok(v) => v,
+                    Err(_) => {
+                        break;
+                    }
+                };
+                let value = match v_ref {
```
max-sixty, line 54: BTW to the extent that `v_ref.to_string()` gets us there, that's fine!

max-sixty:
> The main work will be loading the data into the DBs — the existing code loads CSVs directly, rather than `INSERT INTO` statements. I do think that gives us a bigger corpus of data to work with. And we are generally trying to move all our examples and tests in the direction of this specific dataset. Do all the DBs support loading from CSVs?

Just to reiterate this — there might be some refinements — e.g. excluding some tests that fail — but basically this is the one piece of work remaining

prql-compiler/tests/integration-rdbms/main.rs
```diff
@@ -0,0 +1,166 @@
...
+    use crate::connection::*;
+
+    #[test]
+    fn test_rdbms() {
+        for port in [5432u16, 3306, 1433 /*, 50000*/] {
+            // test is skipped locally when DB is not listening
+            // in CI it fails
+            if !is_port_open(port) {
+                match env::var("CI") {
+                    Ok(v) if &v == "true" => {
```
max-sixty, line 35: I don't think we should delay merging to discuss this — it's very easy to change later — but I'm in a couple of minds about this. On the one hand, it's important that `cargo test` passes without external dependencies. On the other, it might be quite confusing if CI is failing from running `cargo test` while locally it passes.

I thought your previous approach re ignoring and then running it explicitly was interesting. Both have their tradeoffs.

Another approach would be to include the out-of-process connections as part of a non-default feature, such that running `cargo test --all-features` ran it, but running `cargo test` didn't. That would probably be my vote, but I haven't thought about it enough. (and again — totally fine to resolve later)

CC @aljazerzen for any views.

prql-compiler/tests/integration-vendors/docker-compose.yml
Jelenkee: I don't know if dependabot checks these files but the latest version should be fine

prql-compiler/tests/integration-rdbms/connection.rs
Jelenkee: `ToString` is not implemented for `ValueRef`

prql-compiler/tests/integration-rdbms/docker-compose.yml
eitsupi, line 1: Perhaps we should set `volume`s, as the Postgres and MySQL Docker containers automatically create Docker Volumes.

https://github.com/docker-library/postgres/blob/3db64a0e20b456a554b9da8791704cb3e54af7f2/15/alpine/Dockerfile#L163
https://github.com/docker-library/mysql/blob/99f9c4104edb05ce198c5e9282f4d9d702277a60/8.0/Dockerfile.debian#L88

Like that.

```yml
services:
  postgres:
    image: "postgres:alpine"
    volumes:
      - pgdata:/var/lib/postgresql/data

  mysql:
    platform: linux/amd64
    image: "mysql:debian"
    volumes:
      - mysqldata:/var/lib/mysql

volumes:
  pgdata:
  mysqldata:
```

Jelenkee: What is the benefit of this?

eitsupi: If we do not do this, Docker Volumes with random names will be created.
Naming these in the compose file will make it clear where the volumes were configured.

max-sixty: FWIW a reminder on the norms here — a) these sorts of comments are welcome @eitsupi b) no obligation to implement them all @Jelenkee 

From https://prql-lang.org/book/contributing/development.html

> - We use PR reviews to give general context, offer specific assistance, and collaborate on larger decisions.
>   - Reviews around ‘nits’ like code formatting / idioms / etc are very welcome. But the norm is for them to be received as helpful advice, rather than as mandatory tasks to complete. Adding automated tests & lints to automate these suggestions is welcome.

Jelenkee:
Oh man, it's really difficult.
I am about to use the current [queries](https://github.com/PRQL/prql/tree/f7ef223e40ed4f452737dee02f81db3d5bcd761d/prql-compiler/tests/integration/queries) with importing the CSVs into the DBs. After struggling with the import (handling null values, handling different datetime formats, handling different encodings, ...) I get this when running the tests
image


Why does every DB behave so differently? :angry: 

I'm loosing my motivation...

max-sixty:
I definitely empathize...

Can we reduce the scope a bit? Maybe just exclude any query or any database which doesn't pass, and then we'll resolve those in future PRs? Or feel free to add `x | round 0` or `as | int` functions to round the numbers if those are the only problems?

The initial PR was trying to do a lot — a completely new set of integration tests, running in CI, even with its own snapshot framework! ...which is admirable, but often I find it's much easier to do smaller contributions. Even something as small as running the existing tests in CI, or adding another DB, or using your Trait approach in place of the `mod` in the existing tests — any of those would be a nice legible contribution that you or others could build on.

Maybe there are one or two queries that run successfully across most of the DBs? Could we start with those?

prql-compiler/prqlc/src/cli.rs
```diff
@@ -356,13 +356,13 @@ group a_column (take 10 | sort b_column | derive [the_number = rank, last = lag
...
-           │ ──┬─
-           │   ╰─── Unknown name asdf
-        ───╯
+        [31mError:[0m
+           [38;5;246m╭[0m[38;5;246m─[0m[38;5;246m[[0m:1:1[38;5;246m][0m
+           [38;5;246m│[0m
+         [38;5;246m1 │[0m asdf
+         [38;5;240m  │[0m ──┬─
+         [38;5;240m  │[0m   ╰─── Unknown name asdf
+        [38;5;246m───╯[0m
```
max-sixty: I will fix this

Jelenkee:
I ignored some of the queries for some DBs.

On problem is that the github actions are canceled during compiling. I guess because of OutOfMemory (e.g. [here](https://github.com/PRQL/prql/actions/runs/4545784530/jobs/8013582834?pr=2286) and [here](https://github.com/PRQL/prql/actions/runs/4545784530/jobs/8013583054?pr=2286)). Can you find out why?

prql-compiler/prqlc/src/cli.rs
max-sixty: Fixed in #2335, if you discard this change and merge in `main`

max-sixty:
> On problem is that the github actions are canceled during compiling. I guess because of OutOfMemory (e.g. [here](https://github.com/PRQL/prql/actions/runs/4545784530/jobs/8013582834?pr=2286) and [here](https://github.com/PRQL/prql/actions/runs/4545784530/jobs/8013583054?pr=2286)). Can you find out why?

My guess is that the run is canceled because another run started from a new commit. The latest run completes, but with some snapshot errors [here](https://github.com/PRQL/prql/actions/runs/4545784594/jobs/8013582241). Does running `cargo insta test --accept` write the correct snapshots?

prql-compiler/prqlc/src/cli.rs
```diff
@@ -364,13 +364,13 @@ group a_column (take 10 | sort b_column | derive [the_number = rank, last = lag
...
```
max-sixty:
```suggestion
        Error:
           ╭─[:1:1]
           │
         1 │ asdf
           │ ──┬─
           │   ╰─── Unknown name asdf
        ───╯
```

prql-compiler/tests/integration-rdbms/main.rs
Jelenkee: > On the other, it might be quite confusing if CI is failing from running cargo test while locally it passes.

Yes, but it is the same when the CI runs includes ignored tests

Jelenkee:
Nearly every test passed. It seems that dcoker is not installed on macOS per default

max-sixty:
I think we're basically there. I just fixed the CLI test and now all ubuntu tests pass. 

Given how large his already is — if you're up for refining this over the next few weeks, let's merge now and iterate on it in the future?

We get a failure on Mac, and I guess we will on Windows. Shall we just hack around that by conditioning running these tests on being in Linux? It's not permanent but it'll let us merge now.

---

Some follow-ups I can think of for future PRs. We can move into a separate issue:
- Remove previous integration tests, move the data files into this path
- Run in-process DBs even without an external connection (will be quite helpful if we start using this more)
- Add a description & link to https://github.com/prql/prql/blob/7ae804a9d3f42f6ed3d33c04ab910dca164c7efc/web/book/src/contributing/development.md#L310
- Implement whether to run external integration tests with a feature, rather than reading the `CI` env var (that could become a bit confusing otherwise)
- (minor) Possibly output the snapshots as CSVs / something both machine & person-readable, e.g. with headers, rather than one long line
- (minor) Consider whether we keep it in the current CI test vs move to another, possibly in `test-all.yaml`. It currently adds 40 seconds, which is not too bad. Or we condition the `test-rust` tests on a PR's paths, and avoid running it for small doc changes.
- Any others @Jelenkee ?

prql-compiler/tests/integration-rdbms/main.rs
max-sixty: Yes totally. It is a bit more legible if it's being passed as a command than implicit in the rust code. But regardless, I think doing it as a feature would dominate both options, and let us split out the external vs. in-process tests

(but as above, fine to skip over for this merge)

eitsupi:
Given the runner's resources, it would make sense to run tests that use Docker only on Linux runner.

Jelenkee:
> * Any others @Jelenkee ?

Wenn starting docker compose the owner of prql-compiler/tests/integration/data/chinook is changed on Linux. Probably on macOS too. That should be fixed because you cannot edit the files in there.

Jelenkee:
The CI is not running. Did I break something?

max-sixty:
> The CI is not running. Did I break something?

Maybe a GitHub bug... I added a label to see if that would jolt it and it seems to have worked

prql-compiler/tests/integration-rdbms/main.rs
```diff
@@ -0,0 +1,182 @@
...
+
+    use prql_compiler::sql::Dialect;
+    use prql_compiler::Options;
+    use prql_compiler::Target::Sql;
+
+    use crate::connection::*;
+
+    #[test]
+    fn test_rdbms() {
+        if env::var("SKIP_INTEGRATION").is_ok() {
```
max-sixty, line 27: Very reasonable workaround!

Jelenkee:
It has a problem because insta found snaps that belong to no test becuase the tests were skipped.
https://github.com/PRQL/prql/actions/runs/4556292813/jobs/8036462791?pr=2286

.github/workflows/test-rust.yaml
```diff
@@ -54,6 +54,19 @@ jobs:
...
+        if: ${{ inputs.os != 'windows-latest' }}
+      - name: Wait for database
+        uses: ifaxity/wait-on-action@v1
+        with:
+          resource: "tcp:5432 tcp:3306 tcp:1433"
+          timeout: 60000
+        if: ${{ inputs.os != 'windows-latest' }}
+      - name: skip integration if windows
+        run: echo "SKIP_INTEGRATION=true" >> $GITHUB_ENV
+        if: ${{ inputs.os == 'windows-latest' }}
```
max-sixty: Possibly also on Mac?

max-sixty:
Almost there!!

> It has a problem because insta found snaps that belong to no test becuase the tests were skipped.
> https://github.com/PRQL/prql/actions/runs/4556292813/jobs/8036462791?pr=2286

Re the taskfile error —  let's comment out `--unreferenced=auto` in the `Taskfile.yml`, and add a link to this PR? I can look at whether we can restore that in future.

Jelenkee:
For some reason [test-js](https://github.com/PRQL/prql/actions/runs/4557540423/jobs/8039304075?pr=2286) fails

And the [linter](https://github.com/PRQL/prql/actions/runs/4557803700/jobs/8039912024?pr=2286) is mad at me and I don't know why

.github/workflows/test-rust.yaml
```diff
@@ -54,6 +54,19 @@ jobs:
...
+        working-directory: ./prql-compiler/tests/integration-rdbms
+        if: ${{ inputs.os == 'ubuntu-latest' }}
+      - name: Wait for database
+        uses: ifaxity/wait-on-action@v1
+        with:
+          resource: "tcp:5432 tcp:3306 tcp:1433"
+          timeout: 60000
+        if: ${{ inputs.os == 'ubuntu-latest' }}
+      - name: skip integration if not linux
+        run: echo "SKIP_INTEGRATION=true" >> $GITHUB_ENV
```
max-sixty:
```suggestion
        run: echo "SKIP_INTEGRATION=true" >> "$GITHUB_ENV"
```
I think this is what the linter wants

max-sixty:
> For some reason [test-js](https://github.com/PRQL/prql/actions/runs/4557540423/jobs/8039304075?pr=2286) fails

Weird, I think that's just flaky



> And the [linter](https://github.com/PRQL/prql/actions/runs/4557803700/jobs/8039912024?pr=2286) is mad at me and I don't know why

I added a fix

prql-compiler/tests/integration-rdbms/docker-compose.yml
```diff
@@ -0,0 +1,47 @@
...
+    platform: linux/amd64
+    image: "mysql:debian"
+    ports:
+      - "3306:3306"
+    environment:
+      MYSQL_DATABASE: dummy
+      MYSQL_ROOT_PASSWORD: root
+    volumes:
+      - ./conf:/etc/mysql/conf.d
+      - ../integration/data/chinook:/tmp/chinook
```
eitsupi, line 23: > Wenn starting docker compose the owner of prql-compiler/tests/integration/data/chinook is changed on Linux.

I think it's because in Linux, the user in the container is root. In that case, bind mounting as read-only might solve the problem. (I have not tried if this works here!)

The ownership issue on bind mounting does not occur on macOS and Windows, which have different file systems. (Because host OS files are handled as root's)

```suggestion
    volumes: &vol
      - ../integration/data/chinook:/tmp/chinook:ro
  mysql:
    # No arm64 image available; remove when one does become available
    platform: linux/amd64
    image: "mysql:debian"
    ports:
      - "3306:3306"
    environment:
      MYSQL_DATABASE: dummy
      MYSQL_ROOT_PASSWORD: root
    volumes:
      - ./conf:/etc/mysql/conf.d:ro
      - ../integration/data/chinook:/tmp/chinook:ro
```

max-sixty: (I didn't want to add this without checking with @Jelenkee , but it does seem like a good idea)

max-sixty:
I just resolved a merge conflict, then let's merge! I recognize it's been a long road.

I do think it would be quite valuable to refine parts; particularly where it touches the wider library — e.g. removing the old integration tests in favor of these / having a command run the same tests wherever it's run / etc. 

To the extent you're up for opening an issue with the notes we made above, that would be great. No great rush on them, and possible others will pick up bits of it.

We could also add an entry to the changelog announcing the new tests.

Thank you very much indeed @Jelenkee !

prql-compiler/tests/integration-rdbms/docker-compose.yml
eitsupi: Fixed in #2352

PR #2293, opened 2023-03-23
chore(prql-dotnet): Copy library to output directory

vanillajonathan:
Copy the PRQL library to the output directory if it exist.


max-sixty:
Thanks @vanillajonathan

PR #2294, opened 2023-03-23
docs: Update dates in changelog for 2023

k-nut:
I just took a look at the changelog and was a bit confused.

It seems to me that all changes from this year have been accidentally attributed to last year. Time flies I guess 😉


max-sixty:
D'oh! This is my fault. Thanks @k-nut !

PR #2297, opened 2023-03-23
build: Combine `Taskfile`s

max-sixty:
For simplicity, at the cost of some build caching (almost always we should aim for simplicity...)


max-sixty:
Merging b/c I want to make another change, but feedback still welcome

PR #2303, opened 2023-03-24
fix: `Ident::starts_with`

max-sixty:
I'm not sure this was correct prior; at least it didn't pass these tests, and I couldn't work out why it started with `!`.

This is part of the refactoring in #1535

### DIFF
--- a/prql-compiler/src/ast/pl/ident.rs
+++ b/prql-compiler/src/ast/pl/ident.rs
@@ -1,6 +1,5 @@
 use std::fmt::Write;
 
-use itertools::Itertools;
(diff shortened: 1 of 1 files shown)


aljazerzen:
The ! shouldn't have been there.

The new solution does a lot of unnecessary cloning.

max-sixty:
> The new solution does a lot of unnecessary cloning.

Yes, this will go away with the process started by https://github.com/PRQL/prql/pull/2305 (feedback on that v welcome)

> The ! shouldn't have been there.

Yeah, I couldn't quite work this out — tests fail when removing it — I think maybe the `all_equal` is assessing whether the tuples from the zip were equal to each other, rather than the items in each tuple...

But the new way does seem to work!

PR #2308, opened 2023-03-25
fix: Loop in MySQL

Jelenkee:
`loop` is a keyword in MySQL. So it throws an error when executing.
I renamed it to `_loop` in the generated SQL


max-sixty:
Good idea!

PR #2317, opened 2023-03-27
internal: Add some docstrings using GPT-4

max-sixty:
They're not genius insight, but they're pretty good!


aljazerzen (approved):
Looking at these comments, they don't add any insight other than what you can infer by looking at the fields directly.

They would improve public docs, but these are not public functions.

But I see no harm, so I'm +0

max-sixty:
Yeah, I think you might be surprised at how helpful a single sentence which summarizes the code can be when someone is a) new and b) moving quickly. Even if it takes 30 seconds to understand the code and 10 seconds to read the sentence, there is a "cache thrashing" in the cognitive overhead to switch context to understand the code.

That said, some of these are probably going too far `The [SqlTableDecl] struct contains information about a table declaration, including its ID, name, and relation (if it has been defined).` really contains zero information beyond the Struct members, and having a docstring adds to our maintenance burden. Maybe Co-Pilot can generate that on the client side if folks like me want it.

So I'll keep ones like `The AnchorContext struct stores information about tables and columns, and is used to generate new IDs and names.` and discard some others.

max-sixty:
Kept about half and merging

PR #2325, opened 2023-03-28
fix: Don't panic on `group (take)`

max-sixty:
I can't really work out how to do this well — I'm guessing that the error should be handled long before it gets to the point where it's trying to do the `take` on the group?

I asked a couple of questions on Discord re what a `Ty` / `SetExpr` was.

I previously mentioned that RQ was really promising in its ability to reduce the amount of context that was required to develop on the compiler. We briefly touched on whether there were additional IRs that would be useful. I don't think we necessarily need a full IR, but I think it would be useful to have a more structured representation of the intermediate AST that we could use to debug errors like this — is there some stage that this gets to which is correct (beyond PL), where we can see what's obviously wrong by showing the relevant Struct?


prql-compiler/src/tests/test_bad_error_messages.rs
```diff
@@ -49,3 +49,15 @@ fn test_bad_error_messages() {
...
+
+#[test]
+fn test_2270() {
+    assert_display_snapshot!(compile(r###"
+    from tracks
+    group foo (take)
+    "###,
+    )
+    .unwrap_err(), @r###"
+    Error: Expected a function that could operate on a table, but instead found func table<[]> -> table<[]>
```
max-sixty, line 61: Note this is in `bad_error_messages`, but still better than a panic...

aljazerzen (commented):
This error should be caught by type checking. So it should be safe to assume that body is a closure that returns a relation.

max-sixty:
> This error should be caught by type checking. So it should be safe to assume that body is a closure that returns a relation.

i.e. we shouldn't merge this and instead should try and fix this earlier in the compiler?

aljazerzen:
> i.e. we shouldn't merge this and instead should try and fix this earlier in the compiler?

Yes, exactly.

The test is useful, though, we can cherry-pick it later.

aljazerzen:
Well, this is better than panic, so let's just merge.

max-sixty:
> Well, this is better than panic, so let's just merge.

Yeah, probably. It can reduce the urgency while adding complication to the code which shouldn't be there, but yeah I guess on balance panics are quite bad too...

PR #2326, opened 2023-03-28
internal: Add a docstring & question to Module

max-sixty:


prql-compiler/src/semantic/module.rs
```diff
@@ -27,6 +27,9 @@ pub struct Module {
 
     /// List of relative paths to include in search path when doing lookup in
     /// this module.
+    // Q: @aljazerzen — what's the difference between a redirect and a
+    // LayeredModule? Does the data structure contain the destination of the
+    // redirect?
```
aljazerzen, line 32: This is a part of name resolution algorithm. It has not been formally defined yet, so it does not have any docs on it. It is very close to modules RFC, so I may add it there.

Briefly, if you have module structure like this (using made-up syntax):

```
module std {
    func sum ...
    func average ...
}

module default_db {
    let albums = ...
    let tracks = ...
    module sales {
        let orders = ...
    }
}
```

And a query like this:
```
let my_cte = (from default_db.sales.orders | select [average price])
```

Lookup of ident `default_db.sales.orders` is straight forward, as it is already FQ (fully qualified):
- `default_db` is in the same (root) module as the calling code, so it matches,
- `sales` is in `default_db`,
- `orders` is in `default_db`.

Lookup of ident `average` is more complicated, as the root module does not contain `average` name. Lookup steps are:
- look for `average` in root module and find nothing,
- follow redirects in root module,
- because of redirect `std`, so we look for `average` in `std`,
- there is `average` is `std`,
- result of the lookup is FQ ident `std.average`.

max-sixty: Excellent! This is exactly what I've been trying to work through in #1535.

> * follow redirects in root module,
> * because of redirect `std`, so we look for `average` in `std`,

Where does this appear in the structs? `root_module.redirects = [std]`? Is it only the root module that contains redirects?

aljazerzen: > Where does this appear in the structs?

https://github.com/prql/prql/blob/main/prql-compiler/src/semantic/module.rs#L74-L79

> Is it only the root module that contains redirects?

There are also redirects in `_frame` namespace. When you do something `from albums | join tracks []`, there will be a module structure like this:
```
module _frame {
    module albums {
        let name = ...
    }
    module tracks {
        let title = ...
    }
}
```
If we want to match `title` to `_frame.tracks.title`, we need a redirect in root module into `_frame` and a redirect in `_frame` into `tracks`. Same holds for `_frame.albums.name`.

So this is the mechanism that makes sure that `name`, `albums.name` and `_frame.albums.name` are resolved into the same declaration (if there is no other decl for `name` in any of the searched modules.

max-sixty: Excellent, thanks a lot for the explanation

Issue #2330, opened 2023-03-28
Implement link checker for website


max-sixty:
### What's up?

Our current link checker https://github.com/PRQL/prql/blob/0ae2a44ce31c59a2b38c16fdba78fa1838c1a3b2/.github/workflows/pull-request.yaml#L58-L71 checks the static links. But it doesn't build the site and check whether the constructed HTML has broken links.

For that, we'll need something like `htmltest-action`, as described in https://robb.sh/posts/check-links-in-hugo-with-htmltest/.

This would avoid issues such as #2319 / #2313

eitsupi:
I suppose we should use Hugo syntax there instead of Markdown syntax, as like #2331.

```md
See the [press material]({{< ref "/press-material" >}}).
```

https://gohugo.io/content-management/cross-references/

(Of course this link does not work on Markdown source code. This is one of the things I consider a downside of Hugo)

I believe Hugo does a check at build time to see if the links work, so this should work fine.

![image](https://user-images.githubusercontent.com/50911393/228259979-fafd8af0-cf57-48e5-a8c4-eeddcbe5ca59.png)

max-sixty:
Yes, good idea, that is better. It looks like _mostly_ the md links work, but where they're not robust, let's use the hugo `ref` links

ref https://discourse.gohugo.io/t/link-with-ref-relref-why-not-use-markdown-syntax/19290/4

I'll mark this as Postponed — we could add for completeness, but no longer a priority.

PR #2319, opened 2023-03-27
chore: Fix markdownlinkcheck

max-sixty:
Not sure why this is interacting with the linter badly. It's now an external reference, not a huge deal though.

CC @vanillajonathan for info, no need to prioritize

PR #2331, opened 2023-03-28
docs(website): fix link in hugo

eitsupi:
Follow up #2319, related to #2330

Hugo generates websites with a different structure than the hierarchical structure of the source code (Markdown), so Markdown link syntax should not be used here.


max-sixty (approved):
Good idea, thanks!

PR #2336, opened 2023-03-28
docs: Tweak FAQ 

richb-hanover:


web/website/content/faq.md
```diff
@@ -182,14 +188,12 @@ While you should be skeptical of new claims from new entrants
 [commented](https://news.ycombinator.com/item?id=30067406) in a discussion on
 PRQL:
 
-<!-- TODO: the `>` doesn't seem to format on the website  -->
-
-> "FWIW the separate `group_by()` is one of my greatest design regrets with
-> dplyr — I wish I had made `by` a parameter of `summarise()`, `mutate()`,
-> `filter()` etc."
+> FWIW the separate `group_by()` is one of my greatest design regrets with dplyr
```
max-sixty, line 191: Does this now display correctly on the website?

richb-hanover: I believe so. The change to the CSS file in a previous commit switched the selector to be "blockquote p" instead of (I think) "article blockquote" which did _not_ look right. (For example, the "Pretty Reports" post now uses the grey vertical bar...)

Also - I am not sure what happened to first FAQ and some of my other edits to the FAQ. I believe this version is good. But ofc I am happy to accept edits/suggestions.

web/website/content/faq.md
```diff
@@ -6,11 +6,15 @@ Here are some of the most common questions we hear. Have something else you'd
...
 
-{{< faq "Cool story Hansel, but what can I actually do with PRQL now?" >}}
+{{< faq "Cool story Hansel, but what can I actually do with PRQL _now_?" >}}
 
-We're still early, and the opportunities for using PRQL are focused on two
-integrations:
+PRQL builds powerful SQL queries today. In addition, we are focusing on several
+integrations. Here's how you can use PRQL:
 
+- **Do real work with PRQL** Prototype your PRQL queries in the
```
max-sixty, line 14: Maybe move this to the bottom and say something like "Prototype queries"? "Do real work" doesn't seem to be specific to this approach?

web/website/content/faq.md
```diff
@@ -21,8 +25,8 @@ integrations:
   enables querying pandas dataframes, CSVs & Parquet files, and writing the
   output to a dataframe.
 
-Beyond these two integrations, it's very easy to add PRQL to your own apps with
-our [bindings](/#bindings) for Rust, Python & JS.
+It's also very easy to add PRQL to your own apps with our [bindings](/#bindings)
```
max-sixty:
```suggestion
It's also possible to add PRQL to your own apps with our [bindings](/#bindings)
```
(I know this was here before...)

web/website/content/faq.md
```diff
@@ -6,11 +6,15 @@ Here are some of the most common questions we hear. Have something else you'd
 like to ask? Pop by our [Discord](https://discord.com/invite/eQcfaCmsNc) and ask
 away!
 
-{{< faq "Cool story Hansel, but what can I actually do with PRQL now?" >}}
+{{< faq "Cool story Hansel, but what can I actually do with PRQL _now_?" >}}
 
-We're still early, and the opportunities for using PRQL are focused on two
-integrations:
+PRQL builds powerful SQL queries today. In addition, we are focusing on several
+integrations. Here's how you can use PRQL:
```
max-sixty, line 12:
```suggestion
We're still early, and the opportunities for seamlessly writing PRQL are growing. We're focusing on several
integrations:
```

WDYT? I think it's important to stress that we're early and that more is coming....

web/website/content/faq.md
```diff
@@ -6,11 +6,15 @@ Here are some of the most common questions we hear. Have something else you'd
...
 
+- **Do real work with PRQL** Prototype your PRQL queries in the
+  [Playground](https://prql-lang.org/playground/) or the
+  [VS Code extension](https://marketplace.visualstudio.com/items?itemName=PRQL-lang.prql-vscode)
+  and copy/paste the resulting SQL into your database.
 - **[dbt-prql](https://github.com/prql/dbt-prql)** allows writing PRQL in
```
max-sixty, line 18: Unfortunately we could probably remove this from this section now, RIP. Maybe we replace with `prqlr` @eitsupi ?

Or we add a bullet about `prqlc`?

Or `prql-query` @snth ?

(No need to resolve all these before merging)

web/website/content/faq.md
richb-hanover: I'm still casting about to find a way to describe that PRQL _is_ good enough to use right now. (And I think that talking about bindings to other languages makes its relevance seem like a distant possibility.

I'm content with what's here, and I'm running out of time in the near term to devote much more to this page (Tax time in the US (!)). I am completely content to let others continue to edit.

web/website/content/faq.md
richb-hanover: See above

web/website/content/faq.md
richb-hanover: See above.

web/website/content/faq.md
max-sixty: Yes, those disappeared from the previous PR. I thought you might have removed them so we could merge the other things is more easily...

Just checked https://prql-lang.org/faq/ — looks good!

web/website/content/faq.md
max-sixty: > I'm content with what's here, and I'm running out of time in the near term to devote much more to this page (Tax time in the US (!)). I am completely content to let others continue to edit.

Totally empathize, it's important we can make small improvements without being dragged into a total refactor. 

I'll let this sit for a bit to see if others have thoughts; then happy to merge and I'll take a pass?

eitsupi: I agree that we can emphasize that Playground can be used to assist in writing SQL immediately.

I actually use the SQL generated using Playground (after tweaking it), so it would be great if we could emphasize that it can be used right out of the box without installing anything.

(BTW, I look forward to the integration with DBeaver as I often execute SQL in DBeaver.)

richb-hanover: +1... 

On yesterday's Dev Call, I mentioned that VS Code was my preferred environment. It's like the Playground, except I can keep all the PRQL queries in my repo, so they can easily be revised and re-executed.

eitsupi: Perhaps a link to this might explain how PRQL actually works?
https://eitsupi.github.io/querying-with-prql/

web/website/content/faq.md
aljazerzen:
```suggestion
```

richb-hanover: @aljazerzen Hmmm... I had meant to close this and replace with my tweak-faq-2 branch. https://github.com/richb-hanover/prql/blob/tweak-faq-2/web/website/content/faq.md

Let's continue the conversation (about what to do about dbt-prql) over there.

richb-hanover:
I would like to close this in favor of #2474

PR #2341, opened 2023-03-28
docs: Refine "Compiler architecture"

max-sixty:
This is a great doc, I've been using it recently. This refines it slightly, again using GPT-4.

(the only improvement I can think of would be to integrate it with https://github.com/prql/prql/blob/90ec09b2f2f8bc0e3fc81fd49a888e85f4e96e58/prql-compiler/src/lib.rs#L1-L33, and rustdoc could include links, and ensure it doesn't fall out of date. But possibly then it becomes too anchored in the specific code rather than the general approach)

### DIFF
--- a/web/book/src/internals/compiler-architecture.md
+++ b/web/book/src/internals/compiler-architecture.md
@@ -3,7 +3,7 @@
 The PRQL compiler operates in the following stages:
 
 1. **Lexing & Parsing**: PRQL source text is split into tokens with the Chumsky
-   parser named "lexer". Stream of tokens is then parsed into an Abstract Syntax
+   parser named "lexer". The stream of tokens is then parsed into an Abstract Syntax
    Tree (AST).
 
 2. **Semantic Analysis**: This stage resolves names (identifiers), extracts
--- a/web/book/src/internals/compiler-architecture.md
+++ b/web/book/src/internals/compiler-architecture.md
@@ -3,8 +3,8 @@
 The PRQL compiler operates in the following stages:
 
 1. **Lexing & Parsing**: PRQL source text is split into tokens with the Chumsky
-   parser named "lexer". The stream of tokens is then parsed into an Abstract Syntax
-   Tree (AST).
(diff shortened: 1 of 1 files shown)


web/book/src/internals/compiler-architecture.md
```diff
@@ -1,52 +1,55 @@
...
-   - Function calls to transforms (`from`, `derive`, `filter`) are converted
-     from `FuncCall` into `TransformCall`, which is more convenient for later
+# PRQL Compiler Architecture
+
+The PRQL compiler operates in the following stages:
+
+1. **Lexing & Parsing**: The PRQL text is split into tokens and a parse tree is
+   built. This parse tree is then converted into an Abstract Syntax Tree (AST)
+   using the `ast` module. The Chumsky parser is utilized for parsing, and the
+   AST is constructed in `parser.rs`.
```
aljazerzen:
```suggestion
1. **Lexing & Parsing**: PRQL source text is split into tokens with the Chumsky parser named "lexer". Stream of tokens is then parsed into an Abstract Syntax Tree (AST).
```

aljazerzen (approved):

web/book/src/internals/compiler-architecture.md
```diff
@@ -1,52 +1,54 @@
...
+# PRQL Compiler Architecture
 
-Compiler works in the following stages:
+The PRQL compiler operates in the following stages:
 
-1. Lexing & parsing - split PRQL text into tokens, build parse tree and convert
-   into our AST (Abstract Syntax Tree, see `ast` module). Parsing is done using
-   the chumsky parser, AST is constructed in `parser.rs`.
+1. **Lexing & Parsing**: PRQL source text is split into tokens with the Chumsky
+   parser named "lexer". Stream of tokens is then parsed into an Abstract Syntax
```
max-sixty:
```suggestion
   parser named "lexer". The stream of tokens is then parsed into an Abstract Syntax
```

Issue #2079, opened 2023-03-10
`group` & `select`  generates additional field "_expr_0" 


echou:
```sql
from `db.t`
derive d = create_time + 1
group d (
  aggregate [
    n1 = (a | sum),
  ]
)
sort d
select [ d1 = d, n1 ]
```
generates the result:

```sql
SELECT
  create_time + 1 AS d1,
  SUM(a) AS n1,
  create_time + 1 AS _expr_0
FROM
  db.t
GROUP BY
  create_time + 1
ORDER BY
  _expr_0

-- Generated by PRQL compiler version:0.6.0 (https://prql-lang.org)
```

1. if select with group key, the key is added.  (correct)
2. if select without group key, `_expr_0` is added. (correct. but not good)
3. if select with group key alias, `_expr_0` is added. (wrong)

echou:
it seems the `sort` statement cause the bug.

max-sixty:
Thanks a lot for the excellent report @echou

max-sixty:
Even smaller example:

```elm
from foo
derive d = x + 1
sort d
select [col]
```

```sql
SELECT
  col,
  x + 1 AS _expr_0
FROM
  foo
ORDER BY
  _expr_0

-- Generated by PRQL compiler version:0.6.0 (https://prql-lang.org)
```

We're outputting two columns, when we should only be outputting `col`

aljazerzen:
Interesting.

The strange alias seems to be caused by `d` not being selected - it should not even be in the SELECT.

max-sixty:
I haven't got a fix yet but writing down what I have midway through:

At https://github.com/max-sixty/prql/blob/5640989058b213af0d453034e5c7033f385cca3b/prql-compiler/src/sql/gen_query.rs#L254, we have two columns selected;

```
[prql-compiler/src/sql/gen_query.rs:254] pipeline.clone() = [
    Super(
        Select(
            [
                column-1,
                column-3,
            ],
        ),
    ),
    Super(
        Compute(
            Compute {
                id: column-3,
                expr: Expr {
                    kind: ColumnRef(
                        column-0,
                    ),
                    span: Some(
                        span-chars-29-30,
                    ),
                },
                window: None,
                is_aggregation: false,
            },
        ),
    ),
    Super(
        Sort(
            [
                ColumnSort {
                    direction: Asc,
                    column: column-3,
                },
            ],
        ),
    ),
]
```

max-sixty:
[...]

PR #2342, opened 2023-03-29
fix: Raise an error on unavailable excludes

### LINKED ISSUE #2292: `select !` should raise an error if it can't exclude
### What happened?

Currently an invalid `select !` is silently ignored

### PRQL input

```elm
from tracks
select ![milliseconds,bytes]
```

### SQL output

```SQL
SELECT
  *
FROM
  tracks

```

### Expected SQL output

```SQL
Error: we don't know the columns prior to `select ![milliseconds,bytes]`, and we're not using a target that natively supports excluding columns.
```

### MVCE confirmation

- [X] Minimal example
- [X] New issue

### Anything else?

_No response_

max-sixty:
Closes #2292. 

But I now realize was here deliberately (the old code was _more_ work than mine!)

That said I would think it's better to raise here — getting this wrong is still wrong...


max-sixty:
Tests are failing because of https://github.com/PRQL/prql/issues/2079, so I guess we need to solve that first

aljazerzen (approved):
I did not want throw an error here, because we don't have an easy escape path. (see the comment that was removed).

I've pushed a test that exposes this error when using `group (take)`. If you are ok with throwing an error even in this case, we can merge.

For better errors, we'd need to include more spans in RQ.

max-sixty:
> I did not want throw an error here, because we don't have an easy escape path. (see the comment that was removed).

There are two cases I can see:
- The `group 1` — i.e. the test you added — I hadn't thought we had to throw an error there (assuming we wrote perfect code) — is that not correct? That query could compile to `SELECT DISTINCT title FROM tracks`, no?
- The case where someone has specified a column to exclude. I thought that was what the comment was referring to. My vote would be to raise an error there — that the downside of the "no escape path" was lower than the downside of "I asked for the exclude and it just ignored me"...

aljazerzen:
> The group 1 — i.e. the test you added — I hadn't thought we had to throw an error there (assuming we wrote perfect code) — is that not correct? That query could compile to SELECT DISTINCT title FROM tracks, no?

When there is `take 1` it should yes, but `group x (take 2)` should not.

There may be other cases where a column is added and should later be removed without removing the unknown columns, but I cannot think of any right now.

max-sixty:
> When there is `take 1` it should yes, but `group x (take 2)` should not.

Good point. Ideally we'd be able to handle those cases differently, I recognize that will require more work though.

I'll leave this open for a bit and see whether I / someone else gets to it. Otherwise we can close and attempt a broader effort.

PR #2343, opened 2023-03-29
refactor: re-write prql-java

### LINKED ISSUE #1643: Publishing prql-java to maven central

zhicwu:
Closes #1643


.github/workflows/release.yaml
```diff
@@ -166,24 +166,76 @@ jobs:
...
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v3
+      - name: Install targets
+        uses: actions-rs/toolchain@v1
+        with:
+          profile: minimal
+          override: true
+          toolchain: stable
+          target: ${{ matrix.target }}
```
max-sixty, line 200: Unneeded; runners have this already

```suggestion
```

max-sixty:
Hi @zhicwu 

Thanks a lot for the big contribution! Appreciate it.

So I understanding — because we're using JNA rather than JNI — we're no longer linking with the `jni` rust crate; instead we're linking through `prql-lib`, is that correct? 
IIRC when we discussed this, we thought JNA was reasonable given that we don't care much about performance here (though we miss out on help from the `jni` rust crate, possible a worthwhile trade though...)

CC @linux-china for feedback.

.github/workflows/release.yaml
```diff
@@ -166,24 +166,76 @@ jobs:
...
+        uses: samuelmeuli/action-maven-publish@v1
+        with:
+          directory: bindings/prql-java
+          maven_profiles: release
+          maven_args:
+            -Drevision=0.6.1 -q --batch-mode --show-version --strict-checksums
+          gpg_private_key: ${{ secrets.GPG_PRIVATE_KEY }}
+          gpg_passphrase: ${{ secrets.GPG_PASSPHRASE }}
+          nexus_username: ${{ secrets.SONATYPE_USER }}
+          nexus_password: ${{ secrets.SONATYPE_PASSWD }}
```
max-sixty, line 238: Let me know what you want me to set up here

.github/workflows/test-java.yaml
```diff
@@ -13,14 +13,71 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
-  test:
+  build-prql-lib:
```
max-sixty, line 16: It might be worth splitting this out into a separate workflow that both `test-java` and `release` call and the pick up the artifact, to reduce duplication. 

But it's totally fine to add a `TODO` in the code and we can come back to it in a future PR

max-sixty (commented):
(tiny comments, I haven't done a full review of the java)

zhicwu:
> Thanks a lot for the big contribution! Appreciate it.

Sorry it took so long. I just spent too much time playing ChatGPT lately :p

> So I understanding — because we're using JNA rather than JNI — we're no longer linking with the `jni` rust crate; instead we're linking through `prql-lib`, is that correct? 

Yes. And this ensures consistency with all other language bindings, although it's just C# and PHP for now. 

> IIRC when we discussed this, we thought JNA was reasonable given that we don't care much about performance here (though we miss out on help from the `jni` rust crate, possible a worthwhile trade though...)

It's still JNI so the performance is not horrible. Besides, the library is extensible, so user can easily add custom implementation(using jni crate like before or a wrapper of prqlc command-line) as needed. Personally I'd rather build stateless web service using Rust instead of JNI for better throughput and scalability, when performance becoming an issue.

.github/workflows/release.yaml
zhicwu: This was mentioned before. I tried to remove the part but somehow it failed the build. Will try again to see how it goes.

.github/workflows/release.yaml
zhicwu: Yes, please. These are repository secrets required for publishing the signed artifacts to maven central. You have SONATYPE* already and I believe you have GPG key as well.

linux-china:
@zhicwu a question here, how to load prql_lib from file system?  Should I copy prql library into a path?  or packaged all library files into fat jar?

```java
import com.sun.jna.Native;

public class JnaImpl extends Prql.Factory implements Prql {
    private static final PrqlLibrary lib;

    static {
        lib = Native.load("prql_lib", PrqlLibrary.class);
    }

}
```

Refer from  https://developer.okta.com/blog/2022/04/08/state-of-ffi-java

# Java Native Interface (JNI)

### Pros

* Native interface access for C/C++/Assembly
* Fastest solution in Java

### Cons

* Complicated to use and brittle
* Not very secure and could cause memory safety issues
* Overhead and performance loss is possible
* Difficult to debug
* Depends on Java developers to write safe C binding code manually
* You need to compile and ship the C code for each target platform

# Java Native Access (JNA)

### Pros

* Native interface access for C/C++/Assembly
* Simpler to use compared to JNI
* Dynamic binding, no need to write any C binding code manually
* Widely used and mature library
* Better cross-platform support

### Cons

* Uses reflection
* Built on top of JNI
* Has performance overhead and can be slower than JNI
* Difficult to debug


Just my opinion only:  I prefer JNI for prql-java.  

*  JNI performance is better, and it's very important to transpile PRQL to SQL.  
*  With jni-rs, and almost no header file and code generation for JNI, and development is very easy, and no need to write safe C binding code manually.
*  File size of jna-5.13.0.jar is almost 1.8M, why add another dependency in prql-java
* APIs for PRQL just several, and it's not like SQLite with lots of APIs, and flexibility of JNA may be not very important for PRQL.

.github/workflows/test-java.yaml
zhicwu: Thanks. I realized that when copying it from test-java to release 😓  Will try to reuse that in a separate workflow.

zhicwu:
> @zhicwu a question here, how to load prql_lib from file system? Should I copy prql library into a path? or packaged all library files into fat jar?

Native libs are packed into the jar for ease of use. `prql-java` is ~10MB and JNA is 1.8MB.

> Just my opinion only: I prefer JNI for prql-java.
>    JNI performance is better, and it's very important to transpile PRQL to SQL.
>    With jni-rs, and almost no header file and code generation for JNI, and development is very easy, and no need to write >safe C binding code manually.
>    File size of jna-5.13.0.jar is almost 1.8M, why add another dependency in prql-java
>    APIs for PRQL just several, and it's not like SQLite with lots of APIs, and flexibility of JNA may be not very important for PRQL.

Fair enough. To me, JNA is probably easier to start with to unblock some of the work, while we can still add JNI in future release and change default implementation.

zhicwu:
Change to draft as I need to test publishing the artifact to Maven central tonight.

.github/workflows/release.yaml
max-sixty: OK, happy to help if it's not working

max-sixty:
So I understand:

> It's still JNI so the performance is not horrible.

> Just my opinion only: I prefer JNI for prql-java.

> To me, JNA is probably easier to start with to unblock some of the work, while we can still add JNI in future release and change default implementation.

...this is using JNI or JNA? Is it using the JNI interface but not the `jni` crate?

The `jni` crate did seem to make the build very easy — it was a single command to build on any platform, lots of the complexity was encapsulated. But I don't know enough to have a strong view — maybe it's difficult for folks with Java expertise to contribute, and that's worth a lot.

---

> Sorry it took so long. I just spent too much time playing ChatGPT lately :p

Haha, it is magic. I've been using GPT-4 for coding, it's so impressive. I did https://github.com/max-sixty/aoc-gpt a few months ago; next year AoC's is going to be quite interesting.

.github/workflows/release.yaml
max-sixty: Last time we spoke I thought you might be taking the next steps there. But I'm probably misremembering. Do you have the links to hand that I should follow to get these set up? I remember being a bit confused where to go for the Sonotype ones.

zhicwu: Sure. `SONATYPE_USER` and `SONATYPE_PASSWD` are the credentials you used to login https://issues.sonatype.org/ (and https://s01.oss.sonatype.org/). https://issues.sonatype.org/browse/OSSRH-89424 is the ticket for applying Maven central access - I'll request to revoke my access there once I complete testing.

Please refer to [this](https://docs.github.com/en/authentication/managing-commit-signature-verification/generating-a-new-gpg-key) to generate GPG key if you don't have. And use `gpg --keyserver keyserver.ubuntu.com --send-keys <key id>` to publish your key(required to ensure the artifacts are signed by you, not anyone else).

eitsupi:
Is additional work needed to merge this?

eitsupi:
I am curious if there is any reason why this cannot be merged.

max-sixty:
I'm sorry to have not replied for a while — it important to me and to the project that contributions like this receive prompt reviews & merges. Thank you very much to @eitsupi for reminding me.

I don't think I have enough context to have a strong view on whether this approach is better. To some extent, I trust @zhicwu 's judgement — he seems to be very competent indeed... I'd welcome feedback from @linux-china or others.

Echoing my comment above:

> The `jni` crate did seem to make the build very easy — it was a single command to build on any platform, lots of the complexity was encapsulated. But I don't know enough to have a strong view — maybe it's difficult for folks with Java expertise to contribute, and that's worth a lot.

It is a lot of code, and this approach relies on Java more than Rust, which is less available in the team. To the extent @zhicwu would like to remain involved with the project, it's possibly an easy choice to merge it. To the extent this was a generous one-off, then it's possible a more difficult decision — are we able to maintain this?

---

I'm also not sure where we were at re publishing to Maven. (Was anything there on me to do there?)

zhicwu:
Sorry for the late reply. Actually ChatGPT-4 can generate working Java code to use prql-lib, which is pretty much this pull request is about. So I think what really matters are:
1. Publishing compiled PRQL lib for multiple platforms, so that it's easier for all languages to integrate
2. JDBC driver(either use prqlc, the command line, or PRQL lib via JNI/JNA) for IDE integration

IMO, it's probably the best to manage them in separate repos.

I'm occupied lately but I should be able to clean up code and publish a JDBC driver using prqlc in a week or two.

zhicwu:
Just released the JDBC driver, which only uses `prqlc`, the command line. Let's close the PR for now. Perhaps I can add this into JDBCX as a connector, if prql-lib binaries are all published.

PR #2347, opened 2023-03-30
internal: Clarify `TyFunc` docstring

max-sixty:
As discussed on discord


aljazerzen (approved):
I usually include such small changes in other commits, in an effort to keep git history readable.

Issue #1636, opened 2023-01-27
Add `.prql` to linguist


eitsupi:
Of course this does not need to be done right now (and does not meet the submission criteria).

Registering syntax in linguist will activate syntax highlighting on GitHub.
https://github.com/github/linguist/blob/master/CONTRIBUTING.md#adding-an-extension-to-a-language
https://github.com/github/linguist/blob/master/lib/linguist/languages.yml

eitsupi:
This is the search result of the `prql` extension on GitHub.
https://github.com/search?q=path%3A*.prql&type=code

There are 296 files now (linguist requires 2000)

ofek:
Now it's 400.

PR #2348, opened 2023-03-30
docs(book): include PRQL changelog in book

### LINKED ISSUE #2344: Upload the changelog to the web
### What's up?

I am wondering if the changelog can be included in the book or website.

eitsupi:
Close #2344

Since `prql` code blocks in md files included in the Book will evaluated and replace with html, this PR introduce a new Option `no-eval` that modifies the Book preprocessor to not evaluate Changelog's `prql` code blocks.
Also add a pre-commit hook to avoid using bare `prql` code blocks on `CHANGELOG.md`.


CHANGELOG.md
```diff
@@ -233,17 +233,21 @@ This release has 74 commits from 12 contributors. Selected changes:
   formats. _format-arg_ can be `format:csv` or `format:json`. _string-arg_ can
   be a string in any format. (@aljazerzen & @snth, #1514)
 
-  ```prql
+  ```elm
```
max-sixty: Should these never be `prql`?

Maybe we add a grep lint in `pre-commit` if so; otherwise folks are going to forget?

Or are they OK / actually good in `prql`?

eitsupi: Thank you for your review.
The problem here is that in the book the `prql` code blocks are evaluated by the preprocessor and displayed alongside the `sql` code blocks.

If we could disable the evaluation of `prql` code blocks throughout the particular file with ignore settings, special comments or yaml front matter, I think we could continue to use `prql` code blocks there.
Is that possible now?

> Maybe we add a grep lint in `pre-commit` if so; otherwise folks are going to forget?

I did not know of a solution that would make that possible. This is a good method at this stage!

max-sixty: `prql_no_test` will skip over creating a comparison table: https://github.com/prql/prql/blob/13d0060dfe80ee3ab098b85207ef6fae9d367a55/web/book/src/lib.rs#L155

(Though do we sometimes _want_ a comparison table?)

max-sixty:
One thing we want to avoid here is:
- Someone adds an entry to the Changelog with a prql block; the code doesn't compile
- This is only picked up when _building_ the book, not in the book tests, since the book tests only look at the `.md` files in the book. We run the book tests every commit, but we only build the book in `build-web`
- Then `build-web` fails next time someone makes a small change to the web because of that issue, and we have to unravel what caused it, and we're reverting big PRs because of a small changelog problem.

We could run `build-web` on changes to the changelog; I think that would solve it

CHANGELOG.md
eitsupi: I think the issue here is the gap between rendering on GitHub and rendering on Book.
That is, if `prql` code blocks are recognized on GitHub in the future (#1636), Changelog will need to use `prql` code blocks, but may not want to output `sql` on the Book.

eitsupi: Considering that the syntax has changed and past `prql` code block can no longer be compiled, Changelog still needs to disable `prql` code evaluation on a file-by-file basis.

max-sixty: > Considering that the syntax has changed and past `prql` code block can no longer be compiled, Changelog still needs to disable `prql` code evaluation on a file-by-file basis.

We could use a comment such as `prql-book: ignore-file` here and react to that by ignoring the rest of the file: https://github.com/prql/prql/blob/13d0060dfe80ee3ab098b85207ef6fae9d367a55/web/book/src/lib.rs#L111

(I would much rather we did that than manually exclude `CHANGELOG.md`, even though it's more work)

max-sixty: > That is, if `prql` code blocks are recognized on GitHub in the future (#1636), Changelog will need to use `prql` code blocks, but may not want to output `sql` on the Book.

Yeah, we could use something like 

`````
```prql,no-test
derive x = 5
```
`````

...which GitHub would still handle as PRQL (at least it seems to work with `sql` just now)

But yes, we would have to go back and change old language blocks when the syntax no longer works

Comment (eitsupi) [reply CHANGELOG.md]: I remember the docusaurus syntax.

We can add information separated by spaces.
https://docusaurus.io/docs/markdown-features/code-blocks

````md
```jsx title="/src/components/HelloCodeTitle.js"
function HelloCodeTitle(props) {
  return <h1>Hello, {props.name}</h1>;
}
```
````

...is rendered as

```jsx title="/src/components/HelloCodeTitle.js"
function HelloCodeTitle(props) {
  return <h1>Hello, {props.name}</h1>;
}
```

In any case, I am sure we can discuss these new features later.

max-sixty: Yes, I actually thought I'd make a change to use a list, quick PR at https://github.com/PRQL/prql/pull/2350/files

(doesn't affect this PR though)

eitsupi: Wow, wonderfully fast work, thanks!
So can we merge this PR now?

max-sixty: Yes!

max-sixty: What should the instructions for including change blocks here be? "Always use `elm`"? "Only use `prql` if you want a comparison table"?

eitsupi: > What should the instructions for including change blocks here be? "Always use `elm`"? "Only use `prql` if you want a comparison table"?

Good point.

It seems we need to stay with `elm` for now.
For now, it seems that when the `prql` code block is evaluated, a header `PRQL` is added and the display is corrupted.

image

https://github.com/PRQL/prql/blob/fe1b7d487f300d51b72723ad9d3fdac9a84238dc/CHANGELOG.md?plain=1#L315-L327

eitsupi:
I added `no-eval` tag to skip evaluate the code block in 9e89bbbd3d323714e498c65b7bccf9db79b61bed.
This should work the same way as the `elm` on the book, so it can be used as a perfect replacement for the `elm` code block.
(Of course, we could continue to use `elm`.)

max-sixty:
Perfect! 

Should we also add a https://pre-commit.com/#pygrep so no one puts a `prql$` tag in the Changelog?

eitsupi:
> Should we also add a [pre-commit.com/#pygrep](https://pre-commit.com/#pygrep) so no one puts a `prql$` tag in the Changelog?

I added that in 8347d17ddedb7b08d4885f4e9636f355f199f586 and it seems working fine de5c5ff5c56710918f09b59444b29bb08769cd89

eitsupi:
It would be great if we could automatically link to issue numbers and GitHub user accounts like `pkgdown` does (https://pkgdown.r-lib.org/news/index.html), but my skills are not currently up to the task of implementing this...
https://github.com/r-lib/pkgdown/blob/c354aa7e5ea1f9936692494c28c89e5bdd31fc68/R/repo.R#L39-L46

.pre-commit-config.yaml
```diff
@@ -53,6 +53,15 @@ repos:
         types: ["rust"]
         # A regex searching sql files for either `prj-` or 'tgs-'.
         entry: "dbg!"
+  - repo: local
+    hooks:
+      - id: prql-codeblock
+        name: Prevent prql codeblocks evaluating in book
+        description: prql code blocks are evaluated and replaced in the book
```
max-sixty:
```suggestion
        description: prql code blocks are evaluated and replaced in the book; instead use `prql no-eval`
```

max-sixty:
> It would be great if we could automatically link to issue numbers and GitHub user accounts like `pkgdown` does (https://pkgdown.r-lib.org/news/index.html), but my skills are not currently up to the task of implementing this... https://github.com/r-lib/pkgdown/blob/c354aa7e5ea1f9936692494c28c89e5bdd31fc68/R/repo.R#L39-L46

Yes totally. A challenge for the future (...generations...)!

max-sixty:
I hit the button to get this in for 0.7.0, hope that's OK!

max-sixty:
In! Thanks @eitsupi !

eitsupi:
Thanks!

PR #2350, opened 2023-03-30
refactor: Use tags for the language code blocks

max-sixty:
Idea from https://github.com/PRQL/prql/pull/2348


eitsupi:
Great!

I checked CommomMark Spec and found that it mentions spaces as delimiter (info string).
https://spec.commonmark.org/0.30/#example-143

Could the delimiter be a space?

````md
~~~~    ruby startline=3 $%@#$
def foo(x)
  return 3
end
~~~~~~~
````

to

```html
<pre><code class="language-ruby">def foo(x)
  return 3
end
</code></pre>
```

`startline=3 $%@#$` is ignored.

max-sixty:
Changed to spaces!

PR #2358, opened 2023-03-30
fix: Throw error on invalid header args

### LINKED ISSUE #2230: Additional named args in header don't error

Jelenkee:
Closes #2230


max-sixty:
@Jelenkee oh no! https://github.com/PRQL/prql/pull/2245. I did link the issue from there — tell me what I could have done to make it more obvious though!

max-sixty:
I would also favor #2245 because it catches it earlier in the compilation, and generally that means we can get better error messages

Jelenkee:
Damnit! Sorry, didn't see it.
My fault. One should read a ticket before starting. Or issue in this case.

Maybe adding a comment with a open PR could help.

Anyway...

PR #2359, opened 2023-03-31
chore(test): set MySQL's `secure-file-priv=""` option in the command instead of conf file

eitsupi:
`secure-file-priv=""` needs to be set in the mysql container, but since it can be set with the command used when the container is created without using the configuration file, so remove the configuration file and the command is specified in the compose file.

### DIFF
--- a/prql-compiler/tests/integration-rdbms/conf/my.cnf
+++ b/prql-compiler/tests/integration-rdbms/conf/my.cnf
@@ -1,2 +1,0 @@
-[mysqld]
-secure-file-priv = ""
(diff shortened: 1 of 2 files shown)


eitsupi:
It does not seem working.

```log
 failures:

---- tests::test_rdbms stdout ----
thread 'tests::test_rdbms' panicked at 'called `Result::unwrap()` on an `Err` value: MySqlError { ERROR 1290 (HY000): The MySQL server is running with the --secure-file-priv option so it cannot execute this statement }', prql-compiler/tests/integration-rdbms/connection.rs:233:53
```

max-sixty:
Nice!

Issue #2353, opened 2023-03-30
Tweaks for integration tests


Jelenkee:
### What's up?

Some follow-ups for #2286 

* [ ] Remove previous integration tests, move the data files into this path
* [ ] Run in-process DBs even without an external connection (will be quite helpful if we start using this more)
* [ ] Add a description & link to https://github.com/prql/prql/blob/7ae804a9d3f42f6ed3d33c04ab910dca164c7efc/web/book/src/contributing/development.md#L310
* [ ] Implement whether to run external integration tests with a feature, rather than reading the CI env var (that could become a bit confusing otherwise)
- [ ] (minor) Possibly output the snapshots as CSVs / something both machine & person-readable, e.g. with headers, rather than one long line
- [ ] (minor) Consider whether we keep it in the current CI test vs move to another, possibly in test-all.yaml. It currently adds 40 seconds, which is not too bad. Or we condition the test-rust tests on a PR's paths, and avoid running it for small doc changes.
- [x] When starting docker compose the owner of prql-compiler/tests/integration/data/chinook is changed on Linux. Probably on macOS too. That should be fixed because you cannot edit the files in there. #2352

(edited by @max-sixty to add an item that @Jelenkee suggested. And now completed)

max-sixty:
>  Implement whether to run external integration tests with a feature, rather than reading the CI env var

Specifically for this, I was thinking:
- We have a [feature](https://doc.rust-lang.org/cargo/reference/features.html) in `prql-complier` called something like `test-external-dbs`, which is not `default`
- We put the list of tests to run under the feature (or use `cfg-if` if needed), so it's the longer list when that feature is active
- Run in CI with `--all-features`
- Remove the special-case code around CI from the library
- (minor) Also put dependencies behind this feature, so we didn't need to compile them otherwise

This would mean we get dependable, hermetic external tests, _and_ still get the in-process tests by default

PR #2362, opened 2023-04-01
test: Refactor integration tests

Jelenkee:
#2353


prql-compiler/tests/integration/docker-compose.yml
```diff
@@ -1,23 +1,46 @@
...
-        condition: service_healthy
+      POSTGRES_DB: dummy
+      POSTGRES_USER: root
+      POSTGRES_PASSWORD: root
+    volumes: &vol
+      - ./data/chinook:/tmp/chinook:ro
+  mysql:
+    # No arm64 image available; remove when one does become available
+    platform: linux/amd64
+    image: "mysql:debian"
```
eitsupi: Just a question. Why did you choose `debian` instead of `oracle` in the original PR (#2286)?

`oracle` seems to have a history of being added for arm64 support. (docker-library/mysql#680)

```sh
$ docker buildx imagetools inspect mysql:debian
Name:      docker.io/library/mysql:debian
MediaType: application/vnd.docker.distribution.manifest.list.v2+json
Digest:    sha256:01003f288c58133dad270905e2768e6a1d399d709a4a1375bc4f982f747252c3

Manifests:
  Name:      docker.io/library/mysql:debian@sha256:38d6c177f28baa3b93350a59f0aae2c032418f05e2efd979c9da22d0a27d6223
  MediaType: application/vnd.docker.distribution.manifest.v2+json
  Platform:  linux/amd64
```

```sh
$ docker buildx imagetools inspect mysql:oracle
Name:      docker.io/library/mysql:oracle
MediaType: application/vnd.docker.distribution.manifest.list.v2+json
Digest:    sha256:ff3828b105886f496e47eddced1676b2abd1419a40a65ac4ab387ba1c538a09f

Manifests:
  Name:      docker.io/library/mysql:oracle@sha256:60eccd02d676ed2214c07be7292da931717f89efffafc2d1f66456daee7c4dc6
  MediaType: application/vnd.docker.distribution.manifest.v2+json
  Platform:  linux/amd64

  Name:      docker.io/library/mysql:oracle@sha256:2cb5e59c2166e94a3fb7e4da239e09112097af483f9f2bc8f1f87e3394fdf533
  MediaType: application/vnd.docker.distribution.manifest.v2+json
  Platform:  linux/arm64/v8
```

Jelenkee: No specific reason. tbh I copied the section with mysql from the internet :wink:. And that includes the comment

eitsupi: I see.
Perhaps there is worth changing to `oracle` here.

prql-compiler/tests/integration/README.md
```diff
@@ -34,6 +31,8 @@ postgres server that should already have data loaded in.
 
 This will not run as a part of `cargo test`, but can be run with docker-compose.
 
+### TODO
+
```
max-sixty: Maybe all done? :)

prql-compiler/Cargo.toml
```diff
@@ -73,3 +73,6 @@ exactly = 1
 file = "../web/book/src/language-features/target.md"
 replace = 'prql version:"{{version}}"'
 search = 'prql version:"[\d.]+"'
+
+[features]
+test-external-dbs = []
```
max-sixty, line 78: Excellent.

We can also put the external features behind the feature, so we don't need to compile them locally; i.e. something like:

```toml
[dependencies]

[target.'cfg(not(target_family="wasm"))'.dev-dependencies]
chrono = {version = "0.4", features = [], default-features = false}
criterion = "0.4.0"
tokio = {version = "1", features = ["full"]}
tokio-util = { version = "0.7", features = ["compat"] }
pretty_assertions = "1.3.0"
csv = "1.2"
postgres = { version = "0.19.3", optional = true }
mysql = { version = "23", optional = true }
tiberius = { version = "0.12", features = ["sql-browser-tokio", "bigdecimal", "time"], optional = true }
rusqlite = { version = "0.29.0", features = ["bundled", "csvtab"] }
pg_bigdecimal = { version = "0.1", optional = true }

[features]
test-external-dbs = [
    "postgres",
    "mysql",
    "tiberius",
    "pg_bigdecimal",
]
```

prql-compiler/tests/integration/main.rs
```diff
@@ -1,20 +1,43 @@
...
+    use crate::connection::*;
 
     #[test]
-    fn test() {
-        // TODO: we could have a trait rather than mods for each of these?
-        let mut pg_client = postgres::connect();
-        let duckdb_conn = duckdb::connect();
-        let sqlite_conn = sqlite::connect();
+    fn test_rdbms() {
+        if env::var("SKIP_INTEGRATION").is_ok() {
```
max-sixty: (Fine to leave for a different PR)

I was thinking we would remove this, and only rely on whether the feature was enabled, so there's no external context leaking in here. I'll add a comment on how I was thinking about doing that in the CI file in case that's helpful.

prql-compiler/tests/integration/main.rs
```diff
@@ -30,258 +53,127 @@ mod tests {
...
-
-    mod sqlite {
-        use super::path;
-        use rusqlite::{types::ValueRef, Connection};
-
-        pub fn connect() -> Connection {
-            Connection::open(path("data/chinook/chinook.db")).unwrap()
+    fn get_connections(runtime: &Runtime) -> Vec<Box<dyn DBConnection>> {
+        let mut connections: Vec<Box<dyn DBConnection>> = vec![];
+        if let Ok(connection) = duckdb::Connection::open_in_memory() {
```
max-sixty: I'm not so sure about this design — if something can't be enabled it'll pass the tests, but actually we want to fail...

prql-compiler/tests/integration/main.rs
```diff
@@ -1,20 +1,43 @@
 // Re-enable on windows when duckdb supports it
 // https://github.com/wangfenjin/duckdb-rs/issues/62
 #![cfg(not(any(target_family = "windows", target_family = "wasm")))]
+// TODO enable it for all OS
+#![cfg(target_os = "linux")]
+#![cfg(feature = "test-external-dbs")]
```
max-sixty: (fine to leave for another PR)

I was thinking we would have something like:

```rust
fn get_connections() -> Vec<dyn Connection> {
   #[cfg(feature = "test-external-dbs")]
   vec![...]

   #[cfg(not(feature = "test-external-dbs"))]
   vec![...]
}
```

(not exact code, probably they need to be initialized etc)

Totally fine to have a different design — you're the decider — would be great to always run the inmemory DBs regardless of the feature, though.

WDYT?

.github/workflows/test-rust.yaml
```diff
@@ -79,7 +79,9 @@ jobs:
           runner.os == 'Linux') }}
         with:
           command: insta
-          args: test --unreferenced=auto ${{ inputs.target_option }}
+          args:
+            test --unreferenced=auto ${{ inputs.target_option }} --features
```
max-sixty: Instead of using the env var, we could do something like `${{ inputs.os == 'ubuntu-latest' && '--features=test-external-dbs' || '' }}`

max-sixty (commented):
Awesome, thanks a lot @Jelenkee — this is shaping up to be really excellent code. I left a few comments. Happy to merge now and do those changes in another PR — as you wish.

prql-compiler/tests/integration/main.rs
Jelenkee: You're right. That makes sense for in-memory dbs

prql-compiler/tests/integration/main.rs
Jelenkee: Good idea.

prql-compiler/Cargo.toml
Jelenkee: that doesn't work because dev-dependencies cannot be optional

prql-compiler/tests/integration/main.rs
Jelenkee: Should the test fail, if the feature is active and the DBs are not running? Probably yes

.github/workflows/test-rust.yaml
Jelenkee: I don't understand what you mean

prql-compiler/tests/integration/main.rs
max-sixty: Yes, for sure

prql-compiler/Cargo.toml
max-sixty: Ah, sorry!

.github/workflows/test-rust.yaml
max-sixty: sorry, let me add a suggestion

.github/workflows/test-rust.yaml
```diff
@@ -79,7 +76,9 @@ jobs:
...
+            test-external-dbs
```
max-sixty: Here's what I meant in the other comment — using the `os` value to determine whether to include the `--features` option. Does that make sense?

```suggestion
            test --unreferenced=auto ${{ inputs.target_option }}
            ${{ inputs.os == 'ubuntu-latest' && '--features=test-external-dbs' || '' }}
```

i.e. "if the `os` value is ubuntu, then include this string, otherwise don't include any string"

prql-compiler/tests/integration/main.rs
```diff
@@ -30,258 +48,136 @@ mod tests {
...
+            };
+            for (k, v) in results.iter().skip(1) {
+                pretty_assertions::assert_eq!(
+                    *first_result.1,
+                    *v,
+                    "{} == {}: {test_name}",
+                    first_result.0,
+                    k
+                );
             }
```
max-sixty, line 74: Your implementation is probably better at debugging, so feel completely free to keep it.

That said, I think it's possible to simplify it a bit — here's one approach. The result is basically the same (the only difference is that the snapshot file contains the input prql as the expression, because of the third item passed to `assert_snapshot!`)

```rust

    #[test]
    fn test_rdbms() {
        let runtime = Runtime::new().unwrap();
        let mut connections = get_connections(&runtime);

        for con in &mut connections {
            setup_connection(con.as_mut(), &runtime);
        }

        // for each of the queries
        glob!("queries/**/*.prql", |path| {
            // read
            let prql = fs::read_to_string(path).unwrap();

            if prql.contains("skip_test") {
                return;
            }

            for con in &mut connections {
                let vendor = con.get_dialect().to_string().to_lowercase();
                if prql.contains(format!("skip_{}", vendor).as_str()) {
                    continue;
                }
                let result = run_query(con.as_mut(), prql.as_str(), &runtime);
                let result_string = result.iter().map(|row| row.join(",")).join("\n");

                assert_snapshot!("rdbms".to_string(), &result_string, &prql);
            }
        });
    }
```

(hopefully this is helpful feedback, rather than dispiriting that someone is "rewriting your code", let me know either way)

prql-compiler/tests/integration/main.rs
```diff
@@ -1,20 +1,38 @@
 // Re-enable on windows when duckdb supports it
 // https://github.com/wangfenjin/duckdb-rs/issues/62
 #![cfg(not(any(target_family = "windows", target_family = "wasm")))]
+// TODO enable it for all OS
+#![cfg(target_os = "linux")]
```
max-sixty:
```suggestion
```

Together with the CI change, I think we can remove these now (I tested locally, it works great!)

prql-compiler/tests/integration/snapshots/integration__tests__rdbms@invoice_totals.prql.snap
```diff
@@ -0,0 +1,360 @@
+---
+source: prql-compiler/tests/integration/main.rs
+expression: result_string
+input_file: prql-compiler/tests/integration/queries/invoice_totals.prql
+---
+2009-01,2009-01-01,1,2,1.98,2,
```
max-sixty, line 6: Not on you at all, and def not necessary in this PR, but one thing we could do is adjust the queries so they produce more aggregated results — we probably get the same test coverage with more legible outputs...

max-sixty:
Looking great! We can merge whenever you're ready!

prql-compiler/tests/integration/snapshots/integration__tests__rdbms@invoice_totals.prql.snap
Jelenkee: Sure.
And we have to remove the skips for certain dbs. Especially MsSQL is hardly tested.

prql-compiler/tests/integration/README.md
```diff
@@ -24,42 +24,27 @@ cargo build.
...
 
 Steps to run the tests:
 
-1. Compile the integration test:
-
+1. Run docker compose (may take a while on the first time):
    ```
-   $ cargo build --test integration
+   $ cd prql-compiler/tests/integration
+   $ docker-compose up
```
eitsupi:
```suggestion
   $ docker compose up
```

Jelenkee:
@max-sixty Docs have been added. You may have a look at it. If everything is fine, you can merge

max-sixty:
Great — am promoting it from Draft and will merge!

max-sixty:
Great — am promoting it from Draft and will merge!

max-sixty:
Thanks @Jelenkee !

Is there anything else remaining on the infra? I think it's in a very good state.

If you thought https://github.com/PRQL/prql/pull/2362#discussion_r1155358865 was valid, fine to make that change.

We could also start thinking about the current exclusions — was there anything you saw that's a PRQL bug (more than just the DB doesn't support it — in which case maybe we could improve the error messages but can't make it work...)?

Issue #2104, opened 2023-03-11
dbt (jinja) integration been removed from prql-compiler?


eitsupi:
### What happened?

The functionality that was in prql-compiler for dbt integration seems to have been removed in #1722 (moved to prqlc).

https://github.com/PRQL/prql/blob/ed1e896ecbb1c5e18f0f1286639118e3a77b0100/book/src/integrations/dbt.md?plain=1#L156-L158

Perhaps the documentation needs updating?
And it might be better to put it on Changelog as a breaking change.

### PRQL input

```elm
from {{ ref('foo') }}
```

### SQL output

```SQL
None
```

### MVCE confirmation

- [X] Minimal example
- [X] New issue

max-sixty:
Yes; we never actually ended up using this — I couldn't get dbt to run things in the order that I wanted. (And now there are discussions about a deeper integration with dbt).

So yes we should remove these docs. I can do that unless anyone gets there first...

Yes we can add to the changelog too, no harm in doing that.

eitsupi:
Closes because the document has been updated.

PR #2366, opened 2023-04-02
docs(changelog): fix typo and tweak some items related to breaking chage of 0.6.0

eitsupi:
Related to #2104

A set of some minor fixes that I noticed when I looked at the Changelog that is now included in the Book.

- Fix some typos.
- Added some info on some of breaking changes
  - Pass through jinja template (`{{` and `}}`).
  - `switch` -> `case`

I tried to include the admonish note about the change of name from `switch` to `case` in the changelog, but it didn't work (bug?).
So I add it to the `case` page.


max-sixty:
Thanks a lot for spotting those; appreciate it!

max-sixty (approved):

PR #2377, opened 2023-04-03
revert: #2202

max-sixty:
I think #2202 causing this failure: https://github.com/PRQL/prql-vscode/actions/runs/4600865002/jobs/8128008789?pr=170 — `prql-js` is trying to run `wasm-pack` even when it's installed as a dependency.

Possibly there's a better way of doing this where it runs `wasm-pack` when building from scratch, but not when installing as a dependency. I can look more later, unless anyone knows off-hand


max-sixty:
Replaced by #2378

PR #2389, opened 2023-04-04
fix: re-export `MessageKind`

vanillajonathan:
Re-export the `MessageKind` enum so it becomes visible outside the crate.


max-sixty:
Looks very reasonable; requires `cargo fmt` though

vanillajonathan:
Ok, done!

max-sixty:
Thanks!

PR #2399, opened 2023-04-08
fix: column inference on relation literals

### LINKED ISSUE #2398: `select ![]` after `from s"..."` does not work
### What happened?

`select![]` after the `from` S-string does not seem to work.

### PRQL input

```elm
prql target:sql.duckdb

from s"""
SELECT *
FROM foo
"""
select ![bar]
```

### SQL output

```SQL
None
```

### Expected SQL output

```SQL
WITH table_0 AS (
  SELECT
    *
  FROM
    foo
)
SELECT
  * EXCLUDE (bar)
FROM
  table_0 AS table_1
```

### MVCE confirmation

- [X] Minimal example
- [X] New issue

### Anything else?

_No response_

aljazerzen:
Closes #2398

### DIFF
--- a/prql-compiler/src/ast/pl/frame.rs
+++ b/prql-compiler/src/ast/pl/frame.rs
@@ -30,16 +30,8 @@ pub struct FrameInput {
     /// Local name of this input within a query.
     pub name: String,
 
-    /// Source that provides the data for this input.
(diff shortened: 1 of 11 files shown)


max-sixty:
Nice!

PR #2403, opened 2023-04-09
devops: Enable building arm64 devcontainer images

max-sixty:
ref #2404

This fails IIUC; will confirm though. Can close assuming it does.


.github/workflows/devcontainer.yaml
```diff
@@ -16,6 +16,7 @@ concurrency:
 jobs:
   build:
     runs-on: ubuntu-latest
+    timeout-minutes: 14400
```
eitsupi: My understanding is that setting a value beyond 6 hours does not make sense for the GitHub hosted runner.

eitsupi (commented):
It's great that you could build this!

max-sixty:
OK this works! Required some more complication in the code, but it does work... Merging...

PR #2407, opened 2023-04-10
chore(ci): Reset caches

max-sixty:
These are large enough that they're starting to thrash; seeing whether building them fresh reducing the size at all. It may alternatively be that we have enough dependencies that it's too large, and we need to do something like a single cache job that writes for all targets, which every job then reads. It's a shame we can't get more space, even for a fee...

### DIFF
--- a/.github/workflows/build-web.yaml
+++ b/.github/workflows/build-web.yaml
@@ -54,6 +54,7 @@ jobs:
       - name: 💰 Cache
         uses: Swatinem/rust-cache@v2
         with:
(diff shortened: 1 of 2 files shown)


max-sixty:
This does seem to reduce the cache size a bit — likely the cache isn't completely removing dependencies that are no longer used, and so accumulate over time.

By looking at the caches page, a couple of the rust cache sizes went from:
- 1GB->830MB
- 1GB->680MB

PR #2409, opened 2023-04-10
feat: Add `read_{parquet,csv}` functions

max-sixty:
Rendered book content:

image


prql-compiler/src/sql/std_impl.prql
```diff
@@ -25,3 +25,7 @@ func as<scalar> `noresolve.type` column ->  s"CAST({column} AS {type})"
 # String functions
 func lower <text> column -> s"LOWER({column})"
 func upper <text> column -> s"UPPER({column})"
+
+# Source-reading functions, primarily for DuckDB
+func read_parquet<table> source<text> -> s"SELECT * FROM read_parquet({source})"
```
max-sixty, line 30: @aljazerzen would it be possible to confirm that it's correct to have these in both `std` & `std_impl`?

prql-compiler/src/tests/test.rs
```diff
@@ -3446,3 +3446,25 @@ fn test_upper() {
     "###
     );
 }
+
+#[test]
+fn test_read_parquet_duckdb() {
+    assert_display_snapshot!(compile(r#"
+    from (read_parquet 'x.parquet')
```
eitsupi, line 3453: I think we need to test to make sure it works when the string is enclosed in double quotes.

eitsupi (commented):
I'm not sure if targeting only DuckDB SQL is the right direction here. (Of course, this is an experimental phase now, so it's not a problem.)

It seems that Polars SQL requires the use of the `read_csv` function instead of `read_csv_auto`.
https://github.com/pola-rs/polars/blob/2ac1d77973d6e7673ae3cc789c2d2301e0c0d574/polars/polars-sql/src/table_functions.rs#L12-L15

Are you planning to introduce some kind of generic function where the contents of the function also changes depending on the target
 Or would we need to use a different module for a different target?

prql-compiler/src/sql/std_impl.prql
aljazerzen: Yes, the `std` one is used for declarations and the `sql::std_impl` is for SQL implementations.

I've started work on having different impls for different dialects, but we are not there yet.

web/book/src/language-features/standard-library/reading-files.md
```diff
@@ -0,0 +1,18 @@
+# Reading files
+
+We have a couple of functions named `read_*`, which ask the DB to read files,
+designed for DuckDB:
+
+```prql
+from (read_parquet 'artists.parquet')
+join (read_csv 'albums.csv') [==track_id]
+```
```
aljazerzen, line 9: What do you think about not even having `from`?

```suggestion
```prql
read_parquet 'artists.parquet'
join (read_csv 'albums.csv') [==track_id]
```
```

aljazerzen (approved):

aljazerzen:
Nice, looking good.

prql-compiler/src/tests/test.rs
max-sixty: Added! 

(I am continually surprised at the bugs we find, but if that didn't work then I'd be _really_ surprised :) )

web/book/src/language-features/standard-library/reading-files.md
max-sixty: But then we can't use with `join` etc?

(the very first version of PRQL, for like 3 days, didn't even have a `from`, it was just `employees | select`...)

max-sixty:
> Are you planning to introduce some kind of generic function where the contents of the function also changes depending on the target
> Or would we need to use a different module for a different target?

Yes, I think this is an important question.

Ideally we'd have functions with the same signature, but different outputs dependent on the target. If we can't get them to have the same signature, we could have different modules; e.g. `duckdb.read_parquet`...

PR #2414, opened 2023-04-12
refactor: SRQ, sql:preprocess and sql:anchor

aljazerzen:
- refactor SRQ AST
- prqlc sql:preprocess
- prqlc sql:anchor

I've formalized the fact that I've already introduced an intermediate AST representation in `prql_compiler::sql` module. This IR is named "Sql Relation Query" or SRQ.

With this realization it made sense to add cli commands that make it easy to run compiler passes on the AST:

```
$ cargo run -p prqlc -- sql:preprocess _a.prql
$ cargo run -p prqlc -- sql:anchor _a.prql
$ cargo run -p prqlc -- sql:compile _a.prql     # alias for compile
```

Makes work on #2079 and #2244 much easier I think.

### DIFF
--- a/prql-compiler/prqlc/src/cli.rs
+++ b/prql-compiler/prqlc/src/cli.rs
@@ -71,7 +71,15 @@ enum Command {
     #[command(name = "sql:preprocess")]
     SQLPreprocess(IoArgs),
 
+    /// Parse, resolve, lower into RQ & preprocess & anchor SRQ
+    ///
+    /// Only displays the main pipeline.
+    #[command(name = "sql:anchor")]
+    SQLAnchor(IoArgs),
+
     /// Parse, resolve, lower into RQ & compile to SQL
+    ///
+    /// Only displays the main pipeline and does not handle loop.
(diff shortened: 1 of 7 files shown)


prql-compiler/prqlc/src/cli.rs
```diff
@@ -67,8 +67,21 @@ enum Command {
         format: Format,
     },
 
+    /// Parse, resolve, lower into RQ & preprocess SRQ
+    #[command(name = "sql:preprocess")]
```
max-sixty, line 71: Traditionally I guess this would be a subcommand `prqlc sql preprocess`, but totally fine as-is

max-sixty (approved):
Excellent; I'm a big fan of these representations; particularly if we can start having tests from one to the other — then it becomes much easier for someone who doesn't know the whole compiler to focus on just a small part.

This also reminds me that I had a plan to write out the representations of each compiler stage of canonical examples — I can add this output too. I did the very first bit of this in https://github.com/PRQL/prql/pull/2418.

PR #2415, opened 2023-04-12
fix: sort after group

### LINKED ISSUE #2079: `group` & `select`  generates additional field "_expr_0"

aljazerzen:
- fix: excess columns
- fix: forbid anchoring computes after aggregate
- tweak prqlc sql:anchor
- test
- more tests

Closes #2079


web/book/tests/snapshots/snapshot__tests__prql__examples__variables-1.prql.snap
```diff
@@ -1,24 +1,30 @@
...
-  AVG(table_0._expr_0) / 1000 * 1000 AS salary
+  _expr_0 / 1000 AS salary_k,
+  _expr_0 / 1000 * 1000 AS salary
 FROM
   table_1 AS table_0
-  JOIN titles ON table_0.emp_no = titles.emp_no
-GROUP BY
-  titles.title
 LIMIT
   10
```
max-sixty, line 29: This is a good example of the changes — the SQL is a bit less succinct:
- We have an extra CTE
- `FROM table_3 AS table_2`

...but not in a dramatic way. It's possible an inevitable part of making the compiler pipeline more modularized — the whole thing becomes a bit less optimized, and it's probably a good tradeoff

prql-compiler/tests/integration/snapshots/integration__tests__rdbms@switch.prql.snap
```diff
@@ -3,14 +3,14 @@ source: prql-compiler/tests/integration/main.rs
...
-no composer,4884
-no composer,6373
-no composer,6635
-L. Muggerud,7941
-no composer,11650
-L. Muggerud,21211
-unknown composer,29048
-Gilberto Gil,32287
-Chico Science,33149
+Samuel Rosa
```
max-sixty: OK nice, this was wrong before! From https://github.com/prql/prql/blob/33fb9e02348158b1c0d6543514771a81285c09e0/prql-compiler/tests/integration/queries/switch.prql

max-sixty (approved):
Awesome, thanks a lot.

I left some notes on the SQL — obv the fix is incomparably more important though.

prql-compiler/tests/integration/snapshots/integration__tests__rdbms@switch.prql.snap
aljazerzen: Yeah, I didn't even notice :D

web/book/tests/snapshots/snapshot__tests__prql__examples__variables-1.prql.snap
aljazerzen: I don't think that these things will cause performance degradation to any modern optimizer. But it is admittedly less readable.

(My tolerance for badly looking target SQL went way way up since I started working for edgedb :D )

aljazerzen: `FROM table_3 AS table_2` could be fixed probably - it's making a distinction between a "table name" and a "table instance name" (in Postgres they call these two "relation" and "range var"). We could use "table name" as the "table instance name" here. But not in general, because same table, could be instanced in a query twice:

```
WITH tab AS (SELECT ...)
SELECT a.id, b.id
FROM tab a
JOIN tab b ON (a.id = b.parent_id)
```

aljazerzen: The extra CTE is also not required here, but it is required in a very similar case. I didn't think it was worth the time to find a rule that would separate these cases correctly.

In this PR, I've basically declared "computations of newly-aggregated expressions are not allowed". I needed this rule to make another query work, but it was a bit to strict because is needlessly affecting this query.

PR #2418, opened 2023-04-12
test: Test the SQL output of the integration test examples

max-sixty:
More notes inline

It's a big diff only because it removes an unnecessary level of `mod`


prql-compiler/tests/integration/main.rs
```diff
@@ -4,178 +4,191 @@
...
 
-        for con in &mut connections {
-            setup_connection(con.as_mut(), &runtime);
-        }
+#[test]
+fn test_sql_examples() {
+    glob!("queries/**/*.prql", |path| {
+        assert_snapshot!(compile(&fs::read_to_string(path).unwrap()).unwrap())
+    });
+}
```
max-sixty: This is the only material change in this file

PR #2420, opened 2023-04-12
devops: fix cache volume mounts of dev container

max-sixty:
@eitsupi — should we be mounting this? When it's mounted, none of the cargo tools work out of the box for me, such as `cargo insta`.

(This is more of a question than a pull-request — I thought it was easiest to show the diff — but great if we have a solution that means we can close this)

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -19,13 +19,6 @@
       ]
     }
   },
-  "mounts": [
-    {
-      "source": "devcontainer-cargo-cache",
-      "target": "/usr/local/cargo",
-      "type": "volume"
-    }
-  ],
   "postCreateCommand": {
     "install-maturin": "task install-maturin",
     "install-npm-dependencies": "task install-npm-dependencies"
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -21,7 +21,7 @@
   },
   "mounts": [
     {
-      "source": "devcontainer-cargo-cache",
+      "source": "devcontainer-cargo-cache-${devcontainerId}",
       "target": "/usr/local/cargo/registry",
       "type": "volume"
     }
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -19,6 +19,13 @@
       ]
     }
   },
+  "mounts": [
+    {
+      "source": "devcontainer-cargo-cache",
+      "target": "/usr/local/cargo/registry",
+      "type": "volume"
+    }
+  ],
   "postCreateCommand": {
     "install-maturin": "task install-maturin",
     "install-npm-dependencies": "task install-npm-dependencies"


max-sixty:
Check out the caching!

```
 build-devcontainer / build (pull_request) Successful in 14s 
```

eitsupi:
Without it, every rebuild will download many files from crates.io.
See devcontainers/templates#117

eitsupi:
Could you remove the docker volume once and try again?

max-sixty:
On this branch, `cargo insta` works, but on `main`, it doesn't... Does it work for you?

What about only mapping in the `/usr/local/cargo/registry` rather than the whole `/usr/local/cargo/`, which includes `bin`, which contains the tools?

IIUC the registry isn't that big — it's much much less work than caching the compilation of dependencies, for example (though that's not affected by this)

eitsupi:
`cargo insta` seems working on my container.

What I am thinking of as a possibility is that your previous use of the amd64 image has left you with Docker Volume `devcontainer-cargo-cache` with `cargo-insta`, etc. built for the amd64 target.
Could you try to remove the container and `docker volume rm devcontainer-cargo-cache`, then rebuild the container?

max-sixty:
Ah interesting that it works on yours.

Where is the `cargo-insta` binary? On mine it's in `/usr/local/cargo/bin`, so gets hidden when we mount that volume. Maybe yours is somewhere else, maybe in a user-specific path?

eitsupi:
> Where is the `cargo-insta` binary? On mine it's in `/usr/local/cargo/bin`, so gets hidden when we mount that volume. Maybe yours is somewhere else, maybe in a user-specific path?

Same place.

eitsupi:
It is pretty odd that it hides when the volume is mounted.
Is this a macOS specific bug?

eitsupi:
I open it on a new Codespace and cargo-insta is working.

image

max-sixty:
But surely that's what mounting a volume does? It replaces the path in the image with new contents, such that anything that was there is no longer accessible?

How did the `cargo-insta` binary get there? Was it installed after the volume was mounted? Do you have all the other binaries we install with `task install-cargo-tools` in that path?

Does the devcontainer work on this branch, without the volume?

Am v confused!

eitsupi:
> But surely that's what mounting a volume does? It replaces the path in the image with new contents, such that anything that was there is no longer accessible?
> 
> How did the `cargo-insta` binary get there? Was it installed after the volume was mounted? Do you have all the other binaries we install with `task install-cargo-tools` in that path?

My understanding is that when a Docker Volume is first connected to a container, the contents of the container are copied to the volume.
So I suspect that your Docker Volume is the cause, please make sure that you have completely deleted the Docker Volume (with the `docker volume ls` and `docker volume rm` commands) and try to create the container again.

> Do you have all the other binaries we install with task install-cargo-tools in that path?

Yes.

```sh
$ ls -la /usr/local/cargo/bin
total 312136
drwxrwxrwx  2 root   root         4096 Apr 13 14:55 .
drwxrwsrwx  5 vscode rustlang     4096 Apr 13 14:55 ..
-rwxr-xr-x  1 vscode vscode    4505176 Apr 13 00:04 bacon
-rwxrwxrwx 13 root   root     15788704 Mar 28 18:32 cargo
-rwxr-xr-x  1 vscode vscode   10763960 Mar 23 01:44 cargo-audit
-rwxr-xr-x  1 vscode vscode   24843360 Apr 12 23:58 cargo-binstall
-rwxrwxrwx 13 root   root     15788704 Mar 28 18:32 cargo-clippy
-rwxrwxrwx 13 root   root     15788704 Mar 28 18:32 cargo-fmt
-rwxr-xr-x  1 vscode vscode    6517280 Mar 23 11:28 cargo-insta
-rwxrwxrwx 13 root   root     15788704 Mar 28 18:32 cargo-miri
-rwxr-xr-x  1 vscode vscode   14070624 Mar 21 22:06 cargo-release
-rwxrwxrwx 13 root   root     15788704 Mar 28 18:32 clippy-driver
-rwxr-xr-x  1 vscode vscode    2182520 Apr 11 01:39 default-target
-rwxr-xr-x  1 vscode vscode   13374272 Mar  5 00:33 mdbook
-rwxr-xr-x  1 vscode vscode    3254736 Oct 23 14:25 mdbook-admonish
-rwxr-xr-x  1 vscode vscode    3767576 Apr 11 01:41 mdbook-footnote
-rwxr-xr-x  1 vscode vscode    1423776 Feb 14 09:59 mdbook-toc
-rwxrwxrwx 13 root   root     15788704 Mar 28 18:32 rls
-rwxrwxrwx 13 root   root     15788704 Mar 28 18:32 rustc
-rwxrwxrwx 13 root   root     15788704 Mar 28 18:32 rustdoc
-rwxrwxrwx 13 root   root     15788704 Mar 28 18:32 rustfmt
-rwxrwxrwx 13 root   root     15788704 Mar 28 18:32 rust-gdb
-rwxrwxrwx 13 root   root     15788704 Mar 28 18:32 rust-gdbgui
-rwxrwxrwx 13 root   root     15788704 Mar 28 18:32 rust-lldb
-rwxrwxrwx 13 root   root     15788704 Mar 28 18:32 rustup
-rwxr-xr-x  1 vscode vscode    3992040 Feb  2 01:48 wasm2es6js
-rwxr-xr-x  1 vscode vscode    6539160 Feb  2 01:48 wasm-bindgen
-rwxr-xr-x  1 vscode vscode    8137992 Feb  2 01:49 wasm-bindgen-test-runner
-rwxr-xr-x  1 vscode vscode   10949648 Mar 20 02:03 wasm-pack
```

max-sixty:
OK you're completely correct, sorry.

> My understanding is that when a Docker Volume is first connected to a container, the contents of the container are copied to the volume.

This wasn't my understanding, I'll look more — possibly I was just wrong.

---

Though I'm still not sure that mapping in this whole path is useful, as opposed to just the `registry` — it means if we add another tool to the canonical tools, it doesn't get reflected in devcontainers that have already been created. Is that right? 

Is there anything apart from `registry` that would require rebuilding? Could we add that to the docker build?

---

Edit: Yes I was wrong — for paths that don't map to a host path, or where the host path doesn't exist, the volume starts with the image's contents. IIUC this also means it won't update when the image updates — so I think we should limit this path to just the `registry`, or not have it at all, given that we rely on the tools in that path.

Again, almost all the work to compile is when _building_ the dependencies (cached in `prql/target` by default), not on downloading the registry. (Does everyone agree with that point?)

eitsupi:
Indeed, this was originally necessary because cargo install was done after the container was created, and now that they are already installed in the Docker image, it makes sense not to use volume.
It would make sense to change only holding the cache of the registry.

max-sixty:
OK done! I'll merge but lmk any follow ups

.devcontainer/devcontainer.json
```diff
@@ -22,7 +22,7 @@
   "mounts": [
     {
       "source": "devcontainer-cargo-cache",
```
eitsupi:
```suggestion
      "source": "devcontainer-cargo-cache-${devcontainerId}",
```

As pointed out in devcontainers/templates#117, this seemed necessary.

max-sixty: Ah thanks!

eitsupi:
Slight modification. Thanks for trying this out!

PR #2421, opened 2023-04-13
build: Fix build for linux-aarch64 target

max-sixty:
Seems to be the same issue as https://github.com/remacs/remacs/issues/1393

Found when trying to build in the devcontainer.

Until there are GitHub Actions runners on ARM, this isn't easy to test. I'm not sure why the pointer representation is different between linux & mac or ARM...


aljazerzen (approved):
If this works, it's ok.

But it's a sign that we should not be depending on archaic C semantics and type definitions, but should switch to [modern type definitions](https://flatbuffers.dev/) in prql-lib.

PR #2422, opened 2023-04-13
feat: rename operators and & or to && & ||

### LINKED ISSUE #2170: Infix operators — `and` & `or` vs `&&` & `||`
### What's up?

Do we want to keep `and` & `or` and infix operators?

Currently there's almost a clear split:
- Symbols, such as `+`, `*`, are all infix — i.e. they go between operands, `5 + 8`, `4 * 3`
  - Some are prefix, such as `!` or `-` when they're unary
- Functions are all prefix — i.e. they go before operands — i.e. `sum foo`, `in 5..10 bar`
  - And then the pipe symbol lets them go after operands — i.e. `foo | sum`, `bar | in 5..10`, which is how we normally use them

..._except_ for `and` & `or`, which are infix.

So we could instead use `&&` and `||` for these. That's a bit "techie" for my liking, but the consistency would be good.

aljazerzen:
Closes #2170

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@
 - New functions `read_parquet` and `read_csv`, which mirror the DuckDB
   functions, instructing the database to read from files (@max-sixty, #2049).
 
+- Rename `and` to `&&` and `or` to `||` (@aljazerzen, #2422).
+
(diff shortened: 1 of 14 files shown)


max-sixty:
Thanks!

PR #2423, opened 2023-04-13
docs: Fix Operator table to display "||" properly;

richb-hanover:
Had to cheat and use `&#x7c;` because the pipe character confused the table;
Rename section heading to "Operators and Expressions" because it was hard to find the operators

### DIFF
--- a/web/book/src/syntax.md
+++ b/web/book/src/syntax.md
@@ -16,7 +16,7 @@ A summary of PRQL syntax
 | `=`             | [Assigns](transforms/select.md) & [Aliases](transforms/join.md)      | `from e = employees` <br> `derive total = (sum salary)` |
 | `:`             | [Named args & Parameters](queries/functions.md)                      | `interp low:0 1600 sat_score`                           |
 | `[]`            | [Lists](./syntax.md#lists)                                           | `select [id, amount]`                                   |
-| `()`            | [Precedence & Parentheses](./syntax.md#precedence-and-parentheses)   | `derive celsius = (fahrenheit - 32) / 1.8`              |
+| `+ - * /`       | [Operators & Expressions](./syntax.md#operators-and-expressions)     | `derive celsius = (fahrenheit - 32) / 1.8`              |
 | `''` & `""`     | [Strings](language-features/strings.md)                              | `derive name = 'Mary'`                                  |
 | `` ` ` ``       | [Quoted identifiers](./syntax.md#quoted-identifiers)                 | `` select `first name`  ``                              |
 | `#`             | [Comments](./syntax.md#comments)                                     | `# A comment`                                           |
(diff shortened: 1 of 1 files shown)


eitsupi:
> the pipe character confused the table

I suppose escaped by back slash like `\|`

web/book/src/syntax.md
```diff
@@ -16,7 +16,7 @@ A summary of PRQL syntax
 | `=`             | [Assigns](transforms/select.md) & [Aliases](transforms/join.md)      | `from e = employees` <br> `derive total = (sum salary)` |
 | `:`             | [Named args & Parameters](queries/functions.md)                      | `interp low:0 1600 sat_score`                           |
 | `[]`            | [Lists](./syntax.md#lists)                                           | `select [id, amount]`                                   |
-| `()`            | [Precedence & Parentheses](./syntax.md#precedence-and-parentheses)   | `derive celsius = (fahrenheit - 32) / 1.8`              |
+| `+ - * /`       | [Operators & Expressions](./syntax.md#operators-and-expressions)     | `derive celsius = (fahrenheit - 32) / 1.8`              |
```
max-sixty, line 19: We definitely want to reference parentheses here though — the table is a lookup for various forms of syntax, and parentheses is one of them... (It's also required even without operators)

Should we have a different line for these?

max-sixty (commented):
Thanks for fixing. 

I did something similar on L15 for the pipe symbol for reference

web/book/src/syntax.md
richb-hanover: re: escaping the pipe character (e.g.`\|`)... That would work except the "\|\|" is in backticks which don't get masked from Markdown. Fortunately, the problem character is "|" - it won't be obvious that it isn't in a monospace font. I'll change it back to the simpler `\|`

re: Expressions / Operators / Parentheses / Precedence: I think Expressions takes in the three other concepts. Of course, parentheses are important and required for precedence, and it's good to show the priority in that table. But people searching for what operator to use for `and` are more likely to look for "operator" or "expression" than "parentheses"...

To aid with searching, I am considering factoring _syntax.md_ into its own directory, so that all of its disparate concepts can be called out separately in the TOC sidebar. Any reason not to? Thanks.

max-sixty: I think we should probably add one new line for `&&` & `||`. I don't think we really need a line for operators such at `+`, they're so obvious. But no strong view if we want to add them. We already have an `Expressions` line so we could put them there?

We definitely need to keep the line for parentheses — that's the syntax that needs the most documentation, it needs a section. I'm open to almost any change apart from removing Parentheses from being obvious :). If we wanted to move the precedence table out of that section into `Expressions`, I'd be open to that.

Separately, we could also add the fat arrow and link to Case? (which I realize is in a weird order in the index, because it's still in the location for Switch...)

---

Yes, open to it having its own section, you're right it's difficult to navigate at the moment. If we do that, let's do a single PR with only that change, since it'll be a big diff.

richb-hanover: > Yes, open to it having its own section, you're right it's difficult to navigate at the moment. If we do that, let's do a single PR with only that change, since it'll be a big diff.

Yes. I will do this (new directory, single PR) over the weekend.

richb-hanover:
I'm closing this PR in favor of a new PR that factors the syntax.md file into a new /syntax directory

PR #2425, opened 2023-04-14
docs: Factor "syntax.md" into /syntax folder

richb-hanover:
This PR factors the single top-level _syntax.md_ into its own `/syntax` folder, with each of the components described in its own file. 

Extremely small editorial changes to the component files in the `/syntax` folder so the only diff compare is to ensure that I got all the links right.

This PR does _not_ address the Precedence/Parentheses question; that's next, but it's easier to see now that the files are more focused


eitsupi:
This is great because every time I opened the book I was lost as to where the operators were listed!

aljazerzen (approved):
This bothered me too, thanks for factoring

richb-hanover:
@aljazerzen @eitsupi Can either of you see why the test is failing?  Thanks.

eitsupi:
Perhaps the snapshots of the snapshot tests needs to be updated?

max-sixty:
Could we run `task test-rust`?

max-sixty:
This ranks as the most universally appreciated PR @richb-hanover !

aljazerzen:
With many things that were added to the book, we should think about reorganizing it so it reads more like a book and not like a reference manual.

max-sixty:
> With many things that were added to the book, we should think about reorganizing it so it reads more like a book and not like a reference manual.

Yes interesting point.

WDYT the best way of doing that is?

I could imagine being more focused on some canonical examples?

(maybe we open a new issue if necessary)

richb-hanover:
@max-sixty Thanks for merging. You probably saw that the backport failed

Issue #1965, opened 2023-02-26
Type system


aljazerzen:
Up until now, PRQL hasn't defined what "relation" is, at least on the language
level. We've been talking about numbers, arithmetic operations and datetimes,
without really defining them and offloading their definitions to what ever
database each of us has in mind. But their definitions are important, with
the simplest example of expression `4 - 5`, which has different results depending
on whether integers are signed or not.

So goal of this issue is to settle the definition of what our functions are
operating on. This will allow us to define what results of expressions _should
be_ and then try to implement that behavior for each of the DBMSs.

This will obviously (and hopefully) trigger lots of responses.
To structure the discussion better, I've written something resembling a RFC in PR #1964.
I suggest we use review comments to stay on topic.

This issue can remain open as a tracking issue for new type-system-related issues 
and for PRs implementing it.

Ref #381

PR #2428, opened 2023-04-14
refactor: converge RQ exprs to `BuiltInFunction`

aljazerzen:
In preparation for #1965. 

For this input:

```elm
from orders
filter status == 'open'
select [
    item_sum + tax - discount
]
```

... this PR changes RQ:
```diff
33,35c33,36
<     kind: !Binary
<       left:
<         kind: !ColumnRef 0
<       op: Eq
<       right:
<         kind: !Literal
---
>     kind: !BuiltInFunction
>       name: std.eq
>       args:
>       - kind: !ColumnRef 0
>       - kind: !Literal
52,56c51,57
<       kind: !Binary
<         left:
<           kind: !Binary
<             left:
<               kind: !ColumnRef 1
<             op: Add
<             right:
<               kind: !ColumnRef 2
<         op: Sub
<         right:
<           kind: !ColumnRef 3
---
>       kind: !BuiltInFunction
>         name: std.sub
>         args:
>         - kind: !BuiltInFunction
>             name: std.add
>             args:
>             - kind: !ColumnRef 1
>             - kind: !ColumnRef 2
>         - kind: !ColumnRef 3
```

Basically, we are removing `ExprKind::Binary`, `ExprKind::Unary` and `ExprKind::FString` in favor of `ExprKind::BuiltInFunction` (which should maybe be renamed to `ExprKind::FuncCall`).

There are a few minor reasons for this change:
- reduce the surface area of RQ,
- it's a step toward having dialect differences written in PRQL instead of constructing sqlparser AST. To do that, we'd override `std_impl.prql`,
- distinguish between different operator implementations. For example, we may want for operator `+` in PRQL to be able concatenate strings. With this PR, that operator could be translated either to `std.add` or `std.concat` depending on the type of operands.

I've also added a bit of machinery for matching and unpacking these functions in `sql::std`.


prql-compiler/src/sql/gen_expr.rs
```diff
@@ -514,59 +484,66 @@ pub(super) fn translate_select_item(cid: CId, ctx: &mut Context) -> Result<Selec
...
-) -> Result<Option<sql_ast::Expr>> {
-    if matches!(op, BinOp::Eq) || matches!(op, BinOp::Ne) {
-        let expr = if matches!(a.kind, ExprKind::Literal(Literal::Null)) {
-            b.kind.clone()
-        } else if matches!(b.kind, ExprKind::Literal(Literal::Null)) {
-            a.kind.clone()
-        } else {
-            return Ok(None);
-        };
+fn try_into_is_null(expr: Expr, ctx: &mut Context) -> Result<Result<sql_ast::Expr, Expr>> {
```
max-sixty, line 533: Is it worth adding a note why we have wrapped `Result`s? What does it mean to have an error on the inner vs the outer?

web/book/tests/snapshots/snapshot__tests__prql__language-features__f-strings-1.prql.snap
```diff
@@ -1,16 +1,19 @@
...
+    CONCAT(
+      CONCAT(
+        CONCAT(
+          CONCAT(CONCAT(CONCAT('http', tls), '://www.'), domain),
+          '.'
+        ),
+        tld
+      ),
+      '/'
+    ),
```
max-sixty: How viable would it be to have `CONCAT` be a variarg function? This isn't the friendliest SQL. (But not a showstopper by any means...)

I guess it's quite difficult at the moment given the proposed code has the number of args in the compiler (https://github.com/PRQL/prql/pull/2428/files#diff-2540aa8b42227cb2ab1983dcc56db5f2401778a2ba89fb1c33525c31c59d7e6bR97) , and we don't yet support lists?

max-sixty (commented):
Added a couple of quick thoughts, overall looks v reasonable, thanks. I can have another look through later if helpful

web/book/tests/snapshots/snapshot__tests__prql__language-features__f-strings-1.prql.snap
aljazerzen: I've intentionally made functions have predefined number and types of arguments. But limitation is imposed on RQ only, we can do some clever translations in sql backend to output a single CONCAT function call.

prql-compiler/src/sql/gen_expr.rs
aljazerzen: Yes it is, will do.

max-sixty:
Awesome, thanks for the `CONCAT` changes!

PR #2429, opened 2023-04-14
docs(prqlc): Update prqlc README.md to correct `SELECT *`

reivilibre:
On a manual run I get

```
SELECT
  salary
FROM
  employees
WHERE
  has_dog
```
as output. 

`SELECT *` gives a bad first impression because it can be inefficient.


eitsupi (approved):
Thank you for catching this!

PR #2431, opened 2023-04-14
fix: [Early WIP] Allow Idents to contain namespaces

max-sixty:
As discussed, this is me publishing my working code to explore #1535. It's been sitting around for a few weeks.

I expect none of this code will make it into the final version, but it might make it possible for @aljazerzen to say something like "you're on the wrong track, you need to look at this other function"...

Open to feedback on better ways of presenting the code — it's currently literally my working code with `dbg!` statements everywhere; maybe I should add more comments on what I see as going on?


prql-compiler/src/semantic/context.rs
```diff
@@ -196,14 +197,20 @@ impl Context {
             }
         }
 
+        // 1535: I think the thing we need to do is allow this lookup to create
+        // a new module!
```
aljazerzen: A little further down, but generally, correct.

prql-compiler/src/semantic/module.rs
```diff
@@ -83,6 +83,7 @@ impl Module {
     pub fn insert(&mut self, ident: Ident, entry: Decl) -> Result<Option<Decl>> {
         let mut ns = self;
 
+        // 1535: this is where I think we insert
```
aljazerzen, line 86: Correct

prql-compiler/src/semantic/resolver.rs
```diff
@@ -552,31 +554,72 @@ impl Resolver {
...
                     .map(|t| matches!(t, Ty::Table(_)))
-                    .unwrap_or_default();
-
-                is_table
-            });
+                    .unwrap_or_default()
+            }));
 
-        let has_tables = !tables.is_empty();
+        let has_tables = !dbg!(&tables).is_empty();
```
aljazerzen, line 567: This "has_tables" is badly named. It should be "has_relations" and it is true for all functions that take a relation as an arguments: `from`, `select`, `take`...

Code here is injecting `_frame` and `_right` modules into the context, but only for the duration of resolving arguments of this functions call. In effect, this allows you this:

```
FunctionCall:
  args:
  - column foo   <- this foo can "see" relation a and resolve to a.foo
  - relation a
```

... so this is not where you'd want to infer modules in `default_db`.

aljazerzen (commented):
This is the part that does the inference:

https://github.com/PRQL/prql/pull/2431/files#diff-54f0771b2a1409b9bf65752bfcb28d90ff4681d8c8fff92af3581c284cbef0d4R218-R230

It:
- finds the declaration that was resolved from the wildcard (i.e. `default_db.*`),
- assumes that this declaration is a `DeclKind::Infer`, which contains the declaration that should be instanced when a wildcard is resolved (in this case a `DeclKind::TableDecl`),
- inserts that decl into parent module of the infer decl (in this case `default_db`)

prql-compiler/src/semantic/context.rs
max-sixty: But once we hit [this line](https://github.com/PRQL/prql/pull/2431/files#diff-54f0771b2a1409b9bf65752bfcb28d90ff4681d8c8fff92af3581c284cbef0d4R214) `0 => Err(format!("Unknown name {ident}")),`, we raise. 

So presumably we need to add a new module into the root before we hit [this section you reference above](https://github.com/PRQL/prql/pull/2431/files#diff-54f0771b2a1409b9bf65752bfcb28d90ff4681d8c8fff92af3581c284cbef0d4R218-R225) — is that correct?

Or maybe we should we be adding it in a previous pass of that section?

prql-compiler/src/semantic/resolver.rs
max-sixty: OK, I commented out this code

max-sixty: The reason I was thinking to add it here is that presumably we only want to add modules when the reference is somewhere we could expect relations. So for example:

```
from x.y.z # we want to add an x module here
select a.b.c # wee don't want to add an a or b module here
```

But probably we could do that elsewhere?

aljazerzen: Oh, I see the reasoning.

Yes, we could handle this elsewhere, but I don't know where exactly at the moment. Will come back to this.

aljazerzen: My investigation into this issue ended in major change with how declarations are inferred. So I opened a PR that implements this.

aljazerzen:
Superseded by #2563

PR #2443, opened 2023-04-15
docs: Add `&&` & `||` to syntax table

max-sixty:
The diff is bigger than it looks, rendered comparison forthcoming


max-sixty:
Existing:

image


Proposed: 
image

It's not great that it spans two lines. But the existing approach already spanned two lines in other columns, and I'm not sure of a better way — we def don't want to have a line per operator where we announce the meaning of `+`...

max-sixty:
@richb-hanover for your review...

richb-hanover:
Good idea to add `&&` and `||` Merge it. 

I'm still thinking about how to handle parentheses and will build off what's committed at the time.

PR #2449, opened 2023-04-15
feat: Fixed integration tests for more vendors

Jelenkee:
Some queries are still invalid for MsSQL. I added some comments to the queries to explain why they don't work.


prql-compiler/tests/integration/queries/invoice_totals.prql
```diff
@@ -26,7 +25,8 @@ select [
     day,
     num_orders = s"CAST({num_orders} AS INT)",
     num_tracks = s"CAST({num_tracks} AS INT)",
-    total_price = s"ROUND(CAST({total_price} AS FLOAT) * 100) / 100",
+    total_price = s"ROUND(CAST({total_price} AS FLOAT) * 100, 0) / 100",
     running_total_num_tracks = s"CAST({running_total_num_tracks} AS INT)",
     num_tracks_last_week = s"CAST({num_tracks_last_week} AS INT)"
 ]
+take 20
```
max-sixty, line 31: Great idea to reduce the snapshots.

In another PR we could try adding more prql-specific features here, like `take 10..20`, or using `as` rather than the s-string `CAST`...

prql-compiler/tests/integration/queries/invoice_totals.prql
```diff
@@ -1,5 +1,4 @@
-# skip_mysql
-# skip_mssql
+# skip_mssql (error: The function 'LAG' may not have a window frame.)
```
max-sixty, line 1: Excellent notes

prql-compiler/tests/integration/connection.rs
```diff
@@ -29,17 +32,21 @@ pub struct MysqlConnection(pub mysql::Pool);
...
 
 pub trait DBConnection {
-    fn run_query(&mut self, sql: &str, runtime: &Runtime) -> Vec<Row>;
+    fn run_query(&mut self, sql: &str, runtime: &Runtime) -> Result<Vec<Row>>;
 
     fn import_csv(&mut self, csv_name: &str, runtime: &Runtime);
 
     fn get_dialect(&self) -> Dialect;
+
+    fn modify_sql(&self, sql: String) -> String {
```
max-sixty, line 43:
```suggestion
    // We sometimes want to modify the SQL `INSERT` query (we don't modify the
    // SQL `SELECT` query)
    fn modify_sql(&self, sql: String) -> String {
```

Just to make clear that we're testing the `SELECT` without modification. We could also rename the method if we have the patience, but no great stress.

prql-compiler/tests/integration/connection.rs
```diff
@@ -242,45 +243,71 @@ impl DBConnection for MysqlConnection {
...
+        Ok(vec)
     }
 
     fn import_csv(&mut self, csv_name: &str, runtime: &Runtime) {
-        self.run_query(&format!("LOAD DATA INFILE '/tmp/chinook/{csv_name}.csv' INTO TABLE {csv_name} FIELDS TERMINATED BY ',' OPTIONALLY ENCLOSED BY '\"' LINES TERMINATED BY '\n' IGNORE 1 ROWS;"), runtime);
+        // hacky hack for MySQL
+        // MySQL needs a special character in csv that means NULL (https://stackoverflow.com/a/2675493)
+        // 1. read the csv
+        // 2. create a copy with the special character
+        // 3. import the data and remove the copy
```
max-sixty, line 268: Gosh, thanks for doing this, not easy!

prql-compiler/tests/integration/connection.rs
```diff
@@ -242,45 +243,71 @@ impl DBConnection for MysqlConnection {
...
             }
             vec.push(columns);
         }
-        vec
+        Ok(vec)
     }
 
     fn import_csv(&mut self, csv_name: &str, runtime: &Runtime) {
-        self.run_query(&format!("LOAD DATA INFILE '/tmp/chinook/{csv_name}.csv' INTO TABLE {csv_name} FIELDS TERMINATED BY ',' OPTIONALLY ENCLOSED BY '\"' LINES TERMINATED BY '\n' IGNORE 1 ROWS;"), runtime);
+        // hacky hack for MySQL
```
max-sixty, line 264:
```suggestion
        // hacky hack for MySQL
        // TODO: could use https://docs.rs/tempfile/latest/tempfile/struct.NamedTempFile.html
```

(I don't think it's worth doing now, this is more than fine atm)

max-sixty (approved):
Thanks — this is a great demo of why these tests are so helpful — nice improvements in compatibility!

I appreciate there's lots of work to get them to run — e.g. the MySQL stuff...

prql-compiler/tests/integration/connection.rs
Jelenkee: That's not true. the `SELECT` is modified as well ([see here](https://github.com/PRQL/prql/blob/0153056deadae7a573c5f15f365c39bd487fdeda/prql-compiler/tests/integration/main.rs#L183)). It is necessary for casting. For example in MySQL the type `INT` exists, but to cast it to an int you need to use `SIGNED` :facepalm:

prql-compiler/tests/integration/connection.rs
Jelenkee: tempfile won't work because it uses the tmp directory of the OS. But the new file must exist in that directory because it is "mounted" to the docker container

max-sixty: Sorry, you're right. I had thought that possibly it was sending the _contents_ of the file, but it's indeed loading from the file itself.

prql-compiler/tests/integration/connection.rs
max-sixty: Ah, good point, sorry for my misunderstanding. I see we changed from https://github.com/PRQL/prql/blob/712ae15296a6ba41092b7e870595489425bb1d54/prql-compiler/tests/integration/main.rs#L160-L164

I'm a bit concerned about this approach, because it hides the frailties of PRQL. It's possible to make the tests pass by adding things to this function, but PRQL is still arguably producing the incorrect SQL.

I had a look through the examples — they're mostly types — which aren't easy to get precise given the heterogeneity there. Maybe that's OK for the moment — these tests are valuable enough that we don't want to have to exclude everything. 

There's one that's quotes — that's more concerning — I think we should instead fix that if it's from PRQL query. https://github.com/PRQL/prql/blob/0153056deadae7a573c5f15f365c39bd487fdeda/prql-compiler/tests/integration/connection.rs#L278

I think two options from here:
1. Remove these replacements, and put the `CAST` functions in a single example and exclude that for DBs (but maybe for all DBs, in which case less attractive...)
2. Split up the replacements for `schema.sql` — which is totally fine to mangle — and for the queries themselves — and have `TODO`s around the latter. Then in the long term we can think about how to eliminate them.

WDYT?

prql-compiler/tests/integration/queries/invoice_totals.prql
max-sixty: I see we have another example with the more advanced `take` syntax, and I put in https://github.com/PRQL/prql/pull/2462 for the s-strings

prql-compiler/tests/integration/connection.rs
Jelenkee: I removed the replacement of the double quotes because it's not necessary. It was added, when there was table with an uppercase letter in the schema.sql. Btw, we should an integration test for that. It's a tricky subject.

I agree with you that with the current modifying of SQL some issues of the compiler can slip through.
I would prefer the first option.

max-sixty: OK great — do you think we can do the first option? If we exclude `invoice_totals` from most of the dialects, can we remove these replacements from the query path?

Jelenkee: I removed the CASTs from invoice_totals.prql and added it to a separate file

Jelenkee: #2462 won't be necessary anymore

max-sixty:
Code looks good! I hope the float rounding is consistent across DBs. We could comment out that column for the moment if it's not.

Looks like it needs a `cargo insta test --accept` run for the SQL snapshots.

prql-compiler/tests/integration/connection.rs
```diff
@@ -29,17 +32,23 @@ pub struct MysqlConnection(pub mysql::Pool);
...
 pub trait DBConnection {
-    fn run_query(&mut self, sql: &str, runtime: &Runtime) -> Vec<Row>;
+    fn run_query(&mut self, sql: &str, runtime: &Runtime) -> Result<Vec<Row>>;
 
     fn import_csv(&mut self, csv_name: &str, runtime: &Runtime);
 
     fn get_dialect(&self) -> Dialect;
+
+    // We sometimes want to modify the SQL `INSERT` query (we don't modify the
+    // SQL `SELECT` query)
```
max-sixty, line 42: 👍

max-sixty:
Another great PR @Jelenkee , thanks!

PR #2455, opened 2023-04-15
docs: Combine `precedence` syntax pages

max-sixty:
This retains the `()` line in the syntax index, but then combines the content in the `expressions-and-operators` page. Inspired by @richb-hanover's suggestions — I think this could be the best of both worlds?

@richb-hanover for your review!

PR #2460, opened 2023-04-16
docs: Another approach for combining Expressions & Parentheses

richb-hanover:
I, too, merged the "Expressions & Operators" page with the "Precedence and Parentheses" page. It's remarkable about how similar the result is to #2455.

I was never sold on the notion that "precedence" was the most important concept, so I re-wrote the rules to describe when parentheses were necessary.

There are many embedded questions that need to be addressed before this could be merged. I would appreciate your thoughts.

@max-sixty @aljazerzen @eitsupi


web/book/src/syntax/README.md
```diff
@@ -15,7 +15,6 @@ A summary of PRQL syntax:
 | `:`                              | [Named args & Parameters](../queries/functions.md)                      | `interp low:0 1600 sat_score`                           |
 | `[]`                             | [Lists](./lists.md)                                                     | `select [id, amount]`                                   |
 | <code>! && \|\| == +</code>, etc | [Expressions & Operators](./expressions-and-operators.md)               | <code>filter a == b + c \|\| d >= e</code>              |
-| `()`                             | [Precedence & Parentheses](./precedence-and-parentheses.md)             | `derive celsius = (fahrenheit - 32) / 1.8`              |
```
max-sixty, line 18: The thing we really need to keep here is an entry in the table for parentheses!

richb-hanover: Good point. I'll add that later today. What about the other questions in that page?

web/book/src/syntax/expressions-and-operators.md
```diff
@@ -15,21 +15,87 @@ filter circumference > 10 && color != "red"
 
 ## Operator precedence
 
-This table gives formal description of operator precedence. Because function
-calls have the lowest precedence, nested function calls or arguments that start
-or end with an operator require parentheses.
+This table shows operator precedence. Use parentheses `()` to prioritize
+operations and for function calls (see below.) _Question: Are parentheses
+operators? Should they be in this table?_
```
max-sixty: Yes, maybe they should be first?
@aljazerzen ?

web/book/src/syntax/expressions-and-operators.md
```diff
@@ -15,21 +15,87 @@ filter circumference > 10 && color != "red"
...
+
+  _Question: Why not require parentheses around function calls in a list? Why
+  preserve this (confusing) special case?_
+
+- Parentheses are not required for expressions that do not contain function
+  calls, for example: `foo + bar`.
+
+Here's a set of examples of these rules:
+
+```prql no-fmt
```
max-sixty, line 62: Can I confirm there are no changes to this section?

(It can be easier to split the PRs up so the file move is in a different PR to the text changes. But I recognize that can also add overhead to the process...)

web/book/src/syntax/expressions-and-operators.md
```diff
@@ -15,21 +15,87 @@ filter circumference > 10 && color != "red"
...
+|   _parens (?)_ | `()`              |     ??     |   see below   |
+
+## Parentheses
+
+PRQL uses parentheses `()` for several purposes:
+
+- Parentheses group operands to control the order of evaluation, for example:
+  `((1 + x) * y)`
+
+- Parentheses set off an [inner transform](./inner-transforms.md) for the
```
max-sixty: Is the "set off" term clear? I'm not familiar with it.

How about just "are required"?

web/book/src/syntax/expressions-and-operators.md
```diff
@@ -15,21 +15,87 @@ filter circumference > 10 && color != "red"
...
+  # Doesn't require parentheses, because it's in a list (confusing, see footnote)!
+  average_distance = average distance,
+]
+# Requires parentheses because of the minus sign
+sort (-distance)
+# A list is fine too
+sort [-distance]
+```
+
+```admonish note
```
max-sixty, line 101: 👍

web/book/src/syntax/precedence-and-parentheses.md
```diff
@@ -1,64 +0,0 @@
...
-  # Doesn't require parentheses, because it's in a list (confusing, see footnote)!
-  average_distance = average distance,
-]
-# Requires parentheses because of the minus sign
-sort (-distance)
-# A list is fine too
-sort [-distance]
-```
-
-This doesn't work, for example (though it should provide a much better error
```
max-sixty, line 58: I think this was helpful to include — otherwise someone can be confronted by the block of text and not really realize how it can affect them. Here's a very clear example of something you might expect to work and doesn't...

web/book/src/syntax/expressions-and-operators.md
aljazerzen: I'd say they are not, but it doesn't hurt to include them. And yes, they should be first, yes.

web/book/src/syntax/expressions-and-operators.md
```diff
@@ -15,21 +15,87 @@ filter circumference > 10 && color != "red"
...
+  main pipeline, for example: `(column-name | in 0..20)`
+
+- Parentheses wrap a function call that is part of a larger expression, for
+  example: `round 0 (sum distance)` _Question: Is it necessary to state "on the
+  right-hand side of an assignment"? Are there counterexamples?_
+
+  _Question: Why not require parentheses around function calls in a list? Why
+  preserve this (confusing) special case?_
+
+- Parentheses are not required for expressions that do not contain function
```
max-sixty, line 57: 👍

web/book/src/syntax/expressions-and-operators.md
```diff
@@ -15,21 +15,87 @@ filter circumference > 10 && color != "red"
...
+
+- Parentheses set off nested function calls that contain a pipe, either the `|`
+  symbol or a new line. “Nested” means within a transform; i.e. not just the
+  main pipeline, for example: `(column-name | in 0..20)`
+
+- Parentheses wrap a function call that is part of a larger expression, for
+  example: `round 0 (sum distance)` _Question: Is it necessary to state "on the
+  right-hand side of an assignment"? Are there counterexamples?_
+
+  _Question: Why not require parentheses around function calls in a list? Why
```
max-sixty: I def see that perspective. But it would add some syntactic noise to:

```elm
aggregate [
  max salary, # would be invalid
  min salary, # would be invalid
  count,
]
```

I'd not set either way, though not in a rush to change from the current state

web/book/src/syntax/expressions-and-operators.md
```diff
@@ -15,21 +15,87 @@ filter circumference > 10 && color != "red"
...
+
+- Parentheses set off a minus sign for a function argument, for example:
+  `add (-1) (-3)`
+
+- Parentheses set off nested function calls that contain a pipe, either the `|`
+  symbol or a new line. “Nested” means within a transform; i.e. not just the
+  main pipeline, for example: `(column-name | in 0..20)`
+
+- Parentheses wrap a function call that is part of a larger expression, for
+  example: `round 0 (sum distance)` _Question: Is it necessary to state "on the
```
max-sixty: IIUC this is the same point as https://github.com/PRQL/prql/pull/2460/files#r1168000249.

I had this as a footnote so that it was complete. Having the semantics fully listed somewhere means there is some canonical source of truth, which is important if we want to change it. But I agree it's not really relevant for most users — we could still have it in the footnote, or link to a fuller description in the code / in some tests...

max-sixty:
Overall great! Thanks a lot for really caring about this @richb-hanover . I left some comments; I think we can do the easy bits and then merge?

web/book/src/syntax/expressions-and-operators.md
richb-hanover: `()` is now listed at the top. Precedence is listed as "0" - I'd look for guidance here about whether to set its priority to zero, or to set it to 1 and increase all the others.

web/book/src/syntax/expressions-and-operators.md
richb-hanover: The only change to the example is that I changed "it's contains a pipe" to "it contains a pipe"

(Someday I need to find how to regenerate the snapshots of SQL. I currently use `task test-rust`but that sometimes takes 15 minutes in the Dev Container if crates need to be compiled again.)

web/book/src/syntax/expressions-and-operators.md
richb-hanover: If you're asking about it, I guess "set off" is not clear :-) And I don't mean like a firecracker...

I've changed it to "denote" to see if that reads better.

web/book/src/syntax/precedence-and-parentheses.md
richb-hanover: Ahah! I just figured out what that sentence means. I've rewritten it...

web/book/src/syntax/expressions-and-operators.md
richb-hanover: OK. There _are_ counterexamples. I'll include the "on the right hand side..."

web/book/src/syntax/expressions-and-operators.md
richb-hanover: see the change that I just committed re: "on the right-hand side..."

richb-hanover:
I have edited the page to address the questions. I think it can be merged as-is.

Outstanding Questions:

* What precedence should `()` be assigned in the table? (I set it to zero)
* Are there any **other** transforms besides `group()` and `window()` that use parentheses?
* Might it make sense to require parentheses around all function calls in a list? Why preserve this (confusing) special case?

Thanks.

web/book/src/syntax/expressions-and-operators.md
```diff
@@ -94,8 +88,17 @@ sort (-distance)
...
+from employees
+derive total_distance = sum distance # generates the error shown below
+derive other_distance = (sum distance) # works as expected
+```
+
 ```admonish note
 We are seeking feedback about these rules to see if they can be more intuitive.
 We're also planning to make the error messages much better,
-so the compiler is there to help out.
+so the compiler can help out.
```
max-sixty, line 104: 👍

web/book/src/syntax/expressions-and-operators.md
```diff
@@ -42,23 +41,18 @@ PRQL uses parentheses `()` for several purposes:
...
   `((1 + x) * y)`
 
-- Parentheses set off an [inner transform](./inner-transforms.md) for the
-  `group ()` and `window ()` transforms. _Question: Are there any other
-  transforms that use parentheses like this?_
+- Parentheses denote an [inner transform](./inner-transforms.md) for the
+  `group ()` and `window ()` transforms. 
 
-- Parentheses set off a minus sign for a function argument, for example:
+- Parentheses denote a minus sign for a function argument, for example:
```
max-sixty: `denote` defn: "be a sign of; indicate."

I don't think that's quite right....

"delimit"? Kinda formal / clunky but the meaning is closer...

web/book/src/syntax/expressions-and-operators.md
```diff
@@ -15,21 +15,91 @@ filter circumference > 10 && color != "red"
...
+|  function call |                   |     10     |               |
+
+## Parentheses
+
+PRQL uses parentheses `()` for several purposes:
+
+- Parentheses group operands to control the order of evaluation, for example:
+  `((1 + x) * y)`
+
+- Parentheses denote an [inner transform](./inner-transforms.md) for the
```
max-sixty: In a separate PR, we could merge the `inner transform` page into this, it's a bit sparse and lonely on its own...

web/book/src/syntax/expressions-and-operators.md
```diff
@@ -15,21 +15,91 @@ filter circumference > 10 && color != "red"
...
+  `group ()` and `window ()` transforms.
+
+- Parentheses denote a minus sign for a function argument, for example:
```
max-sixty:
```suggestion
- Parentheses denote a minus sign of a function argument, for example:
```
of / for?

web/book/src/syntax/expressions-and-operators.md
```diff
@@ -15,21 +15,91 @@ filter circumference > 10 && color != "red"
...
+sort [-distance]
+```
+
+Note: The `total_distance` statement below generates an error because the
+function is not in a list. (The PRQL compiler should display a better error
+message.)
+
+```prql error
+from employees
+derive total_distance = sum distance # generates the error shown below
```
max-sixty, line 97: 👍

web/book/src/syntax/expressions-and-operators.md
richb-hanover: Yeah, clunky and formal. But let's go with it.

web/book/src/syntax/expressions-and-operators.md
richb-hanover: I wondered about merging it in. I guess `group` and `window` can link to the sub-page.

web/book/src/syntax/expressions-and-operators.md
richb-hanover: of

web/book/src/syntax/expressions-and-operators.md
```diff
@@ -15,21 +15,91 @@ filter circumference > 10 && color != "red"
...
-| mul            | `* / %`           | 4          | left-to-right |
-| add            | `+ -`             | 5          | left-to-right |
-| compare        | `== != <= >= < >` | 6          | left-to-right |
-| coalesce       | `??`              | 7          | left-to-right |
-| and            | `&&`              | 8          | left-to-right |
-| or             | <code>\|\|</code> | 9          | left-to-right |
-| function call  |                   | 10         |               |
+|          Group | Operators         | Precedence | Associativity |
+| -------------: | ----------------- | :--------: | :-----------: |
+|    parentheses | `()`              |     0      |   see below   |
```
max-sixty, line 25: @aljazerzen for your post-merge review. Not really an operator but I thought it did belong in the precedence table, let us know if not

web/book/src/syntax/expressions-and-operators.md
```diff
@@ -15,21 +15,91 @@ filter circumference > 10 && color != "red"
...
+|            add | `+ -`             |     5      | left-to-right |
+|        compare | `== != <= >= < >` |     6      | left-to-right |
+|       coalesce | `??`              |     7      | left-to-right |
+|            and | `&&`              |     8      | left-to-right |
+|             or | <code>\|\|</code> |     9      | left-to-right |
+|  function call |                   |     10     |               |
+
+## Parentheses
+
+PRQL uses parentheses `()` for several purposes:
```
max-sixty, line 39: I think this reads quite well now! Thanks @richb-hanover !

web/book/src/syntax/expressions-and-operators.md
```diff
@@ -15,21 +15,91 @@ filter circumference > 10 && color != "red"
...
+message.)
+
+```prql error
+from employees
+derive total_distance = sum distance # generates the error shown below
+derive other_distance = (sum distance) # works as expected
+```
+
+```admonish note
+We are seeking feedback about these rules to see if they can be more intuitive.
```
max-sixty:
```suggestion
We're continuing to think whether these rules can be more intuitive.
```

(I actually think this is one place where we're not really "seeking feedback" (one of the **only** places) — we know it's over-complicated but need to think how it all fits together. We're seeking genius ideas if someone has them, but it would require lots of context rather than just "this part is awkward")

richb-hanover: Agreed.

web/book/src/syntax/expressions-and-operators.md
aljazerzen: I replied about that somewhere - it's fine even if they are technically not operators. , understandable enough.

PR #2462, opened 2023-04-16
docs: Update examples to use fewer s-strings

max-sixty:
I couldn't get the first few to use s-strings, will open an issue


prql-compiler/tests/integration/snapshots/integration__invoice_totals.prql@invoice_totals.prql.snap
```diff
@@ -20,21 +20,21 @@ WITH table_1 AS (
 SELECT
   month,
   day,
-  CAST(_expr_2 AS INT) AS num_orders,
-  CAST(_expr_1 AS INT) AS num_tracks,
-  ROUND(CAST(_expr_0 AS FLOAT) * 100) / 100 AS total_price,
```
max-sixty: We get a postgres error on this change. I think probably we wait to resolve https://github.com/PRQL/prql/pull/2449/files#r1167783654 before addressing

PR #2463, opened 2023-04-16
test: Force amd64 image for mssql image

max-sixty:
No arm64 available, so this fails on arm64 machines.

Also add a comments re the tags


max-sixty:
(CC @Jelenkee — lmk any post-merge feedback for these)

PR #2473, opened 2023-04-20
docs: impl strategy for type reprs

aljazerzen:


web/book/src/internals/type-system.md
```diff
@@ -308,25 +311,105 @@ Language and does not have to bother with exact physical layout of types.
 
 ### Non-bijection cases between PTS and STS
 
-There are cases where a PTS construct has multiple possible and valid
-representations in some STSs.
+### Type representations
```
max-sixty, line 314: nit: the heading above has no content now

web/book/src/internals/type-system.md
```diff
@@ -308,25 +311,105 @@ Language and does not have to bother with exact physical layout of types.
...
+    created_at: timestamp,
+}
+```
+
+A similar example is an "array of strings type" in PTS that could be represented
+by a `text[]` (if DBMS supports arrays) or `json` or it's variant `jsonb` in
+Postgres. Again, the representation would affect operators: in Postgres, arrays
+would be accessed with `my_array[1]` and json arrays would use
+`my_json_array -> 1`. This example may not be applicable, if we decide that we
+want a separate JSON type in PST.
```
max-sixty, line 347: This is a v interesting idea, and I could imagine it being helpful. It's almost like a third level of type definition — Physical, Logical DB, Logical PRQL.

I find the JSON example good, and could imagine extending this to types that represent the real world, such as `distance` * `distance` = `area`, but `distance` / `time` = `speed`.

That said, it does also seem like a quite advanced feature. To the extent there's some way of signposting this as a "future idea" even beyond the appendix, I think that could help distinguish between "this is will help you understand the PRQL type system" vs. "these are some future ideas".

web/book/src/internals/type-system.md
```diff
@@ -308,25 +311,105 @@ Language and does not have to bother with exact physical layout of types.
...
+(i.e. `std.int8.add`). These functions have typed parameters and a return value.
+If an RQ function call does not match the function declaration in number or in
+types of the parameters, this is considered an invalid RQ AST.
+
+We provide multiple implementations for each RQ function. They are annotated
+with a target (i.e. `#[target(sql.sqlite)]`) and have their params annotated
+with type reprs (i.e. `#[repr(int)]`).
+
+```
+# using a made-up syntax
```
max-sixty, line 378: Good signpost!

web/book/src/internals/type-system.md
```diff
@@ -308,25 +311,105 @@ Language and does not have to bother with exact physical layout of types.
...
+
+#[target(sql.sqlite)]
+func std.int8.add
+    #[repr(int8)] x
+    #[repr(int8)] y
+    -> s"{x} + {y}"
```
max-sixty, line 384: Is there an example of when this would be useful? Maybe `in` for ranges vs. lists?

The ints example below is maybe OK, but not that tethered to what's likely to be helpful for PRQL. No objection to including for the moment, but future versions might want to more closely connect the abstract ideas to something concrete

max-sixty (approved):
Looks good!

web/book/src/internals/type-system.md
aljazerzen: Again, I forgot the include my "big picture" thinking.

This is mostly a framework for dealing with dialects. For example timestamps in SQLite, regex functions and concat are only the first examples. I've started noticing that when I tried moving more tests into the integration tests, either SQLite or DuckDB complained about something.

This framework will probably not be used in it's entirety, but it's needed as a north star of the development on the backend. Initially these were just my notes to clear my head, but then I realized they would be useful to anyone dealing with the compiler.

web/book/src/internals/type-system.md
aljazerzen: SQLite's temporal types.

Arrays (which have many different flavors in different dialects)

Issue #2477, opened 2023-04-20
Docs: How to understand/describe the stdlib notation?


richb-hanover:
### What's up?

 In #2057, @max-sixty used some lingo about function signatures (which makes sense). I see the same kind of notation on the [stdlib](https://prql-lang.org/book/language-features/standard-library/index.html) page, but I don't really understand what the descriptions mean.

I kinda can read them: `func min <scalar || column> column -> null` means that the `min` function takes either a scalar or a column name. But then... what? It returns a column? And what's the skinny-arrow to "null"?

My request: to include a secret keyring decoder that describes this notation the PRQL std functions page. Thanks.

aljazerzen:
Ughh, this a good indication that our syntax for function definition is not obvious.

`<scalar || column> column` is saying "parameter column of type `scalar || column`"
`->` means "maps to"
`null` is the function body, because we don't have a syntax that would say "internal compiler code"

richb-hanover:
This is not a serious problem, just a documentation issue... (I am delighted that the Book includes all the `std` functions, straight from the source code. When a new std function arrives, it automatically gets documented.) We just need to include a description of how to read this notation. 

I'm happy to write something up, but need a Cliff Notes to get me started. NB: I wouldn't be this verbose for each function definition - just selected examples to explain each different bit of syntax.

* From what you wrote above, I think I would write this explanation. 
   > `func min <scalar || column> column -> null`
   > The `min` function takes a parameter that can be either a `scalar` or `column name` and returns a value of another `column name` (the second instance of the word `column`). The `-> null` means that `min` calls internal compiler code to perform this.

   Is this correct?

There are other bits of syntax that I don't understand:

* In `func first<column> offset column -> null`,
   * Why are `first` and `<column>` not separated by a space? What does that signify?
   *  I think I'd write:
      > The `first` function takes a parameter of a column and an (integer) offset and calls internal compiler code to return a column 

   Is this correct?
* In `func round<scalar> n_digits column -> null`
   * Same question about space between `round` and `<scalar>`
   * I think I'd write:
      > The `round` function takes a scalar parameter and calls an internal function to return a column.

   Is this correct?
* In ``func from<table> `default_db.source`<table> -> null`` 
   * Same question about space between `from` and `<table>`
   * I'm not sure about the rest of the syntax
* In this definition:
   ```
   func intersect<table> `default_db.bottom`<table> top<table> -> (
    from t = _param.top
[...]

PR #2480, opened 2023-04-23
docs: Update std.prql to include explanation of notation

richb-hanover:
Adding comment block to head of std.prql to address #2477


max-sixty:
Thanks @richb-hanover !

PR #2481, opened 2023-04-24
fix(prql-compiler): allow `SELECT\n` in s-strings representing a table

### LINKED ISSUE #2430: Problem with parsing s-string in `from`?
### What happened?

```elm
let _tab0 = s"""
  SELECT foo
  FROM bar
"""

from _tab0
```

The above PRQL is processed successfully, but the following PRQL with added newlines seems to cause an error.

### PRQL input

```elm
let _tab0 = s"""
  SELECT
    foo
  FROM bar
"""

from _tab0
```

### SQL output

```SQL
None
```

### Expected SQL output

```SQL
WITH table_0 AS (
  SELECT
    foo
  FROM
    bar
),
_tab0 AS (
  SELECT
    *
  FROM
    table_0 AS table_1
)
SELECT
  *
FROM
  _tab0
```

### MVCE confirmation

- [x] Minimal example
- [x] New issue

### Anything else?

_No response_

eitsupi:
Fix #2430

### DIFF
--- a/prql-compiler/src/sql/gen_expr.rs
+++ b/prql-compiler/src/sql/gen_expr.rs
@@ -400,13 +400,14 @@ pub(super) fn translate_query_sstring(
 ) -> Result<sql_ast::Query> {
     let string = translate_sstring(items, context)?;
 
+    let re = Regex::new(r"(?i)^SELECT(\s|\n|\r)").unwrap();
     let prefix = if let Some(string) = string.trim().get(0..7) {
         string
     } else {
         ""
     };
 
-    if prefix.eq_ignore_ascii_case("SELECT ") {
+    if re.is_match(prefix) {
         if let Some(string) = string.trim().strip_prefix(prefix) {
             return Ok(sql_ast::Query {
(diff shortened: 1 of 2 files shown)


prql-compiler/src/sql/gen_expr.rs
```diff
@@ -400,13 +400,14 @@ pub(super) fn translate_query_sstring(
 ) -> Result<sql_ast::Query> {
     let string = translate_sstring(items, context)?;
 
+    let re = Regex::new(r"(?i)^SELECT(\s|\n|\r)").unwrap();
```
max-sixty, line 403: FYI this can be `r"(?i)^SELECT\b"`

max-sixty:
Thanks!

prql-compiler/src/sql/gen_expr.rs
eitsupi: Thanks for pointing that, I didn't know `\b`...
I will open another PR.

PR #2474, opened 2023-04-20
docs: Tweak FAQ

richb-hanover:
Update What can you do with PRQL to say... "Prototype PRQL queries...";
Update Press Materials link to point to PRQL/prql-brand;
Fix "What's with the 'aggregate'..." heading

PR #2482, opened 2023-04-24
docs: Update link to #example section

richb-hanover:
@aljazerzen I see that this PR is failing because of two sets of bad links:

- I think "#example" in this file should be lower-case
- The former Press Materials referred to the logos directly which have been renamed for consistency (https://github.com/PRQL/prql-brand/pull/12). There's another PR for the FAQ page (#2474) that removes the Press Materials page entirely, in favor keeping all this information in PRQL/prql-brand repo. Merging that would probably allow this one to go forward. Thanks


richb-hanover:
The 503 error on the php-prql link (https://www.php.net/manual/en/book.ffi.php) appears to be temporary. It should work next time...

max-sixty:
Thanks a lot @richb-hanover !

PR #2488, opened 2023-04-25
feat(prqlc): add `get-targets` command and `--traget` option of `compile` command

eitsupi:


Taskfile.yml
```diff
@@ -205,6 +205,8 @@ tasks:
     #   - "**/*.lock"
     #   - "**/*.snap"
     cmds:
+      # Build prqlc first for assert_cmd tests
+      - cargo build -p prqlc
```
eitsupi: I need to build prqlc before testing also in the GitHub Action workflow to use `assert_cmd`...

aljazerzen (approved):
Great work, I cannot believe we didn't have this before... :D

prql-compiler/prqlc/src/cli.rs
```diff
@@ -477,4 +502,19 @@ group a_column (take 10 | sort b_column | derive [the_number = rank, last = lag
...
+            .unwrap()
+            .args(["get-targets"])
+            .assert()
+            .success()
+            .stdout(
+                predicates::str::is_match(r"sql\.[a-z]+\n")
+                    .unwrap()
+                    .count(n_targets),
+            );
+    }
```
aljazerzen: This is failing because it requires a --release binary of prqlc ... which is called from a test in a --test prqlc binary.

We could probably make this work, but I'm not sure it's worth the headache.

aljazerzen:
We need to make sure that when someone clones the repo and runs only:

```
cargo test
```

... the thing works.

eitsupi:
> We need to make sure that when someone clones the repo and runs only:
> 
> ```
> cargo test
> ```
> 
> ... the thing works.

I agree.

For CLI testing, I will open a follow-up issue, so I will remove the test in this PR.

max-sixty:
Thanks — this looks good!

FWIW I think `--list-targets` would be more idiomatic, but not the biggest issue obv

eitsupi:
> FWIW I think `--list-targets` would be more idiomatic, but not the biggest issue obv

Indeed.

However, since this is a reference implementation, I feel it is worthwhile to have the same function name `get-targets` as prql-js and prql-python.

max-sixty:
> However, since this is a reference implementation, I feel it is worthwhile to have the same function name `get-targets` as prql-js and prql-python.

Yeah, I think that's fine for the function name. My sense is that `list` is an idiom in CLIs. And we don't necessarily need the same name between the CLI and the function (though they could both be `list` if folks prefer...)

But I'm not that confident — I'm only going on my own sense, we shouldn't load on it unless others share it.

PR #2490, opened 2023-04-26
build: fix markdown link checker

aljazerzen:


.github/.markdown-link-check.json
```diff
@@ -9,10 +9,9 @@
   ],
   "ignorePatterns": [
     {
-      "pattern": "^https://twitter.com"
-    },
-    {
-      "pattern": "^https://www.php.net"
+      "_comment_1": "Our pipeline should depend only on the work we do.",
```
max-sixty, line 12: Clever way of doing comments!

max-sixty:
This is a difficult tradeoff — blocking unrelated PRs is very annoying. But we need some way to check broken links — it's a poor signal of maintenance if links are broken for a while. Do we have any way of measuring that with the current code?

Previously we only ran checks on *changed* files for PRs, and then ran over everything every night, and I would occasionally check it and fix any breaks.

Then we had a bunch of breaks when someone moved a file, and the links **to** that file weren't checked in the PR, since they weren't part of the diff. So I turned on all files for `build-web`, which runs on changes to `web`. But then I was out of action for a while, and so this stayed broken.

3 options I can see
- Find some way to check whether links are broken on some regular cadence with this config file (every week would be fine, doesn't need ms latency...)
- Turn it back on for `build-web`, see whether that causes more false-positives
- Revert to the old mode, where we only check diffs and someone (probably me) would check it occasionally. 
  - I might be more tempted to revert PRs which break links rather than spend the 20 mins exploring & fixing
- Don't check all links, rely on people finding them out manually

I would probably vote for the reverting to the old mode. I would be -1 on just not ever checking links

aljazerzen:
To be clear: we are currently still checking all links pointing to prql-lang.org/**, github.com/PRQL/** or raw\\.githubusercontent\\.com/PRQL/**. This covers basically everything one can change with a PRQL PR, so the tests check that nothing was broken.

The difference is that with this PR, pipelines won't break when php.net starts returning 500 all of the sudden.

That would be your 4th option.

max-sixty:
> To be clear: we are currently still checking all links pointing to prql-lang.org/**, github.com/PRQL/** or raw\.githubusercontent\.com/PRQL/**. This covers basically everything one can change with a PRQL PR, so the tests check that nothing was broken.

Yes good point, I got that but didn't fully internalize its impact, you're right that covers anything we could change in a PR.

BUT if the php.net link moves (which has happened in the past — e.g. dbt changed their docs and we got a bunch of 404s), how would we catch that?

One 5th option: Have a different (empty?) config file which runs on nightly without these constraints — WDYT?

aljazerzen:
That should we a warning then. A nightly action would accomplish that, yes.

PR #2494, opened 2023-04-26
chore: fixate minijinja version

aljazerzen:


max-sixty:
Thanks. Possibly we can remove this, but possibly we'll find something useful for it too

PR #2499, opened 2023-04-27
test: Add prqlc test

### LINKED ISSUE #177: Test the CLI

max-sixty:
~This is code from #2488. It passes locally, I wanted to see whether it passed in CI too.~

Two points:
- I _think_ the key is to add this as an integration test, not a unit test. The existing method works
- While CI tests were running I changed to use `insta-cmd`'s `get_cargo_bin` function. But actually I think it works without that — the original `assert_cmd` tests passed (and regardless we could use it with that.

Would close #177

What do folks think? @eitsupi ? Would you prefer the `assert_cmd`? It does have more following. But the snapshots do seem like good tests?


eitsupi (approved):
Looks great! Thanks!

PR #2532, opened 2023-05-03
refactor: `unpack` & `try_unpack`

max-sixty:
As I mentioned on Discord. Inspired by [this discussion](https://github.com/PRQL/prql/pull/2458#discussion_r1181413010)
- This is me trying to become more acquainted with the code
- We don't have to merge it (genuinely)
- Though I do find this code simpler to understand
- It's less impressive rust, maybe less abstract, less use of the type system
- It's the same verbosity, as measured by lines of code
- It does more allocating, more `.clone()`
- But it's actually slightly faster on `cargo bench`, about 3% on my machine
  - It doesn't run each `try_into_*` before doing the next one if that fails
  - Maybe it allows for fewer eventual optimizations in the future though?
- It would need a quick pass to add the `regex` from today

GPT-4 helped me a lot!

---
- I added a couple of notes on further changes. We're probably doing too much `if args.len() == X` in lots of places, and could have these as structs & enums in rust, though I'm not sure how deep in the compiler we want them (maybe there's a difference between functions & operators?)


prql-compiler/src/sql/gen_expr.rs
```diff
@@ -76,132 +76,230 @@ pub(super) fn translate_expr(expr: Expr, ctx: &mut Context) -> Result<sql_ast::E
...
-        LtEq,
-        And,
-        Or,
-        StringConcat,
-    ];
-
-    let Some((decl, _)) = try_unpack(&expr, DECLS)? else {
-        return Ok(Err(expr));
+/// Translates into IS NULL if possible
+fn process_null(name: &str, args: &[Expr], ctx: &mut Context) -> Result<sql_ast::Expr> {
```
aljazerzen, line 126:
```suggestion
fn process_is_null(name: &str, args: &[Expr], ctx: &mut Context) -> Result<sql_ast::Expr> {
```

prql-compiler/src/sql/gen_expr.rs
```diff
@@ -76,132 +76,230 @@ pub(super) fn translate_expr(expr: Expr, ctx: &mut Context) -> Result<sql_ast::E
...
+                    if let Some(op) = operator_from_name(name) {
+                        if args.len() == 2 {
+                            let (left, right) = (&args[0], &args[1]);
+                            translate_binary_operator(left, right, op, ctx)?
+                        } else {
+                            super::std::translate_built_in(expr, ctx)?
+                        }
+                    } else {
+                        super::std::translate_built_in(expr, ctx)?
+                    }
```
aljazerzen: This whole part is being repeated. Could it be converted into a fall-through case of the match?

```
ExprKind::BuiltInFunction { ref name, ref args } => {

    // special cases
    match name.as_str() {
        "std.eq" | "std.ne" => {
            if ... {
                return process_null(...);
            }
        },
        "std.neg" | "std.not" => {
            if args.len() == 1 {
                return process_unary(name, &args[0], ctx)?
            }
        }
        _ => {}
    }
    
    ... the fall-trough case ...
}
```

prql-compiler/src/sql/gen_expr.rs
```diff
@@ -76,132 +76,230 @@ pub(super) fn translate_expr(expr: Expr, ctx: &mut Context) -> Result<sql_ast::E
...
+        "std.sub" => Some(Minus),
+        "std.eq" => Some(Eq),
+        "std.ne" => Some(NotEq),
+        "std.gt" => Some(Gt),
+        "std.lt" => Some(Lt),
+        "std.gte" => Some(GtEq),
+        "std.lte" => Some(LtEq),
+        "std.and" => Some(And),
+        "std.or" => Some(Or),
+        "std.concat" => Some(StringConcat),
```
aljazerzen, line 322: Ideally, this would use `std::STD_X` so we have it "linked" with RQ.

aljazerzen (approved):
This is way better, as in way more readable.

I went too deep with the no-clone frenzy :D It's actually slower probably because it was searching for impl of functions in linear search. This PR is now using sub-logarithmic lookups (because it's using `match`).

max-sixty:
> This is way better, as in way more readable.
> 
> I went too deep with the no-clone frenzy :D It's actually slower probably because it was searching for impl of functions in linear search. This PR is now using sub-logarithmic lookups (because it's using `match`).

OK great! Thanks for being open. I was holding that it was "worse but easier rust" — maybe it's actually / actually much worse. I do think this sort of change can make it easier for others to contribute...

I'll update to yesterday's regex addition, resolve hte comments, and leave a couple of notes re strings vs `FunctionDecl`s (i.e. https://github.com/PRQL/prql/pull/2532#discussion_r1183937392)

prql-compiler/src/sql/gen_expr.rs
max-sixty: Good idea; done

prql-compiler/src/sql/gen_expr.rs
max-sixty: Yes, I left some notes [here](https://github.com/PRQL/prql/pull/2532/files#diff-2540aa8b42227cb2ab1983dcc56db5f2401778a2ba89fb1c33525c31c59d7e6bR82-R87), I agree the current state isn't ideal. I'm not sure how much we want to have these deep in the compiler with something like an Enum...

max-sixty:
...and using slice patterns reduces some of the `if args.len...` verbosity: https://github.com/PRQL/prql/pull/2532/commits/e6fc1f232ab572edce075d7b705b5dd6af9beee1

max-sixty:
I'll merge given your ✅ , but happy to do follow-up with any other changes

PR #2535, opened 2023-05-04
feat: Improve regex search operator error message

### LINKED ISSUE #2534: Improve location of displayed error for Regex
### What happened?

Using the new regex facility, this code gives a too-subtle error in the Playground:

```
prql target:sql.sqlite
from foo
filter bar ~= 'love'
```
The error indication is a red squiggly line under the word `prql`, which isn't very obvious.

The text of the error message itself - `"regex functions are not supported by this dialect (or PRQL doesn't yet implement this dialect)"` is fine.

It would be better to place that error message in red (below the query), shown in the way as if the final line was changed to the (syntactically incorrect) `filter bar ~= 'love'x` 

Even better would be to highlight the `~=` as the source of the error.

(PS Do we have a name for that red area that displays error messages?)

### PRQL input

```elm
See above
```

### SQL output

```SQL
N/A
```

### Expected SQL output

```SQL
N/A
```

### MVCE confirmation

- [X] Minimal example
- [X] New issue

### Anything else?

_No response_

max-sixty:
Closes #2534 

Following up with a question about passing the errors through

### DIFF
--- a/prql-compiler/src/sql/gen_expr.rs
+++ b/prql-compiler/src/sql/gen_expr.rs
@@ -103,7 +103,15 @@ pub(super) fn translate_expr(expr: Expr, ctx: &mut Context) -> Result<sql_ast::E
                 "std.concat" => return process_concat(&expr, ctx),
                 "std.regex_search" => {
                     if let [search, target] = args.as_slice() {
-                        return process_regex(search, target, ctx);
+                        return process_regex(search, target, ctx).map_err(|e| {
+                            // Add the span of the expression to the error. But
+                            // we pass back an `anyhow::Error`, so we need to
+                            // try downcasting it first. If that fails, we
+                            // return the original error.
+                            e.downcast_ref::<crate::Error>()
+                                .map(|e| e.clone().with_span(expr.span).into())
+                                .unwrap_or(e)
+                        });
                     }
                 }
                 _ => match try_into_between(expr.clone(), ctx)? {
@@ -222,10 +230,12 @@ fn process_regex(search: &Expr, target: &Expr, ctx: &mut Context) -> Result<sql_
         // TODO: name the dialect, but not immediately obvious how to actually
         // get the dialect string from a `DialectHandler`.
         //
-        // MSSQL doesn't support them, MySQL & SQLite have a different construction.
-        bail!("regex functions are not supported by this dialect (or PRQL doesn't yet implement this dialect)");
+        // MSSQL doesn't support them, MySQL & SQLite have a different
+        // construction.
+        return Err(Error::new(
+            crate::Reason::Simple("regex functions are not supported by this dialect (or PRQL doesn't yet implement this dialect)".to_string())
+        ).into())
     };
-
     let args = [search, target]
         .into_iter()
         .map(|a| {
--- a/prql-compiler/src/sql/gen_expr.rs
+++ b/prql-compiler/src/sql/gen_expr.rs
@@ -103,7 +103,7 @@ pub(super) fn translate_expr(expr: Expr, ctx: &mut Context) -> Result<sql_ast::E
                 "std.concat" => return process_concat(&expr, ctx),
                 "std.regex_search" => {
                     if let [search, target] = args.as_slice() {
-                        return dbg!(process_regex(search, target, ctx)).with_span(expr.span);
+                        return process_regex(search, target, ctx).with_span(expr.span);
                     }
                 }
                 _ => match try_into_between(expr.clone(), ctx)? {
--- a/prql-compiler/src/sql/gen_expr.rs
+++ b/prql-compiler/src/sql/gen_expr.rs
@@ -103,15 +103,7 @@ pub(super) fn translate_expr(expr: Expr, ctx: &mut Context) -> Result<sql_ast::E
                 "std.concat" => return process_concat(&expr, ctx),
                 "std.regex_search" => {
                     if let [search, target] = args.as_slice() {
-                        return process_regex(search, target, ctx).map_err(|e| {
-                            // Add the span of the expression to the error. But
-                            // we pass back an `anyhow::Error`, so we need to
-                            // try downcasting it first. If that fails, we
-                            // return the original error.
-                            e.downcast_ref::<crate::Error>()
-                                .map(|e| e.clone().with_span(expr.span).into())
-                                .unwrap_or(e)
-                        });
+                        return dbg!(process_regex(search, target, ctx)).with_span(expr.span);
                     }
                 }
                 _ => match try_into_between(expr.clone(), ctx)? {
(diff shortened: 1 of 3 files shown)


prql-compiler/src/sql/gen_expr.rs
```diff
@@ -103,7 +103,7 @@ pub(super) fn translate_expr(expr: Expr, ctx: &mut Context) -> Result<sql_ast::E
                 "std.concat" => return process_concat(&expr, ctx),
                 "std.regex_search" => {
                     if let [search, target] = args.as_slice() {
-                        return process_regex(search, target, ctx);
+                        return process_regex(search, target, ctx, expr.span);
```
max-sixty: Ideally we would not pass the span down — we'd raise an error in `process_regex` and then attach the span here.

But I don't think that's possible with our error types?

Here's what I was attempting:

```diff
diff --git a/prql-compiler/src/sql/gen_expr.rs b/prql-compiler/src/sql/gen_expr.rs
index 10626a49..86a2d246 100644
--- a/prql-compiler/src/sql/gen_expr.rs
+++ b/prql-compiler/src/sql/gen_expr.rs
@@ -103,7 +103,8 @@ pub(super) fn translate_expr(expr: Expr, ctx: &mut Context) -> Result<sql_ast::E
                 "std.concat" => return process_concat(&expr, ctx),
                 "std.regex_search" => {
                     if let [search, target] = args.as_slice() {
-                        return process_regex(search, target, ctx, expr.span);
+                        return process_regex(search, target, ctx)
+                            .map_err(|e| e.with_span(expr.span));
                     }
                 }
                 _ => match try_into_between(expr.clone(), ctx)? {
@@ -221,8 +222,7 @@ fn process_regex(
     search: &Expr,
     target: &Expr,
     ctx: &mut Context,
-    span: Option<Span>,
-) -> Result<sql_ast::Expr> {
+) -> Result<sql_ast::Expr, crate::Error> {
     let Some(regex_function) = ctx.dialect.regex_function() else {
         // TODO: name the dialect, but not immediately obvious how to actually
         // get the dialect string from a `DialectHandler`.
@@ -230,8 +230,8 @@ fn process_regex(
         // MSSQL doesn't support them, MySQL & SQLite have a different
         // construction.
         return Err(Error::new(
-            crate::Reason::Simple("regex functions are not supported by this dialect (or PRQL doesn't yet implement this dialect)".to_string())).with_span(span)
-        .into())
+            crate::Reason::Simple("regex functions are not supported by this dialect (or PRQL doesn't yet implement this dialect)".to_string())
+        ).into())
     };
     let args = [search, target]
         .into_iter()
```

But it requires returning our error type from `process_regex`, and I'm not sure we can convert the `anyhow` error type that a different part of the function returns, so we get:

```
error[E0308]: mismatched types
   --> prql-compiler/src/sql/gen_expr.rs:106:32
    |
27  |   pub(super) fn translate_expr(expr: Expr, ctx: &mut Context) -> Result<sql_ast::Expr> {
    |                                                                  --------------------- expected `std::result::Result<sqlparser::ast::Expr, anyhow::Error>` because of return type
...
106 |                           return process_regex(search, target, ctx)
    |  ________________________________^
107 | |                             .map_err(|e| e.with_span(expr.span));
    | |________________________________________________________________^ expected struct `anyhow::Error`, found struct `error::Error`
    |
    = note: expected enum `std::result::Result<_, anyhow::Error>`
               found enum `std::result::Result<_, error::Error>`

error[E0277]: `?` couldn't convert the error to `error::Error`
   --> prql-compiler/src/sql/gen_expr.rs:243:23
    |
225 | ) -> Result<sql_ast::Expr, crate::Error> {
    |      ----------------------------------- expected `error::Error` because of this
...
243 |         .try_collect()?;
    |                       ^ the trait `std::convert::From<anyhow::Error>` is not implemented for `error::Error`
    |
    = note: the question mark operation (`?`) implicitly performs a conversion on the error value using the `From` trait
    = help: the following other types implement trait `FromResidual<R>`:
              <std::result::Result<T, F> as FromResidual<Yeet<E>>>
              <std::result::Result<T, F> as FromResidual<std::result::Result<Infallible, E>>>
    = note: required for `std::result::Result<sqlparser::ast::Expr, error::Error>` to implement `FromResidual<std::result::Result<Infallible, anyhow::Error>>`

Some errors have detailed explanations: E0277, E0308.
```

CC @aljazerzen if you have thoughts (also as part of lowering the skill floor of these sorts of contributions)

aljazerzen: Attaching span into the Err in Result is a sound strategy. As you say, it's a little tricky because you cannot use `anyhow::Error` on the function, but you could have a helper function to "enrich anyhow::Error":

```
fn enrich_error(anyhow::Error, span: Option<Span>) -> anyhow::Error {
    // try to downcast to our error
    // if successful, do .with_span()
    // if not, just return the error as is
}
```

max-sixty: (copying from https://discord.com/channels/936728116712316989/1104523631024930926/1104523631024930926, feel free to respond in either)

When should we use [Errors](https://github.com/PRQL/prql/blob/258dc92df06d0f038ad0d1db6db111082b517dcc/prql-compiler/src/error.rs#L16) vs [ErrorMessages](https://github.com/PRQL/prql/blob/258dc92df06d0f038ad0d1db6db111082b517dcc/prql-compiler/src/error.rs#L96)?

`with_span` etc are implemented on `Error`, but `downcast` goes to an `ErrorMessages`...

aljazerzen: I meant `anyhow::downcast`, not our `error::downcast`. That should give you Error and not ErrorMessages.

max-sixty: Ah great, OK.

(separately, I'm still interested in the Error vs ErrorMessage distinction! Should we try and unify them?)

aljazerzen: They are meant to be an internal detail and a public interface. Also, ErrorMessage contains much more information (location in terms of lines and columns, pretty error display in source code) so it's much larger.

If it bothers you that we have two error types, feel free to merge them.

max-sixty: As discussed on the call — `Error` is the internal struct and `ErrorMessage` is the external interface. So here these should be `Error`. I'll take a TODO to add some comments in the code.

max-sixty: Done in #2552

max-sixty:
Actually waiting on #2553, a general approach to the problem

prql-compiler/src/sql/gen_expr.rs
```diff
@@ -103,7 +103,7 @@ pub(super) fn translate_expr(expr: Expr, ctx: &mut Context) -> Result<sql_ast::E
                 "std.concat" => return process_concat(&expr, ctx),
                 "std.regex_search" => {
                     if let [search, target] = args.as_slice() {
-                        return process_regex(search, target, ctx);
+                        return process_regex(search, target, ctx).with_span(expr.span);
```
max-sixty, line 106: Pretty nice construction now!

PR #2543, opened 2023-05-07
internal: Use `strum` to produce variant names

max-sixty:
I came across this feature, seems like an small but easy simplification


eitsupi (approved):
Thanks!

PR #2549, opened 2023-05-07
fix: website highlighting on pi day post

aljazerzen:


max-sixty:
Is there something we need to run to update the minified version when we cange the original?

aljazerzen:
We don't have an original version committed. I've just added Python so formatting is not broken for the pi day post.

PR #2559, opened 2023-05-10
docs: "What's next", for discussion

max-sixty:
@snth asked where to make comments on this — the doc we used for discussing "What's next" on the call at the weekend.

So adding here — I'm not planning to merge it in its current state, but we can use this PR for any discussions. (And very open to merging something like this, probably more developed).

### DIFF
--- a/web/website/content/posts/2023-05-07-whats-next-for-prql.md
+++ b/web/website/content/posts/2023-05-07-whats-next-for-prql.md
@@ -1,0 +1,118 @@
+# How can we make PRQL successful?
+
+A synthesis of where we're at, and some ideas for focusing the project in a
+successful direction for 2H2023.
+
+## Strengths
+
+- Huge developer engagement and excitement — one of the most successful
+  open-source launches in 2022, as evidenced by GitHub stars, HackerNews
+  reactions, and VC outreach.
+- Without rehashing PRQL's logic — this is a promising area for transformation,
+  as very few people enjoy SQL — separating the flawed language from the robust
+  relational algebra offers significant potential.
+- An good institutional design for the problem space — completely open-source,
+  devoid of commercial interests; this avoids conflicts of interest, especially
+  crucial at the language layer.
+
+## Weaknesses
+
+In summary, we have one major shortcoming: we haven't yet established a flywheel
+consisting of:
+
+- Users — using the language, providing feedback on what works and what doesn't.
+  They then become evangelists for more users, attracting tools that want to
+  integrate PRQL. (Some technical users may even become contributors.)
+- Tools — incorporating PRQL into their offerings, broadening PRQL's audience,
+  and giving us valuable insights into which language features succeed or fail.
+- Developers — building PRQL! Due to the Users & Tools, this feels like an
+  impactful and exciting project rather than a personal hobby (although that's
+  fine too!).
+
+Specifically, on each of those:
+
+- Very few actual users (compared to stars)
+- No outstanding implementation in any data tool. While there are decent
+  bindings, it's unclear where to direct someone who simply wants to write PRQL
+  and view data. Some attempts (e.g., dbt-prql) faltered on technical grounds.
+- Though the Developers aspect of the three-legged stool has consistently been
+  the strongest, even here we've seen a decline in engagement:
+  - The past few months have had fluctuating levels of contributions outside the
+    core team — with a surge in Q1, but slowing since then.
+  - Recently, even the core team's pace has decreased (although there's always
+    variance, and I'm not suggesting a quick fix by us working harder).
+
+## Finding the Flywheel
+
+The most crucial step is enabling people to use PRQL. Much will follow — we'll
+know which features and bugs to tackle, be able to showcase satisfied users, and
+so on.
+
+Some options:
+
+- Revisit `dbt`. I had envisioned this as a particularly fruitful integration —
+  for both PRQL and dbt.
+
+  - The monkeypatch plugin couldn't technically achieve everything needed.
+  - I also failed to have the dbt-core PR merged — it received a brief review
+    before being dropped.
+  - The dbt team remains interested in merging an improved version of the
+    dbt-core plugin (but won't guarantee a merge, which I can understand, though
+    makes it a less compelling investment)
+  - I'm hesitant to invest more time given this track record but open to it if
+    others believe it's a worthwhile bet.
+
+- Enhance language-specific integrations
+
+  - The Jupyter integration is "OK". Python is very popular so if it were
+    possible to make it excellent, then it could be great. But can it be great?
+    Can we get good editor features in a notebook? I _think_ we can get syntax
+    highlighting.
+    - TODO: how much integration can we actually do in Jupyter? Could we have an
+      LSP?
+  - The R extension looks great. I don't have much experience with R and haven't
+    explored it extensively. Do others have opinions? Is the R audience enough
+    to make a project-wide bet on R?
+
+- Create our own tool. This gives us complete control.
+
+  - But unless it's something exclusive to PRQL, are we likely to build
+    something superior to what's available?
+  - We're already asking people to bet on PRQL — is asking them to gamble on
+    both a new language and a new tool likely to work?
+  - Generally, successful things are smaller than we anticipate, not larger.
+  - Is there something PRQL can do that SQL can't, which would provide more
+    justification for building our own tool? Perhaps something related to
+    pipelines and instant feedback (e.g. the playground displaying data results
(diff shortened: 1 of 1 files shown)


web/website/content/posts/2023-05-07-whats-next-for-prql.md
```diff
@@ -0,0 +1,118 @@
...
+    is quite impressive and could be expanded)?
+
+- Fork Rill
+
+  - Rill seems to be an especially good match for PRQL — enabling quick
+    exploration and immediate results.
+  - However, I didn't receive a response from the team in my latest email
+    (though I can follow up again). We had a productive Zoom meeting previously.
+  - We could consider forking the project if necessary, and then using any
+    success there to attempt to merge it to mainline.
```
aljazerzen, line 96: As I noted, forking Rill would be a temporary boost to the project, but would not showcase strengths of PRQL as it would be hard to integrate good highlighting and LSP support.

snth:
Thanks for this. I've had some thoughts written down on Hackmd for a few days but have not been finding the time to transfer them (since I want to edit and expand). I will do my best to get these across as soon as I can.

web/website/content/posts/2023-05-07-whats-next-for-prql.md
```diff
@@ -0,0 +1,118 @@
...
+
+  - The Jupyter integration is "OK". Python is very popular so if it were
+    possible to make it excellent, then it could be great. But can it be great?
+    Can we get good editor features in a notebook? I _think_ we can get syntax
+    highlighting.
+    - TODO: how much integration can we actually do in Jupyter? Could we have an
+      LSP?
+  - The R extension looks great. I don't have much experience with R and haven't
+    explored it extensively. Do others have opinions? Is the R audience enough
+    to make a project-wide bet on R?
```
eitsupi, line 75: In my personal opinion, the most valuable aspect of prqlr is the ability to easily embed PRQL code blocks in Quarto documents.
This also can be done with IPython magic with pyprql, but it is interpreted as a Python code block and does not look very good. (quarto-dev/quarto-cli#4227)

Embedding PRQL code blocks into Quarto documents works very well, as in https://github.com/eitsupi/querying-with-prql, but the problem is that there is no way to run PRQL code interactively.
SQL code blocks can be executed interactively in the RStudio IDE, but if you want PRQL code blocks to work the same way, we will need to add functionality to the RStudio IDE. (Related to rstudio/rstudio#12767)
Also, not even SQL code blocks are currently executable in VSCode. (quarto-dev/quarto#26)
I am thinking it would be more realistic to wait for Quarto to be able to shape IPython Magic's embedded code blocks into pretty code blocks rather than making PRQL code blocks interactively executable in RStudio or VSCode.

snth:
Hi folks,

sorry about the long absence. I will start with a general post with some comments and thoughts and then maybe will open a review with comments on specific points in the text (if I feel that's still going to add value).

@max-sixty 's discussion text frames the problem of adoption in terms of Users, Tools and Developers. My sense is that we should aim for users, probably through availability of PRQL in tools. Developers would make that possible but are probably harder to attract/recruit. I like how Josh Wills described DuckDB in the [Data Stack show podcast](
https://datastackshow.com/podcast/data-collection-secrets-the-search-data-problem-with-josh-wills/) (starting at 46:20) as an "Experiential Good", i.e. something where the experience is so good that anyone that tries it starts becoming an advocate for it. I feel similarly about PRQL and feel that if we could just get people to try PRQL then adoption would follow on its own over time. In that episode they also talk about how Josh was singing the praises of DuckDB long before it was as cool as it is now and I think the experiential good point was about how, even though the number of users was small at the start, the experience had such a pofound effect on people that the growth came eventually, it had to come. I still hold out hope in a similar vein for PRQL.

Overall, I think we should continue to work on making the language as good and useful as it can be while trying to enable integration into more tools. I got a few people to open experimental branches of PRQL into their tools (BuenaVista, Evidence.dev) but failed to convert those into full integrations because of lack of time in my part. That is my bad and a lesson I'm taking away from that is that I will try to create more visibility around this and open Issues on Github for these so others can assist or take over should I drop away.

I feel that building our own tool would be a distraction. It's hard enough for tools like dbt or Rill to get traction and becoming another competitor in that space would just lead to more fragmentation and I also doubt that we have the resources / manpower to be very competitive. I also think that there are still things that can be improved in the language to make the user experience better first so I would start (continue) there.

DuckDB is amazing and the pace at which they continue to innovate, both on the db tech as well as their SQL dialect, is staggering. I therefore totally see the attraction of focusing our efforts on just targeting them as a backend (as I did in my Pi Day blog post or in prql-query). However for the PRQL language I believe that would be a mistake, at least at this point. We've made quite a song and dance about PRQL's database agnosticism as being one of its core features/benefits. To reverse course on that point would not set a good precedent for building trust amongst users that we will stick to our vision and roadmap.

One idea I had to get around the burden of supporting multiple SQL dialects is to utilise the [sqlglot library](https://github.com/tobymao/sqlglot). It's a great pity that this is written in Python instead of Rust. I thought that here perhaps a bit of funding could be well spent. The idea is to have sqlglot translated to Rust. Perhaps with GPT4 assistance this might be a quicker job than it would be otherwise (I have no personal experience of this but that's the impression I'm getting from LLM chatter). It's quite a well contained project and should be relatively easily testable so it might make a good candidate for a GSOC project? Otherwise maybe with a bit of funding someone could be paid to do this as a freelance job? Toby Mao, the sqlglot creator, might also be interested in this since I think sqlglot forms quite an integral part of his sqlmesh project so that would also benefit from the speed and reliability improvements of having sqlglot in Rust.

I think dbt continues to be a worthwhile integration to pursue. **sqlmesh** is the new entrant in that space which is also picking up quite a bit of interest. I haven't had time to investigate it properly yet myself but from what I've seen so far it looks good. A sqlmesh integration could also be great. Especially if we could cooperate on sqlglot.

snth:
Update on my previous post: based on this tweet Toby Mao might not be that enthusiastic to rewrite sqlglot into Rust:

https://twitter.com/Captaintobs/status/1658351633099812865

aljazerzen:
> We've made quite a song and dance about PRQL's database agnosticism as being one of its core features/benefits. To reverse course on that point would not set a good precedent...

I wouldn't say we should start dropping support, just not support new features for dialects that require additional work.

The main point here is that while PRQL right may be interesting to "language people" and people who are really frustrated with SQL, it does not provide a good workflow to any user group. Our recommended way of using PRQL with real data right now, would probably be Jupyter magic, which does not have syntax highlighting or error detection. There are different problems with VS code extension and pq, but my point is this:

*we should focus on one use-case, and perfect that*

This does not mean that we drop support for other use-cases, even contrary: many of the fixes and features we will implement for the focused use-case will be applicable for other use cases too. My thinking here is that we cannot develop the tooling for every use-case where PRQL would excel: there is just too many of them. We need a MVP that people can use, right now, without copy-pasting from the Playground.

tobymao:
> Update on my previous post: based on this tweet Toby Mao might not be that enthusiastic to rewrite sqlglot into Rust:
> 
> https://twitter.com/Captaintobs/status/1658351633099812865

Although there's little chance I'm going to rewrite sqlglot in rust, I'm happy to collaborate. For example, you could leverage sqlglot's multi dialect transpilation, optimizations (predicate/projection pushdowns, boolean algebra) etc. Let me know if you're interested in chatting :)

max-sixty:
> The main point here is that while PRQL right may be interesting to "language people" and people who are really frustrated with SQL, it does not provide a good workflow to any user group. Our recommended way of using PRQL with real data right now, would probably be Jupyter magic, which does not have syntax highlighting or error detection. There are different problems with VS code extension and pq, but my point is this:
> 
> _we should focus on one use-case, and perfect that_

I strongly agree with this (where "one use-case" might be a couple, but it's driving towards a use-case...)

max-sixty:
Something @snth brought up on the recent dev call — a TUI for querying data, similar to the playground:
- Could work with local data, potentially data from an HTTP source
- Would rely on DuckDB for all computation (so less flexible than the `prql-query` vision)
- Similar to the playground, it could show results on every keystroke. So unlike a `prqlc compile | duckdb` approach, it could allow for easy iteration; maybe keeping some history of queries
- [Visidata](https://www.visidata.org/) is a comparison, implemented in python

This has the same drawbacks as other proposals to build our own tools — users have to change both the language _and_ the tool they're using. But the TUI might attract technical users, which are likely a good set of early users. And it might uniquely appeal to PRQL's advantages for exploratory work. Having it in Rust would give access to a bunch of TUI crates.

aljazerzen:
Moved to a HackMD document.

PR #2560, opened 2023-05-10
refactor: split SQL backend into two stages

aljazerzen:
Splits SRQ from RQ AST completely.

I want to do this because currently, SRQ is being created just of sub-trees of RQ and not the whole RQ at once. This means that the compilation process was switching between compilation of RQ -> SRQ and SRQ -> SQL.

This PR separates this abstractions cleanly and opens opportunities to simplify SQL backend code. `sql::Context` and `anchor::Context` are not yet separated cleanly, but I'll leave that for a future PR.

The whole SRQ AST can be shown using `prql sql:anchor` command.


prql-compiler/src/sql/mod.rs
```diff
@@ -75,23 +75,21 @@ pub mod internal {
...
-        Ok(anchor::extract_atomics_naive(pipeline, &mut ctx.anchor))
+    pub fn anchor(query: Query) -> Result<srq::ast::SqlQuery> {
+        let (query, _ctx) = srq::compile_query(query, Some(dialect::Dialect::Generic))?;
+        Ok(query)
     }
 }
 
 #[derive(Debug)]
-struct Context {
+pub(self) struct Context {
```
max-sixty:
```suggestion
struct Context {
```
(or fine as-is)

prql-compiler/src/sql/srq/anchor.rs
```diff
@@ -57,34 +60,16 @@ pub(super) fn extract_atomic(
...
-            }
-        }
-    }
-
-    atomics.reverse();
-    atomics
-}
-
 /// Splits pipeline into two parts, such that the second part contains
 /// maximum number of transforms while "fitting" into a SELECT query.
```
max-sixty, line 64:
```suggestion
/// maximum number of transforms while "fitting" into a SELECT query,
/// and the first part contains the remainder. The input pipeline is consumed.
```

(nothing groundbreaking but for my own notes)

max-sixty, line 64:
```suggestion
/// maximum number of transforms while "fitting" into a SELECT query,
/// and the first part contains the remainder. The input pipeline is consumed.
```

(nothing groundbreaking but for my own notes, feel free to discard)

prql-compiler/src/sql/srq/ast.rs
```diff
@@ -0,0 +1,170 @@
...
+    SString(Vec<InterpolateItem<Expr>>),
+}
+
+#[derive(Debug, Clone, Serialize)]
+pub enum RelationExpr {
+    Ref(TId, Option<String>),
+    SubQuery(SqlRelation, Option<String>),
+}
+
+#[derive(Debug, Clone)]
```
max-sixty: Would be good to be able to serialize SRQ 

(I'm actually not sure how this works without these being serializable?

> The whole SRQ AST can be shown using prql sql:anchor command.

)

prql-compiler/src/sql/srq/ast.rs
```diff
@@ -0,0 +1,170 @@
...
+    Loop {
+        initial: SqlRelation,
+        step: SqlRelation,
+        recursive_name: String,
+    },
+}
+
+/// Similar to [rq::Transform], but more similar to SQL clauses.
+///
+/// Uses a two generic args that allows compiler to do work in multiple stages.
```
max-sixty:
```suggestion
/// Uses two generic args that allows the compiler to work in multiple stages.
```

prql-compiler/src/sql/srq/ast.rs
```diff
@@ -0,0 +1,170 @@
...
+use anyhow::Result;
+use enum_as_inner::EnumAsInner;
+use itertools::Itertools;
+use serde::Serialize;
+
+use crate::ast::pl::{ColumnSort, InterpolateItem, JoinSide, RelationLiteral};
+use crate::ast::rq::{self, Expr, RqFold, TId};
+
+#[derive(Debug, Clone)]
+pub struct SqlQuery {
```
max-sixty, line 15: I wonder if this could just be a `Vec<Cte>`, and the final one gets rendered without the `WITH`

(no need to change obv)

prql-compiler/src/sql/srq/ast.rs
```diff
@@ -0,0 +1,170 @@
...
+pub enum CteKind {
+    Normal(SqlRelation),
+    Loop {
+        initial: SqlRelation,
+        step: SqlRelation,
+        recursive_name: String,
+    },
+}
+
+/// Similar to [rq::Transform], but more similar to SQL clauses.
```
max-sixty:
```suggestion
/// Similar to [rq::Transform], but closer to a SQL clause.
```

prql-compiler/src/sql/srq/ast.rs
```diff
@@ -0,0 +1,170 @@
...
+            _ => self.as_ref(),
+        }
+    }
+
+    pub fn into_super_and<T, F: FnOnce(rq::Transform) -> Result<T, rq::Transform>>(
+        self,
+        f: F,
+    ) -> Result<T, Self> {
+        self.into_super()
+            .and_then(|t| f(t).map_err(SqlTransform::Super))
```
max-sixty, line 116: Maybe basic question, but the error is getting mapped into a `SqlTransform::Super`? How does that work?

prql-compiler/src/sql/srq/ast.rs
```diff
@@ -0,0 +1,170 @@
...
+    }
+}
+
+pub trait SrqFold<RelIn, RelOut, SuperIn, SuperOut>: RqFold {
```
max-sixty, line 120: FWIW I still don't really get what `Super` is. 

I know we briefly exchanged comments on another PR a while ago. To the extent it's possible to add a note somewhere, that would make it clearer... (though don't want to slow you down)

prql-compiler/src/sql/srq/ast.rs
```diff
@@ -0,0 +1,170 @@
...
+            bottom: fold.fold_rel(bottom)?,
+            distinct,
+        },
+        SqlTransform::Select(v) => SqlTransform::Select(v),
+        SqlTransform::Filter(v) => SqlTransform::Filter(v),
+        SqlTransform::Aggregate { partition, compute } => {
+            SqlTransform::Aggregate { partition, compute }
+        }
+        SqlTransform::Sort(v) => SqlTransform::Sort(v),
+        SqlTransform::Take(v) => SqlTransform::Take(v),
```
max-sixty: Many / most of these could just be

```suggestion
        _ => transform
```

...I think, but no great issue, and maybe explicit forces us to consider each case

prql-compiler/src/sql/srq/context.rs
```diff
@@ -45,7 +46,43 @@ pub(super) struct SqlTableDecl {
     /// Relation that still needs to be defined (usually as CTE) so it can be referenced by name.
     /// None means that it has already been defined, or was not needed to be defined in the
     /// first place.
-    pub relation: Option<SqlRelation>,
+    pub relation: RelationStatus,
+}
+
+#[derive(Debug, Clone)]
+pub enum RelationStatus {
```
max-sixty, line 53: I guess this could just be `Option<RelationAdapter>`? (But fine to have it specific ofc)

prql-compiler/src/sql/srq/context.rs
```diff
@@ -45,7 +46,43 @@ pub(super) struct SqlTableDecl {
...
+    /// Table or a common table expression. It can be referenced by name.
+    Defined,
+
+    /// Relation expression which is yet to be defined.
+    NotYetDefined(RelationAdapter),
+}
+
+impl RelationStatus {
+    /// Analogous to [Option::take]
```
max-sixty, line 62: Ah OK, as above

prql-compiler/src/sql/srq/gen_query.rs
```diff
@@ -0,0 +1,272 @@
+//! This module is responsible for translating PRQL AST to sqlparser AST, and
```
max-sixty:
```suggestion
//! This module is responsible for translating PRQL RQ to sqlparser AST, and
```
Is that correct?

prql-compiler/src/sql/srq/mod.rs
```diff
@@ -0,0 +1,86 @@
+//! SQL Relational Query
+//!
+//! This in an internal intermediate representation between RQ and SQL AST.
+//!
+//! For example, RQ does not have a separate node for DISTINCT, but uses [crate::ast::rq::Take] 1 with
+//! `partition`. In [super::preprocess] module, [crate::ast::rq::Transform] take is wrapped into
+//! [SqlTransform], which does have [SqlTransform::Distinct].
```
max-sixty, line 7: Great summary!

prql-compiler/src/sql/srq/mod.rs
```diff
@@ -0,0 +1,86 @@
...
+//!
+//! This module also contains compiler from RQ to SRQ.
```
max-sixty:
```suggestion
//! This module also contains the compiler from RQ to SRQ.
```

prql-compiler/src/sql/srq/mod.rs
```diff
@@ -0,0 +1,86 @@
...
+        // One aggregate, take at the end
+        let prql: &str = r###"
+        from employees
+        filter country == "USA"
+        aggregate [sal = average salary]
+        sort sal
+        take 20
+        "###;
+
+        assert_eq!(count_atomics(prql), 1);
```
max-sixty: Obv don't change, but FYI I've started putting the whole expression in the `assert`s, since then when it fails, it shows the full input (in particular with `insta`)

prql-compiler/src/sql/srq/ast.rs
```diff
@@ -0,0 +1,170 @@
...
+        step: SqlRelation,
+        recursive_name: String,
+    },
+}
+
+/// Similar to [rq::Transform], but more similar to SQL clauses.
+///
+/// Uses a two generic args that allows compiler to do work in multiple stages.
+/// First convert RQ to [SqlTransform<TableRef, rq::Transform>] and at the end
+/// compile that to [SqlTransform<RelationExpr, ()>].
```
max-sixty, line 56: Thanks for explaining. I'm probably 70% of the way to having a good mental model of this

max-sixty (approved):
Overall looks great!

I've tried to spend more time trying to understand this so I don't fall behind on it. I got most of the way there.

I left a spectrum of comments, none require changes. Some ask for a bit more explanation if you have the patience.

One thing that would be good to do is to allow SRQ to be serialized — e.g. a test which prints the yaml. Fine to do in another PR / leave it to someone else though.

prql-compiler/src/sql/srq/ast.rs
aljazerzen: It's using Debug trait. I can add serialize.

prql-compiler/src/sql/srq/ast.rs
aljazerzen: Yes, it probably could, but I intended SRQ to dictate query structure - how many CTEs, subqueries and all that is determined in anchor stage, so the last "translation stage" is one-to-one mapping into SQL AST.

aljazerzen: I'll add a comment about that.

prql-compiler/src/sql/srq/ast.rs
aljazerzen: Yeah, this is not actually an error.

Strum provides this method:

```rust
fn into_select(self) -> Result<Select, Self>
```
... which returns Self in the Err if unwrapping was unsuccessful. Result is used instead of Option because you don't want to drop the value of `self` when unwrapping is unsuccessful.

This method is declared so I can do this:
```rust
sql_transform.into_super_and(|transform| transform.into_loop())
```
... it allows access to `.into` methods within the `Super`.

prql-compiler/src/sql/srq/context.rs
aljazerzen: Yes, I had that, but this seemed more readable.

prql-compiler/src/sql/srq/gen_query.rs
aljazerzen: RQ -> SRQ

This comment is mostly obsolete now.

aljazerzen:
Thanks for the review. I've left quite a few things unfinished and you caught a quite a few of hem.

There is still the question of how to name these compiler stages. "anchor + preprocess" and "translate" are what they are called now (in some places).

There are also duplicate "gen_query" files, which should probably have different names...

prql-compiler/prqlc/src/cli.rs
```diff
@@ -502,4 +517,74 @@ group a_column (take 10 | sort b_column | derive [the_number = rank, last = lag
...
+                      end: 53
+                  - kind:
+                      Literal:
+                        Integer: 0
+                    span:
+                      start: 56
+                      end: 57
+              span:
+                start: 47
+                end: 57
```
max-sixty, line 587: This is great! Very understandable!

prql-compiler/src/sql/srq/ast.rs
```diff
@@ -0,0 +1,189 @@
...
+pub enum SqlTransform<Rel = RelationExpr, Super = rq::Transform> {
+    /// Contains [rq::Transform] during compilation. After finishing, this is emptied.
+    ///
+    /// For example, initial an RQ Append transform is wrapped as such:
+    ///
+    /// ```ignore
+    /// rq::Transform::Append(x) -> srq::SqlTransform::Super(rq::Transform::Append(x))
+    /// ```
+    ///
+    /// During preprocessing, `Super(Append)` is converted into `srq::SqlTransform::Union { .. }`.
```
max-sixty, line 67: 👍

prql-compiler/src/sql/srq/ast.rs
max-sixty: Ah yes, I remember this from the library that makes the `.into_X` now. IIRC it _is_ an error, even if it's not represented as such...

prql-compiler/src/sql/srq/context.rs
max-sixty: Yes cool. I guess something like `RelationDefinition` might be more descriptive (but don't change now)

web/book/tests/snapshots/snapshot__examples__employees-2.snap
```diff
@@ -14,34 +14,34 @@ WITH table_5 AS (
...
   SELECT
-    AVG(table_4._expr_0) AS salary_avg,
-    STDDEV(table_4._expr_0) AS salary_sd,
+    AVG(table_2._expr_0) AS salary_avg,
+    STDDEV(table_2._expr_0) AS salary_sd,
     dm.emp_no
   FROM
-    table_5 AS table_4
-    JOIN dept_emp AS de ON table_4.emp_no = de.emp_no
+    table_4 AS table_2
```
max-sixty, line 23: Irrelevant to this PR — one day we can collapse these!

aljazerzen: Actually a bit relevant.

Without this split, you'd have to assign a name to a table instance before knowing if it will compile to a subquery or a CTE.

aljazerzen: ... which is not possible.

And now there is a time in compilation process where all table instances know whether they will be sub-queries or references to CTEs. So we can infer the instance name from the table reference.

PR #2561, opened 2023-05-11
refactor: nicer Span serialization

aljazerzen:


max-sixty:
Great!

I do like the more concise serializations; I find for me it makes it easier & faster to scan an AST

PR #2562, opened 2023-05-11
feat: propage sort down pipelines

### LINKED ISSUE #1363: Transform row ordering guarantees

aljazerzen:
Closes #1363 

This PR also includes lots of refactoring that was needed to accomplish what I wanted.


prql-compiler/src/sql/srq/anchor.rs
```diff
@@ -226,7 +224,7 @@ pub(super) fn anchor_split(
     ctx: &mut AnchorContext,
     preceding: Vec<SqlTransform<TableRef>>,
     atomic: Vec<SqlTransform<TableRef>>,
-) -> (Vec<SqlTransform<TableRef>>, HashMap<CId, CId>) {
+) -> Vec<SqlTransform<TableRef>> {
```
max-sixty, line 227: (fwiw I do think simplifying signatures is good per se)

prql-compiler/src/sql/srq/postprocess.rs
```diff
@@ -0,0 +1,125 @@
+use std::collections::HashMap;
```
max-sixty, line 1: Would be good to have a summary of what's done in `postprocess` if possible (but don't want to slow you down, maybe this is the sort of thing I can add as a test of my understand!)

prql-compiler/src/sql/srq/anchor.rs
aljazerzen: Yeah, because I store cid_redirects in Context. Which is more complex I'd say.

snth:
👏

PR #2563, opened 2023-05-11
feat!: infering modules and treating them as SQL schemas

### LINKED ISSUE #1535: Resolve opaqueness of idents

aljazerzen:
- Squashed commit of the following
- infer modules
- make table reference an Ident instead of String

Closes #1535

A big commit, I had to tweak how we do inference.

A thought: should we throw an error when a table name ends in `.parquet` or `.csv`? It'd be nice to redirect users toward `read_parquet` or `read_csv`.

### DIFF
--- a/prql-compiler/src/ast/pl/ident.rs
+++ b/prql-compiler/src/ast/pl/ident.rs
@@ -66,6 +66,12 @@ impl Ident {
             .zip(self.iter())
             .all(|(prefix_component, self_component)| prefix_component == self_component)
     }
+
+    pub fn starts_with_part(&self, prefix: &str) -> bool {
+        std::iter::once(prefix)
+            .zip(self.iter())
+            .all(|(prefix_component, self_component)| prefix_component == self_component)
+    }
 }
 
 impl std::fmt::Display for Ident {
--- a/prql-compiler/src/ast/pl/ident.rs
+++ b/prql-compiler/src/ast/pl/ident.rs
@@ -26,6 +26,8 @@ impl Ident {
         }
     }
 
+    /// Remove last part of the ident.
+    /// Result will generally refer to the parent of this ident.
     pub fn pop(self) -> Option<Self> {
         let mut path = self.path;
         path.pop().map(|name| Ident { path, name })
--- a/prql-compiler/src/ast/pl/ident.rs
+++ b/prql-compiler/src/ast/pl/ident.rs
@@ -42,6 +42,12 @@ impl Ident {
         }
     }
 
+    pub fn prepend(self, part: String) -> Ident {
+        let mut parts = vec![part];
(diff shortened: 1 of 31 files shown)


prql-compiler/src/ast/pl/ident.rs
```diff
@@ -40,6 +42,12 @@ impl Ident {
         }
     }
 
+    pub fn prepend(self, part: String) -> Ident {
+        let mut parts = vec![part];
+        parts.extend(self.into_iter());
+        Ident::from_path(parts)
```
max-sixty, line 48: nit if you think concise is better, discard if not
```suggestion
    Ident::from_path(std::iter::once(part).chain(self.into_iter()).collect())
```

prql-compiler/src/ast/pl/ident.rs
```diff
@@ -58,6 +66,12 @@ impl Ident {
             .zip(self.iter())
             .all(|(prefix_component, self_component)| prefix_component == self_component)
     }
+
+    pub fn starts_with_part(&self, prefix: &str) -> bool {
+        std::iter::once(prefix)
+            .zip(self.iter())
+            .all(|(prefix_component, self_component)| prefix_component == self_component)
```
max-sixty:
```suggestion
       self.iter().next().map_or(false, |self_component| self_component == prefix)
```
?

prql-compiler/src/ast/pl/ident.rs
aljazerzen: Nah, i think using vec is more readable than iterators.

max-sixty:
I've been going through my notifications and I found an unsent comment on this — thank you very much indeed @aljazerzen — I found this a pervasive issue for the lang, now solved!

PR #2565, opened 2023-05-11
docs: Update book for sort guarantees

max-sixty:


max-sixty:
Great, much clearer, thanks!

Issue #1910, opened 2023-02-20
Use cargo-dist to create binaries as release assets


max-sixty:
We've had some good progress on creating binaries in #1883 & #762

In addition to those, we could also make lots of binaries with something like https://github.com/axodotdev/cargo-dist. I haven't looked at it in detail, but if it would let us do something like the "Assets" in https://github.com/BurntSushi/ripgrep/releases/tag/13.0.0 (screenshot below), that would be great.

<img width="946" alt="image" src="https://user-images.githubusercontent.com/5635139/220189542-74198000-a93e-4e9a-86bb-9f54bbff8c79.png">

eitsupi:
cargo-dist doesn't support cross compile. axodotdev/cargo-dist#74
So at this point I think it is better not to use cargo-dist.

Another project that might be helpful for CI settings.
https://github.com/Nukesor/pueue

eitsupi:
`dist` (cargo-dist) has made remarkable progress, and now it may even be possible to cross-compile binaries for arm64 Windows (though I'm not sure if that fits with this repository's caching strategy).

max-sixty:
Yes `dist` looks like it's been really successful, +1

In general, I'd be keen to embrace tools like that — we get their enhancements for free and reduce the scope of "non-core" code in PRQL. Ofc requires assessing whether the tool will grow or lose support, which can be difficult ahead of time.

Also there are now [ARM runners in preview for Linux](https://docs.github.com/en/actions/using-github-hosted-runners/using-github-hosted-runners/about-github-hosted-runners#standard-github-hosted-runners-for-public-repositories). But not yet Windows, as you point out

eitsupi:
The problem with adopting dist is that it generates its own GHA workflow file, which would not be compatible with our release strategy and cache.
As for the release, it increases the cache size and may cause the cache to disappear sooner.

max-sixty:
Ah, that's a very good point; I hadn't thought of that. I had a quick look but didn't see a way to get around that in https://opensource.axo.dev/cargo-dist/book/ci/customizing.html

It would be fine if we only ran `dist` on releases, and that created a bunch of cache artifacts. But (I guess?) it also does the "test dry run" runs, which would constantly evict our cache...

PR #2568, opened 2023-05-14
ci: build prqlc on PRs

eitsupi:
Addresses #1910

### DIFF
--- a/.github/actions/build-prqlc/action.yaml
+++ b/.github/actions/build-prqlc/action.yaml
@@ -13,6 +13,9 @@ inputs:
 runs:
   using: composite
   steps:
+    - run: rustup target add ${{ inputs.target }}
+      shell: bash
+
     - uses: Swatinem/rust-cache@v2
       with:
         save-if: ${{ github.ref == 'refs/heads/main' }}
--- a/.github/actions/build-prqlc/action.yaml
+++ b/.github/actions/build-prqlc/action.yaml
@@ -29,23 +29,12 @@ runs:
         sudo apt-get update
         sudo apt-get install -y gcc-aarch64-linux-gnu
         echo 'CARGO_TARGET_AARCH64_UNKNOWN_LINUX_MUSL_LINKER=aarch64-linux-gnu-gcc' >>"$GITHUB_ENV"
-        echo 'RUSTFLAGS=-C link-arg=-lgcc' >>"$GITHUB_ENV"
 
     - name: cargo build
       uses: richb-hanover/cargo@v1.1.0
       with:
         command: build
         args: --package prqlc --release --locked --target=${{ inputs.target }}
-
-    - if: runner.os == 'Windows'
-      shell: powershell
-      run: echo "BIN_NAME=prqlc.exe" >>"$env:GITHUB_ENV"
-
-    - if: runner.os != 'Windows'
-      shell: bash
-      run: |
-        echo "BIN_NAME=prqlc" >>"$GITHUB_ENV"
-        chmod +x target/${{ inputs.target }}/release/prqlc
 
     - name: Upload prqlc
       uses: actions/upload-artifact@v3
--- a/.github/actions/build-prqlc/action.yaml
+++ b/.github/actions/build-prqlc/action.yaml
@@ -31,7 +31,9 @@ runs:
 
     - if: runner.os != 'Windows'
       shell: bash
-      run: echo "BIN_NAME=prqlc" >>"$GITHUB_ENV"
+      run: |
+        echo "BIN_NAME=prqlc" >>"$GITHUB_ENV"
+        chmod +x target/${{ inputs.target }}/release/prqlc
 
     - name: Upload prqlc
       uses: actions/upload-artifact@v3
--- a/.github/actions/build-prqlc/action.yaml
+++ b/.github/actions/build-prqlc/action.yaml
@@ -1,0 +1,40 @@
+name: build-prqlc
+description: >
+  Build prqlc
+inputs:
+  target:
+    description: Build target
+    required: true
+  cross:
+    description: Use cross?
+    required: false
+    default: "false"
+
+runs:
+  using: composite
+  steps:
+    - uses: Swatinem/rust-cache@v2
+      with:
+        save-if: ${{ github.ref == 'refs/heads/main' }}
+        prefix-key: 0.8.1
+
+    - name: cargo build
+      uses: richb-hanover/cargo@v1.1.0
+      with:
+        command: build
+        use-cross: inputs.cross == 'true'
+        args: --package prqlc --release --locked --target=${{ inputs.target }}
+
+    - if: runner.os == 'Windows'
+      shell: powershell
+      run: echo "BIN_NAME=prqlc.exe" >>"$env:GITHUB_ENV"
+
+    - if: runner.os != 'Windows'
+      shell: bash
+      run: echo "BIN_NAME=prqlc" >>"$GITHUB_ENV"
+
+    - name: Upload prqlc
+      uses: actions/upload-artifact@v3
+      with:
+        name: prqlc-${{ inputs.target }}
+        path: target/${{ inputs.target }}/release/${{ env.BIN_NAME }}
--- a/.github/actions/build-prqlc/action.yaml
+++ b/.github/actions/build-prqlc/action.yaml
@@ -5,10 +5,6 @@ inputs:
   target:
     description: Build target
     required: true
-  cross:
-    description: Use cross?
-    required: false
-    default: "false"
 
 runs:
   using: composite
@@ -31,7 +27,6 @@ runs:
       uses: richb-hanover/cargo@v1.1.0
       with:
         command: build
-        use-cross: inputs.cross == 'true'
         args: --package prqlc --release --locked --target=${{ inputs.target }}
 
     - if: runner.os == 'Windows'
--- a/.github/actions/build-prqlc/action.yaml
+++ b/.github/actions/build-prqlc/action.yaml
@@ -23,6 +23,14 @@ runs:
         sudo apt-get update
         sudo apt-get install -y musl-tools
 
+    - if: inputs.target == 'aarch64-unknown-linux-musl'
+      shell: bash
+      run: |
+        sudo apt-get update
+        sudo apt-get install -y gcc-aarch64-linux-gnu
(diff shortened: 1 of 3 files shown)


eitsupi:
It seems working.
https://github.com/PRQL/prql/actions/runs/4971568371
image

.github/actions/build-prqlc/action.yaml
```diff
@@ -0,0 +1,51 @@
...
+
+    - if: runner.os == 'Windows'
+      shell: powershell
+      run: echo "BIN_NAME=prqlc.exe" >>"$env:GITHUB_ENV"
+
+    - if: runner.os != 'Windows'
+      shell: bash
+      run: |
+        echo "BIN_NAME=prqlc" >>"$GITHUB_ENV"
+        chmod +x target/${{ inputs.target }}/release/prqlc
```
eitsupi: This doesn't seem working.

eitsupi:
After a struggle I gave up on building `aarch64-unknown-linux-musl` target on amd64 linux. Could it be a [psm](https://github.com/rust-lang/stacker/tree/master/psm#platform-support) problem?

```sh
$ cargo build -p prqlc --release --target=aarch64-unknown-linux-musl
   Compiling cfg-if v1.0.0
   Compiling once_cell v1.17.1
   Compiling bitflags v1.3.2
   Compiling linux-raw-sys v0.1.4
   Compiling libc v0.2.140
   Compiling lazy_static v1.4.0
   Compiling memchr v2.5.0
   Compiling concolor-override v1.0.0
   Compiling concolor-query v0.3.3
   Compiling utf8parse v0.2.1
   Compiling adler v1.0.2
   Compiling gimli v0.27.2
   Compiling anstyle v0.3.5
   Compiling serde v1.0.157
   Compiling psm v0.1.21
   Compiling pin-project-lite v0.2.9
error: failed to run custom build command for `psm v0.1.21`

Caused by:
  process didn't exit successfully: `/workspaces/prql/target/release/build/psm-bd89755e99a601de/build-script-build` (exit status: 1)
  --- stdout
  OPT_LEVEL = Some("3")
  TARGET = Some("aarch64-unknown-linux-musl")
  HOST = Some("x86_64-unknown-linux-gnu")
  cargo:rerun-if-env-changed=CC_aarch64-unknown-linux-musl
  CC_aarch64-unknown-linux-musl = None
  cargo:rerun-if-env-changed=CC_aarch64_unknown_linux_musl
  CC_aarch64_unknown_linux_musl = None
  cargo:rerun-if-env-changed=TARGET_CC
  TARGET_CC = None
  cargo:rerun-if-env-changed=CC
  CC = None
  RUSTC_LINKER = None
  cargo:rerun-if-env-changed=CROSS_COMPILE
  CROSS_COMPILE = None
  cargo:rerun-if-env-changed=CFLAGS_aarch64-unknown-linux-musl
  CFLAGS_aarch64-unknown-linux-musl = None
  cargo:rerun-if-env-changed=CFLAGS_aarch64_unknown_linux_musl
  CFLAGS_aarch64_unknown_linux_musl = None
  cargo:rerun-if-env-changed=TARGET_CFLAGS
  TARGET_CFLAGS = None
  cargo:rerun-if-env-changed=CFLAGS
  CFLAGS = None
  cargo:rerun-if-env-changed=CRATE_CC_NO_DEFAULTS
  CRATE_CC_NO_DEFAULTS = None
  DEBUG = Some("false")
  CARGO_CFG_TARGET_FEATURE = Some("neon")
  cargo:rustc-cfg=asm
  cargo:rustc-cfg=switchable_stack
  running: "aarch64-linux-musl-gcc" "-O3" "-ffunction-sections" "-fdata-sections" "-fPIC" "-Wall" "-Wextra" "-xassembler-with-cpp" "-DCFG_TARGET_OS_linux" "-DCFG_TARGET_ARCH_aarch64" "-DCFG_TARGET_ENV_musl" "-o" "/workspaces/prql/target/aarch64-unknown-linux-musl/release/build/psm-9cc78c5eb0f1a15f/out/src/arch/aarch_aapcs64.o" "-c" "src/arch/aarch_aapcs64.s"

  --- stderr


  error occurred: Failed to find tool. Is `aarch64-linux-musl-gcc` installed?
```

.github/actions/build-prqlc/action.yaml
```diff
@@ -0,0 +1,54 @@
...
+runs:
+  using: composite
+  steps:
+    - run: rustup target add ${{ inputs.target }}
+      shell: bash
+
+    - uses: Swatinem/rust-cache@v2
+      with:
+        save-if: ${{ github.ref == 'refs/heads/main' }}
+        prefix-key: 0.8.1
```
max-sixty, line 18: Could we add an entry in https://github.com/PRQL/prql/blob/2bb271e9baabc1529649d64434e3ee1768ae1e1a/prql-compiler/Cargo.toml#L114-L115 for this?

.github/actions/build-prqlc/action.yaml
max-sixty: WDYM by it not working? 

If it's that downloading the file from GitHub doesn't retain these permissions — I would guess that's a constraint from the downloading rather than this step (or any download could be marked as executable).

.github/actions/build-prqlc/action.yaml
```diff
@@ -0,0 +1,54 @@
...
+      shell: bash
+      run: |
+        echo "BIN_NAME=prqlc" >>"$GITHUB_ENV"
+        chmod +x target/${{ inputs.target }}/release/prqlc
+
+    - name: Upload prqlc
+      uses: actions/upload-artifact@v3
+      with:
+        name: prqlc-${{ inputs.target }}
+        path: target/${{ inputs.target }}/release/${{ env.BIN_NAME }}
```
max-sixty:
```suggestion
        path: target/${{ inputs.target }}/release/${{ runner.os == 'Windows' && 'prqlc.exe' || 'prqlc' }}
```

...is maybe less explicit but more concise, will mean we can remove most of the previous two steps (your call on which is overall better...)

.github/workflows/pull-request.yaml
```diff
@@ -108,3 +108,27 @@ jobs:
           allowed-failures: check-links
           allowed-skips: test-all, publish-web, nightly
           jobs: ${{ toJSON(needs) }}
+
+  build-prqlc:
```
max-sixty, line 112: Before merging, I think probably we have this in `test-all` rather than `pull-request`  — or are there times it would fail on new code that we'd want to know about?

.github/actions/build-prqlc/action.yaml
```diff
@@ -0,0 +1,54 @@
...
+        sudo apt-get update
+        sudo apt-get install -y musl-tools
+
+    - if: inputs.target == 'aarch64-unknown-linux-musl'
+      shell: bash
+      run: |
+        sudo apt-get update
+        sudo apt-get install -y gcc-aarch64-linux-gnu
+        echo 'CARGO_TARGET_AARCH64_UNKNOWN_LINUX_MUSL_LINKER=aarch64-linux-gnu-gcc' >>"$GITHUB_ENV"
+        echo 'RUSTFLAGS=-C link-arg=-lgcc' >>"$GITHUB_ENV"
```
max-sixty: Is there any chance that `-C link-arg=-lgcc` needs quotes around it? (I don't know this well so please discard if you do...)

max-sixty (approved):
Excellent!! Thanks a lot.

A couple of points:
- No stress about Linux arm64. I see some comments in https://github.com/rust-lang/rust/issues/46651#issuecomment-433611633, not sure if they would help, but fine to leave for another day. 
  - I've also heard about https://github.com/cross-rs/cross, but don't have any experience with it. To the extent this can encapsulate much of the complication around non-default targets, that could work well. I don't see obvious docs for how to use with GHA — not sure whether that's because it's so easy or so difficult! :)
- I'm not sure whether this necessarily needs to be an "action" rather than a normal workflow, especially now that GH has loosened the restrictions on workflows calling other workflows. But it's also fine as an action.

- Do you know what we'd have to do to get these publishing on each release? We may have to auto-release IIUC. That would be a nice use of this.

.github/actions/build-prqlc/action.yaml
eitsupi: I think this is the correct way to specify it, since enclosing it in quotation marks failed.
Searching across GitHub, I can see other projects that do it this way.

.github/workflows/pull-request.yaml
eitsupi: My idea was to provide an easy way to try out features added by pull requests.
For example, I believe duckdb builds and uploads duckdb CLI executables with every pull request.

eitsupi:
Thank you for reviewing this.

I considered cross, but did not try it at this time due to the need for a completely different workflow to use Docker and concerns about slowdowns due to QEMU emulation.

Attaching these to the release would be easy as it is just one (?) additional step.

.github/actions/build-prqlc/action.yaml
eitsupi: I'm sorry, but I didn't understand your intent.

I do not think it is necessary to set a prefix key here, as I believe `Swatinem/rust-cache` will detect changes of Rust source code.
Just copied from the other workflows.

eitsupi:
Thanks to the advice I received in rust-lang/stacker#80, I am now able to build for Linux arm64.
(I have confirmed that the prqlc binary works on my Raspberry Pi 4)

image

Once this PR is merged I will work on attaching these to the releases.

.github/actions/build-prqlc/action.yaml
eitsupi: Oh, I did not understand how the prefix key is automatically updated. I have update the Cargo.toml.

.github/workflows/pull-request.yaml
max-sixty: OK great, we can try it. 

I do notice our CI queue filling up occasionally, and this would almost double the number of jobs. So if it becomes an issue we could always add a `pr-build-artifacts` label and run it based on that.

Definitely fine to try, I agree that would be a nice thing to have access to.

.github/actions/build-prqlc/action.yaml
max-sixty: Thanks a lot!

max-sixty (approved):
Looks great! 

I see there's an unrelated failure — don't let that delay merging this, I'll have a look at that later unless anyone else gets to it

PR #2570, opened 2023-05-15
feat: compile file tree as modules

aljazerzen:
1. We need more tests.
2. Few things are unfinished, namely:
   - resolving names from parent modules,
   - cyclic module references,

Ref #2129


max-sixty:
This looks excellent so far! Thanks for adding a few docstrings...

PR #2575, opened 2023-05-16
feat!: Functions definitions as `VarDef`s

max-sixty:
I think I may delete the rust part of this and start again, as the underlying `VarDef` has changed, and I think I scoped this badly. I was originally intending to change the syntax, and then thought it might be worth making functions another possible type in variable definitions.

Opening PRs has a good track record of getting help, so doing that for now:
- Should functions be a value in `VarDef`s? Or do we need a separate `FuncDef`?
- If not, any thoughts on the easiest way of changing the parser to accept `let`? If we can't have both `let` and `func` I think that's OK.


prql-compiler/src/parser/expr.rs
```diff
@@ -273,6 +274,38 @@ where
...
+    .then_ignore(new_line())
+    // .map(|(((name, return_ty), params), body)| {
+    .map(|(params, body)| {
+        let (pos, nam) = params
+            .into_iter()
+            .map(|((name, ty_expr), default_value)| FuncParam {
+                name,
+                ty_expr,
+                default_value,
+            })
```
aljazerzen, line 301: This part can be mapped earlier, right before `.repeated()`. Not sure if that's more readable, though.

prql-compiler/src/parser/expr.rs
```diff
@@ -96,6 +96,7 @@ pub fn expr() -> impl Parser<Token, Expr, Error = Simple<Token>> + Clone {
             ident_kind,
             case,
             param,
+            lambda_func().map(ExprKind::FuncDef),
```
aljazerzen: I'm not sure if this will work. If it won't, it's probably quite inefficient.

To spot the problem, take a look at Rust call structure:
```
expr() calls lambda_func()
lambda_func() calls expr() and expr_call()
expr_call() calls expr()
```
... which will probably result in stack overflow.

The answer to having recursive parsing expressions in chumsky is to use the `recursive` parser. We have it implemented for `expr()`, which means that within `recursive` closure, all calls to `expr()` could be replaced with the closure param.

(for an example, see this: https://github.com/PRQL/prql/pull/2575/files#diff-3e0ce0ca7ccc5943855871be35a742bf4df9416c5a55979fb5b666a441dde322R21)

The complicated part is that you are constructing lambda_func() parser in a function, which means that you should pass the recursive closure param `expr` into that function, to be used instead of calling `expr()` function again.

In effect, this would be the change:
```diff
- fn lambda_func() -> ... {
+ fn lambda_func<E>(expr: E) -> ...
+ where E: Parser<Token, Expr, ...> + Clone {
```

I'm doing that for `func_call()` already, check it out there.

prql-compiler/src/ast/pl/stmt.rs
```diff
@@ -64,6 +71,7 @@ pub struct FuncParam {
 pub struct VarDef {
     pub name: String,
     pub value: Box<Expr>,
+    pub ty_expr: Option<Expr>,
```
aljazerzen, line 75: Yes, nice.

But we are not parsing this yet, right?

aljazerzen: Wait, we don't need to parse this yet in this PR. Functions will be parsed from this:

```
let a = p -> b
```
into:
```yaml
VarDef:
- name: a,
- value:
    FuncDef:
      positional_params:
      - p
      body:
      - b
```

Now, `FuncDef` can easily be converted into `Closure` in `resolver::closure_of_var_def`.

aljazerzen: Wait, `ExprKind::FuncDef` and `ExprKind::Closure` are now approximately the same thing! 

You don't even need to add `ExprKind::FuncDef` and just parse into `ExprKind::Closure`, which can be renamed into `ExprKind::LambdaFunc` or just `ExprKind::Func`?

aljazerzen (approved):
Awesome! This really feels right - simplification in syntax and semantics without losing any expressive power of the language.

We may want to keep the "func" declarations as I noted somewhere.

aljazerzen:
I've merged main and fixed resolving where you had the comment about the blocker.

aljazerzen:
Oh, and I didn't see the PR message:

> Should functions be a value in VarDefs? Or do we need a separate FuncDef?

Yes. We threat functions as values elsewhere already (via ExprKind::Closure).

prql-compiler/src/ast/pl/stmt.rs
max-sixty: > You don't even need to add `ExprKind::FuncDef` and just parse into `ExprKind::Closure`, which can be renamed into `ExprKind::LambdaFunc` or just `ExprKind::Func`?

Super! That's a great train-of-thought :)

web/book/src/integrations/dbt.md
eitsupi, line 1: The version of PRQL used by dbt-prql is old and I am not sure if this file should be updated.
Or it may be better to simply delete it.

web/website/content/posts/2022-05-19-examples.md
eitsupi, line 1: Should we update past blog posts?
I don't think these queries have been tested to see if they are still viable today.

aljazerzen: That's true, we should not be testing old queries.

web/book/src/integrations/dbt.md
aljazerzen: We can update dbt-prql version, right?

@max-sixty what are your plans on that?

eitsupi: > We can update dbt-prql version, right?

Unfortunately, I don't think it is possible to update dbt-prql beyond PRQL 0.6.0. (#2104)

In reviewing #2104, I noticed that @max-sixty had indicated to deleting the page.

max-sixty: Yeah, unfortunately after lots of work, I don't think `dbt-prql` is viable.

While the extreme hackery to monkey-patch python actually works, we just can't get into the correct place in the dbt compilation pipeline, and so can't support dialects which use backticks (!).

So yes, I think we can remove this as a recommended integration. And then doing this properly is a possible approach from #2559

max-sixty:
@aljazerzen awesome!

Do you want me to take a pass at cleaning it up a bit, or will that cause merge conflicts on your end?

FWIW I would vote to change the website posts unless the content specifically requires the old syntax. Otherwise it's more confusing to have incorrect code out there. An instance we would want old code is "Why we changed the syntax from X to Y", we wouldn't want to update either X or Y, but I don't think we have that here. 

If folks agree I can change those too.

aljazerzen:
Since I reviewed and updated this PR, I've based a lot of my work on in, with care not to touch things that you'd probably come back to clean-up. But it's easier to merge this now and you do the clean-up from main, so I went ahead and merged it anyway.

In short, yes go ahead clean it up, but do it based on the main branch.

Regarding posts: I think it's ok of they are a bit outdated - they are a remnant of the past and not something we have committed to keep up-to-date forever. But I see your point, we should at least have a disclaimer saying "language has changed a bit since this post".

PR #2577, opened 2023-05-16
feat: `into` variable declaration

### LINKED ISSUE #2427: Idea for more PRQLish table creation grammar
### What's up?

Currently, I think the only way to create tables is

```prql
let TableA = (
   from data
   ...
)
```

But it might be useful for a number of reasons to allow some pipeline operator for creating tables like this:
```
from data
...
let TableA = _
```

The point would be to allow components (table definition) to avoid nesting structures that basically block the ability to simply use concat/append as the base operator for constructing pipelines.

So someone might define some vanilla PRQL pipeline and you could then take that and simply extend it more more pipeline logic OR turn it into a table simply by + "let TableA" which sounds like not a big deal but removes a lot of complexity that leaks out of the components when you need to keep track of the type of the query object in order to combine with other query objects.

## UPDATE:

And to be clear, the intention is to enable the following
```
from data
...
let TableA = _

from data
...
join TableA [==something]
...
```

aljazerzen:
Based on #2575

Closes https://github.com/PRQL/prql/issues/2427

### DIFF
--- a/web/book/src/syntax/keywords.md
+++ b/web/book/src/syntax/keywords.md
@@ -2,10 +2,12 @@
 
 At the moment, PRQL uses only four keywords:
 
-- `prql`
 - `let`
-- `func`
+- `into`
 - `case`
+- `prql`
+- `type`
+- `module`
 
 To use these names as columns or relations, use backticks: `` `case` ``.
 
--- a/prql-compiler/prqlc/src/cli.rs
+++ b/prql-compiler/prqlc/src/cli.rs
@@ -595,15 +595,17 @@ group a_column (take 10 | sort b_column | derive [the_number = rank, last = lag 
         //   operations but are always inserted for function calls)
         assert_snapshot!(String::from_utf8(output).unwrap().trim(),
         @r###"
-        from table.subdivision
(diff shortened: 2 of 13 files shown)


web/book/src/syntax/keywords.md
```diff
@@ -2,10 +2,12 @@
 
 At the moment, PRQL uses only four keywords:
```
max-sixty, line 3:
```suggestion
At the moment, PRQL uses only six keywords:
```

max-sixty:
Awesome! Thanks a lot for taking this further. 

I left it in a somewhat messy state — `FuncDef_` with the underscore, lots of things commented out, let me know if I can help clear it up.

aljazerzen:
Don't worry about `FuncDef_`, this PR contains commits from #2575 and will be rebased after that is merged.

web/book/src/syntax/keywords.md
max-sixty: Still six! Will change

max-sixty:
Amazing, great work! I can write some docs...

PR #2578, opened 2023-05-16
ci: add a job to release prqlc binary

### LINKED ISSUE #1910: Use cargo-dist to create binaries as release assets

eitsupi:
Close #1910

### DIFF
--- a/.github/actions/build-prqlc/action.yaml
+++ b/.github/actions/build-prqlc/action.yaml
@@ -50,7 +50,7 @@ runs:
         cd target/${{ matrix.target }}/release
         tar czf "../../../${ARTIFACT_NAME}" prqlc
 
-    - name: Create artifact for Linux and macOS
+    - name: Create artifact for Windows
       shell: bash
       if: runner.os == 'Windows'
       run: |
--- a/.github/actions/build-prqlc/action.yaml
+++ b/.github/actions/build-prqlc/action.yaml
@@ -5,6 +5,10 @@ inputs:
   target:
     description: Build target
     required: true
+outputs:
+  artifact-name:
+    description: The name of the artifact
+    value: ${{ steps.echo-artifact-name.outputs.artifact-name }}
 
 runs:
   using: composite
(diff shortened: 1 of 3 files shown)


.github/workflows/release.yaml
```diff
@@ -24,6 +24,49 @@ jobs:
...
+    steps:
+      - name: 📂 Checkout code
+        uses: actions/checkout@v3
+      - uses: ./.github/actions/build-prqlc
+        with:
+          target: ${{ matrix.target }}
+      - name: Create artifact for Linux and macOS
+        if: runner.os != 'Windows'
+        run: |
+          echo "ARTIFACT_NAME=${{ env.ARTIFACT_BASE_NAME }}.tar.gz" >>"$GITHUB_ENV"
```
max-sixty: We could encapsulate the name if the `action.yaml` had it as an output: https://docs.github.com/en/actions/creating-actions/metadata-syntax-for-github-actions#example-declaring-outputs-for-composite-actions

max-sixty (approved):
Really great — I hadn't realized we could append them on. So we can do this without changing our release process!

When this merges I could release 0.8.2 to test it...

.github/workflows/release.yaml
eitsupi: Thanks!

Sorry, but I ran it on my fork and it is clear that this does not work. I will fix it based on your suggestion.

image

The next version release should be 0.9.0 because of the breaking changes and I think we need to resolve #2581 before the release, I think.
(I will of course work on this PR so this can be merged before release)

eitsupi:
It seems working fine on my fork!

First, make the release state from pre-release to release.

image

Then, artifacts that have been built are added successively.

image

image

image

max-sixty:
> It seems working fine on my fork!

Looking great!

.github/actions/build-prqlc/action.yaml
```diff
@@ -37,10 +41,30 @@ runs:
...
+        tar czf "../../../${ARTIFACT_NAME}" prqlc
+
+    - name: Create artifact for Windows
+      shell: bash
+      if: runner.os == 'Windows'
+      run: |
+        export ARTIFACT_NAME="prqlc-v${{ github.ref_type == 'tag' && github.ref_name || 0 }}-${{ matrix.target }}.zip"
+        echo "ARTIFACT_NAME=${ARTIFACT_NAME}" >>"$GITHUB_ENV"
+        cd target/${{ matrix.target }}/release
+        7z a "../../../${ARTIFACT_NAME}" prqlc.exe
```
max-sixty, line 60: Great impl!

PR #2579, opened 2023-05-16
docs: Remove note from docs

max-sixty:
This now works! Added as a test...

(It's quite nice when we do some internal work and we get a simpler language as a result!)

### DIFF
--- a/prql-compiler/src/tests/test.rs
+++ b/prql-compiler/src/tests/test.rs
@@ -2166,7 +2166,20 @@ fn test_table_alias() {
       LEFT JOIN salaries ON salaries.emp_no = e.emp_no
     GROUP BY
       e.emp_no
-    "###)
+    "###);
+
+    assert_display_snapshot!((compile(r###"
+    from e=employees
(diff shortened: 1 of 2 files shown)


prql-compiler/src/tests/test.rs
```diff
@@ -2166,7 +2166,20 @@ fn test_table_alias() {
       LEFT JOIN salaries ON salaries.emp_no = e.emp_no
     GROUP BY
       e.emp_no
-    "###)
+    "###);
+
+    assert_display_snapshot!((compile(r###"
+    from e=employees
```
eitsupi, line 2172:
```suggestion
    from e = employees
```

max-sixty: I've been doing it the `e=employees` way! I'm fine standardizing on either. 

I think the reason I had sided with the `e=employees` approach is that we don't have commas between each item, so having items contiguous helps to visually group them.

(I'll merge this now but totally open to doing whatever the consensus is)

PR #2582, opened 2023-05-17
ci(prql-python): add linux arm64 wheel and source wheel builds

### LINKED ISSUE #2457: prql-python arm64 linux wheel
### What's up?

Currently prql-python is not available via PyPI for arm64 Linux.

It would be great if we could also do binary releases for arm64 by setting up a workflow like python-polars.
https://github.com/pola-rs/polars/blob/main/.github/workflows/create-python-release.yml

eitsupi:
Close #2457

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -122,6 +122,8 @@ jobs:
       - name: 📂 Checkout code
         uses: actions/checkout@v3
       - uses: ./.github/actions/build-prql-python
+        with:
+          target: ${{ matrix.target }}
 
   publish-prql-python:
     runs-on: ubuntu-latest
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -108,8 +108,16 @@ jobs:
   build-python-wheels:
     runs-on: ${{ matrix.os }}
     strategy:
+      fail-fast: true
       matrix:
-        os: [ubuntu-latest, macos-latest, windows-latest]
+        os: [macos-latest, windows-latest]
+        include:
+          - os: ubuntu-latest
+            target: x86_64
+          - os: ubuntu-latest
+            target: aarch64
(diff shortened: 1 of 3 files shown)


eitsupi:
@max-sixty Could you take a look at this?

max-sixty:
Sorry @eitsupi , I must have missed this. Thanks for pinging. Looking now

.github/workflows/release.yaml
```diff
@@ -108,12 +108,22 @@ jobs:
...
       matrix:
-        os: [ubuntu-latest, macos-latest, windows-latest]
+        os: [macos-latest, windows-latest]
+        include:
+          - os: ubuntu-latest
+            target: x86_64
+          - os: ubuntu-latest
+            target: aarch64
+          - os: ubuntu-latest
+            target: source
```
max-sixty, line 120: FWIW I'm not personally familiar with how these work — is `source` an alternative to `aarch64` — maybe wheels can either contain architecture-specific code or just the python source code? Great if so (and even if not, no need to explain it to me! Though if others are confused, adding a comment would be helpful)

max-sixty (approved):
Looks super! Thank you!

.github/workflows/release.yaml
eitsupi: This "source" is not actually a target, but a keyword to build the maturin source distribution. Please check https://github.com/eitsupi/prql/blob/9c4e52625d928e98e4e8897b9728d0bb3c9a44bc/.github/actions/build-prql-python/action.yaml#L17-L21

Also note that the built wheels are attached to the results of this PR's CI.
The source distribution is simply source code.

max-sixty: I see — this goes to an sdist, which is the source code. (I had seen how we were using it, just was less familiar with why the distinction from the python side). Thank you!

Issue #2124, opened 2023-03-12
Syntax for lists and arrays


aljazerzen:
This is continuation of discussion in https://github.com/PRQL/prql/pull/1964#discussion_r1125719903

As of the latest update to RFC about type system (see linked PR), PRQL will have two container types:

- tuples, which have a known number of entries that can have different types,

- arrays, which have an unknown number of entries all of the same type.

Currently PRQL has "lists" with same semantics as the ones described for tuples. To limit confusion in this thread, let's use terms "tuple", "array" and "list" as I described it here and be clear when proposing changes to the names.

With terminology out of the way, let's talk about the problem: syntax.

As I see it, there are two options:

- Option 1 (`{}` for arrays). Try to preserve current syntax. Use `[]` for tuples, `{}` for arrays,
- Option 2 (JSON-like). Use `{}` for tuples and `[]` for arrays.

max-sixty:
How are you thinking about using Arrays? Are they just the underlying datatype? Or they would have some function in the PRQL language itself?

Are you thinking a map / dict is also a tuple? i.e. "(JSON-like)"?

---

For me the central tradeoff is:
- Almost every language uses parentheses `()` for tuples (is there any that doesn't? Maybe this is the same question as map / dict above then)
- We already have way too much resting on parentheses! https://prql-lang.org/book/syntax.html#precedence-and-parentheses

aljazerzen:
> How are you thinking about using Arrays?

Primarily, we need them to be the building block of what a "relation" is.

In the future, we will probably want to support actual arrays as a [column data type](https://www.postgresql.org/docs/current/arrays.html).

> Are you thinking a map / dict is also a tuple? i.e. "(JSON-like)"?

No, a tuple has known number of entries at the compile time. Map / dict would be another primitive data structure, which would need its own syntax.

I don't see a use-case for maps now, but I do think that we should support them eventually. Which means it would be wise to think about their syntax now.

> For me the central tradeoff is:

Exactly. I'm leaning toward violating the "Almost every language uses parentheses", but I recognize that it's far from ideal.

max-sixty:
As of #2585, we have:
- `[tuple]`
- `{array}` (not used much, only just introduced)
- no struct / map / dict / record type (I'll use the term "record")

The main issue I see here is that these are very different from almost every other language, for both tuples & arrays! And there's an open question of what we can do when we want to represent a struct.

One hard-backward-incompatible change would be to do:
- `{tuple}`
- `[array]`
[...]

PR #2585, opened 2023-05-18
feat: arrays and type system consolidation

### LINKED ISSUE #286: Notation for creating sample data

### LINKED ISSUE #1787: Relation and row literals
(just opening a new issue as #286 has been closed and contains enough discussion about string functions)

```elm
[[first_name: "tim", hobby: "HI", age: 30]]
select [hobby, age]
```

This can be postponed until we formalize type system, because the syntax of relation literals will probably reflect syntax of type definitions.

aljazerzen:
This PR contains many small changes mostly in the frontend of the compiler:

- parsing arrays
- use arrays in the type system
- refactor: Ty::Function -> TypeExpr::Function
- refactor Ty -> TyKind
- refactor minor AST
- refactor TyKind::Infer -> None
- ty name
- refactor TyKind::Table -> Expr.lineage
- refactor TyExpr -> TyKind
- refactor ExprKind::Type -> ExprKind::Set
- refactor std.table -> std.relation
- make relation literal an array of lists
- revert ExprKind::Set -> ExprKind::Type
- array constants to relation literals

I've added a non-breaking syntax for arrays:

```elm
let a = {1, 2, 3}
```

They cannot be directly translated into SQL, but are used in the type system a lot (they are used to implement relations).

I've refactored the data structure that tracks how a relation is manipulated and renamed it from "Frame" into "Lineage".

And I've made sure that relations are actually just arrays of tuples:

```elm
let row1 = [a = 5, b = false]
let row2 = [a = 6, b = true]
let my_relation = {row1, row2}

my_relation | filter b == true
```

... which make great one-liners for constructing relations on-the-fly:

```elm
{[a = 5, b = false], [a = 6, b = true]}
filter b == true
```

Ref #2124 #1965 

Closes #1787

Closes #286

### DIFF
--- a/prql-compiler/prqlc/src/cli.rs
+++ b/prql-compiler/prqlc/src/cli.rs
@@ -9,7 +9,7 @@ use std::process::exit;
 use std::str::FromStr;
 
 use prql_compiler::semantic::{self, reporting::*};
-use prql_compiler::{ast::pl::Frame, pl_to_prql};
+use prql_compiler::{ast::pl::Lineage, pl_to_prql};
 use prql_compiler::{downcast, Options, Target};
 use prql_compiler::{pl_to_rq_tree, prql_to_pl, prql_to_pl_tree, rq_to_sql, FileTree, Span};
 
@@ -326,7 +326,7 @@ impl Command {
     }
 }
 
-fn combine_prql_and_frames(source: &str, frames: Vec<(Span, Frame)>) -> String {
+fn combine_prql_and_frames(source: &str, frames: Vec<(Span, Lineage)>) -> String {
     let source = Source::from(source);
     let lines = source.lines().collect_vec();
     let width = lines.iter().map(|l| l.len()).max().unwrap_or(0);
--- a/prql-compiler/src/ast/pl/expr.rs
+++ b/prql-compiler/src/ast/pl/expr.rs
@@ -3,6 +3,7 @@ use std::fmt::{Display, Write};
 
 use anyhow::{anyhow, Result};
 use enum_as_inner::EnumAsInner;
+use itertools::Itertools;
 use semver::VersionReq;
 
 use serde::{Deserialize, Serialize};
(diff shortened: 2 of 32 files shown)


aljazerzen:
Note: to actually translate arrays into SQL, we'd want to have wildly different implementations of common array-handling-operations for different dialects. I don't think we have proper infrastructure to do this now, we want wait for compiler to understand [representations](https://github.com/PRQL/prql/blob/6b109fb9b27dcdb9232514c41507b24b1baec0f6/web/book/src/internals/type-system.md#type-representations).

max-sixty:
Awesome velocity of changes!

My concern with the `{}` syntax for arrays is how we would represent a struct / record type? 

I do think this is a sticking point; there some chance we end up in a totally different place to other languages because of initial decisions.

(I'm responding quickly because of the pace here, but I'll look later for previous comments on this; I remember there being no easy answer)

max-sixty:
> ```elm
> {[a = 5, b = false], [a = 6, b = true]}
> filter b == true
> ```

This is really awesome, great work. Better than the `from_text` imo!

aljazerzen:
> My concern with the {} syntax for arrays is how we would represent a struct / record type?

We already use `[]` for a struct / record type - which we currently name list :D

max-sixty:
Docs at https://github.com/PRQL/prql/pull/2605

PR #2594, opened 2023-05-21
fix(prqlc): allow `prqlc compile | duckdb`

### LINKED ISSUE #2569: On Windows, I can't get out with `ctrl-d` from the input screen opened via `prqlc compile`
I tried prqlc on PowerShell 7 on Windows Terminal, and I realized that I can't quit from the `prqlc compile` screen.

eitsupi:
Fix comments displayed in interactive mode and make interactive mode output valid SQL output.
(It also fixes #2569)

This PR allows PRQL to be executed interactively by piping `prqlc compile` to `duckdb`, `sqlite3`, etc.

I have also added a note to the prqlc README on how to use the `prqlc compile` command, which I believe is the minimum required to try prqlc.
And this README also has been changed to appear in the book.


web/book/src/SUMMARY.md
```diff
@@ -57,6 +57,7 @@
 
 - [Bindings](./bindings/README.md)
 
+  - [CLI](./bindings/cli.md)
```
eitsupi: @max-sixty Is this the right place for this page?
Or should be "Integrations"?

eitsupi:
(By the way, I think a combination of prqlc and the DuckDB CLI seems more Unix-like and a better path than using the prql-query CLI with DuckDB built in......) (Related to PRQL/prql-query#22)

```sh
prqlc compile --target sql.duckdb | duckdb --json >output.json
```

prql-compiler/prqlc/README.md
```diff
@@ -32,3 +40,46 @@ FROM
 WHERE
   has_dog
 ```
+
+PRQL query can be executed with CLI tools working with SQL, such as
+[DuckDB CLI](https://duckdb.org/docs/api/cli.html).
+
+```sh
+$ curl -sL https://raw.githubusercontent.com/mwaskom/seaborn-data/master/penguins.csv -o penguins.csv
```
eitsupi: Simply I used the same file as below. Please replace with a more appropriate file.
https://github.com/PRQL/prql/blob/5c6491dfdd9542bbde9e5efbb12e5e81d860c956/web/book/src/integrations/jupyter.md?plain=1#L29

prql-compiler/prqlc/README.md
```diff
@@ -2,15 +2,19 @@
...
 
 ```sh
+# From crates.io
 cargo install prqlc
 ```
 
-...or built from source:
-
 ```sh
+# From GitHub
```
max-sixty: I think we do want to indicate that it's from local files — i.e. it's not installing from github.com. 

How about "from a local PRQL repo"?

prql-compiler/prqlc/README.md
```diff
@@ -2,15 +2,19 @@
 
 ## Installation
 
-`prqlc` can be installed with `cargo`:
+`prqlc` is a single, dependency-free binary that compiles PRQL into SQL.
+precompiled binaries are available for Linux, macOS, and Windows on the
+[PRQL release page](https://github.com/PRQL/prql/releases).
+
+If you have Rust toolchain installed, you can install `prqlc` via `cargo`:
```
max-sixty:
```suggestion
`prqlc` can be installed via `cargo`:
```

I'm trying to keep a similar voice throughout the docs, and one way to do that is to not use the 2nd person ("you")...

prql-compiler/prqlc/README.md
```diff
@@ -22,6 +26,10 @@ brew install prql/prql/prql-compiler
 
 ## Usage
 
+### prqlc compile
+
+This command is working as a filter to compile PRQL string into SQL string.
```
max-sixty, line 31:
```suggestion
This command compiles a PRQL string into a SQL string.
```
(I wasn't sure what "filter" means — feel free to have another pass if my change isn't what you were aiming for)

prql-compiler/prqlc/README.md
```diff
@@ -32,3 +40,46 @@ FROM
...
+$ prqlc compile
+```
+
+As with using it as a filter, you can pass the SQL string output to the DuckDB
+CLI, etc.
+
+```sh
+$ prqlc compile | duckdb
+from `penguins.csv`
+take 3
```
max-sixty, line 75: Hmmm, this doesn't work so well on mine — does it for you?

I think maybe we write the `Ctrl-d` to stdout?

```
prqlc compile | duckdb
from `penguins.csv`
take 3
Error: near line 1: Parser Error: syntax error at or near "Enter"
LINE 1: Enter PRQL, then ctrl-d:
        ^
```

(Is there a way of retaining the query? I worry this approach is OK but loses the query on each run. But it's still reasonable to show the example)

prql-compiler/prqlc/src/cli.rs
```diff
@@ -292,7 +292,7 @@ impl Command {
         // it's confusing whether it's waiting for input or not. This
         // offers the prompt.
         if input.is_stdin() && atty::is(atty::Stream::Stdin) {
-            println!("Enter PRQL, then ctrl-d:\n");
+            println!("# Enter PRQL, then press ctrl-d or ctrl-z (windows) to compile:\n");
```
max-sixty, line 295:
```suggestion
            eprintln!("# Enter PRQL, then press ctrl-d or ctrl-z (windows) to compile:\n");
```

...I think this will solve the issue above

max-sixty: FYI we could use this approach to print the correct string: https://github.com/PRQL/prql/blob/e32c5864e3ca04987849e2397d956a0c7a599447/prql-compiler/prqlc/src/cli.rs#L508

web/book/src/SUMMARY.md
max-sixty: I think it's probably more "integrations" — the bindings are for machines, the integrations are for people

max-sixty (approved):
Excellent, thanks a lot!

I added a couple of comments & suggestions, but otherwise good to go

max-sixty:
> (By the way, I think a combination of prqlc and the DuckDB CLI seems more Unix-like and a better path than using the prql-query CLI with DuckDB built in......)

I do think this is a great pattern, agree.

aljazerzen (approved):
Looks good.

prql-compiler/prqlc/README.md
eitsupi: It seems that the `prqlc` you are using was not built from this PR. (the prompt message is different)

prql-compiler/prqlc/README.md
eitsupi: I meant this. https://en.wikipedia.org/wiki/Filter_(software)#Unix
Makes sense?

prql-compiler/prqlc/README.md
eitsupi: > Is there a way of retaining the query?

I believe the query will remain visible in the terminal as shown in this README, and I have verified that it works with bash and fish on Linux and PowerShell on Windows.

eitsupi:
Thanks for your reviews!

I think I have addressed most of the comments, so I will merge this.
If additional updates are needed, I will open another PR.

prql-compiler/prqlc/README.md
max-sixty: > It seems that the `prqlc` you are using was not built from this PR. (the prompt message is different)

Sorry, you're correct, apologies

max-sixty: > I believe the query will remain visible in the terminal as shown in this README, and I have verified that it works with bash and fish on Linux and PowerShell on Windows.

Yes!

A total aside — something @snth and I discussed on the dev call was building a TUI, similar to the playground, where typing a query would pipe through to duckdb on each keystroke, could be a good way of exploring a dataset from the terminal

prql-compiler/prqlc/README.md
max-sixty: Interesting, I wasn't so familiar with the terminology, but you're right! TIL

PR #2597, opened 2023-05-22
docs: Add docs for `into`

max-sixty:
From #2577. We could probably market it more aggressively through the docs too


aljazerzen (approved):

snth:
Main discussion around this feature was in #2427 .

PR #2598, opened 2023-05-22
internal: Some cleanups from #2575

max-sixty:
- Cleaned up some of the code that I commented out
- Changed some of the historical posts to use the new syntax — which I think is appropriate — there is no use for the old syntax; and it'll be confusing. (If it were a post _about syntax_ that would be different)

### DIFF
--- a/prql-compiler/src/ast/pl/expr.rs
+++ b/prql-compiler/src/ast/pl/expr.rs
@@ -95,7 +95,7 @@ pub enum ExprKind {
 
     /// a placeholder for values provided after query is compiled
     Param(String),
-    FuncDef(FuncDef_),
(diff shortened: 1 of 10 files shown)


eitsupi:
> * Changed some of the historical posts to use the new syntax — which I think is appropriate — there is no use for the old syntax; and it'll be confusing. (If it were a post _about syntax_ that would be different)

Is it possible to display the update date on the posts?
If the update date is shown, it becomes predictable to the users that the latest PRQL code is included, but if not, it could be confusing to use syntax that should not exist at the post date.

eitsupi:
Maybe this useful?
https://gohugo.io/variables/git/#lastmod

PR #2603, opened 2023-05-22
fix(prqlc): Slightly better prompt

max-sixty:
Thanks to @eitsupi for #2594. Here are a couple of suggestions for small improvements to the prompt to flup.


eitsupi (approved):
Awesome! Thanks!

PR #2605, opened 2023-05-23
docs: Add docs for Relation literals

max-sixty:
I've combined it with the `from_text` page, as I think they're similar (and I put these first since I think they're generally better).

One question — do we want any keyword for them — e.g. `from_lit {...}`? I don't think we necessarily do, but wanted to raise it, since it's the only construction which can be directly in line now.


aljazerzen:
If you want, you can use:

```
from {[a = 3]}
select a
```
... because `from` is already just a noop function. Its only utility is "pulling from database tables" instead of normal name resolution.

I'd say let's not add too much unnecessary functions here.

web/book/src/language-features/relation-literals.md
```diff
@@ -0,0 +1,96 @@
+# Inline tables
```
aljazerzen:
```suggestion
# Inline relations
```

web/book/src/language-features/relation-literals.md
```diff
@@ -0,0 +1,96 @@
...
+
+It's often useful to make a small table inline, for example when exploring how a
```
aljazerzen:
```suggestion
It's often useful to make a small inline relation, for example when exploring how a
```

web/book/src/language-features/relation-literals.md
```diff
@@ -0,0 +1,96 @@
...
+{[a = 5, b = false], [a = 6, b = true]}
+filter b == true
+select a
+```
+
+```prql no-fmt
+let my_artists = {
+  [artist="Miles Davis"],
+  [artist="Marvin Gaye"],
+  [artist="James Brown"]
```
aljazerzen:
```suggestion
  [artist="James Brown"],
```

Let's cultivate ideomatic prql with a trailing comma.

aljazerzen (approved):

max-sixty:
> If you want, you can use:
> 
> ```
> from {[a = 3]}
> select a
> ```
> 
> ... because `from` is already just a noop function. Its only utility is "pulling from database tables" instead of normal name resolution.

I'm going back & forth on this!
- `from` seems incongruent applied to arrays — they're not table names — and `from` brings tables into the query.
- But having it bare also seems out-of-place, (image below) — there's no transform bringing the array into the query....
- But then we introduce another transform `from_lit`?! 

image

---

FWIW having `from` was a question we went back & forth on in the very early days of PRQL — I originally just had `artists`. I think we were correct to make the change — queries look more cohesive with `from artists` than just `artists`.

Any thoughts from others? 

Otherwise I guess we go with the existing state and see how we feel... I don't have much beyond aesthetics here, so we shouldn't wait at all.

---

FWIW `let` makes it look more reasonable (but we obv can't have different semantics inside & outside a declaration)

image

eitsupi:
My preference is starting with `from`. (The word "lit" does not seem very clear to me.)

Given things like `from foo.csv` and `from df` (df is pandas.DataFame on Python) in DuckDB, I wouldn't mind putting something other than table name after `from`.

max-sixty:
> My preference is starting with `from`. (The word "lit" does not seem very clear to me.)
> 
> Given things like `from foo.csv` and `from df` (df is pandas.DataFame on Python) in DuckDB, I wouldn't mind putting something other than table name after `from`.

This resonates with me...

max-sixty:
Merging this; with a note on us considering `from`...

PR #2609, opened 2023-05-23
internal: Remove BQ special-case

max-sixty:
Because of the new module system, we don't need a special-case for BigQuery any longer!

(At least, I can't find how we'd need it, but I might be wrong — lmk if you suspect I might be...)


aljazerzen (approved):
Yeah!

PR #2611, opened 2023-05-23
feat(prqlc): add the `prqlc shell-completion` command to generate shell completions

eitsupi:


prql-compiler/prqlc/tests/test.rs
```diff
@@ -75,3 +76,1145 @@ fn test_compile() {
     ----- stderr -----
     "###);
 }
+
+#[test]
+fn test_shell_completion() {
+    assert_cmd_snapshot!(Command::new(get_cargo_bin("prqlc")).arg("shell-completion").arg("bash"), @r###"
```
max-sixty: Possibly given the length of this, we don't put inline? If we remove the `@""`, insta makes a snapshot file.

max-sixty (approved):
Awesome!! Very cool.

One idea is to put the instructions into `--help` rather / in addition to the docs, so they're closer to their usage. But not sure how viable that is.

And a separate follow-up might be to add this to the brew recipe so it gets added automatically.

eitsupi:
Thanks for your review!

I think we can improve this more in the future as like your idea, but for now I would like to merge this and allow others to work on that.

prql-compiler/prqlc/tests/test.rs
eitsupi: Thanks, I didn't know that!

PR #2612, opened 2023-05-23
feat!: swap array and tuple syntax

### LINKED ISSUE #2124: Syntax for lists and arrays

aljazerzen:
Closes #2124

TODO:

- [x] Rename list to tuple in the docs
- [x] Fix all the remaining tests

If anyone is keen on writing docs, please help with the effort :D


prql-compiler/examples/compile-files/queries/variables.prql
```diff
@@ -1,20 +1,20 @@
...
+derive {                                     # This adds columns / variables.
   gross_salary = salary + payroll_tax,
   gross_cost = gross_salary + benefits_cost  # Variables can use other variables.
-]
+}
 filter gross_cost > 0
-group [title, country] (                     # For each group use a nested pipeline
-  aggregate [                                # Aggregate each group to a single row
+group {title, country} (                     # For each group use a nested pipeline
+  aggregate {                                # Aggregate each group to a single ro
```
eitsupi:
```suggestion
  aggregate {                                # Aggregate each group to a single row
```

Seems deleted by accident.

max-sixty:
I looked through all the `no-eval` blocks, which are the ones least likely to be checked

web/book/src/syntax/tuples.md
```diff
@@ -0,0 +1,30 @@
...
+  a = x,
+  b = y                # Optional trailing comma
+}
+select {
+  c,                   # Individual item
+  d = b,               # Assignment
+}
+```
+
+Most transforms can take either a list or a single item, so these are
```
eitsupi:
```suggestion
Most transforms can take either a tuple or a single item, so these are
```

Maybe?

web/book/src/syntax/tuples.md
```diff
@@ -28,3 +28,7 @@ select {first_name}
 from employees
 select first_name
 ```
+
+```admonish note
+Tuples were previously named Lists, and represented with `[]` syntax. There may still be references to the old naming.
```
eitsupi: I think it would be helpful to specify that this has changed in version 0.9.0.

max-sixty:
As I was writing the Changelog — one quick question — I realize it's too late for this but also it's going to be much too late soon.

*Could we actually do this with parentheses?*

Would the Python rule work — "A tuple has parentheses with commas between items; otherwise the parentheses are just precedence"?

I'm hesitant to further load parentheses with syntax _but_ if this works, then it'll be consistent with other langs, which would be highly preferable, given that we're naming them tuples.

(Sorry in advance for the last minute point! And no need to delay merging this — we can merge now to avoid conflicts — even if the parentheses idea does have legs and we come back to it)

max-sixty:
@aljazerzen I know you prefer rebases, but this was messy, hopefully the squash is satisfactory...

aljazerzen:
> Could we actually do this with parentheses?

That would be nice. One parenthesis to rule them all.

I think it's possible and actually practical, because we can merge the two constructs a bit:

```
(1 | sqrt, 2.3 | floor, 4)
```

The only downside it this non-linearity:

```
()        -> tuple with 0 elements
(x)       -> just x, no tuple
(x, y)    -> tuple with 2 elements
(x, y, z) -> tuple with 3 elements
```

It can be solved by trailing commas:

```
(,)        -> tuple with 0 elements
(x,)       -> tuple with 1 element
(x, y,)    -> tuple with 2 elements
(x, y, z,) -> tuple with 3 elements
```

But other languages have this problem too, see Python and Rust for example.

I don't think having "{tuple}" is that much of a downside, because we can say "think of an JavaScript object or a C++ struct, but named tuple".

max-sixty:
Yes great, thanks, the parallel examples are excellent.

So it's at least possible... I would summarize the tradeoff as:
- Parentheses are much more familiar
- ~Brackets~ Braces are more distinct, giving a less fragile language with better error messages

Re the familiarity — this concern weighs higher for me:

> I don't think having "{tuple}" is that much of a downside, because we can say "think of an JavaScript object or a C++ struct, but named tuple".

My model is that folks are quick to dismiss things that are unfamiliar. Indeed — if we go with braces, possibly that weighs on calling them Records rather than Tuples (but also I think there's a good theoretical background with Tuples).

Re "fragile language" — while everything is expressible in both, overusing one form of syntax gives us a more fragile language — a _slightly_ incorrect query doesn't converge to a correct query, because it's unclear what the intention is.

For example, in python I've sometimes mistakenly left a comma somewhere:

```python
foo = (1, 5, func(2 + 3))),

foo[0] + 1   # "This should be 1 + 1, he thought

---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
Cell In[8], line 1
----> 1 foo[0] + 1

TypeError: can only concatenate tuple (not "int") to tuple
```

...and then trying to work out what that means

A couple of approaches to mitigate this:
- Rust gets around this by forbidding implicit tuples — they always require parentheses.
- It would be possible to go further by _forcing_ a comma — an empty tuple would be `(,)`, like your example, so adding or removing parentheses can never create different objects (I think)

WDYT? What about others @snth @eitsupi ?

I guess this is something we shouldn't go back & forth on after we release, if at all possible...

eitsupi:
If I understand correctly, are you saying that the parentheses have two meanings?

I'm concerned that the syntax of something like `group` might be confusing.

max-sixty:
Yes, here's an example:

Parentheses:
```elm
group (e.emp_no, e.gender) (
  aggregate (
    emp_salary = average salaries.salary
  )
)
```

- `(e.emp_no, e.gender)` is a tuple
- The other two items are not tuples (we can call them scalars)
  - The `aggregate` must be a scalar
  - (more complex case) FWIW the `(emp_salary = average salaries.salary)` would be semantically equivalent[^1] (but not syntactically equivalent) as a tuple or a scalar, because `aggregate` can take either:

[^1]: all (almost all?) of our transforms would be semantically equivalent, since they can take either a 1-tuple or a scalar — e.g. `select foo` & `select [foo]`, using current syntax

```diff
group (e.emp_no, e.gender) (
  aggregate (
-    emp_salary = average salaries.salary
+    emp_salary = average salaries.salary,
  )
)
```

---

With braces:


```diff
- group (e.emp_no, e.gender) (
+ group {e.emp_no, e.gender} (
  aggregate (
    emp_salary = average salaries.salary
  )
)
```

It's less fragile on the ambiguity — to change from a scalar to a tuple requires actually adding braces:

```diff
group {e.emp_no, e.gender} (
-  aggregate (
+  aggregate {
    emp_salary = average salaries.salary
-  )
+ }
)
```

...and whether there's a comma after `salary` doesn't affect the syntax of the expression.

max-sixty:
Unless someone thinks "yes parentheses could be *great*", I would vote we should merge this, with braces...

While the parentheses proposal is cool, I'm worried it's too ambitious and puts so much weight on the familiarity point. 

The familiarity is almost the only thing going for it, and I worry a bit that it's easy to appreciate now, whereas the downsides — error messages and fragility — will become more salient in the future.

snth:
Hi, weighing in with some half baked ideas just because the discussion and changes seem to be happening so fast. (I've only read the discussion so far and have not looked at the contents of the PR yet.)

What was the original motivation for calling these tuples rather than records or structs?

I got brought up to speed by @max-sixty on the dev call and from what I understood it seemed to me to be safe to translate "tuples" internally (in my mind) to records and that's what I went with. I like the braces `{}` for records. 

Not a fan of what I've seen of the parentheses proposal so far. I agree that parentheses are more familiar for tuples but are these really tuples or are they records?

aljazerzen:
> What was the original motivation for calling these tuples rather than records or structs?

They are called tuples in relational algebra.

> are these really tuples or are they records?

Well, when I tried to formulate what would be the difference between records, stucts and tuples, I realized that they are the same thing. Try forgetting the syntax from other languages and think about what actually *are* tuples and records, and you'll see that they are conceptually equivalent.

snth:
I just skimmed over the PR and the examples and I really like the new `{}` syntax! I think it looks actually really beautiful (in the mathematical sense) - I think it really highlights where you are constructing **ordered** tuples (which answers my previous question) in a way that wasn't obvious with the previous `[]` syntax.

One thing that also pops out from this (to me at least) is that `join` should not take `{}` as an argument as @aljazerzen already pointed out in https://github.com/PRQL/prql/issues/2124#issuecomment-1559079527 because there is no ordering in the join conditions as `&&` is commutative. It also gives preferential status to `&&` over `||` which seems unjustified.

richb-hanover:
Excellent! I get to teach my brain a new way of working! :-) From a practical standpoint:

1. In reading PRQL queries, do I simply say "tuple" where I used to say "list" or "stuff between square brackets"?
2. In converting existing queries, is `s/[/{/g` and s`/]/}/g` sufficient?
3. Is there anything else?

Thanks.

max-sixty:
> In converting existing queries, is `s/[/{/g` and s`/]/}/g` sufficient?

Yes!

> In reading PRQL queries, do I simply say "tuple" where I used to say "list" or "stuff between square brackets"?

It's a good question, I think we're still deciding. We like the theoretical basis of "tuple", but it's a bit techie...

max-sixty:
> One thing that also pops out from this (to me at least) is that `join` should not take `{}` as an argument as @aljazerzen already pointed out in [#2124 (comment)](https://github.com/PRQL/prql/issues/2124#issuecomment-1559079527) because there is no ordering in the join conditions as `&&` is commutative. It also gives preferential status to `&&` over `||` which seems unjustified.

I'm OK with this — it could take a single item (though `derive` could technically take a single item...)

Issue #2599, opened 2023-05-22
Release to Windows Package Manager Community Repository (and set CI to release automatically)


eitsupi:
Previous Discussions #2043

Thanks to #2578, after the next release, we will be able to register a manifest to https://github.com/microsoft/winget-pkgs and install prqlc.exe via `winget install` command.

Once the initial release on the repo is complete, I believe the followings could be used to send the manifests at release time automatically.
https://github.com/vedantmgoyal2009/winget-releaser
https://github.com/microsoft/winget-create

eitsupi:
(I recently noticed that Task, which are also used in this repository, can be installed via winget.)

```sh
❯ winget search task.task
名前 ID        バージョン ソース
---------------------------------
Task Task.Task 3.24.0     winget
```

https://github.com/microsoft/winget-pkgs/blob/6f8ab54231b14bd1528957bba14e29814c0818fb/manifests/t/Task/Task/3.24.0/Task.Task.installer.yaml

eitsupi:
microsoft/winget-pkgs#113133

eitsupi:
After ~~microsoft/winget-pkgs#113133~~ microsoft/winget-pkgs#114768 is merged, we can use https://github.com/vedantmgoyal2009/winget-releaser to release automatically.

vanillajonathan:
Chocolatey might also be of interest.
https://docs.chocolatey.org/en-us/create/create-packages

eitsupi:
Blocked by #3080

eitsupi:
To automate the release of winget using Komac/winget-releaser, a Personal Access Token must be created and a PR must be created by a GitHub account. (like microsoft/winget-pkgs#112690)
https://github.com/vedantmgoyal2009/winget-releaser#token-token

@max-sixty Do you think it is possible to create a PAT of @prql-bot and store it in the secret of this repository?

max-sixty:
> @max-sixty Do you think it is possible to create a PAT of @prql-bot and store it in the secret of this repository?

Definitely! Lmk if you want me to do anything...

PR #2614, opened 2023-05-23
docs(prqlc): add packaging status badge and minor tweaks

eitsupi:
It is currently installable via Homebrew and will be installable via winget in the future (#2599), so the badge would be helpful.


max-sixty (approved):
Excellent!

PR #2618, opened 2023-05-23
docs: Reword "Ordering guarantees" in Book

max-sixty:
Made a few changes:
- Tried to lead with what it means for the user
- Kept the theory but moved it down
- Tried to cut words where possible
- Changed the SQL example to one that DBs will actually get "wrong" rather than just an theoretical "SQL standard" one


web/book/src/transforms/sort.md
```diff
@@ -42,39 +42,75 @@ sort [s"substr({first_name}, 2, 5)"]
...
+Here, PRQL pushes the `sort` down the pipeline, compiling the `ORDER BY` to the
+_end_ of the query. Consequently, most relation transforms retain the row order.
 
-For example, this query:
+The explicit semantics are:
+
+- `sort` introduces a new order,
+- `group` resets the order,
+- `join` maintains the order of the left relation,
+- database tables don't have a defined order.
```
aljazerzen: Well, they do. Postgres will return the same order of rows, no matter how many times you execute a query (if you are not modifying the table in any way).

It's just that order is unknown.

(I'm a bit pedantic here, but people reading "semantics" are usually like that)

web/book/src/transforms/sort.md
```diff
@@ -42,39 +42,75 @@ sort [s"substr({first_name}, 2, 5)"]
...
+
+Comparatively, in SQL, relations possess no order, being orderable solely within
+the context of the query result, CTE, or window functions. The lack of inherent
```
aljazerzen:
```suggestion
the context of the query result, `TAKE` clause, or window functions. The lack of inherent
```

Relations of a CTEs do don't retain order. If they have ORDER BY, strictly speaking, it has no meaning.

aljazerzen (approved):
Much more readable!

web/book/src/transforms/sort.md
max-sixty: Pedantic is good here! I switched it

web/book/src/transforms/sort.md
max-sixty: This is "`LIMIT` statement"? Or it's something else?

web/book/src/transforms/sort.md
max-sixty: Pedantic is good here! I switched it

web/book/src/transforms/sort.md
aljazerzen: Oh yes of course, LIMIT, not take. Brainfart.

PR #2621, opened 2023-05-24
build: Bump toolchain, test `prqlc` for `msrv`

max-sixty:
`rust-analyzer` is now raising mistaken errors for 1.65, as they only support the latest version https://github.com/rust-lang/rust-analyzer/issues/12751#issuecomment-1182887933

We don't want to bump the required version because of https://github.com/PRQL/prql/pull/1561, but I think this approach:
- Lets us work on an updated version
- Tests `prql-compiler` & `prqlc` to ensure they don't fail to support 1.65
- Doesn't let us use any new features in `prql-compiler` or `prqlc` until we bump the required version, but that's completely fine (is there even _anything_ we'd use?)

### DIFF
--- a/.github/workflows/test-all.yaml
+++ b/.github/workflows/test-all.yaml
@@ -144,9 +144,12 @@ jobs:
         with:
           crate: cargo-msrv
         # Note this currently uses a manually maintained key in
-        # `prql-compiler/Cargo.toml`, because of
+        # `prql-compiler/Cargo.toml` (and `prql-compiler/prqlc/Cargo.toml` below), because of
         # https://github.com/foresterre/cargo-msrv/issues/590
-      - name: Verify minimum rust version
+      - name: Verify minimum rust version — prql-compiler
         # Ideally we'd check all crates, ref https://github.com/foresterre/cargo-msrv/issues/295
         working-directory: prql-compiler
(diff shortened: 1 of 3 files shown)


eitsupi (approved):
Makes sense, thanks!

.github/workflows/test-all.yaml
```diff
@@ -144,9 +144,12 @@ jobs:
...
-        # `prql-compiler/Cargo.toml`, because of
+        # `prql-compiler/Cargo.toml` (and `prql-compiler/prqlc/Cargo.toml` below), because of
         # https://github.com/foresterre/cargo-msrv/issues/590
-      - name: Verify minimum rust version
+      - name: Verify minimum rust version — prql-compiler
         # Ideally we'd check all crates, ref https://github.com/foresterre/cargo-msrv/issues/295
         working-directory: prql-compiler
         run: cargo msrv verify
+      - name: Verify minimum rust version — prqlc
+        working-directory: prqlc
```
eitsupi:
```suggestion
        working-directory: prql-compiler/prqlc
```

aljazerzen (approved):

aljazerzen:
> is there even anything we'd use?

No, I've checked.

PR #2624, opened 2023-05-24
docs: remove the root Dockerfile in favor of pre-built devcontainer image, and some development docs improvements

eitsupi:
The pre-built image seems to work well, so I think we can remove the previous Dockerfile with duplicate roles.
I have also simplified and slightly improved the development documentation accordingly.

I could have written some tips for running the PRQL Book using only `docker` commands with the pre-built image, as in the previous developing-with-docker page, but I removed them because they did not seem very necessary.

@richb-hanover Could you take a look at this?


richb-hanover:
I agree it is interesting to remove the Dockerfile - the Dev Container works great. I am jammed up right now and will be away this weekend, so I won't be able to review these changes for a week or so. 

In parallel, I have been working on a minor change to the Taskfile.yml to use "install-npm-depencies-if-needed" task_ to avoid re-running `npm install...` each time I run `task run-playground`. 

I would appreciate another set of eyes on it: https://github.com/richb-hanover/prql/tree/cache-npm

eitsupi:
Thanks for your reply, and no problem. We can can be revert at any time, even after this PR merged...

> In parallel, I have been working on a minor change to the Taskfile.yml to create a Update Taskfile.yml to use "install-npm-depencies-if-needed" task to avoid re-running npm install... each time I run task run-playground.
>
> I would appreciate another set of eyes on it: richb-hanover/prql@cache-npm

Great. I think it is absolutely worthwhile to do that, as the tasks on the Taskfile in this repository contain very few status checks.

As a side note, it should be possible to ignore status checks with the `--force` option, so there may be no need to separate tasks.

max-sixty:
Yes I think this probably makes sense. 

The original Dockerfile served us well, thank you @richb-hanover. It remains in our git history and PRQL's history, even if not in the file tree!

max-sixty (approved):

PR #2635, opened 2023-05-26
ci: Remove nightly docker test

max-sixty:


.github/workflows/nightly.yaml
```diff
@@ -92,31 +92,34 @@ jobs:
...
-
-      # https://aschmelyun.com/blog/using-docker-run-inside-of-github-actions/
-      - name: Test
-        uses: addnab/docker-run-action@v3
-        with:
-          image: prql:latest
-          options: -v ${{ github.workspace }}/:/src
-          run: task test-rust
+  # We now use the devcontainer. TODO: is it possible to have a similar test for
+  # that? Or that would require VSCode to install the dependencies?
```
max-sixty, line 96: @eitsupi would you have any thoughts here?

No great stress. But I generally find that if we don't test something, it breaks, and the further away from the cause it breaks, the more difficult it is to diagnose!

max-sixty:
Merging with the comment-ed out part and question, so nightly tests pass

.github/workflows/nightly.yaml
eitsupi: I think we can use the `docker run` command as before or use the devcontainer CLI to run JS and Python tasks in addition to Rust.
However, these take time to execute, so we need to think carefully about which ones to test.

max-sixty: What tests could we run?

Because we build the base image in CI, but other commands — like `task` — are only built when builds the devcontainer in VSCode.

Ideally we would run `task test-rust`, like we did in this test.

max-sixty: I don't think we need to worry much about runtime, since this only runs once a day

PR #2643, opened 2023-05-26
refactor: AST cleanup

aljazerzen:


prql-compiler/src/semantic/resolver.rs
```diff
@@ -631,8 +655,27 @@ impl Resolver {
...
+                    ty: self.fold_ty_or_expr(p.ty)?,
+                    ..p
+                })
+            })
+            .try_collect()?;
+        closure.return_ty = self.fold_ty_or_expr(closure.return_ty)?;
+        Ok(closure)
+    }
+
+    fn cast_built_in_function(&mut self, closure: Func) -> Result<Result<Expr, Func>> {
```
max-sixty, line 673: If we can add notes to signatures with things like double `Result`s, that would be great — I do think signatures are a convenient abstraction, and when those are ambiguous it becomes harder to understand (I realize it's not a new func though, and we should balance the docstrings against slowing you down...)

max-sixty (commented):
Looks good!

PR #2644, opened 2023-05-26
feat!: tuple type checking and remove tuple coersion for join

### LINKED ISSUE #2641: Join tuple equivalence
### What's up?

Currently, the second arg of join takes either:
```
# - scalars
join rel cond1

# - tuples
join rel {cond1, cond2, cond3}
```

In #469 we changed `filter` not to accept tuples and force the user to either use `&&` or multiple filters.

The reasoning for this change is based on a move away from implicit `&&` that are inserted into the tuple to convert it into a boolean expression. One may incorrectly assume the tuple means "any of the conditions" and not "all of the conditions".

@snth left [a comment about this here](https://github.com/PRQL/prql/issues/2124#issuecomment-1563462793).

aljazerzen:
Mostly deals with type checking tuples in-depth and also closes #2641


CHANGELOG.md
```diff
@@ -697,7 +697,7 @@ We've had to make some modest breaking changes for 0.3:
   ```diff
   from employees
   -join positions [id]
-  +join positions {==id}
+  +join positions [==id]
```
max-sixty, line 700: Should this be ()?

prql-compiler/src/semantic/std.prql
```diff
@@ -33,6 +30,7 @@ type array = [null]
 
 # Scalar
 type scalar = int || float || bool || text || date || time || timestamp || null
+type scalar_tuple = {scalar..}
```
max-sixty, line 33: Do I understand correctly this this is "1 dimensional", i.e. no nesting?

web/book/tests/snapshot.rs
```diff
@@ -23,54 +23,50 @@ use walkdir::WalkDir;
...
+    for Example { name, tags, prql } in examples {
+        let result = compile(&prql, &Options::default().no_signature());
+        let should_succeed = !tags.contains(&LangTag::Error);
+
+        match (should_succeed, result) {
+            (true, Err(e)) => errs.push(format!(
+                "
+---- {name} ---- ERROR
+  Use `prql error` as the language label to assert an error compiling the PRQL.
+  -- Original PRQL --
```
max-sixty, line 38: Thanks, this is better!

max-sixty (commented):
Great!

CHANGELOG.md
aljazerzen: This is a revert of a changelog update.

max-sixty: OIC, yes you're correct ofc

prql-compiler/src/semantic/std.prql
aljazerzen: This name is not ideal - it should be tuple_of_scalars.

I've added `{type..}` as notation for tuple of many `type`s. It was needed to throw an error when a function was passed instead of evaluating to a column.

PR #2648, opened 2023-05-28
docs(book): add the `compiler` section

eitsupi:
- Change the title of prqlc's README to "PRQL compiler CLI". (Related to #2599, I think it would be easier to understand if changing the package name like this. In other words, when we see something like `prqlc` and `prqlr` side by side, it is difficult to guess that `prqlc` is a CLI.)
- Add the `compiler` section to the book.
  - The `prql-compiler` page and `prqlc` page are moving to the section.
  - Add the "Compile target" page to describe the priority of the target.


max-sixty:
(will add a couple of comments when home!)

prql-compiler/prqlc/README.md
```diff
@@ -1,6 +1,7 @@
-# prqlc
+# PRQL compiler CLI
```
aljazerzen: Let's keep `prqlc` name here, it's the name of the cli after all.

web/book/src/SUMMARY.md
```diff
@@ -55,6 +55,13 @@
...
   - [Target & Version](./language-features/target.md)
 
+- [Compiler](./compiler/README.md)
+
+  - [Reference implementation](./compiler/rust/README.md)
+    - [CLI](./compiler/rust/cli.md)
+  - [Compiler features](./compiler/compiler-features/README.md)
+    - [Target & Version](./compiler/compiler-features/target.md)
+
 - [Bindings](./bindings/README.md)
```
aljazerzen, line 65: Bindings are technically "Language bindings to the Rust prql-compiler library", so they should maybe be under "Compiler" section too.

Or maybe, "Bindings" section should be titled "Compiler" and contain a page about the CLI and a page about "Target & version"

aljazerzen (approved):
prqlc name is following javac, rustc, gcc convention for compiler CLIs. We had it named `prql-compiler`, but that's too long for a command line.

This looks good. I have a few structuring suggestions, but feel free to dismiss them.

prql-compiler/prqlc/README.md
eitsupi: Sorry for the lack of explanation, `prqlc` is a great program name and I don't think it needs to be changed.

My point is to use "PRQL compiler CLI" as the package name that appears in winget manifests, etc.

For example, GitHub CLI (https://github.com/cli/cli) is a binary named `gh`, but the package name on the README and in the winget manifest is `GitHub CLI`.
https://github.com/microsoft/winget-pkgs/blob/03bf176f0ada12b0723f6d59f4b7ad6f5fba1630/manifests/g/GitHub/cli/2.29.0/GitHub.cli.locale.en-US.yaml#L2-L18

aljazerzen: Oh, sure - PRQL compiler CLI is a good description of what this is.

max-sixty: I agree `prqlc` is insufficient here. But I still think it will be referred to by its name `prqlc` — how about:

```suggestion
# PRQL compiler CLI — `prqlc`
```

web/book/src/SUMMARY.md
```diff
@@ -67,7 +74,6 @@
 
 - [Integrations](./integrations/README.md)
 
-  - [CLI](./integrations/cli.md)
```
max-sixty: I agree this doesn't really fit at the moment.

But I think we need some list of "How do I use PRQL?", which this is very much part of. To some extent we get that from the https://prql-lang.org/faq/, but having a list here would be good too. The proposed change better reflects how we produce these but arguably is worse on how they're consumed.

One idea would be changing this section to "Tools & Integrations", retaining this as a link, even if the page here just points to the https://github.com/PRQL/prql/pull/2648/files#diff-5209855e27f70bffaa6997cd87e9aed958ae9c4bb768f841bfff586b0c55562fR1.

WDYT?

web/book/src/compiler/README.md
```diff
@@ -0,0 +1,16 @@
...
+
+The reference compiler implementation of PRQL is written in Rust. And there is a
+CLI for it.
+
+- [Reference implementation](./rust/README.md)
+  - [CLI](./rust/cli.md)
+
+## Other implementations
+
+### Go PRQL
```
max-sixty: I'm very up for promoting things outside the official org. But is this still being developed? I see the most recent commit was almost a year ago. If it was a quick effort which is no longer being worked on, I'm not sure it makes sense to link to...

eitsupi:
@max-sixty Thanks for your review. I think your points make sense, so I have updated these.

I think there is still room for improvement regarding the structure of the book, but can we merge with this for now?

aljazerzen:
Sure, it's an improvement for sure.

eitsupi:
@max-sixty Could you take a look at this?

web/book/src/compiler/compiler-features/target.md
```diff
@@ -0,0 +1,29 @@
...
+and `sql.duckdb` in the `--target` option of the `prqlc compile` command. In
+this case, `sql.duckdb` takes precedence and the SQL output is based on the
+DuckDB dialect.
+
+```sh
+echo 'prql target:sql.generic
+      from foo' | prqlc compile --target sql.duckdb
+```
+
+To use the target described in the query, a special target `sql.any` can be
```
max-sixty, line 22: Isn't this the default too? If so, I think we should say that; the current version might not be clear on what happens when nothing is supplied.

web/book/src/compiler/compiler-features/target.md
```diff
@@ -0,0 +1,29 @@
...
+
+```admonish note
+See also the [Target & Version](../../../language-features/target.md) language features page.
+```
+
+## Priority of targets
+
+The compile target of a query is defined in the top of the query or compiler's
+option. If different targets are specified in these two places, the option
+specified in the compiler takes precedence.
```
max-sixty:
```suggestion
The compile target of a query is defined in the query's header or as an
argument to the compiler.
option. The argument to the compiler takes precedence.
```

web/book/src/compiler/compiler-features/target.md
```diff
@@ -0,0 +1,29 @@
...
+this case, `sql.duckdb` takes precedence and the SQL output is based on the
+DuckDB dialect.
+
+```sh
+echo 'prql target:sql.generic
+      from foo' | prqlc compile --target sql.duckdb
+```
+
+To use the target described in the query, a special target `sql.any` can be
+specified in the compiler option.
```
max-sixty, line 23:
```suggestion
The default target is `sql.any`, which defers to the query header. It's possible (but not required) to
specify this explicitly:
```

web/book/src/SUMMARY.md
```diff
@@ -55,6 +55,13 @@
   - [F-strings](./language-features/f-strings.md)
   - [Target & Version](./language-features/target.md)
 
+- [Compiler](./compiler/README.md)
+
+  - [Reference implementation](./compiler/rust/README.md)
+    - [CLI](./compiler/rust/cli.md)
+  - [Compiler features](./compiler/compiler-features/README.md)
+    - [Target & Version](./compiler/compiler-features/target.md)
```
max-sixty, line 63: These make sense, but the span is quite low — 5 items in 3 levels. What about moving this section down a bit, and integrating it with `Internals`? Or at least moving a page like "Compiler architecture" here?

(I don't feel that strongly though, feel free to merge as-is and we can iterate)

max-sixty:
> @max-sixty Thanks for your review. I think your points make sense, so I have updated these.
> 
> I think there is still room for improvement regarding the structure of the book, but can we merge with this for now?

Sorry I missed this, I really try hard to not have things wait on me. Thank you for pinging me the next day. Yes we can merge! I added a couple comments but fine to merge this as-is.

PR #2649, opened 2023-05-29
fix: file loading and lookup of main pipeline

aljazerzen:


max-sixty:
@aljazerzen I notice you're trying to fix the Windows build in https://github.com/PRQL/prql/pull/2658/files. Remember totally fine to revert and relax...

PR #2654, opened 2023-05-30
fix: array typing

aljazerzen:


prql-compiler/src/tests/test.rs
```diff
@@ -3646,7 +3646,7 @@ fn test_array() {
        │
      2 │     let a = [1, 2, false]
        │                    ──┬──
-       │                      ╰──── array expected types of all of its elements to be int, but found bool
+       │                      ╰──── array expected type `int`, but found type `bool`
```
max-sixty, line 3649: Nice, that is better!

aljazerzen: It's better? Before I used a message handcrafted to this specific situation, but now I've defaulted to the general type-checking message. I though the error message was less informative :D

max-sixty: The disadvantage of the previous one was that it pointed to `false` and gave a message about the array. I do think having a tight connection between the message and the thing it's pointing to is important.

Rust does this well with the borrow checker, showing two locations — the location of the first and second move — I guess the ideal here (which I'm not expecting us to do for a while) is to have another error message saying "here's the array, it starts with `int`s, so we inferred it to be an array<int>"

PR #2657, opened 2023-05-30
feat: resolving names from parent modules

aljazerzen:


max-sixty:
Ah, this conflicts with the changes I added to https://github.com/PRQL/prql/pull/2659. I didn't see this when adding that.

I marginally prefer the naming in #2659 so I won't close that one, but your call @aljazerzen

aljazerzen:
Hmm, I do prefer to `is_superset_of_array`, because it clearly states what's happening in regards to our type system.

But this change should go into #2659 instead.

PR #2663, opened 2023-05-31
internal: Remove extra gitignore entries

max-sixty:


.gitignore
```diff
@@ -12,12 +12,6 @@ lcov.info
 .idea
 
 /*.prql
-_*.prql
-**/project/**/*.prql
```
max-sixty, line 16: These are presumably ignoring committed files?

.gitignore
```diff
@@ -12,12 +12,6 @@ lcov.info
...
-/*.rq
-_*.rq
-/*.pl
-_*.pl
```
max-sixty, line 20: Not sure why these are here; looks like I added them. Using `foo.rq.yaml` makes more sense

.gitignore
aljazerzen: I was testing the cli and had a few directories like this. Now that one of them is commited, we can remove this, yes.

aljazerzen (approved):

aljazerzen:
Oh, I still have a bunch of `_a.prql`, `_b.prql` test files all over the repo. I'll add `_*.prql` back in.

max-sixty:
Actually if we make it `Foo.prql` then it's no longer necessary!

But I definitely think we should avoid having any files that are ignored-but-already-tracked, could make for difficult bugs

aljazerzen:
No, I mean, these are individual files I have, not part of a project.

PR #2666, opened 2023-05-31
fix: Refine some error messages

max-sixty:
`INTERCEPT` vs `INTERSECT`, remove use of 2nd person, align `Hint` & `Error`

### DIFF
--- a/prql-compiler/src/error.rs
+++ b/prql-compiler/src/error.rs
@@ -137,7 +137,7 @@ impl Display for ErrorMessage {
 
             writeln!(f, "{}Error: {}", code, &self.reason)?;
             if let Some(hint) = &self.hint {
-                writeln!(f, " Hint: {}", hint)?;
+                writeln!(f, "Hint: {}", hint)?;
             }
         }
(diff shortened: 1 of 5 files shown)


prql-compiler/src/error.rs
```diff
@@ -137,7 +137,7 @@ impl Display for ErrorMessage {
 
             writeln!(f, "{}Error: {}", code, &self.reason)?;
             if let Some(hint) = &self.hint {
-                writeln!(f, " Hint: {}", hint)?;
+                writeln!(f, "Hint: {}", hint)?;
```
max-sixty, line 140: @aljazerzen lmk if there was a reason for this and I'll revert

aljazerzen: I've added that for visual grouping of non-composed errors:

```
Error: blah
Hint: bleh
Error: blah
Hint: bleh
```
... looks like these are 4 separate messages.

max-sixty: Added #2675 as a synthesis...

PR #2658, opened 2023-05-30
fix: prqlc tests on Windows

aljazerzen:

PR #2667, opened 2023-05-31
fix: Another attempt at debugging Windows tests

max-sixty:
Parallel effort to #2658 without mangling that PR


prql-compiler/prqlc/src/cli.rs
```diff
@@ -398,14 +398,15 @@ mod clio_extended {
             if path == "-" {
                 Ok(Self::std())
             } else {
-                let file = File::open(path)?;
-                if file.metadata()?.is_dir() {
-                    return Ok(Input::Directory(path.to_os_string()));
+                let pathbuf = PathBuf::from(path);
+                if pathbuf.is_dir() {
```
max-sixty, line 402: The difference is here — `File::open(path)` on a directory on Windows seems to panic, but is fine on Mac & Linux...

aljazerzen: Oh god, thank you!

I was just about to drop support for Windows. Debugging in GH actions is hell and dealing with a bug for Windows specifically is nerve wracking.

Thank you again.

max-sixty: I was half way to renting a Windows VM tbh...

PR #2669, opened 2023-05-31
feat: revamp PRQL codegen framework

### LINKED ISSUE #1420: PRQL code generation & formatting

aljazerzen:
Closes #1420


prql-compiler/src/codegen.rs
```diff
@@ -0,0 +1,590 @@
...
+
+impl std::fmt::Display for pl::Ty {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        let opt = WriteOpt::new_width(u16::MAX);
+        f.write_str(&self.write(opt).unwrap())
+    }
+}
+
+pub trait WriteSource {
+    /// Converts self to it's source representation according to specified
```
max-sixty, line 39:
```suggestion
    /// Converts self to its source representation according to specified
```

prql-compiler/src/codegen.rs
```diff
@@ -0,0 +1,590 @@
...
+    let needs_escape = s.is_empty()
+        || s.starts_with(forbidden_start)
+        || (s.len() > 1 && s.chars().skip(1).any(forbidden_subsequent));
+
+    if needs_escape {
+        format!("`{s}`")
+    } else {
+        s.to_string()
+    }
+}
```
max-sixty, line 323: GPT4 wrote this as a regex:

```suggestion
use regex::Regex;
use once_cell::sync::Lazy;

static IDENT_REGEX: Lazy<Regex> = Lazy::new(|| Regex::new(r"^[_$a-z][_a-z\d]*$").unwrap());

pub fn write_ident_part(s: &str) -> String {
    if IDENT_REGEX.is_match(s) {
        s.to_string()
    } else {
        format!("`{}`", s)
    }
}
```

I actually though I did this before, but probably it was for something different...

max-sixty: Should we also allow `*`, ref https://github.com/PRQL/prql/pull/2669/files#diff-107555a4ee015fdfa938922785f2d69c79432e2debb074e8270db0b05c9195c7R7 ?

max-sixty:
Extremely impressive @aljazerzen !

prql-compiler/src/codegen.rs
max-sixty: I remembered where I did this before — is this the same as https://github.com/prql/prql/blob/aa3ed6196596e38b3dfffad63d0b590e9eeed535/prql-compiler/src/sql/gen_expr.rs#L783 ? (fine to have it twice unless it's definitely going to remain the same)

aljazerzen: This is the problem with GPT: it looks likes it's correct, but it's not. The regex does not quote empty strings, but it should.

In anycase, this snippet was copied from the `gen_expr.rs` a while back and was living on `codegen` branch for last few months.

If you feel like it needs to be improved, go ahead. But add tests if you are using GPT.

max-sixty: Very good point!! 

Yeah, agree it has that frailty. I have found it helpful for these sorts of "chore" tasks; agree tests become more important...

max-sixty: Though in this case is it actually wrong? An empty string does get quoted by the regex.

aljazerzen: No, it's actually correct. An empty string does get quoted by the regex, which is correct. The logic has also been inverted.

But that's my point: it's more work validating the output as opposed to rewriting it. And things like this don't even need a rewrite or any more time spent on them.

max-sixty: Yes I agree re focus

max-sixty: (I had already put https://github.com/PRQL/prql/pull/2678 in — which I think does use the version that gets the `*` correct, but not 100% sure)

PR #2678, opened 2023-06-01
refactor: Consolidate regex ident checks

max-sixty:
From https://github.com/PRQL/prql/pull/2669#discussion_r1212792233

The behavior change is on handling `*`, which I _think_ is correct — at least it seems to compile OK.

### DIFF
--- a/prql-compiler/src/codegen.rs
+++ b/prql-compiler/src/codegen.rs
@@ -1,4 +1,4 @@
-use crate::ast::pl;
+use crate::{ast::pl, utils::VALID_IDENT};
 
 pub fn write(stmts: &Vec<pl::Stmt>) -> String {
     let mut r = String::new();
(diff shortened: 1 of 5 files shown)


aljazerzen (commented):
I trust you that this is correct.

max-sixty:
> I trust you that this is correct.

TBC when I say "I think is correct", I mean that this is the correct PRQL: https://github.com/PRQL/prql/pull/2678/files#diff-5557efedcdc69e2fbea057c573b9fb68fb96c2f5b8a087b0ba8b25531bfd949dR9-R10 — i.e. that we should not be quoting `*` like we do now.

aljazerzen:
Oh, no that is not correct.

Well, produced PRQL will still be correct, but the backticks are not required in `` a.`*` ``.

max-sixty:
> Oh, no that is not correct.
> 
> Well, produced PRQL will still be correct, but the backticks are not required in `` a.`*` ``.

Sorry — to confirm — is the proposed code correct or the existing code? 

I think the proposed code is correct, which removes the backticks — right?

aljazerzen:
I'm embarrassed how many times I've made a mistake here. Yes, removing backticks is correct.

PR #2681, opened 2023-06-01
feat: formalize RQ operators

### LINKED ISSUE #746: `include` other prql files & module system
Currently, seems we only have 1 [stdlib.prql](https://github.com/prql/prql/blob/main/prql-compiler/src/sql/stdlib.prql)

which will be included when parse every other prql file. maybe we can make this "include" as a feature.

so, I can define some custom functions in separate prql file:

```
func day_of_week col -> s' ......'
func date_trunc  col unit -> s'.....'
```

while in the main prql, I can first include them, and then use those functions:

```
prql include:datetime_util

from table1
derive[ a = day_of_week(datetime_col)]
```

we may support specify library search path for different dialects, like:

```
--- current folder
    --   datetime_util.prql (default one)
    --   mysql folder
         --- datetime_util.prql (overload one)
```

aljazerzen:
Closes #746

Changes are mostly internal, but could be exposed an extension interface, where extensions could provide `.prql` files visible to users and also RQ operator implementations that would implement custom RQ operators.


prql-compiler/src/sql/std.sql.prql
```diff
@@ -44,5 +51,34 @@ let lte = l r -> null
 let and = l r -> null
 let or = l r -> null
 let coalesce = l r -> s"COALESCE({l}, {r})"
+let regex_search = text pattern -> s"REGEXP({text}, {pattern})"
 let neg = l -> null
 let not = l -> null
+
+module postgres {
```
max-sixty, line 58: Awesome!

max-sixty (commented):
Quite the refactor!

V impressive for the target-specific modules. Renders some of the upstream work with regex operators moot — probably I should have just used strings originally by dialect.

aljazerzen:
Your work on regex inspired me a bit on how to organize this. So it was not moot :D

PR #2684, opened 2023-06-01
feat: int division operator

### LINKED ISSUE #2426: Introduce integer division operator `//`
### What's up?

DuckDB seems to introduce a new division operator `/` and  the current division operator `/` will be moved to `//` (duckdb/duckdb#7082).

I'm wondering if the same thing could be done with PRQL.

In my opinion, PRQL should not do the division itself as it currently does, but should convert `//` to `/` for targets that do not have two division operators.

> Integer Division: SQLite, Postgres, SQL Server
> Floating Point Division: MySQL, SparkSQL, BigQuery, Snowflake, ClickHouse

Note that My SQL seems to require the `DEV` operator for integer division.

aljazerzen:
Closes #2426

Also makes sure that float division is always producing floats.
This is done by casting to float before division. This can be removed
when match is implemented and we can check types and use only the
appropriate operation.


prql-compiler/src/sql/std.sql.prql
```diff
@@ -38,7 +38,8 @@ let read_parquet = source -> s"read_parquet({source})"
 let read_csv = source -> s"read_csv_auto({source})"
 
 let mul = l r -> null
-let div = l r -> null
+let div_i = l r -> s"ROUND({l} / {r} - 0.5)"
+let div_f = l r -> s"(CAST({l} AS FLOAT) / {r})"
```
eitsupi: I am not sure that these are always correct.

- If `ROUND` is "Bankers' rounding", `ROUND(x - 0.5)` would not be `FROOR(x)`.
- If `l` is a type like FLOAT64, I think `CAST(l AS FLOAT)` is not correct. (See PEP 238 cited in this comment https://github.com/PRQL/prql/issues/2426#issuecomment-1554694777)

aljazerzen: > If ROUND is "Bankers' rounding", ROUND(x - 0.5) would not be FLOOR(x).

I used FLOOR before, but it was giving wrong answers for `-13 / 5`. I'll run tests on all databases locally now.

---

>If l is a type like FLOAT64, I think CAST(l AS FLOAT) is not correct.

Their argument is that it would not work for complex numbers. I don't think we should concern ourselves with that yet.

aljazerzen: > If ROUND is "Bankers' rounding", ROUND(x - 0.5) would not be FLOOR(x).

I used FLOOR before, but it was giving wrong answers for `-13 / 5`. I'll run tests on all databases locally now.

---

>If l is a type like FLOAT64, I think CAST(l AS FLOAT) is not correct.

Their argument is that it would not work for complex numbers. I don't think we should concern ourselves with that yet.

aljazerzen: On MySQL, `x * 1.0` somehow produces different results than `CAST(x AS FLOAT)`. I think it's converting for decimal instead of float.

prql-compiler/src/sql/std.sql.prql
```diff
@@ -38,8 +38,9 @@ let read_parquet = source -> s"read_parquet({source})"
 let read_csv = source -> s"read_csv_auto({source})"
 
 let mul = l r -> null
-let div = l r -> null
-let mod = l r -> null
+let div_i = l r -> s"FLOOR(FLOOR(ABS({l} / {r})) * SIGN({l} / {r}))"
+let div_f = l r -> s"({l} * 1.0 / {r})"
```
max-sixty, line 42: One downside of s-strings is that if `l` is `5 + 5`, then it doesn't understand precedence, and becomes `(5 + 5 * 1.0 / 2)`, so the `/ 2` only operates on `5` rather than `10`.

I guess we handle this, I can check further to confirm.

prql-compiler/src/sql/std.sql.prql
```diff
@@ -56,29 +57,33 @@ let neg = l -> null
 let not = l -> null
 
 module postgres {
-    let regex_search = text pattern -> s"{text} ~ {pattern}"
+    let div_i = l r -> s"trunc({l} / {r})::bigint"
+    let round = n_digits column -> s"ROUND(({column})::numeric, {n_digits})"
+    let regex_search = text pattern -> s"({text} ~ {pattern})"
```
max-sixty, line 62: So good that we have the dialect modules now — adding something goes from "difficult rust" to "easy PRQL"!

prql-compiler/src/tests/test.rs
```diff
@@ -1142,7 +1142,7 @@ fn test_nulls() {
     "###).unwrap()), @r###"
     SELECT
       *,
-      COALESCE(amount + 2, 15) AS amount
+      COALESCE(amount + 2, 3 * 5) AS amount
```
max-sixty, line 1145: Out of interest, why did we remove these? (no normative view, just out of interest)

prql-compiler/tests/integration/queries/arithmetic.prql
```diff
@@ -0,0 +1,27 @@
+[
+    { x_int =  13, x_float =  13.0, k_int =  5, k_float =  5.0 },
+    { x_int = -13, x_float = -13.0, k_int =  5, k_float =  5.0 },
+    { x_int =  13, x_float =  13.0, k_int = -5, k_float = -5.0 },
+    { x_int = -13, x_float = -13.0, k_int = -5, k_float = -5.0 },
```
max-sixty, line 5: I'm guessing this isn't auto-formatted? It's beautiful! 

(Maybe that's a side project — align relation literals like this automatically!)

prql-compiler/src/sql/std.sql.prql
max-sixty, line 42: @aljazerzen how would you feel about:
- Generic dialect is just `l / r`
- MySQL & MsSql use `s"({l} * 1.0 / {r})"`

That way, the default SQL is cleaner, and we keep the guarantees when a dialect is specified. I do still place some weight on having the introductory examples be really easy to read and compare.

The downside is that we don't guarantee the type of division on a generic dialect.

(I'm happy to make the change if you agree)

web/book/tests/snapshots/snapshot__queries__functions__piping__2.snap
```diff
@@ -4,7 +4,7 @@ expression: "let fahrenheit_to_celsius = temp -> (temp - 32) / 1.8\nlet interp =
 ---
 SELECT
   *,
-  ((temp_c - 32) / 1.8 - 0) / 100 AS boiling_proportion
+  ((temp_c - 32 * 1.0 / 1.8) - 0 * 1.0 / 100 - 0) AS boiling_proportion
```
max-sixty, line 7: Re the comment above re generic dialect — I do think this is now harder to read

prql-compiler/src/sql/std.sql.prql
max-sixty: Actually this comes up in https://github.com/PRQL/prql/pull/2692

```
(started + unfinished * 1.0 / started)
# and
* 1.0 / 1 - (started ...
```

prql-compiler/src/sql/std.sql.prql
max-sixty: Yes, this is correct — MySQL seems to have Decimal division rather than Float division...

prql-compiler/src/sql/std.sql.prql
aljazerzen: We need more parenthesis!

Maybe move the "binding strength" calculation and parenthesis generation into here. Annotate the strength using some new annotation syntax?.

prql-compiler/src/tests/test.rs
aljazerzen: I didn't feel like validating all of the evaluation rules I had. So I just deleted them.

prql-compiler/src/sql/std.sql.prql
aljazerzen: I don't like that. Character `/` in PRQL means something else than `/` in SQL. I don't think it should be problematic to emit code that looks differently.

All of this would be way more ergonomic if we displayed query results in the book, instead of SQL.

max-sixty: > I don't like that. Character `/` in PRQL means something else than `/` in SQL.

I agree with the principle — if it's different then that's bad.
But it's only different in some minor dialects. My suggestion is to support those iff they're explicitly specified.

Another approach would be to default to DuckDB dialect (or Postgres, or Snowflake, i.e. one which has reasonable behavior here), rather than have the default try to cater to every dialect. Then we could use `/` safely.

> All of this would be way more ergonomic if we displayed query results in the book, instead of SQL.

Showing results would be useful too, but many people are much more used to SQL, and it's much clearer to them to see the actual SQL rather than see the results of a query on a table they're not familiar with.

I get that there's a tradeoff there — to what extent are we "producing SQL" vs "producing query results"? When I speak to people about PRQL, they do say they value the legible SQL output. And if I didn't know about PRQL, I would care about that — it's too difficult to have trust or learn the language otherwise. 

I think possibly we're on different sides of this discussion, I hope we can find a synthesis of our perspectives.

prql-compiler/src/tests/test.rs
max-sixty: OK!

prql-compiler/src/sql/std.sql.prql
max-sixty: I added one option in https://github.com/PRQL/prql/issues/2694...

aljazerzen: If we could be emitting simpler SQL that still passes the tests, we should definitely simply it. This is not hard now.

I don't strongly feel we should switch to showing query results instead of SQL, just a suggestion.

But I do feel strongly that if we decide on different semantics than SQL, we should compile to something that does implement those semantics correctly, in all edge cases. Otherwise our argument that "PRQL can be used with any SQL database" is simply not true.

We could present this as "PRQL handles edge cases for you".

prql-compiler/tests/integration/docker-compose.yml
```diff
@@ -1,3 +1,4 @@
+version: "2.4"
```
eitsupi, line 1: This line seems added by accident?

prql-compiler/src/sql/std.sql.prql
```diff
@@ -56,29 +57,33 @@ let neg = l -> null
...
 
 module bigquery {
     let regex_search = text pattern -> s"REGEXP_CONTAINS({text}, {pattern})"
 }
 
 module duckdb {
-    let regex_search = text pattern -> s"REGEXP_MATCHES({text}, {pattern})"
-}
+    # TODO: uncomment when updated on latest DuckDB version
+    # let div_i = l r -> s"({l} // {r})"
```
eitsupi, line 86: DuckDB's `//` does not work for floats, so I think we can't use this.
duckdb/duckdb#7558

prql-compiler/tests/integration/docker-compose.yml
aljazerzen: Nope, it's intentional. 

Docker-compose version 3.x was rejecting the file, saying mssql contains entries that are not allowed.

eitsupi: That is weird. Do you know what version of docker compose you are using?

- Currently the latest docker compose is 2.18.1 (See https://github.com/docker/compose)
- docker compose 2.X is based on the compose spec, and `version` is deprecated in the compose spec. https://github.com/compose-spec/compose-spec/blob/master/spec.md#version-top-level-element

aljazerzen:
```
$ docker-compose --version
docker-compose version 1.25.0, build unknown
```

prql-compiler/src/sql/std.sql.prql
max-sixty: I think we agree on the principles:
- PRQL should be correct (not approximately correct)
- We should have legible SQL

...and here they're in conflict. 

I probably weigh Legibility more highly than you?

Is there a synthesis of these tradeoffs where we can do both? 
- Default compile to DuckDB dialect?
- Default compile to DuckDB on the website & playground?
- Have a small list of supported dialects that work with the generic dialect, and others need to explicitly specify the dialect

WDYT?

prql-compiler/tests/integration/docker-compose.yml
eitsupi: > $ docker-compose --version
> docker-compose version 1.25.0, build unknown

Oh, your docker-compose version is outdated. compose version 1 is no longer supported, and compose version 2 is widely used these days.
compose v1 is written in Python, and compose v2 is fully rewritten in Go as Docker plugin.
Please uninstall compose v1 and install v2.
https://docs.docker.com/compose/install/

PR #2688, opened 2023-06-01
fix: lineage of unnamed columns

### LINKED ISSUE #2392: `select` works after `from_text` but `derive` fails
### What happened?

I found the following error when using `derive` as in the attached query.

```
Error: 
   ╭─[:1:23]
   │
 1 │ ╭─▶ from_text format:json """{
   ┆ ┆   
 6 │ ├─▶ }"""
   │ │          
   │ ╰────────── This table contains unnamed columns that need to be referenced by name
   │     
   │     Help: The name may have been overridden later in the pipeline.
───╯
```

Is updating columns by derive not allowed?

### PRQL input

```elm
from_text format:json """{
  "columns": ["a"],
  "data": [
    [1]
  ]
}"""
derive a = a + 1
```

### SQL output

```SQL
None
```

### Expected SQL output

```SQL
WITH table_0 AS (
  SELECT
    1 AS a
)
SELECT
  a + 1 AS a
FROM
  table_0 AS table_1
```

### MVCE confirmation

- [X] Minimal example
- [X] New issue

### Anything else?

_No response_

### LINKED ISSUE #2627: `from_text` seems to have unorthogonal behavior
### What happened?

`from_text` seems to raise an error when a later derive overrides its column.

I've listed the title as "unorthogonal", since that's the most confusing part of it — this works fine:

```elm
from foo
select a
derive a = a
```

but we see the table differently from the `derive` line if it's instead from `from_text`, as below

### PRQL input

```elm
from_text """
a
1
2
3
"""
derive a = a
```

### SQL output

```SQL
Error: 
   ╭─[:1:11]
   │
 1 │ ╭─▶ from_text """
   ┆ ┆   
 6 │ ├─▶ """
   │ │         
   │ ╰───────── This table contains unnamed columns that need to be referenced by name
   │     
   │     Help: The name may have been overridden later in the pipeline.
───╯
```

### Expected SQL output

_No response_

### MVCE confirmation

- [X] Minimal example
- [X] New issue

### Anything else?

_No response_

aljazerzen:
Closes #2627
Closes #2392

### DIFF
--- a/prql-compiler/src/tests/test.rs
+++ b/prql-compiler/src/tests/test.rs
@@ -3770,6 +3770,7 @@ fn test_lineage() {
     );
 }
 
+#[test]
 fn test_type_as_column_name() {
     // #2503
     assert_display_snapshot!(compile(r#"
--- a/prql-compiler/src/tests/test.rs
+++ b/prql-compiler/src/tests/test.rs
@@ -3714,3 +3714,58 @@ fn test_double_stars() {
     "###
     );
 }
+
+#[test]
+fn test_lineage() {
+    // #2627
+    assert_display_snapshot!(compile(r#"
+    from_text """
+    a
+    1
+    2
+    3
+    """
+    derive a = a
+    "#).unwrap(),
+        @r###"
+    WITH table_2 AS (
+      SELECT
+        '    1' AS a
+      UNION
+      ALL
+      SELECT
+        '    2' AS a
+      UNION
+      ALL
+      SELECT
+        '    3' AS a
+    )
+    SELECT
+      a,
+      a
+    FROM
+      table_2 AS table_1
+    "###
+    );
+
+    // #2392
+    assert_display_snapshot!(compile(r#"
+    from_text format:json """{
+        "columns": ["a"],
+        "data": [[1]]
+    }"""
+    derive a = a + 1
+    "#).unwrap(),
+        @r###"
+    WITH table_2 AS (
+      SELECT
+        1 AS a
+    )
+    SELECT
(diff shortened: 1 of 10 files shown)


prql-compiler/src/tests/test.rs
```diff
@@ -3714,3 +3714,58 @@ fn test_double_stars() {
...
+fn test_lineage() {
+    // #2627
+    assert_display_snapshot!(compile(r#"
+    from_text """
+    a
+    1
+    2
+    3
+    """
+    derive a = a
```
max-sixty, line 3728:
```suggestion
    derive b = a
```

Otherwise the result is a bit confusing since it has duplicate column names

prql-compiler/src/tests/test.rs
```diff
@@ -3714,3 +3714,58 @@ fn test_double_stars() {
...
+      SELECT
+        '    2' AS a
+      UNION
+      ALL
+      SELECT
+        '    3' AS a
+    )
+    SELECT
+      a,
+      a
```
max-sixty, line 3745:
```suggestion
      a AS b
```

max-sixty (commented):
Thanks!

prql-compiler/src/tests/test.rs
aljazerzen: Overriding the name was the cause of the error.

And the resulting SQL is even wrong a little, because the first column should be named something else than `a`.

max-sixty: Sorry, yes

PR #2692, opened 2023-06-01
docs: Update website queries

max-sixty:
I also realize there's a small bug in the current implementation, will add a link

### DIFF
--- a/web/website/content/_index.md
+++ b/web/website/content/_index.md
@@ -216,7 +216,7 @@ showcase_section:
         from weather
         select temp_f = (fahrenheit_from_celsius temp_c)
       sql: |
-        SELECT 
+        SELECT
           (temp_c * 9 * 1.0 / 5) + 32 AS temp_f
         FROM
           weather
--- a/web/website/content/_index.md
+++ b/web/website/content/_index.md
@@ -169,8 +169,8 @@ showcase_section:
         SELECT
           *,
           started + unfinished AS finished,
-          (started + unfinished) / started AS fin_share,
-          (started + unfinished) / started / (1 - (started + unfinished) / started) AS fin_ratio
+          (started + unfinished * 1.0 / started) AS fin_share,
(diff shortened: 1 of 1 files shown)


web/website/content/_index.md
```diff
@@ -169,8 +169,8 @@ showcase_section:
         SELECT
           *,
           started + unfinished AS finished,
-          (started + unfinished) / started AS fin_share,
-          (started + unfinished) / started / (1 - (started + unfinished) / started) AS fin_ratio
+          (started + unfinished * 1.0 / started) AS fin_share,
```
max-sixty, line 172: This produces incorrect results; it's a manifestation of https://github.com/PRQL/prql/pull/2684#discussion_r1213656459.

We could use parentheses; we'll get too many, but that might be OK. Particularly if we only do this on explicit dialects.

web/website/content/_index.md
```diff
@@ -169,8 +169,8 @@ showcase_section:
...
+          ( (started + unfinished * 1.0 / started) * 1.0 / 1 - (started + unfinished * 1.0 / started) ) AS fin_ratio
```
max-sixty, line 173: Re the decision of whether we do this coercion on a generic dialect:

Overall I think this looks much worse to a new potential user — it's not immediately obvious to them why all these `* 1.0`, even if it does produce more consistent results. (I still think the int division is great tbc, always a tradeoff between precision and easy SQL...!)

max-sixty:
To the extent the s-string / parenthesis issue is going to cause a lot of work — we can reduce support for some dialects if that's easier — we'd discussed this a while ago and I think were all generally open to it...

max-sixty:
I'll merge so tests pass BUT I don't think we should release like this — it's incorrect and verbose. Check out https://github.com/PRQL/prql/issues/2694 for one idea on some of the verbosity.

(But the overall approach of having dialect-specific stdlib is excellent, we should obv keep that)

PR #2693, opened 2023-06-02
fix: Add parentheses to new division s-strings

max-sixty:
This is required for correctness — see https://github.com/PRQL/prql/pull/2692 for an example.

But it really pollutes the resulting expressions with parentheses, so I'm not sure if this is the right approach.

### DIFF
--- a/prql-compiler/src/sql/std.sql.prql
+++ b/prql-compiler/src/sql/std.sql.prql
@@ -38,9 +38,12 @@ let read_parquet = source -> s"read_parquet({source})"
 let read_csv = source -> s"read_csv_auto({source})"
 
 let mul = l r -> null
-let div_i = l r -> s"FLOOR(FLOOR(ABS({l} / {r})) * SIGN({l} / {r}))"
-let div_f = l r -> s"({l} * 1.0 / {r})"
-let mod = l r -> s"{l} % {r}"
+# If we want to raise an error when division by zero occurs, we can use the
+# following:
+# let div_i = l r -> s"CASE WHEN {l} * {r} < 0 THEN -FLOOR(ABS({l} / {r})) ELSE FLOOR(ABS({l} / {r})) END"
+let div_i = l r -> s"FLOOR(FLOOR(ABS(({l}) / ({r}))) * SIGN(({l}) / ({r})))"
+let div_f = l r -> s"({l} * 1.0 / ({r}))"
+let mod = l r -> s"({l}) % ({r})"
 let add = l r -> null
 let sub = l r -> null
 let eq = l r -> null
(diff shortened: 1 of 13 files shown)


prql-compiler/src/sql/std.sql.prql
```diff
@@ -38,9 +38,15 @@ let read_parquet = source -> s"read_parquet({source})"
...
-let div_f = l r -> s"({l} * 1.0 / {r})"
-let mod = l r -> s"{l} % {r}"
+# If we want to raise an error when division by zero occurs, we can use the
+# following:
+# let div_i = l r -> s"CASE WHEN {l} * {r} < 0 THEN -FLOOR(ABS({l} / {r})) ELSE FLOOR(ABS({l} / {r})) END"
+let div_i = l r -> s"FLOOR(FLOOR(ABS(({l}) / ({r}))) * SIGN(({l}) / ({r})))"
+# Should this even be:
+#   let div_f = l r -> s"(({l}) * 1.0 / ({r}))"
+# ...given that we can have higher precedence than mul?
+let div_f = l r -> s"({l} * 1.0 / ({r}))"
```
aljazerzen, line 48: The problem is that we can have *lower* precedence: `(3 + 4 * 1.0 / (5 + 6))`

prql-compiler/src/sql/std.sql.prql
```diff
@@ -38,9 +38,15 @@ let read_parquet = source -> s"read_parquet({source})"
 let read_csv = source -> s"read_csv_auto({source})"
 
 let mul = l r -> null
-let div_i = l r -> s"FLOOR(FLOOR(ABS({l} / {r})) * SIGN({l} / {r}))"
-let div_f = l r -> s"({l} * 1.0 / {r})"
-let mod = l r -> s"{l} % {r}"
+# If we want to raise an error when division by zero occurs, we can use the
+# following:
+# let div_i = l r -> s"CASE WHEN {l} * {r} < 0 THEN -FLOOR(ABS({l} / {r})) ELSE FLOOR(ABS({l} / {r})) END"
```
aljazerzen, line 43: If r = 0, there is still division by zero in the else branch.

Let's not concern ourselves with division by zero for now.

aljazerzen (approved):

aljazerzen:
While you're at it, can you remove these lines: https://github.com/PRQL/prql/pull/2693/files#diff-f6d964f7215b03bfdb696d81cf4258db7cf5686558bf44fa627d031f2f474c2eR92-R93 ?

@eitsupi [mentioned that we cannot use](https://github.com/PRQL/prql/pull/2684#discussion_r1214209273) the new DuckDB integer division

PR #2696, opened 2023-06-02
fix: Allow codegen to work on escaped strings

max-sixty:


max-sixty:
Only a couple of unintentional `no-fmt` remain! (Though I guess there are cases we haven't found yet)

prql-compiler/src/codegen.rs
```diff
@@ -585,4 +585,26 @@ mod test {
...
+        from tracks
+        filter (name ~= "\\(I Can't Help\\) Falling")
+        "#,
+        )
+        .unwrap()
+        .into_iter()
+        .exactly_one()
+        .unwrap();
+
+        // TODO: should this require `trim`? Possibly a linebreak is being added.
```
aljazerzen, line 604: It would be better not to generate additional `\n` if we don't want them. Or is trim here to mitigate snapshot formatting by `cargo fmt`?

aljazerzen (approved):

prql-compiler/src/codegen.rs
max-sixty: I'm not sure, I can look. It's possible we're generating an additional line break...

PR #2703, opened 2023-06-03
test(prql-compiler): add tests for `Target::names`

eitsupi:


max-sixty:
Looks good. It is quite a lot of output — we might not need to snapshot them... (but no great stress)

eitsupi:
Sorry for lots of snapshots.
I did not know what test would be best. Feel free to rewrite it.

max-sixty:
No prob! I slimmed it down slightly (hope that's OK — your code was fine — tell me if this feels like an intrusion!)

PR #2704, opened 2023-06-03
test: Slim down `Target::names` slightly

max-sixty:
(Removes the snapshotting; @eitsupi was OK with the change from #2703)


eitsupi:
Thanks!

PR #2709, opened 2023-06-03
feat: s-string format, annotations and SQL binding strength

### LINKED ISSUE #2694: Division operators / s-strings
### What's up?

We've recently added an int division operator, and moved some logic from the compiler to s-strings.

Moving logic to s-strings is great – it means when we get an issue "this dialect doesn't work with X", it's really easy for someone to make a contribution.

But it does have a couple downsides — in particular because they're only strings, they don't know about precedence. That means that need to add defensive parentheses per https://github.com/PRQL/prql/pull/2693, meaning we end up with quite an expression (both because of the float coercion and the parentheses, let's just focus on the parentheses here though):

```diff
-  ((temp_c - 32) / 1.8 - 0) / 100 AS boiling_proportion
+  ((temp_c - 32 * 1.0 / (1.8)) - 0 * 1.0 / (100 - 0)) AS boiling_proportion
```

I'm particularly concerned about this with the generic SQL translation that folks first encounter when they're approaching the language — I think it's quite important to signal that PRQL is easy to engage with, inspectable, etc.

One option — **parse the s-string into a SQL AST** with `sqlparser-rs`, replacing the variables with `$`. So

```sql
ROUND(ABS(($l) / ($r)) - 0.5) * SIGN(($l) / $r)
```
...gets parsed into[^1]:

<details>

```json
{
  "BinaryOp": {
    "left": {
      "Function": {
        "name": [
          {
            "value": "ROUND",
            "quote_style": null
          }
        ],
        "args": [
          {
            "Unnamed": {
              "Expr": {
                "BinaryOp": {
                  "left": {
                    "Function": {
                      "name": [
                        {
                          "value": "ABS",
                          "quote_style": null
                        }
                      ],
                      "args": [
                        {
                          "Unnamed": {
                            "Expr": {
                              "BinaryOp": {
                                "left": {
                                  "Nested": {
                                    "Value": {
                                      "Placeholder": "$l"
                                    }
                                  }
                                },
                                "op": "Divide",
                                "right": {
                                  "Nested": {
                                    "Value": {
                                      "Placeholder": "$r"
                                    }
                                  }
                                }
                              }
                            }
                          }
                        }
                      ],
                      "over": null,
                      "distinct": false,
                      "special": false,
                      "order_by": []
                    }
                  },
                  "op": "Minus",
                  "right": {
                    "Value": {
                      "Number": [
                        "0.5",
                        false
                      ]
                    }
                  }
                }
              }
            }
          }
        ],
        "over": null,
        "distinct": false,
        "special": false,
        "order_by": []
      }
    },
    "op": "Multiply",
    "right": {
      "Function": {
        "name": [
          {
            "value": "SIGN",
            "quote_style": null
          }
        ],
        "args": [
          {
            "Unnamed": {
              "Expr": {
                "BinaryOp": {
                  "left": {
                    "Nested": {
                      "Value": {
                        "Placeholder": "$l"
                      }
                    }
                  },
                  "op": "Divide",
                  "right": {
                    "Value": {
                      "Placeholder": "$r"
                    }
                  }
                }
              }
            }
          }
        ],
        "over": null,
        "distinct": false,
        "special": false,
        "order_by": []
      }
    }
  }
}
```

</details>

...and then we can replace the `$l` & `$r` with the arguments. On constructing the SQL, the precedence is handled without parentheses.

The downside of this is that an s-string like `s"{x}_column"` no longer works — `x` needs to be a self-contained expression. (We could have two types of s-string, or two types of variables (e.g. `$` vs `{}`, but then that's more complicated.)

What do folks think?

[^1]: 
    That comes from this rust:
    ```rust
    fn main() {
        use sqlparser::dialect::GenericDialect;
        use sqlparser::parser::Parser;
        let sql = "ROUND(ABS(($l) / ($r)) - 0.5) * SIGN(($l) / $r)";

        let parser = Parser::new(&GenericDialect);
        let mut ast = parser.try_with_sql(sql).unwrap();
        println!(
            "{}",
            serde_json::to_string_pretty(&ast.parse_expr().unwrap()).unwrap()
        );
    }
    ```

aljazerzen:
Closes #2694

Closes #2693

Adds parsing of s-string formats:
```
s"Hello {first_name:format}"
```
... where `format` can be any string and does not yet have a user-facing function.

Adds parsing of annotations:
```
#[my_annotation arg1 arg2]
let a = ...
```
... where `arg1`, `arg2` and so on can be any expression. They don't yet have any user-facing function.

### DIFF
--- a/prql-compiler/prqlc/src/cli.rs
+++ b/prql-compiler/prqlc/src/cli.rs
@@ -735,6 +735,7 @@ group a_column (take 10 | sort b_column | derive {the_number = rank, last = lag 
                         - y
               ty_expr: null
               kind: Main
+            annotations: []
         source_ids:
           1: ''
         "###);
--- a/prql-compiler/src/ast/pl/expr.rs
+++ b/prql-compiler/src/ast/pl/expr.rs
@@ -252,7 +252,10 @@ pub struct Pipeline {
 #[derive(Debug, PartialEq, Eq, Clone, Serialize, Deserialize)]
 pub enum InterpolateItem<T = Expr> {
     String(String),
-    Expr { expr: Box<T>, format: Option<String> },
+    Expr {
+        expr: Box<T>,
+        format: Option<String>,
+    },
 }
 
 /// Inclusive-inclusive range.
--- a/prql-compiler/src/ast/pl/expr.rs
+++ b/prql-compiler/src/ast/pl/expr.rs
@@ -252,7 +252,7 @@ pub struct Pipeline {
 #[derive(Debug, PartialEq, Eq, Clone, Serialize, Deserialize)]
 pub enum InterpolateItem<T = Expr> {
     String(String),
-    Expr(Box<T>),
+    Expr { expr: Box<T>, format: Option<String> },
 }
 
 /// Inclusive-inclusive range.
--- a/prql-compiler/src/ast/pl/fold.rs
+++ b/prql-compiler/src/ast/pl/fold.rs
@@ -176,7 +176,10 @@ pub fn fold_interpolate_item<F: ?Sized + AstFold>(
 ) -> Result<InterpolateItem> {
     Ok(match interpolate_item {
         InterpolateItem::String(string) => InterpolateItem::String(string),
-        InterpolateItem::Expr(expr) => InterpolateItem::Expr(Box::new(fold.fold_expr(*expr)?)),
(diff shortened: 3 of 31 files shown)


max-sixty:
(edit: moving questions to https://github.com/PRQL/prql/issues/2694)

PR #2710, opened 2023-06-03
test(docs): add tests for examples in website hero and repo README

eitsupi:
This PR renames the website tests to docs tests and adds more tests for examples in docs.

### DIFF
--- a/prql-compiler/tests/docs/main.rs
+++ b/prql-compiler/tests/docs/main.rs
@@ -51,6 +51,7 @@ fn website_hero_example() -> String {
 #[test]
 fn test_readme_examples() {
     let contents = fs::read_to_string("../README.md").unwrap();
+    // Similar to code at https://github.com/PRQL/prql/blob/65706a115a84997c608eaeda38b1aef1240fcec3/web/book/tests/snapshot.rs#L152, but specialized for the Readme.
     let re = Regex::new(r"(?s)```(elm|prql)\n(?P<prql>.+?)\n```").unwrap();
     assert_ne!(re.find_iter(&contents).count(), 0);
     for cap in re.captures_iter(&contents) {
--- a/prql-compiler/tests/docs/main.rs
+++ b/prql-compiler/tests/docs/main.rs
@@ -52,6 +52,7 @@ fn website_hero_example() -> String {
 fn test_readme_examples() {
     let contents = fs::read_to_string("../README.md").unwrap();
     let re = Regex::new(r"(?s)```(elm|prql)\n(?P<prql>.+?)\n```").unwrap();
+    assert_ne!(re.find_iter(&contents).count(), 0);
     for cap in re.captures_iter(&contents) {
         let prql = &cap["prql"];
         compile(prql).unwrap();
--- a/prql-compiler/tests/docs/main.rs
+++ b/prql-compiler/tests/docs/main.rs
@@ -49,6 +49,16 @@ fn website_hero_example() -> String {
 }
 
 #[test]
+fn test_readme_examples() {
+    let contents = fs::read_to_string("../README.md").unwrap();
+    let re = Regex::new(r"(?s)```(elm|prql)\n(?P<prql>.+?)\n```").unwrap();
+    for cap in re.captures_iter(&contents) {
+        let prql = &cap["prql"];
+        compile(prql).unwrap();
+    }
+}
+
+#[test]
 fn test_website_examples() {
     for example in website_examples() {
         let prql = example.get("prql").unwrap().as_str().unwrap();
--- a/prql-compiler/tests/docs/main.rs
+++ b/prql-compiler/tests/docs/main.rs
@@ -14,20 +14,38 @@ fn sql_normalize(sql: &str) -> String {
     re.replace_all(sql, " ").trim().to_string()
 }
 
-fn website_examples() -> Vec<Value> {
-    let website_contents = fs::read_to_string(WEBSITE_TOPPAGE)
+fn website_contents() -> Value {
+    let contents = fs::read_to_string(WEBSITE_TOPPAGE)
         .unwrap()
         .replace("---", "");
-    let yaml = serde_yaml::from_str::<Value>(&website_contents).unwrap();
-    let examples = yaml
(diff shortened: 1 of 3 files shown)


eitsupi:
Wow, it seems the new test to have suddenly discovered an error in the README example!

```sh
$ cargo run -p prqlc compile
   Compiling prql-compiler v0.8.1 (/workspaces/prql/prql-compiler)
   Compiling prqlc v0.8.1 (/workspaces/prql/prql-compiler/prqlc)
    Finished dev [unoptimized + debuginfo] target(s) in 41.87s
     Running `target/debug/prqlc compile`
Enter PRQL, then press ctrl-d to compile:

from employees
filter country == "USA"                       # Each line transforms the previous result
aggregate {                                   # `aggregate` reduces each column to a value
  max salary,
  min salary,
  count,                                      # Trailing commas are allowed
}
Error:
   ╭─[Root.prql:6:3]
   │
 6 │   count,                                      # Trailing commas are allowed
   │   ──┬──
   │     ╰──── function std.aggregate, param `columns` expected type `scalar`, but found type `array -> int`
   │
   │ Help: Have you forgotten an argument to function std.count?
   │
   │ Note: Type `scalar` expands to `int || float || bool || text || date || time || timestamp || null`
───╯
```

@aljazerzen Any thoughts?

aljazerzen:
Yeah, I changed `count` to require the `non_null` argument. I don't like it and it breaks a lot of queries, so I intend to revert that before release.

eitsupi:
It seems that the default argument used to be set to `*`.
Is there any reason why this cannot be restored? (Just curious)

https://github.com/PRQL/prql/blob/26cd9af589e4ce21ddf9700ecfbeeabc50f0f878/prql-compiler/src/sql/std_impl.prql#L7

aljazerzen:
I needed it like that because I was changing how types work, and when a function is deemed a window function. It's an implementation thing.

prql-compiler/tests/docs/main.rs
```diff
@@ -0,0 +1,75 @@
...
+        .get("prql_example")
+        .unwrap()
+        .as_str()
+        .unwrap()
+        .to_string()
+}
+
+#[test]
+fn test_readme_examples() {
+    let contents = fs::read_to_string("../README.md").unwrap();
```
max-sixty, line 53: I had thought that these lines were somehow doing this: https://github.com/PRQL/prql/blob/99b82a3afb3deda0168ba4d79226574e0fab8bed/prql-compiler/src/lib.rs#L244-L246

But I guess they aren't... (I'm not even sure how it would be parsing the PRQL. But then I'm also not sure what that code is doing, in that case...)

prql-compiler/tests/docs/main.rs
```diff
@@ -0,0 +1,75 @@
...
+    let re = Regex::new(r"(?s)```(elm|prql)\n(?P<prql>.+?)\n```").unwrap();
+    assert_ne!(re.find_iter(&contents).count(), 0);
```
max-sixty, line 56: Clever regex-ing!

In a way this is similar to https://github.com/PRQL/prql/blob/65706a115a84997c608eaeda38b1aef1240fcec3/web/book/tests/snapshot.rs#L152, but easier to have this separate given they're in different places

prql-compiler/tests/docs/main.rs
```diff
@@ -0,0 +1,75 @@
...
+        .unwrap()
+        .as_str()
+        .unwrap()
+        .to_string()
+}
+
+#[test]
+fn test_readme_examples() {
+    let contents = fs::read_to_string("../README.md").unwrap();
+    let re = Regex::new(r"(?s)```(elm|prql)\n(?P<prql>.+?)\n```").unwrap();
```
max-sixty, line 55:
```suggestion
    // Similar to code at https://github.com/PRQL/prql/blob/65706a115a84997c608eaeda38b1aef1240fcec3/web/book/tests/snapshot.rs#L152, but specialized for the Readme.
    let re = Regex::new(r"(?s)```(elm|prql)\n(?P<prql>.+?)\n```").unwrap();
```

prql-compiler/tests/docs/main.rs
max-sixty: I added a comment with a ref

prql-compiler/tests/docs/main.rs
```diff
@@ -0,0 +1,75 @@
+use prql_compiler::Options;
```
max-sixty, line 1: FWIW Rust unfortunately has poor performance with making lots of test binaries. So it would be faster if we could have this as a file within this test https://github.com/PRQL/prql/blob/65706a115a84997c608eaeda38b1aef1240fcec3/web/book/tests/snapshot.rs (which would change to a module, like this file is)

See the link at https://github.com/PRQL/prql/blob/f4eb8c73918b139aa2e49d1c4e82e1652cf956f4/prql-compiler/tests/integration/README.md#L54-L55

(but we can merge this and make the change / I'll make the change, I don't want to slow you down!)

max-sixty (approved):

prql-compiler/tests/docs/main.rs
eitsupi: I am hardly familiar with the Rust documentation, but perhaps this is a feature for testing Rust code blocks?

prql-compiler/tests/docs/main.rs
eitsupi: I was initially thinking of code block analysis using `pulldown-cmark`, but changed because it seemed like overkill. (And extracting code blocks with `pulldown-cmark` was too difficult for me!)

prql-compiler/tests/docs/main.rs
max-sixty: Ah sorry, that's pulling in the `prql-compiler/README.md` and testing that rust!

prql-compiler/tests/docs/main.rs
max-sixty: Yeah I def think writing _another_ parser wouldn't be great!

This code is a very good balance IMO

PR #2700, opened 2023-06-02
feat: refactoring and type checking operators

aljazerzen:
- fix: lineage of unnamed columns
- fix
- fix
- refactor: resolver
- refactor
- TableDecl.column RelationColumn -> TupleField
- TableDecl columns -> ty
- renaming relations
- refactor: BinOp, UnOp -> RqOperator
- typecheck operators
- types types types
- tests, fmt and clippy

PR #2713, opened 2023-06-03
fix: Try removing `compile-files` for SO

max-sixty:
Something is causing a Stack Overflow on Windows; maaaybe it could be this line? https://github.com/PRQL/prql/pull/2700/files#diff-2e1623f80d59e3d48345bf175c77cecd15f88c8c7edb81a03ead35194b839720R16

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -814,13 +814,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "compile-files"
-version = "0.8.1"
-dependencies = [
- "prql-compiler",
-]
-
-[[package]]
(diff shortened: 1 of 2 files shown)


max-sixty:
This does seem to be the source. Given we don't use the example much, I've disabled `compile-files` while we resolve the `count *` issue. 

I'm happy to have a look at the query more if that's helpful (but I know @aljazerzen is already planning to make changes with the `count *`)

eitsupi:
ref https://github.com/PRQL/prql/pull/2700#issuecomment-1575206792

PR #2719, opened 2023-06-04
test: Consolidate test of examples in a single test binary

max-sixty:
As discussed in https://github.com/PRQL/prql/pull/2710, a small reorg.

Thanks to @eitsupi for writing the code!

(this looks like a big diff but it's actually very small — just moving flies and a couple touch-ups)


max-sixty:
(this looks like a big diff but it's actually very small — just moving flies and a couple touch-ups)

Issue #2622, opened 2023-05-24
Ordering, sorting and nulls


aljazerzen:
The question is: where should `null`s values end up in when `sort`-ing a relation?

A followup for #2617

This justification was made by dissecting "what sorting is" down to formal mathematical (unwritten) definition of relations and PRQL. So bear with me, it may look like it's getting way to broad and abstract, but I believe that semantics of the language will remain as concise and orthogonal as they are now only if we really know what we are designing.

----

As a start, I'd suggest that instead of thinking about where the NULLs should be, we design this by figuring out what comparison operation should return for null values and derive the behavior of sort from that.

So, these are the key questions:
```
1    < 1     ->   false  (obviously)
1    < 2     ->   true   (obviously)
1    < null  ->   ?
2    < 1     ->   false  (obviously)
2    < 2     ->   false  (obviously)
2    < null  ->   ?
null < 1     ->   ?
null < 2     ->   ?
null < null  ->   ?
```

There are three ways of filling this out:
1. `null < x` for every non-null x,
1. `x < null` for every non-null x,
1. `false` for all comparisons to null.

Option 1. and 2. are straight forward: if null is less than every other value, this translates to DESC NULLS LAST and ASC NULLS FIRST. If null is greater than every other value, it's the other way around.

But option 3 is interesting. What happens if null is not less, greater or equal to any number?

This is (of course) well researched in math and it is named a (strict) partial ordering. The `<` operator (they call it 'relation', but I'm not going to do that to the reader) must be [ireflexive, anti-symmetric and transitive](https://en.wikipedia.org/wiki/Partially_ordered_set#Partial_order). With these properties it makes sense to define that a *"sorted sequence"* has every pair `a < b` in such positions that a is before b.

But this partial ordering does not imply that there is a single unique *sorted sequence*.

For example, take a set of products and a operator "is made from". You can have a chain "wood < wooden board < desk". But there is also "wood < wooden pillar < house". So these are two separate sorted chains with a common predecessor. But there is no relation between desk and house, so they cannot be merged into a single chain. Nevertheless, you can construct a *sorted sequence* such that every element comes after all its components:

```
wood, wooden board, desk, wooden pillar, house
```

The *sorted sequence* of this ordering is not unique as you could also have: 

```
wood, wooden board, wooden pillar, desk, house
```

... and the desired property would still hold.

[...]

PR #2720, opened 2023-06-04
fix: tweak div operators

eitsupi:
Based on the discussion of #2426.
If my understanding is correct, some definitions need to be modified and some can be simplified.

These definitely need integration tests......

- ~~In SQL, `ROUND` seems rounding off (not banker's rounding), so `ROUND({l} / {r} - 0.5)` is sufficient for truncated div for generic. (Of course, this does not mean anything other than SQL!)~~ -> Sorry, I am mistaken...
- In Postgres, `TRUNC({l} / {r})` seems always returning int, not float.
- In MySQL, the `DIV` operator can be used for int div.
- In MySQL or DuckDB 0.8, the `/` operator can be used for float div.


eitsupi:
DuckDB's `TRUNC` function needs duckdb 0.8... (duckdb/duckdb#7036)

max-sixty:
> DuckDB's `TRUNC` function needs duckdb 0.8... ([duckdb/duckdb#7036](https://github.com/duckdb/duckdb/pull/7036))

😬 😬

😀 

Maybe we discuss the broader issue on the call tomorrow. If we can decide on a way forward on #2622, then we can upgrade...

---

So we can test this, can we merge use the generic implementation for DuckDB?

eitsupi:
I don't know if this is an issue that needs to be corrected.
(Of course, at least the test needs to be updated.)

```log
failures:

---- test_rdbms stdout ----
thread 'test_rdbms' panicked at 'assertion failed: `(left == right)`: duckdb == mysql: arithmetic

Diff < left / right > :
 [
     [
<        "2.6",
<        "2.6",
<        "2.6",
<        "2.6",
>        "2.6000",
>        "2.6000",
>        "2.60000",
>        "2.60000",
         "2",
         "2",
         "2",
         "2",
         "3",
         "3",
         "3",
         "3",
         "13",
         "13",
         "13",
         "13",
     ],
     [
<        "-2.6",
<        "-2.6",
<        "-2.6",
<        "-2.6",
>        "-2.6000",
>        "-2.6000",
>        "-2.60000",
>        "-2.60000",
         "-2",
         "-2",
         "-2",
         "-2",
         "-3",
         "-3",
         "-3",
         "-3",
         "-13",
         "-13",
         "-13",
         "-13",
     ],
     [
<        "-2.6",
<        "-2.6",
<        "-2.6",
<        "-2.6",
>        "-2.6000",
>        "-2.6000",
>        "-2.60000",
>        "-2.60000",
         "-2",
         "-2",
         "-2",
         "-2",
         "3",
         "3",
         "3",
         "3",
         "13",
         "13",
         "13",
         "13",
     ],
     [
<        "2.6",
<        "2.6",
<        "2.6",
<        "2.6",
>        "2.6000",
>        "2.6000",
>        "2.60000",
>        "2.60000",
         "2",
         "2",
         "2",
         "2",
         "-3",
         "-3",
         "-3",
         "-3",
         "-13",
         "-13",
         "-13",
         "-13",
     ],
 ]
```

https://github.com/PRQL/prql/actions/runs/5167751885/jobs/9308839628?pr=2720#step:11:222

eitsupi:
MySQL results seem to be affected by system settings:

> In division performed with [/](https://dev.mysql.com/doc/refman/5.7/en/arithmetic-functions.html#operator_divide), the scale of the result when using two exact-value operands is the scale of the first operand plus the value of the [div_precision_increment](https://dev.mysql.com/doc/refman/5.7/en/server-system-variables.html#sysvar_div_precision_increment) system variable (which is 4 by default). For example, the result of the expression 5.05 / 0.014 has a scale of six decimal places (360.714286).

https://dev.mysql.com/doc/refman/5.7/en/arithmetic-functions.html

prql-compiler/tests/integration/docker-compose.yml
```diff
@@ -1,4 +1,3 @@
-version: "2.4"
```
eitsupi, line 1: ref: https://github.com/PRQL/prql/pull/2684#discussion_r1215076974

max-sixty:
> MySQL results seem to be affected by system settings:

Can we set these for the integration tests? Assuming it's between 2.60 & 2.600, that's fine for overriding in the tests...

eitsupi:
> Can we set these for the integration tests? Assuming it's between 2.60 & 2.600, that's fine for overriding in the tests...

Yes, it can be set up like 796b68a3638075809553d5e28c6cc995597932f8.
But I don't think we can match all the results with DuckDB, etc. in all tests because of the effect of the the first operand of digits...

It would be nice if the results could be compared numerically rather than in a string representation.

eitsupi:
Is the easiest way is excluding trailing zeros?

max-sixty:
> It would be nice if the results could be compared numerically rather than in a string representation.

Yes...

To the extent there's a way to hack around the test without discarding the result (i.e. we're OK between 26.0 and 26.00, but not between 26.0 and 26.5), that's very acceptable — e.g. we could add a function converting to strings in the SQL. Even if it's not elegant, it would work...

eitsupi:
It seems to have passed the test by removing the trailing 0.
I think this is change something like `26.0` to `26.` in the tests, but probably not a problem.

prql-compiler/tests/integration/main.rs
```diff
@@ -212,3 +214,15 @@ fn replace_booleans(rows: &mut Vec<Row>) {
...
     }
 }
+
+// MySQL may adds 0s to the end of results of `/` operator
+fn remove_trailing_zeros(rows: &mut Vec<Row>) {
+    let re = Regex::new(r"^(|-)\d+\.\d+0+$").unwrap();
+    for row in rows {
+        for col in row {
+            if re.is_match(col) {
+                *col = Regex::new(r"0+$").unwrap().replace_all(col, "").to_string();
```
max-sixty, line 224: Looks reasonable as-is. Can it just be a single `\.0+$` regex for simplicity, without an `if re.is_match`?

max-sixty:
Looks good! (I haven't thought through the exact operations but lmk if you want me to)

prql-compiler/tests/integration/docker-compose.yml
aljazerzen: Thanks!

prql-compiler/src/sql/std.sql.prql
```diff
@@ -49,13 +49,13 @@ let read_csv = source -> s"read_csv_auto({source:0})"
...
-let div_i = l r -> s"FLOOR(FLOOR(ABS({l:11} / {r:11})) * SIGN({l:11} / {r:11}))"
+let div_i = l r -> s"FLOOR(ABS({l:11} / {r:11})) * SIGN({l:0}) * SIGN({r:0})"
 
 #[binding_strength 11]
-let div_f = l r -> s"({l} * 1.0 / {r})"
+let div_f = l r -> s"({l:11} * 1.0 / {r:11})"
 
 #[binding_strength 11]
-let mod = l r -> s"{l} % {r}"
+let mod = l r -> s"{l:11} % {r:11}"
```
aljazerzen: When an interpolation does not have the binding strength specified it defaults to binding strength of the parent.

This was meant for simple cases like this one, where the whole s-string is just one binary operation.

Is behavior confusing? We can remove this defaulting and require all interpolations to have the strength specified. Or we can default to 100, which would insert parenthesis?

eitsupi: Sorry for unnecessary change. I reverted it.

prql-compiler/tests/integration/main.rs
eitsupi: I think it is worth clarifying that only numbers are covered, but I have no strong opinion.

prql-compiler/tests/integration/main.rs
```diff
@@ -212,3 +214,15 @@ fn replace_booleans(rows: &mut Vec<Row>) {
         }
     }
 }
+
+// MySQL may adds 0s to the end of results of `/` operator
+fn remove_trailing_zeros(rows: &mut Vec<Row>) {
```
aljazerzen, line 219: This is almost cheating... The `/` operator in MySQL returns DECIMAL, which appends the 0. It does transparently convert to INTEGER when needed, at least in all of my testing.

If sometime later we find that MySQL has a problem related to decimals, this is it.

eitsupi: I don't know if we should dare to use the less accurate FLOAT type.
Shouldn't the comparison be based on the accuracy of the actual numbers rather than the display method, and if the issue here is only display, then why not just use the DECIMAL type?

max-sixty: Is it a decimal vs a float, or a decimal vs an integer?

The float / decimal distinction seems less significant (though I don't have much experience with decimals...)

max-sixty: And we can always reduce the support level of some of these dialects if they're taking time away from other things — MySql & MsSql are top of the chopping block!

eitsupi: Personally, I think it would be great if MySQL is supported since ClickHouse and Detabend are MySQL compatible.
(By the way I just found out today that Detabend has the `//` operator. Haven't checked to see how it works.)

I feel that MSSQL has a significant disadvantage to support because it cannot use `USING`...

prql-compiler/src/sql/std.sql.prql
aljazerzen: No, no, I was asking for feedback on the new std.sql.prql format. It may have sounded passive aggressive, but it wasn't meant that way.

eitsupi: Ah, thanks for the explanation.
I simply did not understand the mechanism when I made this change, but it certainly seems safe and reasonable to set the default to 100.
I do not have a strong opinion, however.

prql-compiler/tests/integration/main.rs
aljazerzen: Oh, I though this was int vs decimal, but it's float vs decimal. My bad.

This is the related rule, (probably) https://github.com/PRQL/prql/pull/2720/files#diff-f6d964f7215b03bfdb696d81cf4258db7cf5686558bf44fa627d031f2f474c2eR125

The more experimentation I did, the more I find that MySQL prefers using decimals over floats. Nothing wrong with that, but it's different than most other engines.

(decimals are basically fixed-point numbers. stored as integers + the point offset)

Well in anycase, this should not be a problem in the near future, just something to keep in mind.

max-sixty:
On the dev call we decided to simplify the Generic implementation, particularly for `/`, so the examples on the website are easy to understand.

Shall we merge this and then I can make that change?

Issue #2711, opened 2023-06-03
Sorting does not seem to be considered when use `DISTINCT ON`


eitsupi:
### What happened?

From https://github.com/PRQL/prql/pull/2708#discussion_r1215611454

I used the latest dev version (5886a758b66422a1a398dcbad7c7deaa43022b90).

### PRQL input

```elm
prql target:sql.duckdb

from employees
group department (
  sort age
  take 1
)
```

### SQL output

```SQL
SELECT
  DISTINCT ON (department) *
FROM
  employees
```

### Expected SQL output

```SQL
SELECT
  DISTINCT ON (department) *
FROM
  employees
ORDER BY
  department,
  age
```

### MVCE confirmation

- [X] Minimal example
- [X] New issue

aljazerzen:
Was distinct on even tested in a database?

max-sixty:
This is my fault! I'm looking now

max-sixty:
Closed by #2715

eitsupi:
Thanks!

PR #2722, opened 2023-06-04
test: DISTINCT ON

aljazerzen:
Ref #2711

Adds a test for DISTINCT ON. It also uncovered two bugs:
- When `group x (sort y | take 1)` is followed by a `sort`, the outer sort would
  override the inner one. This case is now split into two atomics and has two
  ORDER BY clauses: one for the DISTINCT ON and one for the resulting relation.
- ORDER BY that is associated with DISTINCT ON has to include the partition columns.
  Postgres throws an error if this is not the case, while DuckDB silently ignores
  the ORDER BY.


prql-compiler/src/tests/test.rs
```diff
@@ -1507,6 +1507,7 @@ fn test_distinct_on() {
     FROM
       employees
     ORDER BY
+      department,
```
max-sixty, line 1510: Thanks, I didn't realize this was required. On reflection I should have added an integration test given I'm less familiar with this construct. (And I think that's a very reasonable thing for us to ask each other to do, so feel free to prod me on that sort of thing)

aljazerzen: Yes, I agree - let's be more diligent here.

We should also be converting the units tests we have into integration tests. Last few times that I did that it unearthed new bugs.

max-sixty: Yes great.

(i do think the unit tests are still useful for the quick iterations — it's much easier to debug why bad SQL is being produced than why we're getting bad query results — but agree that there's plenty of room to expand the integration tests)

PR #2729, opened 2023-06-05
refactor: Change annotation syntax to use `@`

max-sixty:
This is not great code — likely needs some error handling, and maybe some changes to how it's implemented in Chumsky. But it works!


snth:
Sorry, there's no linked issue so asking here: was there a discussion around `#[...]` vs `@{...}` syntax for annotations?

No preference from my side at this stage but just curious what the considerations would be.

aljazerzen (approved):
Makes sense. Annotations can now be arbitrary expressions - literals, function calls, tuples, lists.

And SQL backend expects only tuples.

aljazerzen:
We mentioned here: #2694

snth:
Some thoughts that came to me on this are that

- to the extent that the annotations are key-value pairs or tuples, it makes that we use our `{}` syntax for those, and
- if you think of them as ATtributes, then it makes sense to use the `@` symbol as a useful mnemonic.

So `@{...}` seems like a great choice.

web/book/tests/documentation/snapshots/documentation__book__language-features__regex__regex-expressions__3.snap
```diff
@@ -7,5 +7,5 @@ SELECT
 FROM
   tracks
 WHERE
-  name ~ '\(I Can''t Help\) Falling'
+  (name) ~ ('\(I Can''t Help\) Falling')
```
eitsupi, line 10: These changes are intentional?

max-sixty: No, I saw these too — I'm not sure why they happen...

max-sixty:
I cleaned up the code a bit, and merging. 

I don't know why we're getting these extra parentheses at https://github.com/PRQL/prql/pull/2729#discussion_r1217878059 — I don't even see where it could be affecting this! Not super important so I won't spend time on it, lmk any dissent

web/book/tests/documentation/snapshots/documentation__book__language-features__regex__regex-expressions__3.snap
aljazerzen: You've missed a few annotations.

https://github.com/PRQL/prql/pull/2729/files#diff-f6d964f7215b03bfdb696d81cf4258db7cf5686558bf44fa627d031f2f474c2eR100-R150

max-sixty: Ah sorry!

(I blame the comment syntax :D )

vanillajonathan:
In the future could there be other types of annotations?
Could custom annotations be declared?

Maybe `@Binding{Strength=4}` would be preferred over `@{binding_strength=11}` in that case.

The syntax is rather unorthodox, hence I think using square brackets instead of curly brackets would be preferred as it would be more familiar.

aljazerzen:
Syntax for annotations is just:

```
Annotation :== '@' Expr
```
... where `Expr` can be any expression. We are just using a tuple here. You could use `[]` and have arrays.

In any case, this was added mostly as an internal detail. It is not documented and especially `binding_strength` is not part of the language, so it can be changed without notice. So there is no need to make it future proof.

PR #2739, opened 2023-06-05
feat: Use simple syntax for generic dialect

max-sixty:
As discussed on the dev call and at https://github.com/PRQL/prql/pull/2720#event-9428609734

Unfortunately this will fail tests at the moment, since I don't think integration tests use generic-specific dialects. We'll need to update that first.

### DIFF
--- a/web/book/tests/documentation/snapshots/documentation__book__queries__functions__piping__2.snap
+++ b/web/book/tests/documentation/snapshots/documentation__book__queries__functions__piping__2.snap
@@ -4,9 +4,7 @@ expression: "let fahrenheit_to_celsius = temp -> (temp - 32) / 1.8\nlet interp =
 ---
 SELECT
   *,
-  (
-    (((temp_c - 32) * 1.0 / 1.8) - 0) * 1.0 / (100 - 0)
-  ) AS boiling_proportion
+  ((((temp_c - 32) / 1.8) - 0) / (100 - 0)) AS boiling_proportion
 FROM
   kettles
 
--- a/web/website/content/_index.md
+++ b/web/website/content/_index.md
@@ -166,13 +166,11 @@ showcase_section:
         }
 
       sql: |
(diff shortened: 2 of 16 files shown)


eitsupi:
> Unfortunately this will fail tests at the moment, since I don't think integration tests use generic-specific dialects.

I think this is because the integration test includes some dialects that use Generic's div_f.

I set div_f for all the dialects in 231d6a4. (And I sorted them for clarity.)

web/website/content/_index.md
```diff
@@ -169,8 +175,10 @@ showcase_section:
         SELECT
           *,
           started + unfinished AS finished,
-          ((started + unfinished) * 1.0 / started) AS fin_share,
-          ( ((started + unfinished) * 1.0 / started) * 1.0 / (1 - ((started + unfinished) * 1.0 / started)) ) AS fin_ratio
```
eitsupi, line 173: I replaced this with the current results in e355722 because I was concerned about the space.

max-sixty:
@eitsupi that's much much better, thank you!

web/book/tests/documentation/snapshots/documentation__book__queries__functions__piping__2.snap
```diff
@@ -4,9 +4,7 @@ expression: "let fahrenheit_to_celsius = temp -> (temp - 32) / 1.8\nlet interp =
 ---
 SELECT
   *,
-  (
-    (((temp_c - 32) * 1.0 / 1.8) - 0) * 1.0 / (100 - 0)
-  ) AS boiling_proportion
+  ((((temp_c - 32) / 1.8) - 0) / (100 - 0)) AS boiling_proportion
```
max-sixty, line 7: I know we removed arithmetic from PRQL, but this would be cool to simplify

("cool", not necessarily "important"!)

PR #2741, opened 2023-06-05
test: Add comments re CRLF, make test robust

max-sixty:
- Removes the CI hack in Elixir tests, as that should no longer be needed, and hide the test failing for folks developing on Windows
- Makes the Readme tests robust to either scheme
- Adds some comments

Overall not sure what's best to do here — make tests robust to either vs. mandate one. Probably the first is better, so we don't depend on local configs, but also more work. The current state is very reasonable AFAIU.


eitsupi:
I sometimes use Windows, but I turn off the meddlesome automatic LF and CRLF conversion by Git.

The automatic conversion between CRLF and LF really doesn't help, since software like Notepad in current Windows works fine with LF instead of CRLF, I think.

PR #2744, opened 2023-06-06
test: Snapshot integration test queries refinements

max-sixty:
[edit: Renamed from "test: Snapshot integration test queries by dialect"]


prql-compiler/tests/integration/snapshots/integration__sql_DuckDb@pipelines.prql.snap
```diff
@@ -0,0 +1,28 @@
+---
+source: prql-compiler/tests/integration/main.rs
+expression: "# skip_mssql (https://github.com/PRQL/prql/issues/2448)\n# skip_mysql (regex not yet implemented)\n# skip_sqlite (regex not yet implemented)\n# booleans cannot be selected directly. instead CASE is needed (https://dba.stackexchange.com/a/2774)\nfrom tracks\n\nfilter (name ~= \"Love\")\nfilter ((milliseconds / 1000 / 60) | in 3..4)\nsort track_id\ntake 1..15\nselect {name, composer}\n"
```
eitsupi: This test can be enabled for mysql now?

https://github.com/PRQL/prql/blob/8128a134f7faa4bf474a2ab51a36652ae03b1f12/prql-compiler/src/sql/std.sql.prql#L156

aljazerzen:
This is quite a lot of snapshots. Are we sure that we even need snapshots here? It gets tedious accepting ~100 reviews when changing some minor SQL generation aspect.

max-sixty:
> This is quite a lot of snapshots. Are we sure that we even need snapshots here? It gets tedious accepting ~100 reviews when changing some minor SQL generation aspect.

I somewhat agree. But I also see us changing functions and not being able to see the impact of the changes when reviewing.

I also thought about only showing the diffs from the generic output. Or we could put all dialects in a single doc? Do any of those sound better?

(We can also merge and then later undo if we don't find them useful after a while)

aljazerzen:
Could we just show the generated SQL in the error message of the failing test?

max-sixty:
> Could we just show the generated SQL in the error message of the failing test?

If it fails! But don't we also want to see the output of the SQL?

Or maybe not — maybe the result passing is sufficient?!

aljazerzen:
This is "what's the output of using PRQL: SQL or query results?" again :D And I vote for query results only.

max-sixty:
OK, let's try for a bit longer without each dialect. I reverted that while keeping some internal changes

prql-compiler/src/sql/dialect.rs
```diff
@@ -75,12 +75,32 @@ impl Dialect {
         }
     }
 
+    pub fn support_level(&self) -> SupportLevel {
```
eitsupi, line 78: Love this!

max-sixty:
OK merging!

This was overly refine-y, but I actually learnt a bit about how to use Traits / Structs / Enums together, I think the rust code is pretty good there now (if it's not, lmk, otherwise I'm going to take this lesson elsewhere!)

PR #2745, opened 2023-06-06
test: Enable MySql regex tests

max-sixty:
Can't enable Sqlite, notes inline


eitsupi:
Seems working fine! 🎉

PR #2748, opened 2023-06-06
feat: Use code indentation of 2

max-sixty:
As discussed on the dev call...

No religious view (generally I find 4 good in many langs), though I do think with compact queries, 2 is very elegant


eitsupi:
Note that we can use `.editorconfig` to enforce indent size to all `.prql` files. (And check by MegaLinter on CI)

max-sixty:
> Note that we can use `.editorconfig` to enforce indent size to all `.prql` files. (And check by MegaLinter on CI)

And hopefully a pre-commit step of `prqlc fmt`!

PR #2751, opened 2023-06-07
feat: Allow `prqlc` to print backtraces on errors

max-sixty:
This can be helpful when debugging. There are some notes inline on the implementation; feedback welcome.


prql-compiler/prqlc/tests/test.rs
```diff
@@ -186,10 +186,12 @@ fn project_path() -> PathBuf {
 
 fn prqlc_command() -> Command {
     let mut cmd = Command::new(get_cargo_bin("prqlc"));
-    // We set this in CI to force color output, but we don't want `prqlc` to
+    // We set `CLICOLOR_FORCE` in CI to force color output, but we don't want `prqlc` to
     // output color for our snapshot tests. And it seems to override the
     // `--color=never` flag.
     cmd.env_remove("CLICOLOR_FORCE");
+    // We don't want the tests to be affected by the user's `RUST_BACKTRACE` setting.
+    cmd.env_remove("RUST_BACKTRACE");
```
aljazerzen, line 194: can you also add `RUST_LOG`?

prql-compiler/prqlc/src/cli.rs
```diff
@@ -27,6 +29,17 @@ pub fn main() -> color_eyre::eyre::Result<()> {
 
     if let Err(error) = cli.command.run() {
         eprintln!("{error}");
+        // TODO: this is clunky — we're using `backtrace.status()` to check if
+        // backtraces are enabled, but then printing the backtrace from the
+        // error. Is there a way to check whether backtrades are enabled without
+        // capturing another one; e.g. from the error itself?
```
aljazerzen: So this is equivalent to looking up the `RUST_BACKTRACE` env var?

max-sixty: Ah, yes, probably. I thought this might be exposed directly, or encapsulated somewhere, but maybe we should just check that

prql-compiler/prqlc/tests/test.rs
max-sixty: Done

prql-compiler/prqlc/src/cli.rs
aljazerzen: Looking at the source of `backtrace::Backtrace::enabled()`, it does check another env var, but we'd probably be find by just checking `RUST_BACKTRACE`.

max-sixty: Ah, I somehow missed that — let's use that!

edit: Oh, it's private. I'll vendor it

PR #2755, opened 2023-06-07
fix: Consistent color handling in outputs

max-sixty:
I started trying to get `prqlc fmt` to show colors correctly, went yak shaving, and now stopping to commit what I have. I'm not really sure the best way of handling this:
- Ideally we want to use the auto detection that `concolor` provides
- Ideally we also want to be able to set the color mode explicitly to `prql-compiler`; e.g. from bindings
- Ariandne takes a bool — but also promises to use `concolor`, so I'm not sure how it that's supposed to work
- We currently take a bool (not `Option<bool>`), which doesn't leave space for "auto" mode, and that's a supported interface in our bindings. We could set the global option on each call, seems inelegant, but would probably work


max-sixty:
Usurped by #2773

PR #2756, opened 2023-06-07
docs(book): update some sections about std lib

eitsupi:
Minor updates to documentation.

I really wanted to write about the division operator, but I couldn't decide where to put it because the pages about operator were separated into `language-features/standard-library` and `syntax/expressions-and-operators.md`.

Perhaps it might be a good to rename `syntax/expressions-and-operators.md` and move the information on operators to `language-features/standard-library`.

For example, in the DuckDB documentation, operators are introduced with functions.
https://duckdb.org/docs/sql/functions/overview

### DIFF
--- a/web/book/src/language-features/standard-library/README.md
+++ b/web/book/src/language-features/standard-library/README.md
@@ -7,14 +7,13 @@ Currently s-strings are an escape-hatch for any function that isn't in our
 standard library. If we find ourselves using them for something frequently,
 raise an issue and we'll add it to the stdlib.
 
-```admonish note
-Currently the stdlib implementation doesn't support different DB implementations itself;
-those need to be built deeper into the compiler. We'll resolve this at some point. Until
-then, we'll only add functions here that are broadly supported by most DBs.
-```
-
 Here's the source of the current
 [PRQL `std`](https://github.com/PRQL/prql/blob/main/prql-compiler/src/semantic/std.prql):
+
+```admonish note
+PRQL 0.9.0 has started supporting different DB implementations for standard library functions.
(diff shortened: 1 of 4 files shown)


web/book/src/language-features/standard-library/README.md
```diff
@@ -7,15 +7,14 @@ Currently s-strings are an escape-hatch for any function that isn't in our
...
-those need to be built deeper into the compiler. We'll resolve this at some point. Until
-then, we'll only add functions here that are broadly supported by most DBs.
-```
-
 Here's the source of the current
 [PRQL `std`](https://github.com/PRQL/prql/blob/main/prql-compiler/src/semantic/std.prql):
 
+```admonish note
+PRQL 0.9.0 has started supporting different DB implementations for standard library functions.
+The source is the [`std.sql`](https://github.com/PRQL/prql/blob/main/prql-compiler/src/sql/std.sql.prql).
```
max-sixty, line 15: Nice, thank you!

max-sixty (approved):

max-sixty:
Do we have int division somewhere? We could add to https://github.com/PRQL/prql/blob/65706a115a84997c608eaeda38b1aef1240fcec3/web/book/src/syntax/README.md#L26?



> Perhaps it might be a good to rename `syntax/expressions-and-operators.md` and move the information on operators to `language-features/standard-library`.

We went back & forth on this a few times a few months ago. The content on that page — Precedence & Parentheses — don't really fit with the standard library, though it would fine to rename it; maybe it's something like "syntax details" etc?

eitsupi:
Thank you for taking a look at this.
I will merge this for now and take a look at it latter. (Feel free to change it if I didn't get to it)

PR #2759, opened 2023-06-07
ci: Try upgrading docker-compose in CI

max-sixty:
From discord:

This is caused by docker-compose v2.18.0, which is on the new version of runners

https://docs.docker.com/compose/release-notes/

v2.18.1 fixes it. But not sure how that gets released to the runners.


max-sixty:
@PRQL/prql-team  FYI I had to disable the external DB integration tests — we can reenable, hopefully soon, when the GHA runner upgrades.

Unless someone wants to get docker compose upgraded — but seems like a low productivity task (and not fun...)

PR #2767, opened 2023-06-08
docs: Update main crate docstring

max-sixty:
I think `prql-compiler-macros` is now published. And I refined the style a bit.

Does the `compile-files` example need a macro within it? It complicates the example slightly — we're saying "if you want to compile files, then use `prql-compiler`, rather than `prql-compiler-macros` but write your own macro"??

### DIFF
--- a/prql-compiler/src/lib.rs
+++ b/prql-compiler/src/lib.rs
@@ -49,7 +49,7 @@
 //!
 //! - Compile PRQL queries to SQL at build time.
 //!
-//!     Use the `prql-compiler-macros` crate; for example:
+//!     For inline strings, use the `prql-compiler-macros` crate; for example:
 //!     ```ignore
 //!     let sql: &str = prql_to_sql!("from albums | select {title, artist_id}");
 //!     ```
--- a/prql-compiler/src/lib.rs
+++ b/prql-compiler/src/lib.rs
@@ -1,5 +1,5 @@
-//! Compiler for PRQL language.
-//! Targets SQL and exposes PL and RQ abstract syntax trees.
+//! Compiler for PRQL language. Targets SQL and exposes PL and RQ abstract
+//! syntax trees.
 //!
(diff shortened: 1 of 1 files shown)


prql-compiler/src/lib.rs
```diff
@@ -49,23 +47,21 @@
 //!     # }
 //!     ```
 //!
-//! - ... compile PRQL queries to SQL at build time.
+//! - Compile PRQL queries to SQL at build time.
 //!
-//!     Use `prql-compiler-macros` crate (unreleased), which can be used like
-//!     this:
+//!     Use the `prql-compiler-macros` crate; for example:
```
aljazerzen:
```suggestion
//!     For inline strings, use the `prql-compiler-macros` crate; for example:
```

prql-compiler/src/lib.rs
```diff
@@ -49,23 +47,21 @@
...
 //!     ```ignore
 //!     let sql: &str = prql_to_sql!("from albums | select {title, artist_id}");
 //!     ```
 //!
-//! - ... compile .prql files to .sql files at build time.
+//!     Alternatively, to compile `.prql` files to `.sql` files at build time,
+//!     call `prql-compiler` from `build.rs`. See [this example
+//!     project](https://github.com/PRQL/prql/tree/main/prql-compiler/examples/compile-files).
```
aljazerzen:
```suggestion
//!     For compiling whole files (`.prql` to `.sql`), call `prql-compiler`
//!     from `build.rs`.
//!     See [this example project](https://github.com/PRQL/prql/tree/main/prql-compiler/examples/compile-files).
```

aljazerzen (approved):
Nice.

prql-compiler/src/lib.rs
max-sixty: Good suggestions!

PR #2768, opened 2023-06-08
docs: Move Terminology

max-sixty:
I'm not sure where this should go, but its current place I don't think is that coherent any longer.

I also removed the index in the book for Internals — I think better to have no index than an out-of-date one. Would be great if there were a tool to build this automatically!

### DIFF
--- a/web/book/src/internals/README.md
+++ b/web/book/src/internals/README.md
@@ -3,10 +3,22 @@
 This chapter explains PRQL's semantics: how expressions are interpreted and
 their meaning. It's intended for advanced users and compiler contributors.
 
-- [Name resolving](./name-resolving.md)
-- [Functions](./functional-lang.md)
-- [Syntax highlighting](./syntax-highlighting.md)
-
 It's also worth checking out the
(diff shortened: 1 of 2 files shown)


web/book/src/internals/README.md
```diff
@@ -3,10 +3,22 @@
...
+
+[_Relation_](<https://en.wikipedia.org/wiki/Relation_(database)>): Standard
+definition of a relation in context of databases:
+
+- An ordered set of tuples of form `(d_0, d_1, d_2, ...)`.
+- Set of all `d_x` is called an attribute or a column. It has a name and a type
+  domain `D_x`.
+
+_Frame_: descriptor of a relation. Contains list of columns (with names and
+types). Does not contain data.
```
aljazerzen:
```suggestion
```

aljazerzen: This it no longer named frame, but lineage. And is in the process of being merged with `Ty`, so it can be removed from here.

aljazerzen (approved):
Eventually, we may want to collect formal specifications of PRQL and include this too. For now, this is the best place for it.

PR #2777, opened 2023-06-10
fix: `first` & `last` functions

### LINKED ISSUE #2775: LAST_VALUE() and FIRST_VALUES() doesn't take "offset" argument
### What happened?

Hi,

I appear, at least on postgresql, LAST_VALUE() AND FIRST_VALUE() window functions doesn't accept an offset argument like https://github.com/PRQL/prql/blob/main/prql-compiler/src/sql/std.sql.prql#L31 suggest.

### PRQL input

```elm
from t
derive diff = (last index) - index
```

### SQL output

```SQL
Error:
    ╭─[Root.prql:11:16]
    │
 11 │ derive diff = (last index) - index
    │                ─────┬────
    │                     ╰────── function std.sub, param `left` expected type `int || float || timestamp || date`, but found type `array -> infer`
    │
    │ Help: Have you forgotten an argument to function std.last?
────╯
```

### Expected SQL output

```SQL
SELECT
  *,
  LAST_VALUE(index) OVER (ROWS BETWEEN UNBOUNDED PRECEDING AND 1 PRECEDING) - index AS diff
FROM
  t
```

### MVCE confirmation

- [X] Minimal example
- [X] New issue

### Anything else?

Also wonder if it's possible to workaround by defining custom window function as prql functions using s-strings, but  I didn't manage to get it working.

max-sixty:
Closes #2775. But we should add some integration tests too at some point...


prql-compiler/src/sql/std.sql.prql
```diff
@@ -27,8 +27,8 @@ let count_distinct = column -> s"COUNT(DISTINCT {column:0})"
 # Window functions
 let lag = offset column -> s"LAG({column:0}, {offset:0})"
 let lead = offset column -> s"LEAD({column:0}, {offset:0})"
-let first = offset column -> s"FIRST_VALUE({column:0}, {offset:0})"
-let last = offset column -> s"LAST_VALUE({column:0}, {offset:0})"
+let first = column -> s"FIRST_VALUE({column:0})"
+let last = column -> s"LAST_VALUE({column:0})"
```
eitsupi, line 31: Maybe the `RESPECT NULLS` and `IGNORE NULLS` options are needed?

> The SQL standard defines a RESPECT NULLS or IGNORE NULLS option for lead, lag, first_value, last_value, and nth_value. This is not implemented in PostgreSQL: the behavior is always the same as the standard's default, namely RESPECT NULLS. Likewise, the standard's FROM FIRST or FROM LAST option for nth_value is not implemented: only the default FROM FIRST behavior is supported. (You can achieve the result of FROM LAST by reversing the ORDER BY ordering.)

https://www.postgresql.org/docs/current/functions-window.html

max-sixty: WDYM by "needed"? as in they'd be helpful or we need to add them now?

eitsupi: Sorry, of course there is no need to add it now, a follow up issue.
But I think this option is very important and related to #2622.

max-sixty: OK great! Sorry for misunderstanding at first

Issue #2639, opened 2023-05-26
Could we reduce the cost of the `prqlc` build?


max-sixty:
### What's up?

I do think building prqlc artifacts is a good idea — and it's already found one bug. 

But it is more than half the CI capacity of each PR. It also causes the cache to churn a bit more (but I don't think a single run's cache is bigger than our capacity yet, which would cause pathological behavior, so not urgent like that would be). I've noticed PRs take much longer to test when it's busy — e.g. this afternoon — since the queue builds up.

A couple of ideas to reduce the cost:
- Compile in debug rather than release mode for PRs?
- If we do that, share the standard rust cache for that platform, which should mean compiling is almost free, which is almost all the runtime.
- If we don't compile in debug mode, switch to running on `test-all` only, and then we can use the `pr-test-all` label to generate the binaries?

CC @eitsupi (but not super important or urgent, or necessarily your responsibility to solve!)

eitsupi:
Sorry for bothering you.

Currently this job is running for all PRs.
This is the same of the `test-rust` job.

https://github.com/PRQL/prql/blob/d786a8b4969a2f7423cbffe4bffef2bc84b982fa/.github/workflows/pull-request.yaml#L4-L37

In my opinion, there is no need to run these jobs if the prql-compiler source code has not been modified.
How about separating `test-rust` and `build-prqlc` from this workflow and changing condition to run only if the prql-compiler code is changed in the PR?
This would reduce the number of workflows and execution time for small changes such as pre-release chores.

max-sixty:
> How about separating `test-rust` and `build-prqlc` from this workflow and changing condition to run only if the prql-compiler code is changed in the PR?
> This would reduce the number of workflows and execution time for small changes such as pre-release chores.

This would be helpful on the margin, yes. 

Though the majority of PRs do change some rust code, so I don't think it solves everything.

What do you think about the other ideas — e.g. compiling in debug mode & sharing a cache?

eitsupi:
> What do you think about the other ideas — e.g. compiling in debug mode & sharing a cache?

Do you think that means that the build is done by different commands in the releases and PRs?

Perhaps PR should build in debug mode and build-all and release should build in release mode?

max-sixty:
On running only the rust test workflows on rust changes — unfortunately I don't think this is possible:
- If a job is marked as "Required", it's always required, and nothing can merge without it passing.
- We want to have that job be required, so we can click "Enable auto-merge"
[...]

PR #2778, opened 2023-06-10
ci: Build `prqlc` in debug mode, share cache

max-sixty:
This reduces the pressure on the GHA cache, speeds up CI, and reduces the GHA queue.


max-sixty:
Closes #2639

eitsupi:
This probably needs to be tested to make sure it works correctly in the release?

max-sixty:
I'm fairly confident it'll work, and we can always re-run release pipelines if it doesn't. But ofc more testing is good, feel free to have a go...

eitsupi:
Ok, it seems woking.

image

```log
    Compiling prqlc v0.8.1 (D:\a\prql\prql\prql-compiler\prqlc)
    Finished release [optimized] target(s) in 6m 31s
Run export ARTIFACT_NAME="prqlc-v0.8.2-x86_64-pc-windows-msvc.zip"
  export ARTIFACT_NAME="prqlc-v0.8.2-x86_64-pc-windows-msvc.zip"
  echo "ARTIFACT_NAME=${ARTIFACT_NAME}" >>"$GITHUB_ENV"
  cd target/x86_64-pc-windows-msvc/release
  7z a "../../../${ARTIFACT_NAME}" prqlc.exe ../../../LICENSE ../../../prql-compiler/prqlc/README.md
  shell: C:\Program Files\Git\bin\bash.EXE --noprofile --norc -e -o pipefail {0}
  env:
    CACHE_ON_FAILURE: false
    CARGO_INCREMENTAL: 0

7-Zip 22.01 (x64) : Copyright (c) 1999-2022 Igor Pavlov : 2022-07-15

Scanning the drive:
3 files, 8934266 bytes (8725 KiB)

Creating archive: ../../../prqlc-v0.8.2-x86_64-pc-windows-msvc.zip

Add new data to archive: 3 files, 8934266 bytes (8725 KiB)
```

.github/workflows/pull-request.yaml
```diff
@@ -117,10 +117,12 @@ jobs:
...
-          - os: ubuntu-latest
-            target: aarch64-unknown-linux-musl
-          - os: macos-latest
-            target: x86_64-apple-darwin
+          # TODO: Until we have tests for these, we don't have a cache for them.
+          # If we can add tests, then re-enable them. They run on `release.yaml`
+          # regardless.
+          #
+          # - os: ubuntu-latest target: aarch64-unknown-linux-musl
+          # - os: macos-latest target: x86_64-apple-darwin
```
eitsupi, line 125: My understanding is that `x86_64-apple-darwin` is tested on GitHub Actions and `aarch64-apple-darwin` is not.
Is it really necessary to disable these?

max-sixty: Do you think we could add these to the `test-all` or `nightly` tests? Then they'll share the cache and be v fast.

We can only use this new approach of cache-sharing if they're also run in tests, because the caches are only written in tests. Or we could a) revert — though this way will lead to far fewer cache misses, or b) do something like a label, where it's longer, but not required on every PR, or c) have them here but they'll take a long time.

max-sixty: I hadn't realized that Mac was on x86_64. I'll switch it (doesn't affect the discussion above)

eitsupi: > Do you think we could add these to the `test-all` or `nightly` tests? Then they'll share the cache and be v fast.

I think it makes sense to add it to nightly. (In short, one of the purposes of this is to prevent people from realizing that their builds are broken only at release time, so I think we should reduce the number of targets that are only built at release time.)

I am not sure I fully understand your explanation, but are you assuming that if we add this to nightly we can re-enable the build in PR?

max-sixty: > I am not sure I fully understand your explanation, but are you assuming that if we add this to nightly we can re-enable the build in PR?

Yes!

eitsupi: Ok, thanks. I'll take a look at later!

max-sixty: The one thing to watch for is that they should use a different cache key from the other targets on the OS, and I'm not sure how the toolchain gets into the key; maybe setting it with `rustup` before running the cache action will cause it to be different....

PR #2779, opened 2023-06-10
style: Use `foo=bar` style in codegen

max-sixty:


prql-compiler/tests/integration/snapshots/integration__fmt@group_sort.prql.snap
```diff
@@ -4,9 +4,9 @@ expression: "from tracks\nderive d = album_id + 1\ngroup d (\n    aggregate {\n
 input_file: prql-compiler/tests/integration/queries/group_sort.prql
 ---
 from tracks
-derive d = album_id + 1
-group d (aggregate {n1 = (track_id | sum)})
+derive d=album_id + 1
```
eitsupi: Is this easy to read?

```elm
derive d=album_id + 1
```

I feel that it would be difficult to understand which of the following calculation order is used.

```elm
derive (d=album_id) + 1
```

```elm
derive d=(album_id + 1)
```

max-sixty: No, I agree, that seems bad.

What are some options for that?

```
derive d=(album_id + 1)
```

...might be reasonable — i.e. compound expressions go in paretheses?

max-sixty: We could have very tight expressions; i.e. 

```
derive d=album_id+1
```

...but I don't think that'll scale well?

eitsupi: I prefer the followings:

```elm
derive d=(album_id + 1)
```


```elm
derive d = (album_id + 1)
```

(I prefer to have spaces before and after the equals first of all.)

aljazerzen: On second though, I'd vote for:

```
derive d = album_id + 1
```

It looks a bit confusing for join, but it's not that bad:

```
join s = salaries (==employee_id)
join m = manager e.manager_id == m.id
```

aljazerzen:
@max-sixty what's the status here?

max-sixty:
> @max-sixty what's the status here?

Sorry for the delay.

We need to decide on the format! 

Two main options I can see:
- `derive d = album_id + 1`
- `derive d=(album + 1)`

The advantage of the latter is that it partially solves the issues in https://github.com/PRQL/prql/pull/2851#issuecomment-1644487097 (CC @richb-hanover ...)

aljazerzen:
I vote for `derive d = album_id + 1` with +0.5

Mainly because of conventions from other languages.

max-sixty:
The diff got mangled in merges, and I _think_ the current implementation is correct already, but I'll continue reviewing

PR #2790, opened 2023-06-11
ci: Use explicit target in cargo tests

max-sixty:
Using an implicit vs explicit target uses a different cache, even when they're the same

### DIFF
--- a/.github/workflows/pull-request.yaml
+++ b/.github/workflows/pull-request.yaml
@@ -30,7 +30,7 @@ jobs:
       matrix:
         # We use a matrix here to mirror `test-all`, so they share a cache.
         os: [ubuntu-latest]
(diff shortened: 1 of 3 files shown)


max-sixty:
FYI @eitsupi I had to change `musl` to `gnu` because it couldn't seem to run the tests with `musl`. I have no view betwene them; feel free to switch back if you prefer + can get it working...

eitsupi:
> FYI @eitsupi I had to change `musl` to `gnu` because it couldn't seem to run the tests with `musl`. I have no view betwene them; feel free to switch back if you prefer + can get it working...

My understanding is that it is less portable on gnu and more likely to not work despite being a single binary.
From a user perspective there is no advantage in choosing gnu. We at least need to release musl binaries.

PR #2792, opened 2023-06-11
ci: fix rust build target settings in CI

eitsupi:
Follow up #2790

### DIFF
--- a/.github/workflows/test-rust.yaml
+++ b/.github/workflows/test-rust.yaml
@@ -79,5 +79,5 @@ jobs:
           # - Unreferenced snapshots - `--unreferenced=auto`.
           args:
             test --target=${{ inputs.target }} ${{ inputs.target==
-            'x86_64-unknown-linux-gnu' &&
-            '--unreferenced=auto --features=test-external-dbs' || '' }}
+            'x86_64-unknown-linux-gnu' && '--unreferenced=auto
+            --features=test-external-dbs' || '' }}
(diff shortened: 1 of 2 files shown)


.github/workflows/test-rust.yaml
```diff
@@ -22,7 +22,7 @@ jobs:
     steps:
       - name: 📂 Checkout code
         uses: actions/checkout@v3
-      - if: inputs.os == 'Linux'
+      - if: ${{ contains(inputs.target, 'musl') }}
```
max-sixty, line 25: 👍

max-sixty (approved):

eitsupi:
Sorry, Prettier doesn't work with my VS Code and I can't format the yaml file.

max-sixty:
It's weird that it now just works! Sorry for the change, it seems like it wasn't actually a problem.

eitsupi:
I have reverted to musl only for prqlc builds. The tests are still on gnu.

As you know, the default target on Linux is gnu, and this seems to be because musl requires additional tools to be installed, etc.? (you should be able to find various articles on this if you look it up)

max-sixty:
Ah that explains it! 

That doesn't share the cache, so this undoes some of the work. But the ubuntu builds are fast anyway, so it only takes the total time from 2.5 to 3.5 minutes. If you want to try and get the tests working on musl then that could be good. But overall we've done a lot here and it's in a good state, we can leave it for a while.

Thank you!

max-sixty:
(and FYI I downloaded it and tested it — it works. Kinda weird the tests don't work then; I'm fairly confident I had added the `apt-get` command)

Issue #2717, opened 2023-06-04
The `count` function needs `s"*"` for the argument in current dev version


eitsupi:
### What happened?

From https://github.com/PRQL/prql/pull/2710#issuecomment-1575011845, related to #2713

We are currently using workarounds in the following following locations:

https://github.com/PRQL/prql/blob/69571f1bd44e90e77e5a514a72d218ba9947caeb/README.md?plain=1#L34
https://github.com/PRQL/prql/blob/69571f1bd44e90e77e5a514a72d218ba9947caeb/web/website/content/_index.md?plain=1#L26
https://github.com/PRQL/prql/blob/69571f1bd44e90e77e5a514a72d218ba9947caeb/prql-compiler/examples/compile-files/queries/variables.prql#L16

### PRQL input

```elm
from a
derive {cnt = count}
```

### SQL output

```SQL
NA
```

### Expected SQL output

```SQL
SELECT
  *,
  COUNT(*) AS cnt
FROM
  a
```

### MVCE confirmation

- [X] Minimal example
- [X] New issue

aljazerzen:
While working on the compiler, I had to change the rule which triggers the `OVER` clause for window functions to not look at the result type of the function, but instead look at the parameter type. This works well, but it has a problem: some window functions don't have any parameters:
- `count` (there was `non_null:`, but that's a named param)
- `rank`,
- `row_number`,

For the sake of language design this should not matter, but it lead me to the following question:

How does the `count` function work, if it receives no parameters? What's the input of its computation? Does it just "pull from the relation"? How do `rank` and `row_number` know how many resulting rows should they output?

This was my justification for making them have a parameter. `count` now has a positional `non_null`, and `rank` and `row_number` have a `column` param that is not used in compilation.

This is obviously a bit inconvenient, so how could we design the language to work around this? To have the arguments as they were before, but don't use the "pull from the relation" bullshit justification?

I've hoped to find a quick fix, but then I found the similar issue of #2723 and wanted to clear that first, so the two do not get mixed up.

max-sixty:
> How does the `count` function work, if it receives no parameters? What's the input of its computation? Does it just "pull from the relation"? How do `rank` and `row_number` know how many resulting rows should they output?

Yes, I think we can think about it as one of:
- it looks at the whole relation, rather than just one column.
- it looks _any_ column (and at least in the case of count, also counts nulls, unlike when a column is selected, as you describe)

[...]

PR #2796, opened 2023-06-11
tweak: Fix Readme alignment

max-sixty:


max-sixty:
Actually picked a different example to not lead with our current `count` syntax until it's sorted

max-sixty:
Re https://github.com/PRQL/prql/issues/2717 for the count syntax

PR #2797, opened 2023-06-11
ci: Use `ci` profile

max-sixty:
I thought I did this last night, but found this unmerged branch. It seems that the profile overrides the env var, so we need to turn off `debuginfo` in a profile. 

So this switches from using the env vars to a profile

### DIFF
--- a/.github/actions/build-prqlc/action.yaml
+++ b/.github/actions/build-prqlc/action.yaml
@@ -45,7 +45,7 @@ runs:
       with:
         command: build
         args:
-          --package prqlc --profile ${{ inputs.profile }} --locked --target=${{
+          --package=prqlc --profile=${{ inputs.profile }} --locked --target=${{
           inputs.target }}
 
     - name: Create artifact for Linux and macOS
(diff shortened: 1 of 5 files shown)


max-sixty:
Waiting on https://github.com/mitsuhiko/insta/issues/381

max-sixty:
This would be nicer on the margin but actually isn't the cause of the bloat — actually since we now pass `--target`, the setting isn't respected for some binaries; from https://doc.rust-lang.org/cargo/reference/config.html#buildrustflags

> If the --target flag (or [build.target](https://doc.rust-lang.org/cargo/reference/config.html#buildtarget)) is used, then the flags will only be passed to the compiler for the target. Things being built for the host, such as build scripts or proc macros, will not receive the args. Without --target, the flags will be passed to all compiler invocations (including build scripts and proc macros) because dependencies are shared. If you have args that you do not want to pass to build scripts or proc macros and are building for the host, pass --target with the [host triple](https://doc.rust-lang.org/cargo/appendix/glossary.html#target).

I'm not sure why that adds back almost all the cache size though.

I think I've spent enough time on this, and more yak shaving is unlikely to be productive. I'll add a note with a reference to this issue.

An aside — I'm not sure why behavior is so different when passing `--target=(default-target)` vs nothing — this has been the cause of lots of `rustc` issues; e.g. the elixir / caching issues...

Issue #1836, opened 2023-02-14
Replace `wasm-pack`


max-sixty:
Not the most urgent item, but would be good to keep our build steps as uncomplicated as possible, and remove dependencies that aren't maintained.

Currently [we use `wasm-pack` to bundle `prql-js`](https://github.com/PRQL/prql/blob/3b865282e4c2f8fdd309123ecfce970fe0eb824d/prql-js/package.json#L21-L24). IIUC we're using it because 
- we previously wanted to create a simple npm package, which `wasm-pack` does well
- then we decided to also create bundler & web targets
- so we made our own npm package which contained all three
- ...but Instead of using `wasm-bindgen` alone, we kept `wasm-pack` and added `package.json` scripts to move the files it output (link above)

For context, `wasm-pack` is a wrapper of `wasm-bindgen` and isn't really maintained, whereas `wasm-bindgen` remains well-supported.

As discussed in https://rustwasm.github.io/docs/wasm-bindgen/reference/deployment.html, I think it should be possible to remove this middle layer and use `wasm-bindgen` alone; likely using a similar set of steps to what we do now in `package.json` scripts.

This would be a good contribution for someone who's somewhat familiar with JS, and would like an early PR before diving into the PRQL compiler.

max-sixty:
There's now a library that will do build a wasm package as part of the build:

Quoting from https://github.com/PRQL/prql/issues/2158

> https://crates.io/crates/substrate-wasm-builder
>
> It requires nightly, which we won't ever require, but also possibly it might be moving to not require nightly? https://github.com/paritytech/substrate/pull/13580
>
> We're 2-3 months behind on the toolchain, so assuming this works, we could use this in a few months to replace wasm-pack. That would simplify the build a lot as well as solving this perf issue.

So I'll mark this as postponed, but leave it open, and we can implement this when it's available. It'll make the builds faster, much much faster when there are no changes, and reduce our use of unsupported crates.

max-sixty:
Update:

- `substrate-wasm-builder` is cool but not that practical for us — it doesn't generate the JS file. It compiles a `.wasm` artifact by creating a whole new crate at build-time, building it, and then copying the `.wasm` artifact back [^1]
- https://github.com/rustminded/xtask-wasm looks good as a replacement but doesn't seem to handle generating the `package.json` etc, which `wasm-pack` does
- More understanding in https://github.com/rustwasm/wasm-bindgen/discussions/3494#discussioncomment-6245992
[...]

PR #2801, opened 2023-06-12
build: Check to see if npm dependencies are already up to date;

richb-hanover:
New task "check-npm-cache" that avoids unnecessary long startup times.
Save a .npmtimestamp file in the node_modules directory after `npm install`. If a subsequent call to check-npm-cache shows the package.json is older, then 
there's no need to re-run `npm install`...


Taskfile.yml
```diff
@@ -300,12 +297,25 @@ tasks:
...
+    # This avoids long start times when no dependencies have changed
+    # Create node_modules/.npmtimestamp after running 'npm install'
+    # Rerun npm install if package.json is newer than the timestamp
+
+    # Must set `install-links=false` in the playground's `npm install` to
+    # install prql-js as the regular dependency, instead of creating a
+    # symlink. Refer to https://github.com/PRQL/prql/pull/1296.
+    desc: Check to see if package.json has changed since we ran npm install
+    dir: web/playground
+    status: # if both are true, don't rebuild
```
max-sixty: Can we replace this all with `sources:`? https://taskfile.dev/usage/#by-fingerprinting-locally-generated-files-and-their-sources

richb-hanover: Probably. My concern is that the `node_modules` folder has ~50,000 files. Computing a checksum over all those files might take nearly as much time as the `npm install` I'll check the`timestamp` method to see how it performs.

max-sixty: Can we use it on `package.json` though? Like we're doing here?

richb-hanover: Remarkably, the checksum option is fast - less than a second. It's also indistinguishable from the `timestamp` method. So this commit switches to using checksums.

My test process was to modify `package.json` (add a blank line) and run `task run-playground` which calls `npm install ...`

I then ^C, and re-ran `task run-playground` and the Task reports, _task: Task "check-npm-cache" is up to date_

Can you suggest any other tests? Thanks.

max-sixty: I'm actually not sure if it's using the files under `generates`; I couldn't completely decipher the docs:

> For the checksum (default) or timestamp method to work, it is only necessary to inform the source files. When the timestamp method is used, the last time of the running the task is considered as a generate.

Taskfile.yml
```diff
@@ -300,12 +297,25 @@ tasks:
...
     cmds:
-      # Must set `install-links=false` in the playground's `npm install` to
-      # install prql-js as the regular dependency, instead of creating a
-      # symlink. Refer to https://github.com/PRQL/prql/pull/1296.
-      - npm install --install-links=false
+      - task: check-npm-cache
       - npm start
-#
+
+  check-npm-cache:
```
max-sixty: Overall this is great, very nice addition.

Can we rename this to something like `install-playground`? It's doing more than just checking the cache.

~Then let's merge!~ (edit: one note added below)

max-sixty:
Actually one question — do we need to reinstall when rust files in `prql-compiler` change? Will this forget to pick that up?

Even if that is the case, we can still have a task like `run-playground-cached` which does this.

richb-hanover:
re: renaming - a good idea

re: `prql-compiler`changes - I bet that would be important. It complicates the sources/generates statements, since they're not in the same directory. (At least my simple-minded first cut of adding a test for `prql-compiler/src/*` didn't work.) 

I'll keep bulling ahead...

max-sixty:
No prob — why don't we merge this with something like `run-playground-cached`, add a `desc:` field, and then you're not waiting around?

And then we can refactor this all when we upgrade rust and can do this as part of the standard build, which is cached itself.

richb-hanover:
I went with a task name of "install-playground-npm-dependencies". 

Then we need to figure out how to trigger the `npm install` when prql-compiler sources change as well. Thanks.

Taskfile.yml
```diff
@@ -300,12 +297,25 @@ tasks:
     desc: Build & serve the playground.
     dir: web/playground
     cmds:
-      # Must set `install-links=false` in the playground's `npm install` to
-      # install prql-js as the regular dependency, instead of creating a
-      # symlink. Refer to https://github.com/PRQL/prql/pull/1296.
-      - npm install --install-links=false
+      - task: install-playground-npm-dependencies
```
max-sixty, line 300: I was thinking this wouldn't be the default under `run-playground` — there would be a separate task `run-playground-npm-dependencies` 

The default needs to pick up the full path, otherwise it would be v confusing that it's using old rust code. 

I'm not sure it worth investing much in us detecting the rust changes, since it might be difficult and it will be replaced by the proper solution when we can do https://github.com/PRQL/prql/issues/1836

richb-hanover: Hmmm... This PR is meant as a convenience for people doing documentation. The current `main` branch _always_ runs the full npm install (taking ~1 minute), for both `task run-web` and `task run-playground`, even if I ^C the one, start the other, with no changes to any of the source files.

On further inspection, I see there's a `task run-website` that does what I want. It simply serves out the (previously-built) static site. I have created a similar `task run-pg` that people can use to avoid the npm step completely.

I believe it's useful to keep the separate `install-playground-npm-dependencies` task because it factors out the (unexpected) `--install-links` option into a single function.

Potential To-do's:

1. Detect changes to `prql-compiler` source file in `install-playground-npm-dependencies`
2. Regularize the task names
   - `build-web` - builds the website (the current `build-web`)
   - `build-run-web` - builds and runs the website (the current `run-web`)
   - `run-web` - run previously-built static site (the current `run-website`)
   - similar names for build-playground, build-run-playground, run-playground
   - similar names for run-book, just for completeness

max-sixty: (responding quickly, forgive me for brevity!)

> * `build-web` - builds the website (the current `build-web`)
> 
>     * `build-run-web` - builds and runs the website (the current `run-web`)
> 
>     * `run-web` - run previously-built static site (the current `run-website`)

I've been referring to `web` as all three — book / playground / website — and `website` as just the static site (at prql-lang.org). Does that change your view at all?

> Detect changes to `prql-compiler` source file in `install-playground-npm-dependencies`

As I say above — if you want to do this then you're welcome to, but it'll go away shortly when #1836 is unblocked (by Debian rust version...)



> On further inspection, I see there's a `task run-website` that does what I want. It simply serves out the (previously-built) static site.

The `run-` steps _also_ build things IIUC. But yes OK; we could have `run-` skip the building and have `build-run` do both? (Though, again, soon building will be very cheap, so it might be a smaller change to just add your new task as a suffix of `-cached` and then we refactor it out when we do #1836?)

Taskfile.yml
```diff
@@ -300,6 +300,12 @@ tasks:
       - task: install-playground-npm-dependencies
       - npm start
 
+  run-pg:
```
max-sixty, line 303: This is OK! I would have voted for something clearer — not obvious how `pg` is different from `playground` (though it is elegant in a way! Smaller and less descriptive, just like running vs building!!)

max-sixty:
If it's OK, I'm going to merge, and then we do any follow-ups in a separate PR?

richb-hanover:
@max-sixty This is great. I'll be delighted when building is cheap. In the mean time, I don't have to be grumpy waiting for `npm install` every. single. time. I. run. :-) Thanks.

Issue #2757, opened 2023-06-07
Getting `fmt` to the finish-line


max-sixty:
### What's up?

After @aljazerzen built the foundations of this, I have filled a few gaps, and now all the examples format into correct PRQL. But we still have a few gaps in getting to something we can use everywhere.

In order to roll this out fully, it's quite important to be close to 100%, since all PRQL will be formatted this way — in the book, whenever someone saves a file in VS Code, whenever someone runs `pre-commit` (both in our repo and others'), etc. Without being close to 100%, it's not that instrumentally useful.

- [ ] We currently remove comments!
- [x] Do we want to elide parentheses around function calls in `{x = (sum foo)}` (from https://github.com/PRQL/prql/pull/2803#discussion_r1231216571). This in the docs [here](https://github.com/PRQL/prql/blob/bb61a5ce0d49426475d70c6b852591614de8d686/web/book/src/syntax/expressions-and-operators.md#L85). It might require understanding the _grandparent_ node (need to think more)
- [x] `module` breaks, which breaks the standard library (though I actually thought we were only going to do files as modules @aljazerzen ?) https://github.com/PRQL/prql/pull/2949
- [x] Do we want `from t=tracks` or `from t = tracks`? I marginally preferred the former — given that we have fairly few separators, having tighter expressions groups things better. (but this is not a strong view, maybe +0.3, and only aesthetics) https://github.com/PRQL/prql/pull/2779
- [x] Do we want to break inner transforms at all? e.g. this is a lot on one line. But I think my reaction is mostly based on what I'm used to, and I don't think there would be a good rule that lets us handle inner transforms differently. Though we could linebreak earlier (e.g. line break at 40) https://github.com/PRQL/prql/blob/31c3e71b55a0c6ca383e9c1367438352f024fa6b/prql-compiler/tests/integration/snapshots/integration__fmt@distinct_on.prql.snap#L8  https://github.com/PRQL/prql/pull/2950
- [x] `{(-foo)}` has extra parentheses, : https://github.com/PRQL/prql/blob/31c3e71b55a0c6ca383e9c1367438352f024fa6b/prql-compiler/tests/integration/snapshots/integration__fmt@distinct_on.prql.snap#L8-L9 — https://github.com/PRQL/prql/pull/2803

aljazerzen:
> module breaks, which breaks the standard library

We use it only internally, because I don't want to have 7 different files with 5 lines in each.


> Do we want from t=tracks or from t = tracks?

[...]

PR #2803, opened 2023-06-12
feat: Format `UnOp` correctly

max-sixty:
Writing out my thinking on the UnOp precedence. The code here is not really worth looking at, it's just me debugging, but hopefully the description is engagable.

We have a trilemma:

```
derive a = -b
# derive a = (-b) # wrong
# derive (a = -b) # wrong
```

- `-` has a higher binding strength than `=`, and so doesn't need parentheses
- `=` has a higher binding strength than its parent (the funccall), and so doesn't need parentheses

```
mul 5 (-3)
# mul 5 -3 # wrong
```

- `-` has a lower binding strength than its parent  (the func call), and so *does* need parentheses

So we get this trillemma:

- `-` > `=`
- `=` > funccall
- funccall > `-`

Am I thinking about this correctly? Are there any ways around it?


prql-compiler/src/codegen.rs
```diff
@@ -263,10 +276,54 @@ impl WriteSource for pl::ExprKind {
     }
 }
 
+const EQ_STRENGTH: i32 = 0;
```
aljazerzen: You probably want to name this after the operation (alias / assign), not the character (equals).

prql-compiler/src/codegen.rs
```diff
@@ -277,6 +334,8 @@ fn write_expr(expr: &pl::Expr, parent: &pl::ExprKind, opt: WriteOpt) -> Option<S
 
 fn binding_strength(expr: &pl::ExprKind) -> i32 {
     match expr {
+        // For example, if it's an Ident, it's basically infinite — a simple
+        // ident never needs parentheses around it.
         pl::ExprKind::Ident(_) => 10,
         pl::ExprKind::All { .. } => 10,
```
aljazerzen: I use value 100 for these cases in the SQL backend. Should we use it here too?

prql-compiler/src/codegen.rs
```diff
@@ -295,8 +354,8 @@ fn binding_strength(expr: &pl::ExprKind) -> i32 {
             pl::BinOp::Or => 1,
             pl::BinOp::Coalesce => 2,
         },
-        pl::ExprKind::Unary { .. } => 9,
-        pl::ExprKind::FuncCall(_) => 0,
+        pl::ExprKind::Unary { .. } => 6,
+        pl::ExprKind::FuncCall(_) => 7,
         pl::ExprKind::Func(_) => 0,
 
         _ => 11,
```
aljazerzen, line 362: Why do other nodes have stronger binding than idents?

prql-compiler/src/codegen.rs
```diff
@@ -295,8 +354,8 @@ fn binding_strength(expr: &pl::ExprKind) -> i32 {
             pl::BinOp::Or => 1,
             pl::BinOp::Coalesce => 2,
```
aljazerzen, line 350: Something is not right here. Looking at this https://github.com/prql/prql/blob/main/prql-compiler/src/parser/expr.rs#L191-L197, the precedence should be:

```
unary > range > mul,div,mod > add,sum > compare > coalesce > and > or
```

(I've tested that it does add some unnecessary parenthesis)

prql-compiler/src/codegen.rs
```diff
@@ -547,9 +606,26 @@ impl WriteSource for pl::TupleField {
...
+    use similar_asserts::assert_eq;
 
     use super::*;
 
+    fn assert_fmt_matches(input: &str) {
+        let stmt = generate_single_stmt(input);
+        assert_eq!(input.trim(), stmt.trim());
+    }
+
+    // TODO: Is there a library function that can do this? (Guessing there is...)
```
aljazerzen: There is not. Parsing just one statement is not that useful for most usecases.

prql-compiler/src/codegen.rs
```diff
@@ -263,10 +276,54 @@ impl WriteSource for pl::ExprKind {
...
+    let mut r = String::new();
+    // If `-1` has an alias `foo`, we're going to write `foo = -1`, so we want
+    // to take the strength of `=`, not of `-1`
+    let strength_self = EQ_STRENGTH;
+    let strength_parent = binding_strength(parent);
+    dbg!(&expr, strength_self, &parent, strength_parent);
+
+    if strength_parent >= strength_self {
+        r += "(";
+        r += &write_ident_part(&expr.alias.clone().unwrap());
```
aljazerzen:
```suggestion
        r += &write_ident_part(&expr.alias.as_ref().unwrap());
```

aljazerzen (changes requested):
Looking good.

The bindings strengths have to be adjusted, though.

max-sixty:
> Looking good.
> 
> The bindings strengths have to be adjusted, though.

Are you less concerned than me about the issues in the PR description? I said there that the code was almost not worth looking at but there was a big issue with our semantics! :)

aljazerzen:
Oh, again I missed the PR description (I think GH notification opens the code changes directly), sorry.

----

I don't understand the problem here.

First, I think the second example is valid:
```
derive a = -b     # valid
derive a = (-b)   # also valid
# derive (a = -b) # wrong
```

> So we get this trillemma:
> 
>   -  `-` > `=`
>   -  `=` > funccall
>   -  funccall > `-`

The last point should not conflict with the first two, because there is a node in-between. We can process this as:

```
FuncCall:
  AliasAssign:
    UnOp:
      Ident: {}
```
... here none of the parent-child pairs have an inverted binding strength, so no parenthesis are needed.

```
FuncCall:
  UnOp:
    Ident: {}
```
... here, FuncCall < UnOp, which means we need parenthesis.


Oh, I get the problem now: we cannot just assign numbers to the strengths, because they are not an ordering, i.e. they are not linear.

Well, the assign will need some special treatment. When viewed as a child, it has binding strength one more than func call (which has 0), when viewed as a parent it has binding strength of one more than unop (which has 9).

max-sixty:
> ... here none of the parent-child pairs have an inverted binding strength, so no parenthesis are needed.

Yeah, if I understand your comments correctly — this isn't a problem with parsing, but with producing nicely formatted PRQL. Otherwise the current state is fine.

> Well, the assign will need some special treatment. When viewed as a child, it has binding strength one more than func call (which has 0), when viewed as a parent it has binding strength of one more than unop (which has 9).

Yes, I think that's right.

It makes me wonder whether we're doing this correctly, though. Is it normal to have this sort of special-casing — where there's no strict order of precedence — in formatters?

prql-compiler/src/codegen.rs
max-sixty: I told you not to look at the code!! :D

aljazerzen:
> Yeah, if I understand your comments correctly — this isn't a problem with parsing, but with producing nicely formatted PRQL. Otherwise the current state is fine.

Yes.

> It makes me wonder whether we're doing this correctly, though. Is it normal to have this sort of special-casing — where there's no strict order of precedence — in formatters?

That I don't know - I've not written many formatters. To be frank, this is the only one :D

max-sixty:
I think it's more complicated than we've described, and separating `=` by parent / child isn't sufficient.

Even if we only take cases where `=` is a parent:

```
# `-` (unop) > `=`
derive a = -b

# `=` > funccall
derive x = (add a b)

# But funcall > `==`(unop)
join x (==foo)
```

Possibly we need different strengths for a funcall being a parent or child — that as a parent in `join x (==foo)`, it's stronger than unop, but as a child in `derive x = (add a b)`, it's weaker than both unop and `=`...

max-sixty:
It's not as pretty as I'd like, but it seems to work.

(one option is for `WriteSource::write` to take an optional `Parent`, but that probably is too much blast radius, I'll have another look soon)

prql-compiler/src/codegen.rs
```diff
@@ -123,13 +123,13 @@ impl WriteOpt {
 
 impl WriteSource for pl::Expr {
     fn write(&self, opt: WriteOpt) -> Option<String> {
-        let mut r = String::new();
-        if let Some(alias) = &self.alias {
-            r += &write_ident_part(alias);
-            r += " = ";
+        // TODO: awkwardly designed; there's another branch on alias in the
+        // `write_expr` function; I think we can make this simpler.
```
aljazerzen, line 127: Probably yes. But good enough for now.

prql-compiler/tests/integration/snapshots/integration__fmt@genre_counts.prql.snap
```diff
@@ -1,10 +1,10 @@
...
-expression: "let genre_count = (\n    from genres\n    aggregate {a = count name}\n)\n\nfrom genre_count\nfilter a > 0\nselect a = -a\n"
+expression: "# clickhouse:skip (ClickHouse prefers aliases to column names https://github.com/PRQL/prql/issues/2827)\n# mssql:test\nlet genre_count = (\n    from genres\n    aggregate {a = count name}\n)\n\nfrom genre_count\nfilter a > 0\nselect a = -a\n"
 input_file: prql-compiler/tests/integration/queries/genre_counts.prql
 ---
-let genre_count = (from genres | aggregate {a = count name})
+let genre_count = (from genres | aggregate {a = (count name)})
 from genre_count
 filter a > 0
-select a = (-a)
+select a = -a
```
aljazerzen, line 9: Nice.

prql-compiler/tests/integration/snapshots/integration__fmt@distinct_on.prql.snap
```diff
@@ -1,10 +1,10 @@
...
-expression: "# skip_duckdb (sort in group is ignored. I think this is a bug in duckdb)\n\nfrom tracks\nselect {genre_id, media_type_id, album_id}\ngroup {genre_id, media_type_id} (sort {-album_id} | take 1)\nsort {-genre_id, media_type_id}"
+expression: "# skip_duckdb (sort in group is ignored. I think this is a bug in duckdb)\n\nfrom tracks\nselect {genre_id, media_type_id, album_id}\ngroup {genre_id, media_type_id} (sort {-album_id} | take 1)\nsort {-genre_id, media_type_id}\n"
 input_file: prql-compiler/tests/integration/queries/distinct_on.prql
 ---
 from tracks
 select {genre_id, media_type_id, album_id}
-group {genre_id, media_type_id} (sort {(-album_id)} | take 1)
-sort {(-genre_id), media_type_id}
+group {genre_id, media_type_id} (sort {-album_id} | take 1)
+sort {-genre_id, media_type_id}
```
aljazerzen, line 9: Nice.

prql-compiler/tests/integration/snapshots/integration__fmt@genre_counts.prql.snap
```diff
@@ -1,10 +1,10 @@
 ---
 source: prql-compiler/tests/integration/main.rs
-expression: "let genre_count = (\n    from genres\n    aggregate {a = count name}\n)\n\nfrom genre_count\nfilter a > 0\nselect a = -a\n"
+expression: "# clickhouse:skip (ClickHouse prefers aliases to column names https://github.com/PRQL/prql/issues/2827)\n# mssql:test\nlet genre_count = (\n    from genres\n    aggregate {a = count name}\n)\n\nfrom genre_count\nfilter a > 0\nselect a = -a\n"
 input_file: prql-compiler/tests/integration/queries/genre_counts.prql
 ---
-let genre_count = (from genres | aggregate {a = count name})
+let genre_count = (from genres | aggregate {a = (count name)})
```
aljazerzen, line 6: These parenthesis are not needed, idiomatic PRQL should not have them. This can be improved in a latter PR.

aljazerzen (approved):
This work really shows how last 10% are 90% of the work.

prql-compiler/tests/integration/snapshots/integration__fmt@genre_counts.prql.snap
max-sixty: This is the issue between whether an assign is in a tuple or not — do we want different canonical styles based on that?

(edit: this isn't a good way of expressing it, I restated in #2757)

PR #2804, opened 2023-06-12
refactor(test): separate protocol and dialect in the RDBMS integration tests

eitsupi:
The current implementation allows only one dialect test to be run per DB client protocol. However, it may be more convenient to allow multiple dialects to use the same protocol, such as ClickHouse and Detabend, which implement the MySQL protocol.

This PR separates the protocol from the dialects, allowing multiple dialect tests to be run with a single protocol.


prql-compiler/tests/integration/connection.rs
```diff
@@ -8,339 +8,366 @@ use chrono::{DateTime, Utc};
...
-                        Err(_) => "".to_string(),
-                    },
-                    &Type::INT8 => match row.try_get::<usize, i64>(i) {
-                        Ok(v) => v.to_string(),
-                        Err(_) => "".to_string(),
-                    },
-                    &Type::TEXT | &Type::VARCHAR | &Type::JSON | &Type::JSONB => {
-                        match row.try_get::<usize, String>(i) {
+    pub fn run_query(&mut self, sql: &str, runtime: &Runtime) -> Result<Vec<Row>> {
+        match self.protocol {
```
max-sixty, line 78: My initial reaction is that these huge match blocks for every method of `DBConnection` are an indicator that this should be a trait implemented on each of instances, like the previous approach.

I agree it would be good to separate the protocol from the dialect eventually (though it's also fine to wait until we have the DBs :) ). But can we do that by returning a `DBConnection` here? https://github.com/PRQL/prql/pull/2804/files#diff-61c6edc4c83a1ee02e8c718cd4c3fe7957d91b357a60b89daed30876c1544b3bR44. We'd have something like `Dialect::DataBend => mysql::Pool::new("databend://root:root@localhost:3307/dummy").unwrap()`

Indeed — isn't `DBConnection` the `Protocol` now?

eitsupi: Thanks for checking this.

Since only `run_query` can be shared between some dialects (e.g. MySQL and ClickHouse), something like `import_csv` may be separated by dialect, I believe it is necessary here to return a structure with dialect information, not a protocol.

Does this answer your question?

max-sixty: But if `import_csv` doesn't work for that dialect — e.g. the querying is the same protocol but not the loading — then we can make a new Struct and implement different code on it, forwarding to `MySql::pool` where necessary.

I don't want to be discouraging, and it looks like you've done a lot of rust, which is great. But I still think that the big match on each method is some evidence that this is not the best data model. And I'm not sure what we're getting for the extra layer of indirection? I'm very open to being wrong if @aljazerzen has a perspective?

I don't think it's _bad_ tbc, I don't quite agree with the implementation, that's all, I hope that's reasonable...

I do think that the current `.get_dialect` is bad, let me take a few minutes to try and remove that. (which will cause a small merge conflict but I hope is OK)

eitsupi: I agree that this does not seem to be the best thing to do.
It is simply the only way I could have done it.

If you could rewrite this for me that would be great. (Of course I had hoped this would not be too much of a burden on you, but I am not up to the task)

max-sixty: > If you could rewrite this for me that would be great. (Of course I had hoped this would not be too much of a burden on you, but I am not up to the task)

I really didn't mean to be discouraging! It's overall been very impressive how far your rust has come...

It's very possible I'm wrong here — generally each person has something right and something wrong — that's why I tagged others.

I had a go in #2806, let me know your thoughts.

eitsupi: Thank you very much. I don't think I am being discouraged. Don't worry about it.

> I had a go in #2806, let me know your thoughts.

Awesome, that makes more sense!

aljazerzen (changes requested):
I see what you are trying to do, that would be useful.

Just as max, I also think that the separation between different protocols was clearer when they were traits instead of enums. That's mostly because before you had everything concerning one protocol together which was easier to debug when some integration stops working correctly.

If the protocols shared a lot of common behavior that this may have been justifiable, but I don't think this is the case.

I'd say that this is what we want:
```rust
pub struct DBConnection {
    pub protocol: Box<dyn Protocol>,
    pub dialect: Dialect
}

pub trait Protocol {
    ...
}

struct PostgresProtocol;

impl Protocol for PostgresProtocol {
    ...
}
```

eitsupi:
Close as conflicts seems to be unfixable.

prql-compiler/tests/integration/connection.rs
eitsupi: I looked at this and couldn't figure out how to implement it without using match.
I've created an issue #2813 and will leave it to someone else.

PR #2805, opened 2023-06-12
ci: set manual trigger for some GitHub Actions workflows

eitsupi:
Since GitHubActions does not trigger GitHubActions, I noticed that the duckdb-rs cache was not present in the PR immediately after the duckdb-rs version was raised and the test took a long time in #2804.

In such cases, it would be useful to be able to manually trigger GitHubActions in the main branch to update the cache. It may also be possible to trigger any workflow on any branch without creating a PR, thus avoiding unnecessary test runs.

### DIFF
--- a/.github/workflows/lint-megalinter.yaml
+++ b/.github/workflows/lint-megalinter.yaml
@@ -7,6 +7,7 @@ name: lint-megalinter
 
 on:
   workflow_call:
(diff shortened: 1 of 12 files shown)


max-sixty:
Great! Most of these would be covered by just the `test-all.yaml`, but I guess no real harm in having them there.

Thanks

Issue #2813, opened 2023-06-12
ClickHouse integration test


eitsupi:
I tried that with #2804 and #2811 and failed.
Perhaps refactoring to separate `import_csv` etc from `DbConnection` first as in #2804, then we can add ClickHouse as in #2811.

PR #2814, opened 2023-06-12
refactor(test): separate protocol and dialect in the RDBMS integration tests

eitsupi:
Part of #2813, continued from #2804

### DIFF
--- a/prql-compiler/tests/integration/main.rs
+++ b/prql-compiler/tests/integration/main.rs
@@ -40,6 +40,37 @@ trait IntegrationTest {
     fn modify_sql(&self, sql: String) -> String;
 }
 
+impl DbConnection {
+    fn setup_connection(&mut self, runtime: &Runtime) {
+        let setup = include_str!("data/chinook/schema.sql");
+        setup
+            .split(';')
+            .map(|s| s.trim())
+            .filter(|s| !s.is_empty())
+            .for_each(|s| {
+                self.protocol
+                    .run_query(self.dialect.modify_sql(s.to_string()).as_str(), runtime)
+                    .unwrap();
+            });
+        let tables = [
+            "invoices",
+            "customers",
+            "employees",
+            "tracks",
+            "albums",
+            "genres",
+            "playlist_track",
+            "playlists",
+            "media_types",
+            "artists",
+            "invoice_items",
+        ];
+        for table in tables {
+            self.dialect.import_csv(&mut *self.protocol, table, runtime);
+        }
+    }
+}
+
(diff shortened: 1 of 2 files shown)


eitsupi:
@max-sixty Sorry, but I could not figure out how to implement `import_csv` without using match as you wrote here https://github.com/PRQL/prql/pull/2804#discussion_r1226086555.

prql-compiler/tests/integration/main.rs
```diff
@@ -74,21 +89,124 @@ impl IntegrationTest for Dialect {
...
+                    ),
+                })
+            }
             _ => None,
         }
     }
 }
 
+impl SetUpData for DbConnection {
+    fn import_csv(&mut self, csv_name: &str, runtime: &Runtime) {
```
max-sixty: I think this _does_ probably require the Big Match Statement — the alternative would be a Struct per Dialect, which I think would be too much structure given we already have a Struct per Protocol.

---

I think it's not worth changing (even if my thought is correct, which may not be), but one question — could this be part of `IntegrationTest` trait, and have the methods receive a `protocol`? Generally implementing on the smallest unit keeps it simple. `modify_sql` doesn't even take a `protocol` arg.

prql-compiler/tests/integration/main.rs
```diff
@@ -119,16 +237,16 @@ fn test_fmt_examples() {
 fn test_rdbms() {
     let runtime = &*RUNTIME;
 
-    let mut connections: Vec<(Dialect, Box<dyn DbConnection>)> = Dialect::iter()
+    let mut connections: Vec<DbConnection> = Dialect::iter()
```
max-sixty, line 263: This is nicer...

prql-compiler/tests/integration/main.rs
```diff
@@ -184,14 +303,15 @@ fn test_rdbms() {
     })
 }
 
-fn setup_connection(con: &mut dyn DbConnection, runtime: &Runtime) {
+fn setup_connection(con: &mut DbConnection, runtime: &Runtime) {
```
max-sixty: I guess this could be part of `DbConnection` (but this is me being very refine-y now, no need to change)

max-sixty (commented):
Nice!

prql-compiler/tests/integration/main.rs
eitsupi: That makes sense, thanks.
I updated that.

max-sixty:
Looking good!

eitsupi:
Thank you for your review. I think it is much better now.

PR #2811, opened 2023-06-12
test: clickhouse integration test

eitsupi:
Based on #2804

PR #2815, opened 2023-06-12
test: clickhouse integration tests

### LINKED ISSUE #2813: ClickHouse integration test

eitsupi:
Close #2813, based on #2814

Same as #2811

### DIFF
--- a/prql-compiler/src/sql/dialect.rs
+++ b/prql-compiler/src/sql/dialect.rs
@@ -81,11 +81,12 @@ impl Dialect {
             | Dialect::SQLite
             | Dialect::Postgres
             | Dialect::MySql
-            | Dialect::MsSql => SupportLevel::Supported,
+            | Dialect::MsSql
+            | Dialect::ClickHouse => SupportLevel::Supported,
             Dialect::Generic | Dialect::Ansi | Dialect::BigQuery | Dialect::Snowflake => {
                 SupportLevel::Unsupported
             }
-            Dialect::Hive | Dialect::ClickHouse => SupportLevel::Nascent,
+            Dialect::Hive => SupportLevel::Nascent,
         }
     }
 
--- a/prql-compiler/tests/integration/docker-compose.yml
+++ b/prql-compiler/tests/integration/docker-compose.yml
@@ -58,6 +58,6 @@ services:
       # https://github.com/ClickHouse/ClickHouse/blob/01c7d2fe719f9b9ed59fce58d5e9dec44167e42f/docker/server/entrypoint.sh#L7-L9
       CLICKHOUSE_DO_NOT_CHOWN: "1"
     volumes:
-      # ClickHouse can load csv only from user_files_path (default `/var/lib/clickhouse/user_files/`
+      # ClickHouse can load csv only from user_files_path (default `/var/lib/clickhouse/user_files/`)
       # https://clickhouse.com/docs/en/operations/server-configuration-parameters/settings#server_configuration_parameters-user_scripts_path
       - ./data/chinook:/var/lib/clickhouse/user_files/chinook/:ro
--- a/prql-compiler/tests/integration/docker-compose.yml
+++ b/prql-compiler/tests/integration/docker-compose.yml
@@ -46,3 +46,18 @@ services:
       LC_ALL: en_US.UTF-8
       MSSQL_COLLATION: Latin1_General_100_CS_AI_SC_UTF8
     volumes: *vol
+  clickhouse:
+    image: "clickhouse/clickhouse-server"
+    ports:
+      # 9004 is MySQL emulation port
+      # https://clickhouse.com/docs/en/guides/sre/network-ports
+      - "9004:9004"
+    environment:
+      CLICKHOUSE_DB: dummy
+      # Skip `chown` to user_files_path
+      # https://github.com/ClickHouse/ClickHouse/blob/01c7d2fe719f9b9ed59fce58d5e9dec44167e42f/docker/server/entrypoint.sh#L7-L9
+      CLICKHOUSE_DO_NOT_CHOWN: "1"
+    volumes:
+      # ClickHouse can load csv only from user_files_path (default `/var/lib/clickhouse/user_files/`
+      # https://clickhouse.com/docs/en/operations/server-configuration-parameters/settings#server_configuration_parameters-user_scripts_path
+      - ./data/chinook:/var/lib/clickhouse/user_files/chinook/:ro
--- a/prql-compiler/src/sql/std.sql.prql
+++ b/prql-compiler/src/sql/std.sql.prql
@@ -114,6 +114,8 @@ module clickhouse {
   # https://clickhouse.com/docs/en/sql-reference/functions/arithmetic-functions#divide
   @{binding_strength=11}
   let div_f = l r -> s"({l} / {r})"
+
+  let regex_search = text pattern -> s"match({text:0}, {pattern:0})"
 }
 
 module duckdb {
--- a/prql-compiler/tests/integration/connection.rs
+++ b/prql-compiler/tests/integration/connection.rs
@@ -1,15 +1,12 @@
-use std::env::current_dir;
-use std::fs;
-use std::path::PathBuf;
 use std::time::SystemTime;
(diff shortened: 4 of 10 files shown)


eitsupi:
Oh, I see. The order of the Clickhouse results seems to be random.
https://stackoverflow.com/questions/54786494/clickhouse-query-row-order-behaviour

I have observed that the order of the results changes each time I run the following query.

```sql
WITH table_0 AS (
  SELECT
    13 AS x_int,
    13.0 AS x_float,
    5 AS k_int,
    5.0 AS k_float
  UNION
  ALL
  SELECT
    -13 AS x_int,
    -13.0 AS x_float,
    5 AS k_int,
    5.0 AS k_float
  UNION
  ALL
  SELECT
    13 AS x_int,
    13.0 AS x_float,
    -5 AS k_int,
    -5.0 AS k_float
  UNION
  ALL
  SELECT
    -13 AS x_int,
    -13.0 AS x_float,
    -5 AS k_int,
    -5.0 AS k_float
)
SELECT
  (x_int / k_int)
FROM
  table_0
```

@max-sixty @aljazerzen Do you think we can add a `sort` at the end of every PRQL query so that even systems like ClickHouse match the results?

prql-compiler/tests/integration/docker-compose.yml
```diff
@@ -46,3 +46,18 @@ services:
...
+      # 9004 is MySQL emulation port
+      # https://clickhouse.com/docs/en/guides/sre/network-ports
+      - "9004:9004"
+    environment:
+      CLICKHOUSE_DB: dummy
+      # Skip `chown` to user_files_path
+      # https://github.com/ClickHouse/ClickHouse/blob/01c7d2fe719f9b9ed59fce58d5e9dec44167e42f/docker/server/entrypoint.sh#L7-L9
+      CLICKHOUSE_DO_NOT_CHOWN: "1"
+    volumes:
+      # ClickHouse can load csv only from user_files_path (default `/var/lib/clickhouse/user_files/`
```
eitsupi:
```suggestion
      # ClickHouse can load csv only from user_files_path (default `/var/lib/clickhouse/user_files/`)
```

max-sixty:
> @max-sixty @aljazerzen Do you think we can add a `sort` at the end of every PRQL query so that even systems like ClickHouse match the results?

Yes I think that's OK — it doesn't limit any SQL functionality. Feel free to merge this with most queries excluded and then follow-up with changed queries unexcluded.

I also think we could have the Unsupported dialects look for e.g. `# run_clickhouse`, rather than the absence of e.g. `# skip_clickhouse`, if that lets us add queries without worrying about the tail of dialects. Supported dialects would still require explicit skipping.

(Though maybe that's harder than just adding `# skip_clickhouse` to the others, feel free to do whatever is easier)

prql-compiler/src/sql/dialect.rs
```diff
@@ -81,11 +81,12 @@ impl Dialect {
             | Dialect::SQLite
             | Dialect::Postgres
             | Dialect::MySql
-            | Dialect::MsSql => SupportLevel::Supported,
+            | Dialect::MsSql
+            | Dialect::ClickHouse => SupportLevel::Supported,
```
max-sixty, line 85: Depending on how many exclusions we need, we could move this down a rung. (Probably we should do that for MsSql too)

max-sixty (approved):
Great!

eitsupi:
Note: ClickHouse doesn't have `lag`/`lead` functions.
https://clickhouse.com/docs/en/sql-reference/window-functions

eitsupi:
The problem that we are currently experiencing is that the results of the following query.

https://github.com/PRQL/prql/blob/f35c7e5901552fdb9e617f6f61f712cf3fa588c0/prql-compiler/tests/integration/queries/switch.prql#L1-L8

This is not a connection problem, as the following SQL via MySQL CLI reproduces it.

```sql
WITH table_0 AS (
  SELECT
    CASE
      WHEN composer IS NOT NULL THEN composer
      WHEN genre_id < 17 THEN 'no composer'
      ELSE 'unknown composer'
    END AS display,
    milliseconds
  FROM
    tracks
  ORDER BY
    milliseconds
  LIMIT
    10
)
SELECT
  display
FROM
  table_0
ORDER BY
  milliseconds
```

```
+---------------+
| display       |
+---------------+
| Samuel Rosa   |
|               |
|               |
|               |
| L. Muggerud   |
|               |
| L. Muggerud   |
|               |
| Gilberto Gil  |
| Chico Science |
+---------------+
```

eitsupi:
I expected it to be a case when issue, but the following example works.......

```elm
from [
    {x = '3'},
    {x = '1'},
]
select x = case {
    x == '3' => 'fiz',
    true => x,
}
```

```
MySQL [dummy]> WITH table_0 AS (
    ->   SELECT
    ->     '3' AS x
    ->   UNION
    ->   ALL
    ->   SELECT
    ->     '1' AS x
    -> )
    -> SELECT
    ->   CASE
    ->     WHEN x = '3' THEN 'fiz'
    ->     ELSE x
    ->   END AS x
    -> FROM
    ->   table_0;
+------+
| x    |
+------+
| fiz  |
| 1    |
+------+
2 rows in set (0.025 sec)
```

eitsupi:
Oh, I understand. Perhaps this is being read as an empty character rather than `NULL`...

```sql
WITH table_0 AS (
  SELECT
    CASE
      WHEN composer != '' THEN composer
      WHEN genre_id < 17 THEN 'no composer'
      ELSE 'unknown composer'
    END AS display,
    milliseconds
  FROM
    tracks
  ORDER BY
    milliseconds
  LIMIT
    10
)
SELECT
  display
FROM
  table_0
ORDER BY
  milliseconds
```

```
+------------------+
| display          |
+------------------+
| Samuel Rosa      |
| no composer      |
| no composer      |
| no composer      |
| L. Muggerud      |
| no composer      |
| L. Muggerud      |
| unknown composer |
| Gilberto Gil     |
| Chico Science    |
+------------------+
```

eitsupi:
~~https://clickhouse.com/docs/en/operations/settings/formats#format_csv_null_representation~~
-> Seems not related to...

It seems to be correctly recognized as NULL when reading csv. Perhaps a problem with the table type definition?
https://clickhouse.com/docs/en/sql-reference/data-types/nullable

```sql
> SELECT composer FROM file('/var/lib/clickhouse/user_files/chinook/tracks.csv') limit 5;
+------------------------------------------------------------------------+
| composer                                                               |
+------------------------------------------------------------------------+
| Angus Young, Malcolm Young, Brian Johnson                              |
| NULL                                                                   |
| F. Baltes, S. Kaufman, U. Dirkscneider & W. Hoffman                    |
| F. Baltes, R.A. Smith-Diesel, S. Kaufman, U. Dirkscneider & W. Hoffman |
| Deaffy & R.A. Smith-Diesel                                             |
+------------------------------------------------------------------------+
5 rows in set (0.010 sec)
```

eitsupi:
Ok, `Nullable` works...

```sql
> WITH table_0 AS (
    ->   SELECT
    ->     CASE
    ->       WHEN composer IS NOT NULL THEN composer
    ->       WHEN genre_id < 17 THEN 'no composer'
    ->       ELSE 'unknown composer'
    ->     END AS display,
    ->     milliseconds
    ->   FROM
    ->     tracks
    ->   ORDER BY
    ->     milliseconds
    ->   LIMIT
    ->     10
    -> )
    -> SELECT
    ->   display
    -> FROM
    ->   table_0
    -> ORDER BY
    ->   milliseconds;
+------------------+
| display          |
+------------------+
| Samuel Rosa      |
| no composer      |
| no composer      |
| no composer      |
| L. Muggerud      |
| no composer      |
| L. Muggerud      |
| unknown composer |
| Gilberto Gil     |
| Chico Science    |
+------------------+
10 rows in set (0.011 sec)
```

eitsupi:
It seems working now 🎉

max-sixty:
Great work!!

Issue #2723, opened 2023-06-04
Window functions, revisited


aljazerzen:
Working on the type checking lately, I've realized that window functions don't quite fit into the current type system.

If we ignore that we coerce scalars into tuples of scalars (`select x` into `select {x}`), the type of select would be:
```
let select = columns<{scalar..}> rel<relation> -> <relation> ...
```
... which says that it takes a tuple of any scalars and a relation to produce another relation.

But if you have:
```
select {x, sum x, lag 1 x}
```
... the type of `x` and `sum x` would be `scalar`, but type of `lag 1 x` would be `[scalar]` (array of scalars).

Also, what's the type of `x`? I just said that it is scalar, but how can it then be passed into `sum` and `lag`, which both expect an array of scalars?

----

What I propose to do:

- change window function to work similar to `select` and `derive`:
  ```elm
  select {x}
  window {lag 1 x}
  ```
- ban using aggregation or window functions in all functions except aggregate and window,

This would mean that:

- in select, column references resolve to scalars and the resulting tuple contains scalars,
- in window, column references resolve to arrays of scalars and the resulting tuple contains arrays of scalars
- in window, you cannot do `x + y`, because `x` is an array and `y` is an array,
- in aggregate, column references resolve to arrays of scalars and the resulting tuple contains scalars,
- in aggregate, you cannot do `sum (x + y)`, because `x` is an array and `y` is an array, but you could do `(sum x) + (sum y)`, because the sums are scalars.

Essentially, `select` and `derive` would operate row-wise, `window` would operate column-wise and `aggregate` would do something in-between.

Examples:
```diff
  from tracks
- derive {normalized_price = price / average price}
+ window {avg_price = average price}
+ derive {normalized_price = price / avg_price
```

```diff
  from tracks
- derive {rnk = rank}
+ window {rnk = rank}
```

This would also resolve the unexpected ["windowing by default"](https://prql-lang.org/book/transforms/window.html#windowing-by-default) behavior.

eitsupi:
To me this seems like a more consistent change, but I am concerned about the conflict between a function that takes only one argument, such as `lag`, and a `window` that takes a range.

```elm
from orders
sort day
# Collect
window rows:-3..3 {centered_weekly_average = average value}
```

```elm
from orders
sort day
# Collect?
window rows:-3..0 {lag_value = lag value}
```

```elm
from orders
sort day
# Error?
window rows:-3..3 {lag_value = lag value}
```

snth:
[...]

PR #2816, opened 2023-06-12
feat: evaluator

aljazerzen:
This is a PRQL interpreter. It does not allow reading any data from outside sources, but only computes the values of expressions, defined in PRQL.

For example:
```
1 + 2
```
... is evaluated into:
```
3
```

I've implemented the following things:
- feat: basics of eval
- pipelines
- select, derive, filter
- aggregate
- arg closures
- window and columnar

The purpose of this a tool to aid language design, issue #2723 in particular. I've implemented behavior of "Solution 2", as a trial to see how feasible it is.

Adapting this evaluator is much easier than adapting the whole compiler, so this is a nice tool for experimentation with new language features.


prql-compiler/prqlc/src/cli.rs
```diff
@@ -59,6 +60,12 @@ enum Command {
         input: clio_extended::Input,
     },
 
+    /// Parse & evaluate & generate PRQL code back
+    // TODO: this is not meant to be used apart from language design helper
+    // it is not polished (will panic a lot), so should we prefix with `debug:`?
```
max-sixty: Yeah if we can put it behind a `beta` group, I've seen that pattern work well. (Though we don't really have groups yet; we could do that with the `Sql` items too, in the meantime `debug:` seems fine)

prql-compiler/src/semantic/eval/mod.rs
```diff
@@ -0,0 +1,558 @@
...
+            [std.floor (3.5 + 2.9) + 3, 3]
+        ").unwrap(),
+            @"[9, 3]"
+        );
+    }
+
+    #[test]
+    fn tuples() {
+        assert_display_snapshot!(eval(r"
+              {{a_a = 4, a_b = false}, b = 2.1 + 3.6, c = [false, true, false]}
```
max-sixty, line 500: I didn't know we supported nested types yet!

max-sixty (commented):
No objections, looks like impressive code; I'm not sure what we'll use it for yet, vs. using DuckDB as an engine, but you have much more context than me....

prql-compiler/src/semantic/eval/mod.rs
aljazerzen: Well, the compiler does not support it, but you can parse them :D

Which is enough for the evaluator.

aljazerzen:
No, this is not an engine that we'd actually use - it's for trying language features out without having to implement the whole compilation to SQL. For testing out the corner cases and finding the cases of ambiguity.

Think of it as a "semantics specification" of everything that PRQL should be capable of - nested arrays, tuples, weird functional programming patterns, actual recursive functions. But encapsulated into PRQL source, no outside data sources.

aljazerzen:
Aside: I'm actually quite impressed with language design, as the core of the language semantics fits into 400 SLOC.

snth:
Great idea for iterating on ideas!

Can you give an example of how I would actually use it?

So I run
```bash
prqlc eval "{{a_a = 4, a_b = false}, b = 2.1 + 3.6, c = [false, true, false]}"
```
and then what do I get back?

aljazerzen:
So idea is that some PRQL expressions are values are some are not:
- all literals are values,
- arrays of values are values,
- tuples of values are values,
- functions are values,
- function calls are not values,
- operations are not values.

Eval takes an arbitrary expression and converts it into a value. So, evaluating function calls and operations.

Your example is already a value, so it would not be changed. But you could pipe it trough `derive  {d = (floor b) + 2}`, which would return:
```
{{a_a = 4, a_b = false}, b = 2.1 + 3.6, c = [false, true, false], d = 7}
```

PR #2819, opened 2023-06-12
revert: CI split, #2788

max-sixty:
I realized I was wrong here and found the actual cause of the issue (where the issue was it repeating compilation after splitting the compile & test step)

### DIFF
--- a/.github/workflows/test-rust.yaml
+++ b/.github/workflows/test-rust.yaml
@@ -63,11 +63,10 @@ jobs:
         with:
           command: test
           args:
-            --no-run --locked ${{ inputs.target== 'x86_64-unknown-linux-gnu' &&
-            '--features=test-external-dbs' || '' }}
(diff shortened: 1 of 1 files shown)


.github/workflows/test-rust.yaml
```diff
@@ -66,18 +75,13 @@ jobs:
         with:
           resource: "tcp:5432 tcp:3306 tcp:1433"
           timeout: 60000
-        if: ${{ inputs.target== 'x86_64-unknown-linux-gnu' }}
-        # We previously split up the test compilation as recommended in
-        # https://matklad.github.io/2021/09/04/fast-rust-builds.html, but the
-        # binary is different for `cargo insta test`, so it ends up compiling
-        # twice.
```
eitsupi: Is this change correct?
I believe this prerequisite is necessary because it seems that if the Docker container does not exist it will not succeed.

max-sixty: Sorry! The comment is correct to remove; the `if` is wrong

PR #2826, opened 2023-06-13
test(integration): ensure sort on every query

eitsupi:
Follow up for #2815

ClickHouse does not have a fixed order of rows in the results.
Therefore, this PR adds a `sort` to all queries to guarantee the order of rows across DBs.

### DIFF
--- a/prql-compiler/tests/integration/README.md
+++ b/prql-compiler/tests/integration/README.md
@@ -14,8 +14,8 @@ to 112kB, but that would require a preprocessing step before running
 
 ## Query
 
-For databases like ClickHouse, where the order of results is ambiguous,
-please use `sort` for test queries to to guarantee the order of rows across DBs.
+For databases like ClickHouse, where the order of results is ambiguous, please
+use `sort` for test queries to to guarantee the order of rows across DBs.
 
 For example, instead of the following query:
 
--- a/prql-compiler/tests/integration/README.md
+++ b/prql-compiler/tests/integration/README.md
@@ -12,6 +12,24 @@ tables are stored as CSV files. Their current size is 432kB, it could be gzip-ed
 to 112kB, but that would require a preprocessing step before running
 `cargo test`.
 
+## Query
+
+For databases like ClickHouse, where the order of results is ambiguous,
+please use `sort` for test queries to to guarantee the order of rows across DBs.
+
+For example, instead of the following query:
+
+```elm
+from albums
+```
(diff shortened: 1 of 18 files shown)


eitsupi:
Now it return empty on ClickHouse

https://github.com/PRQL/prql/blob/3c2860b7756f030841ce74c9960820b61920b919/prql-compiler/tests/integration/queries/genre_counts.prql#L2-L9

```sql
MySQL [dummy]> WITH genre_count AS (
    ->   SELECT
    ->     COUNT(name) AS a
    ->   FROM
    ->     genres
    -> )
    -> SELECT
    ->   - a AS a
    -> FROM
    ->   genre_count
    -> WHERE
    ->   a > 0;
Empty set (0.037 sec)
```

It works.

```elm
let genre_count = (
    from genres
    aggregate {a = count name}
)

from genre_count
```

```sql
MySQL [dummy]> WITH genre_count AS (
    ->   SELECT
    ->     COUNT(name) AS a
    ->   FROM
    ->     genres
    -> )
    -> SELECT
    ->   a
    -> FROM
    ->   genre_count;
+------+
| a    |
+------+
|   25 |
+------+
1 row in set (0.010 sec)
```

Perhaps the order of execution of WHERE and SELECT is different only for ClickHouse?

eitsupi:
This SQL

```sql
WITH _tab AS (
  select 1 AS x
)

SELECT -x AS x
FROM _tab
WHERE x == 1
```

returns empty in ClickHouse but returns this in DuckDB:

```
D WITH _tab AS (
>   select 1 AS x
> )
>
> SELECT -x AS x
> FROM _tab
> WHERE x == 1;
┌───────┐
│   x   │
│ int32 │
├───────┤
│    -1 │
└───────┘
```

Changing `-x AS x` to `-x AS y` works as expected in ClickHouse.

```sql
WITH _tab AS (
  select 1 AS x
)

SELECT -x AS y
FROM _tab
WHERE x == 1
```

```
MySQL [dummy]> WITH _tab AS (
    ->   select 1 AS x
    -> )
    -> 
    -> SELECT -x AS y
    -> FROM _tab
    -> WHERE x == 1;
+------+
| y    |
+------+
|   -1 |
+------+
1 row in set (0.007 sec)
```

I believe this is a ClickHouse bug.

eitsupi:
I reported this issue ClickHouse/ClickHouse#50921

PR #2829, opened 2023-06-13
test: enable `DISTINCT ON` integration test for DuckDB

eitsupi:
It seems working now.


max-sixty:
Thanks!

One thing we could do is test that the `skip` examples fail — that's work worked well for the other tests — it means we enable them as soon as they start working. Otherwise we often forget about them.

PR #2830, opened 2023-06-13
devops(devcontainer): add sqlite3 and duckdb cli

eitsupi:
It is useful to have these installed when checking prqlc behavior.


eitsupi:
(I originally wanted to install usql, but unfortunately there was no binary release for arm64 Linux)

max-sixty:
Does DuckDB take ages to compile & install? Or it copies a binary? 

Guessing the latter given your comment re `usql`. Otherwise we could add to the base image...

eitsupi:
Yes, simply download the latest version of the single binary from the GitHub release page. It should take less than a few seconds.

max-sixty:
Great!

max-sixty (approved):

PR #2835, opened 2023-06-13
docs: Update dialect description in book

### LINKED ISSUE #2506: Reduce MSSQL support?
### What's up?

(This will probably fall into the broader discussions we're having re finding PRQL's short-term focus. This provides a case of where giving up some guarantees would allow us to move faster on the margin)

MSSQL is the most difficult to support dialect:
- Doesn't support `USING`, which would give us natural joins #723 (I'm a big supporter of these, but actually I'm not sure everyone is?)
- Doesn't have any regex operator, re https://github.com/PRQL/prql/pull/2458
- Has issues such as https://github.com/PRQL/prql/issues/2448

Now we have the integration tests — which are really excellent — we either need to make lots of small examples so we can test "around" MSSQL, or just exclude it from any example with the regex operator.

I'm also not sure how popular it is? (Though FWIW have had positive experiences with it!)

Re the wider discussions — one option between "only support DuckDB" and "support everything" is we have tiers [like we do for Bindings](https://prql-lang.org/book/bindings/index.html). We'd run integration tests on only the Supported ones, and the others would be best-effort, and MSSQL would drop down.

max-sixty:
We're much better than we used to be, we should reflect that in the book.

Also demotes MsSql support level, which closes #2506

### DIFF
--- a/prql-compiler/src/sql/dialect.rs
+++ b/prql-compiler/src/sql/dialect.rs
@@ -81,16 +81,16 @@ impl Dialect {
             | Dialect::SQLite
             | Dialect::Postgres
             | Dialect::MySql
-            | Dialect::MsSql
+            | Dialect::Generic
             | Dialect::ClickHouse => SupportLevel::Supported,
-            Dialect::Generic | Dialect::Ansi | Dialect::BigQuery | Dialect::Snowflake => {
+            Dialect::MsSql | Dialect::Ansi | Dialect::BigQuery | Dialect::Snowflake => {
                 SupportLevel::Unsupported
             }
             Dialect::Hive => SupportLevel::Nascent,
         }
     }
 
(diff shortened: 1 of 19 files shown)


prql-compiler/src/sql/dialect.rs
```diff
@@ -81,16 +81,16 @@ impl Dialect {
             | Dialect::SQLite
             | Dialect::Postgres
             | Dialect::MySql
-            | Dialect::MsSql
```
eitsupi, line 84: I suppose this change stopping MSSQL tests.
https://github.com/PRQL/prql/blob/d614a5286a63f496eec1a56ba209ec2d1119e644/prql-compiler/tests/integration/main.rs#L288

max-sixty: Good point. I've redone how these are handled...

PR #2837, opened 2023-06-14
refactor!: remove the `sql.hive` target

eitsupi:
From #2831

Since no special features are implemented for the Hive dialect, it probably makes sense to remove this now.

## Note

Remember to update prql-vscode later.
https://github.com/PRQL/prql-vscode/blob/bba543afcbbbd1bc3066cb760e30240d088c5a8a/package.json#L154

-> Done PRQL/prql-vscode#239

### DIFF
--- a/bindings/prql-elixir/lib/prql.ex
+++ b/bindings/prql-elixir/lib/prql.ex
@@ -16,7 +16,6 @@ defmodule PRQL do
           | :ansi
           | :bigquery
           | :clickhouse
-          | :hive
           | :sqlite
           | :snowflake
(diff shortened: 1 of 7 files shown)


max-sixty (approved):

max-sixty:
Thanks! A changelog note would be great given the breaking change!

eitsupi:
> Thanks! A changelog note would be great given the breaking change!

Definitely.
However, the changes regarding the target are extensive, so I will update Changelog in a separate PR.

eitsupi:
I updated changelog in #2843

Issue #2841, opened 2023-06-14
select Multiple Columns


whoarusty:
### What's up?

I tried searching my issue, but I can't find anything so far.

Using prqlc 0.8.1 I'm trying to compile the example in the PRQL Book to select multiple columns, but I get an error:

```
prqlc-x86_64-pc-windows-msvc\prqlc-v0-x86_64-pc-windows-msvc>prqlc compile
Enter PRQL, then press ctrl-z to compile:

from e=employees
select [e.first_name, e.last_name]

Error:
   ╭─[Root.prql:2:8]
   │
 2 │ select [e.first_name, e.last_name]
   │        ─────────────┬─────────────
   │                     ╰─────────────── unexpected `[_frame.e.first_name, _frame.e.last_name]`
   │
   │ Help: this is probably a 'bad type' error (we are working on that)
───╯
```

Does the "Help" message mean this is known issue?

I hope I'm not missing something on my end.

max-sixty:
This is a 0.8.1 vs 0.9.0 issue (though the version will still say 0.8.1, since we bump the version just before we release at the moment).

The new syntax is `{e.first_name, e.last_name}`. 

There's no online book with the new syntax, so unless you're up for building it locally with `task run-book`, it's possible to kinda see it on GitHub, with imperfect formatting https://github.com/PRQL/prql/tree/main/web/book/src/syntax. That said, not _that_ much has changed in simple queries (#2689 for tracking) — mostly just the tuple / list syntax — so working from the old book should be OK.

I added this as a bad error message in https://github.com/PRQL/prql/pull/2842

whoarusty:
OK, thanks.

I look forward to the 0.9.0 release and will start referencing the syntax link you provided before I ask other questions ;)

max-sixty:
No prob — questions always welcome!

PR #2842, opened 2023-06-14
test: Bad error message from #2814

max-sixty:


eitsupi:
Maybe #2841? (not #2814)

PR #2849, opened 2023-06-16
docs: Update CHANGELOG.md - array & tuple discussion

richb-hanover:
I think this explanation/justification for switching from [] to {} is less confusing than the original

Also - are Arrays the same as "Relation Literals"? If so, the latter section should make a bigger nod toward using the new [] syntax.


max-sixty:
That is indeed better, thanks.

> Also - are Arrays the same as "Relation Literals"? If so, the latter section should make a bigger nod toward using the new [] syntax.

Relations consist of an array of tuples, and a relation literal is an array of tuples defined inline / literally. At the moment, that's the only place they show up.

But in theory arrays are more general; e.g. https://www.postgresql.org/docs/current/arrays.html

PR #2851, opened 2023-06-16
feat: elide parentheses in aliases in tuples

max-sixty:
This doesn't work at the moment. Need to think more about how to design it, some notes are inline


max-sixty:
To put some notes here — copying from https://github.com/PRQL/prql/issues/2757#issuecomment-1593617692

--- 

As a reminder, the reason we require parentheses around function calls in `select x = (sum foo)` but not in `{x = sum foo}` is that the latter has an `=` after the first ident, and so isn't a function call.

`x` is aliased to `sum foo`
```elm
select x = (sum foo)
```

...but here, `x` is aliased to `sum` and then `select` receives a single arg of `foo`...
```elm
select x = sum foo  # wrong!
```

`e` is aliased to employees (not to `employees (==id)`)
```
join e=employees (==id)
```

Here, `x` can't be a function call, since after the initial ident there's a `=`
```
derive {
  x = sum foo
}
```

max-sixty:
The main issue is that both `derive x = (sum foo)` and `derive {x = sum foo}` represent the alias identically, so the only distinction in PL is that the latter is in a tuple.

```
echo 'from x
   derive x = (sum foo)
   derive {x = sum foo}' | cargo run -p prqlc -- parse
```

I've labelled the non-tuple & tuple case here:

```yaml
sources:
  '':
  - name: main
    VarDef:
      value:
        Pipeline:
          exprs:
          - FuncCall:
              name:
                Ident:
                - from
              args:
              - Ident:
                - x
          - FuncCall:
              name:
                Ident:
                - derive
              args:
              - FuncCall:  # Non-tuple
                  name:
                    Ident:
                    - sum
                  args:
                  - Ident:
                    - foo
                alias: x
          - FuncCall:
              name:
                Ident:
                - derive
              args:
              - Tuple:
                - FuncCall:  # Tuple case
                    name:
                      Ident:
                      - sum
                    args:
                    - Ident:
                      - foo
                  alias: x
      ty_expr: null
      kind: Main
    annotations: []
source_ids:
  0: ''
```

max-sixty:
One option would be to give the writer much more access to the AST — e.g. it can find the current node and inspect its parents. Would this be viable? Or is it indicative of overly complicated requirements?

richb-hanover:
@max-sixty @aljazerzen 

> As a reminder, the reason we require parentheses around function calls in select x = (sum foo) but not in {x = sum foo} is that the latter has an = after the first ident, and so isn't a function call.
> 
> x is aliased to sum foo
> 
> select x = (sum foo)
> ...but here, x is aliased to sum and then select receives a single arg of foo...
> 
> select x = sum foo  # wrong!
> e is aliased to employees (not to employees (==id))
> 
> join e=employees (==id)
> Here, x can't be a function call, since after the initial ident there's a =
> 
> derive {
>   x = sum foo
> }

This still feels confusing: I don't get the distinction between when something is or isn't a "function". 

From the description above, it feels to me that the question is how PRQL determines the "end" of an alias. Here's my attempt to unwind the current state of affairs:

* In PRQL `x = y` is an "alias" and means "x is a (literal) alias for y", not an "assignment" as in C or Python. True?

* The PRQL rule for an alias is that the next identifier _(correct term?)_ after `=` is the replacement. True?

* In `select x = sum foo # wrong`, the problem is that `x` becomes the alias for `sum`, the first identifier, not the entire remainder of the line. True?

* In `select x = (sum foo) # OK`, `x` becomes an alias for `(sum foo)`. True?

* in `select { x = sum foo } # OK` or `select { x = sum foo, ... } # OK` the `{}` or `,` delimiters mark the end of the alias. Parentheses are not required in either of these cases. True? _I think not. See my next comment_

* in `join e=employees (==id) # OK`, the `e` is an alias for `employees`. True?

* in the `derive` case (below), the EOL also marks the end of the alias. True?
* alternatively, the `{}` delimits the alias. True?  _I think not. See my next comment_
   ```
  derive {
      x = sum foo # OK
   }
   ```
* In all cases, it is allowed to wrap function calls in `( ... )` to make the intent clear. True?

Thanks for listening.

richb-hanover:
Hmmm... I thought I was out of the woods, but playing with the newest commit (db5bd5c4), I see that things are more complicated...

* `select x = sum foo + 5 # wrong` generates an unknown name error for `foo` since the alias is solely for `sum` (same as above) True?

* `select { x = sum foo + 5 } # wrong` generates an error, I think because PRQL tries to consume the full expression into the alias, and `sum` can only take a column name, not an expression. True?

* `select { x = (sum foo) + 5 }  # OK` works as expected. True!

Thanks

aljazerzen:
> "x is a (literal) alias for y", not an "assignment" as in C or Python. True?

True, but you can think of it an assignment. How would you differentiate between the two?

> that the next identifier (correct term?)

*next expression

> in the derive case (below), the EOL also marks the end of the alias. True?

Technically not - the `}` marks the end of the assigned expression.

> In all cases, it is allowed to wrap function calls in ( ... ) to make the intent clear. 

---

I think that all of the other statements are true.

aljazerzen:
Are things clear now? Does the syntax make sense?

max-sixty (commented):
Great approach! Clever...

max-sixty:
Great set of cases @richb-hanover ! 

This one was the most interesting:

> * `select { x = sum foo + 5 } # wrong` generates an error, I think because PRQL tries to consume the full expression into the alias, and `sum` can only take a column name, not an expression. True?

The rule is that parentheses are required around a function call, unless the function call is the only expression in a tuple item. So `select {x = sum foo}` is OK, but `select {x = sum foo + 5}` requires `select {x = (sum foo) + 5}` because it has multiple expressions.

Are there others that are still unclear?

FWIW as I wrote in the book — even if this is coherent, it's still not simple. But I don't think there's a much simpler way. Maaaybe mandating parentheses around function calls, but would add lots of syntactic noise.

richb-hanover:
I'm still struggling with the basics, specifically how to define things, and then how to explain them. I'm using the `main` branch (db5bd5c4) to test this. Random thoughts follow:

- assign vs alias: It seems that `=` "does the right thing" depending on context:
   - in `from` and `join`, the line `t = ...` assigns an alias. An alias just saves typing: PRQL in effect substitutes the full table name wherever `t` appears. True?
   - in `select` and `define` the `=` assigns a name to a column. It also evaluates the expression on the right-hand side to give the value for each row of the result. True?
   - Are there any other cases?
   - I agree we can call this process "assignment"

- Another sticking point is explaining the difference between these cases:
   - 1 `select x = sum             # Error - Have you forgotten ...?`    
   - 2 `select x = sum foo         # Error - unknown name` 
   - 3 `select x = sum foo bar     # Error - Too many arguments to function select`
   - 4 `select { x = sum }         # Error - Have you forgotten ...?`    
   - 5 `select { x = sum foo }     # OK` 
   - 6 `select { x = sum foo bar } # Error - Too many arguments to function sum`

   This leads to the observation that the second case is the "odd man out" - it seems to be handled entirely differently from all the others:
   - In the first case, the parser knows that `sum` takes an argument
   - In the second case, it appears the parser just passes `sum`, as if it doesn't know that it takes an argument;
   - In the third case, something detects an error. I'm not sure what, since the message mentions `select`...
   - In the fourth case, the parser knows that `sum` takes an argument
   - In the fifth case, the parser just passes everything up to the closing `}` or `,` which works correctly;
   - In the sixth case, PRQL can detect that there are too many arguments to `sum`.
   - Do I have this right?

- Part of the "explanation conundrum" is that the error message for the second case is infuriating. PRQL gives "unknown name", but as a newcomer, I am certain that my table contains a column named "foo". 

- And in the third case, it's astonishing to see PRQL claiming `select` has too many arguments.

_Wild speculation follows:_ Why not just pass everything after the `=`, and let subsequent error processing handle problems? PRQL can count the passed-arguments, and has to handle errors in the remainder of the expression. 

Thanks for listening.

richb-hanover:
@max-sixty Our messages crossed in the aether... What do you think of my wild specuation? Thanks.

max-sixty:
> _Wild speculation follows:_ Why not just pass everything after the `=`, and let subsequent error processing handle problems? PRQL can count the passed-arguments, and has to handle errors in the remainder of the expression.

Does the example at https://github.com/PRQL/prql/pull/2851#issuecomment-1595181712 make sense? i.e. `join e=employees (==id)` wouldn't work — we're not aliasing `employees (==id)` to `e`.

---

I _think_ Alias vs Assign isn't a difference any longer, and I'm just misusing words. So everything should be Alias now...

---

Those are a great set of examples, and I agree that the example you point out is the most confusing, since whether or not it's inside a tuple is salient, which seems like it shouldn't be.

And then the error messages are bad, but we can improve those. Feel free to add them to `test_error_messages.rs` if they're not there already...

richb-hanover:
I see what you're saying about `join e=employees (...)`. But I'm still unsatisfied with case 2 being the only one that doesn't recognize that `sum` has arguments.

Here's an alternative definition that, I think, frees PRQL from knowing whether an expression is inside a tuple or not:

- For `from` and `join`, use the first identifier after `=` as the table name. And anything after that identifier in a `join` must be the join criteria
- For `derive` and `select`, use the expression up to the closing punctuation (EOL, `}` or `,` or `#` )

Would that work? I'm casting about to find a way to employ the same machinery used within a tuple when there's no tuple... Thanks.

richb-hanover:
We also need to consider the case where we're not using `std.sum` but perhaps a function that has multiple arguments. 

PRQL seems to know how to count arguments, for example try permutations of this:

```
let fn = v1 v2 -> v1*v2
from invoices
select x = fn foo bar     # gives error
select { x = fn foo bar } # gives expected result
```

max-sixty:
> Here's an alternative definition that, I think, frees PRQL from knowing whether an expression is inside a tuple or not:
> 
>     * For `from` and `join`, use the first identifier after `=` as the table name. And anything after that identifier in a `join` must be the join criteria
> 
>     * For `derive` and `select`, use the expression up to the closing punctuation (EOL, `}` or `,` or `#` )

The problem is that we don't know whether the name of the functions when we're parsing. It's just a word at that point.

One potential solution is to hardcode functions into the parser. But then the compiler becomes much more coupled — it would become extremely non-[context-free](https://en.wikipedia.org/wiki/Context-free_grammar). So that's not a good path.

richb-hanover:
I'm certainly not advocating that we add more coupling between parser and compiler.

I'm grasping for a solution to a user trap. Newcomers will puzzle over PRQL's error messages for these two seemingly identical cases: 

```
select x = sum foo         # Error - unknown name foo
select { x = sum foo }     # OK
```
In the first case, the newcomer will be scrambling around trying to figure out why it's unknown when _they are certain_ their table contains a column named foo. (And of course, that's disastrously misleading because PRQL doesn't know anything about the table...)

I'm really leaning into this because it was one of the first places that the illusion began to crack for me. "Oh wait, PRQL doesn't know everything?..." I have been pretty motivated to "make PRQL work", so early on I may have internalized the rule, "Just put everything into a list (now tuple) and avoid the hassle" of this wonky treatment of the special case.

As I write this, a couple alternative ways to handle the conundrum above come to mind:

- Just require `{...}` on `select` and `derive`. 
- Treat everything between `=` and closing punctuation as the expression, and let subsequent processing give the error message (my earlier suggestion)
- Come  up with better diagnostic message than "unknown name foo". It should describe more correctly what's wrong with "foo" since (we know) it's not unknown

I will also note that I'm confused the current state where `select x = sum` offers the helpful hint: `Have you forgotten an argument to function std.sum?` How does PRQL know that `sum` needs an argument here? And why does it reject an argument when it's present?

As always, thanks for listening.

max-sixty:
> I'm grasping for a solution to a user trap. Newcomers will puzzle over PRQL's error messages for these two seemingly identical cases:
> 
> ```
> select x = sum foo         # Error - unknown name foo
> select { x = sum foo }     # OK
> ```

Yes, tbc, I couldn't agree with you more, the current state pains me. So I'm definitely not saying everything is good! I just can't see the way through yet.



> Just require `{...}` on `select` and `derive`.

Interesting. 

At the parsing stage, we wouldn't know `select x = sum foo` isn't `join x = sum (foo)`. Anything that is "handle `select` and `join` differently at the parsing stage" is a no-go. 

But we could have `select` only take a tuple, and raise an error in the later stage of compilation if passed a scalar; that might ameliorate it. But at the cost of some syntactic noise

> Treat everything between `=` and closing punctuation as the expression, and let subsequent processing give the error message (my earlier suggestion)

I think this makes the parser too weak — basically pass everything through as a blob. While our current error messages are often bad, this change would make them quite harder to improve

> Come up with better diagnostic message than "unknown name foo". It should describe more correctly what's wrong with "foo" since (we know) it's not unknown

Definitely we can do this, possibly I can have a go.

---

One nuclear option is to again split aliases vs assignments in syntax — e.g. use `:` for `join e:employees` while retaining `select x = sum foo`. I'm less keen, since these are conceptually quite similar, and I somewhat trust the general momentum of the project to combine alias & assignments.

richb-hanover:
I have studiously avoided looking at the grammars the PRQL project uses. My last compiler class was about 50 years ago (really!), and I didn't even get a good grade :-)  But I had rather expected that there would be different production rules for `select` and `join` that would determine how much of the input to consume in each circumstance. Maybe not.

I'm going to leave this in your hands. I've expressed (and you understand) my concerns. The PRQL team is far better at this kind of code, so I'll watch for a good solution, if nothing else, a better error message than "unknown name foo". Thank you.

PR #2856, opened 2023-06-17
docs: Update examples.js so the PRQL Playground examples use tuples

richb-hanover:
Checked to ensure none of the other queries in the Playground return errors

(This was much easier than I thought - only _examples.js_ needed to be touched)

### DIFF
--- a/web/playground/src/examples.js
+++ b/web/playground/src/examples.js
@@ -3,25 +3,25 @@ const examples = {
     "sql",
     `from invoices
 filter invoice_date >= @1970-01-16
-derive [                        # This adds columns
+derive {                        # This adds columns
   transaction_fees = 0.8,
   income = total - transaction_fees  # Columns can use other columns
-]
+}
 filter income > 1     # Transforms can be repeated.
 group customer_id (   # Use a nested pipeline on each group
-  aggregate [         # Aggregate each group to a single row
+  aggregate {         # Aggregate each group to a single row
     average total,
     sum_income = sum income,
-    ct = count,
-  ]
+    ct = count sum_income,
+  }
 )
-sort [-sum_income]    # Decreasing order
+sort {-sum_income}    # Decreasing order
 take 10               # Limit to top 10 spenders
 join c=customers (==customer_id)
 derive name = f"{c.last_name}, {c.first_name}"
-select [              # Select only these columns
+select {              # Select only these columns
   c.customer_id, name, sum_income
-]
+}
 derive db_version = s"version()" # S-string, escape hatch to SQL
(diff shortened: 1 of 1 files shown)


richb-hanover:
**Note:** The two files named _introduction.prql_ (at the top of the Playground and at the very bottom (in "Introduction")) are different. Do we care?

**Top**

```
from invoices
filter invoice_date >= @1970-01-16
derive {                        # This adds columns
  transaction_fees = 0.8,
  income = total - transaction_fees  # Columns can use other columns
}
filter income > 1     # Transforms can be repeated.
group customer_id (   # Use a nested pipeline on each group
  aggregate {         # Aggregate each group to a single row
    average total,
    sum_income = sum income,
    ct = count sum_income,
  }
)
sort {-sum_income}    # Decreasing order
take 10               # Limit to top 10 spenders
join c=customers (==customer_id)
derive name = f"{c.last_name}, {c.first_name}"
select {              # Select only these columns
  c.customer_id, name, sum_income
}
derive db_version = s"version()" # S-string, escape hatch to SQL
```

**Bottom**
```
from employees
filter start_date > @2021-01-01               # Clear date syntax
derive {                                      # `derive` adds columns / variables
  gross_salary = salary + (tax ?? 0),         # Terse coalesce
  gross_cost = gross_salary + benefits_cost,  # Variables can use other variables
}
filter gross_cost > 0
group {title, country} (                      # `group` runs a pipeline over each group
  aggregate {                                 # `aggregate` reduces each group to a value
    average gross_salary,
    sum_gross_cost = sum gross_cost,          # `=` sets a column name
  }
)
filter sum_gross_cost > 100_000               # `filter` replaces both of SQL's `WHERE` & `HAVING`
derive id = f"{title}_{country}"              # F-strings like Python
derive country_code = s"LEFT(country, 2)"     # S-strings allow using SQL as an escape hatch
sort {sum_gross_cost, -country}               # `-country` means descending order
take 1..20                                    # Range expressions (also valid here as `take 20`)
```

max-sixty:
Great, thanks @richb-hanover ! Merging and will respond to your other question

max-sixty:
I think we could combine those queries; they serve a similar purpose.

- The first's advantage is that it works with the sample data we have (i.e. it runs in the playground)
- The second's advantage is that it has a wider range of features, doesn't repeat concepts, has better comments describing the operations ("Transforms can be repeated" doesn't mean anything to someone seeing PRQL for the first time)

To the extent you wanted to rewrite something close to the second example to target our sample data, unify the queries, and replace both examples with that — that could be great!

Issue #2840, opened 2023-06-14
Update grammars for 0.9


max-sixty:
### What's up?

We should update the grammars for 0.9. This might just be fixing the tuple / list syntax. They're listed at https://prql-lang.org/book/internals/syntax-highlighting.html

(FWIW some of them could do with a refresh on keywords, or how they handle strings etc, but that's by no means necessary for 0.9)

- [x] Lezer (#2868)
- [x] Handlebars (#2871)
- [x] Monarch (#2874)
- [x] VS Code / Textmate (PRQL/prql-vscode#244)
- [ ] Tree-SItter (PRQL/tree-sitter-prql#13)

eitsupi:
The remaining task is tracked in another repo PRQL/tree-sitter-prql#13, so I close this for now.

PR #2868, opened 2023-06-20
fix(grammars/lezer): update syntax for 0.9.0

eitsupi:
Part of #2840

### DIFF
--- a/grammars/prql-lezer/src/prql.grammar
+++ b/grammars/prql-lezer/src/prql.grammar
@@ -17,9 +17,9 @@ Pipeline { expr_call (pipe+ expr_call)* }
 
 pipe { "|" | ~ambig_newline newline }
 
-List { "[" newline? list_item (("," newline? ) list_item)* ","? newline? "]" }
+Tuple { "[" newline? tuple_item (("," newline? ) tuple_item)* ","? newline? "]" }
 
-list_item { Assign_call | expr_call | Case_branch }
+tuple_item { Assign_call | expr_call | Case_branch }
 
 expr_call { Expr | Func_call }
 // Ideally we would force a space after `Ident` to prevent an invalid s-string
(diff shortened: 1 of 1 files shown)


grammars/prql-lezer/src/prql.grammar
```diff
@@ -96,7 +97,7 @@ Op_bin { Op_bin_only | Op_unary }
 
 Table_def { @specialize<ident_part, "let"> ident_part "=" Nested_pipeline ( newline+ | end ) }
 
-Func_def { @specialize<ident_part, "func"> Func_def_name Func_def_param* "->" expr_call ( newline+ | end ) }
+Func_def { @specialize<ident_part, "let"> Func_def_name "=" Func_def_param* "->" expr_call ( newline+ | end ) }
```
max-sixty: Does this handle the ambiguity with Table_def OK? https://github.com/PRQL/prql/pull/2868/files#diff-b9ff79077300e8e401d4b8b3fe1f55d89a0ea58ad1582012ef7f6ce867667f57R98 

I wasn't sure if you've tried it in the playground. If so, then great. If not, we could them consolidate to `Def`, and offer an option of `lambda` as a RHS, which would be `Func_def_param* "->" expr_call ( newline+ | end )`...

max-sixty (dismissed):
Looks good!

Did you try it with a couple of sample queries in the playground?

(I just pushed https://github.com/PRQL/prql/pull/2869 with a link, as I realize that wasn't documented)

grammars/prql-lezer/src/prql.grammar
eitsupi: This is a far more difficult task than I thought it would be, and it is out of my hands.

Added comment as TODO.
Currently this does not support function definitions and relation literals.
Unfortunately I could not find a solution to this after reading your comment.

You can merge this not working or close this and wait for someone else to work on this.

max-sixty (approved):
Yes, fine. Not ideal but agree getting closer to 0.9.0 is better than keeping this up to date!

grammars/prql-lezer/src/prql.grammar
max-sixty: No problem! I just Approved before seeing your comment, I agree it's better to move forward!

PR #2871, opened 2023-06-20
fix(grammars/handlebars): update syntax for 0.9.0

eitsupi:
Part of #2840


max-sixty (approved):
Great! This one is easier!

PR #2874, opened 2023-06-20
fix(grammars/monarch): update syntax for 0.9.0

eitsupi:
Part of #2840


max-sixty:
Great — interesting this doesn't have lists / tuples — makes it easier!

Issue #2282, opened 2023-03-22
Improving error messages


max-sixty:
### What's up?

Some of our error messages could be much better. We now have the Chumsky parser, on top of the ASTs, which should allow us to make these much great. At the moment, they're anywhere from "much better than SQL" to "a bit worse"...

Here's an example of a really bad error message:

```elm
from film
group
```

```
Error:
    ╭─[:?:?]
    │
    │
    │
    │ Help: are you missing `from` statement?
 ───╯
```

---

Probably one-issue-per-error-message is not a good way of managing them — there would be a huge number of issues, and it would be difficult to see what's current. So instead, I created a file [`test_bad_error_messages`](https://github.com/prql/prql/blob/main/crates/prql_compiler/src/tests/test_bad_error_messages.rs) where we can record bad error messages and seeded it with the above.

Two forms of contributions would be very helpful:
- Add bad error messages
- Take a bad error message, make it good, and move the test to [`test_error_messages`](https://github.com/prql/prql/blob/52fbc00c90d8c12375c73414c04fd190fcf554cd/prql-compiler/src/tests/test_error_messages.rs). The messages there are kept up-to-date with tests on every commit, so there's no risk of wasted effort, and the change can't regress. I left some quick notes for how to approach this, and it's also very encouraged to ask questions.
  - Link back to this issue in the PR, so other folks can we can gradually get a few examples of how these are done

khoa165:
Hi @max-sixty, I am pretty new to Open source in general so I would love to give this a try as it seems doable. Do I just make a PR that references this issue?

max-sixty:
Hi @khoa165 — very much so re opening a PR. 

There are more details on our contribution process [here](https://prql-lang.org/book/contributing/development.html).

How experienced with Rust are you? Some of the error messages will be easy-moderate difficulty. Some will be quite hard! And I don't have a confident view on which are which. So feel free to ask for help, to post something half-formed which doesn't yet work, etc.

If you're less experienced, I'd also encourage starting with any small contribution; I've found that a great way to work up to bigger contributions.

khoa165:
Hi @max-sixty, definitely first time seeing in Rust syntax so a bit I'm definitely "rusty" on this 😅 I worked with some other languages so hopefully I can pick up Rust as I start poking around and contributing more.

khoa165:
I have a question. Your existing function `array_instead_of_tuple` seems to indicate that the following snippet would result in error 
```
from e=employees
[...]

PR #2876, opened 2023-06-20
test: Add a bad error msg to track list

khoa165:
# Changes
- Add a test for a bad error message when query sort decreasing order but without square bracket

## Type
- Dev 

## Issue
https://github.com/PRQL/prql/issues/2282


prql-compiler/src/tests/test_bad_error_messages.rs
```diff
@@ -71,6 +71,20 @@ fn test_bad_error_messages() {
        │     Note: Type `relation` expands to `[tuple_of_scalars]`
     ───╯
     "###);
+
+    // It's better if we can tell them to put in [] bracketß
```
max-sixty:
```suggestion
    // It's better if we can tell them to put in {} braces
```

max-sixty (approved):
Perfect first PR. Great example of a terrible error message...

prql-compiler/src/tests/test_bad_error_messages.rs
khoa165: Is the book outdated too? Cuz I see square bracket being used there (screenshot below)
Screenshot 2023-06-20 at 1 16 44 AM

max-sixty: Yes, all the docs are on 0.8.1; then `main` has the new syntax

max-sixty: `task run-book` to see the current book locally...

PR #2880, opened 2023-06-20
devops: `run-playground` task

max-sixty:
I hadn't realized that #2801 stopped building the rust code by default. This changes it slightly so running `run-playground-cached` avoids building the rust code.

It removes `run-pg`, since the JS build without the rust build takes just a few seconds

### DIFF
--- a/Taskfile.yml
+++ b/Taskfile.yml
@@ -297,13 +297,14 @@ tasks:
     desc: Build & serve the playground.
     dir: web/playground
     cmds:
-      - task: install-playground-npm-dependencies
+      - npm install --install-links=false
       - npm start
 
-  run-pg:
-    desc: Serve the playground without building it
+  run-playground-cached:
+    desc: Build & serve the playground, without rebuilding rust code.
     dir: web/playground
     cmds:
+      - task: install-playground-npm-dependencies
       - npm start
 
   install-playground-npm-dependencies:


eitsupi:
I still feel like spending 2 hours doing a Docker build every time we edit an unrelated task on the Taskfile is a waste of time..... (from #2297)

max-sixty:
> I still feel like spending 2 hours doing a Docker build every time we edit an unrelated task

To ensure I understand — that's just CI time though, right? Or is there time for the user?

If it's CI time — that costs less than $0.01. So if it saves a bit of complexity in the repo, it pays for itself with a few orders of magnitude to spare...

No? Or do you see it differently? Am very open!

eitsupi:
I have a completely different perspective. My mention is CI time, which I believe actually consumes user time and storage too.
This means that an image with a completely different layer is pushed to the registry even though virtually nothing has changed, consuming time and storage each time the user pulls.

max-sixty:
OK I agree there's some tradeoff! I'm trying to understand what the relative weights are though... 

I just pulled the latest image — you're correct the final layer required pulling — the new layer was only 162MB, and took 16s including the extraction... There have been [11 builds on main in the past two months](https://github.com/PRQL/prql/actions/workflows/build-devcontainer.yaml?query=branch%3Amain), which push a new image.

If you feel very strongly (or others want to weigh in), then fine... My reservation is that each thing can be made a bit more efficient by adding complexity. But the complexity affects the whole project, while the efficiency only affects the individual thing.

An aside — if I think about the inefficiency of CI, it's terrible — we build the whole project and run every test, just to check a couple of new lines! It's so wasteful — and it's _still_ really valuable. This is one of the least wasteful parts 😁  (which doesn't mean we shouldn't try and improve it, tbc...)

eitsupi:
Note that if a layer is changed, all subsequent layers will be rebuilt.
This means that the user must rebuild the Python and node.js layers installed on top of the base image via the Dev Container CLI.

BTW, I don't think splitting the Taskfile is that much added complexity.
I consider it almost like splitting `foo.rs` from `main.rs`. (and the Taskfile of this repository is huge)

max-sixty:
Yes, OK, you make good points.

> Note that if a layer is changed, all subsequent layers will be rebuilt.
> This means that the user must rebuild the Python and node.js layers installed on top of the base image via the Dev Container CLI.

Can I ask one question — when does VS Code do a rebuild? Does it proactively pull a new image if one exists on `latest`? Or only if we were to add specific tags to the `.devcontainer.json`?

If it proactively pulls, we could reduce that by only pushing images outside of releases or the stable branch (currently named `web`).

---

That said, if you feel strongly, feel free to split up the files — I don't want to stand in your way of making improvements. The file is fairly stable now, and so we're less likely to have to manage the mapping of which parts of the file affect the job, which was some of the complexity.

We can put the sub-files into `.config` to reduce clutter in the root...

eitsupi:
> Can I ask one question — when does VS Code do a rebuild? Does it proactively pull a new image if one exists on latest? Or only if we were to add specific tags to the .devcontainer.json?

I think this is the same as a normal `docker build`, when the local image is updated, i.e., when the layer pointed to by `ghcr.io/prql/prql-devcontainer-base:latest` is changed by the user with the `docker pull` command.

https://github.com/PRQL/prql/blob/bc0f81072bda0ce18d2cc5df0c0bc915cf50887f/.devcontainer/devcontainer.json#L3

max-sixty:
> I think this is the same as a normal `docker build`, when the local image is updated, i.e., when the layer pointed to by `ghcr.io/prql/prql-devcontainer-base:latest` is changed by the user with the `docker pull` command.

IIRC, while `docker pull` will pull a more recent image at the same tag, `docker build` by default will not if it already has a base image at that tag — that requires `docker build --pull`

PR #2881, opened 2023-06-20
build: Replace `wasm-pack`

max-sixty:
This is the beginning of #1836, replacing `wasm-pack` with https://crates.io/crates/substrate-wasm-builder. `wasm-pack` requires additional binaries to be installed outside of the normal `cargo build` process, has been the source of lots of issues, makes builds slower — in particular small code changes cause a very long iteration cycle in playground builds — and is only lightly maintained now.

Using `wasm-builder` is a bit harder than I thought:
- We need to create the paths `bundler`, `node`, `web` paths ourselves, replacing [these lines](https://github.com/PRQL/prql/blob/5db61f45ad31ff5673fdfb7890cc3699d47940f9/bindings/prql-js/package.json#L20-L21), by calling `wasm-bindgen debug/wbuild/prql-js/prql_js.wasm --target=...`.
- We probably need some way of getting the `.wasm` file location, by reading `include!(concat!(env!("OUT_DIR"), "/wasm_binary.rs"));` for the `WASM_BINARY` or `WASM_BINARY_BLOATY` constants. Ideally we'd do this at build time, all in the `build.rs` script; I'm not sure whether that's possible though.
- Possibly I'm figuring too much of this out on my own and it's done elsewhere? The precedent for this approach seems to be blockchain rather than JS applications... I'll ask around to see if I'm missing anything.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -85,6 +85,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "ansi_term"
+version = "0.12.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "d52a9bb7ec0cf484c551830a7ce27bd20d67eac647e1befb56b0be4ee39a55d2"
+dependencies = [
+ "winapi",
+]
+
+[[package]]
 name = "anstream"
 version = "0.3.2"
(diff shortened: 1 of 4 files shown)


max-sixty:
Postponing given context at #1836, somewhat replaced by #2907

PR #2884, opened 2023-06-20
ci: build release artifacts nightly

eitsupi:
ref #2639, PRQL/pyprql#209

Set up nightly builds of prqlc and prql-python wheels to be the same as at the time of release.
Doing this with every push was clearly excessive.

### DIFF
--- a/.github/workflows/nightly.yaml
+++ b/.github/workflows/nightly.yaml
@@ -184,3 +184,27 @@ jobs:
 
       - name: Perform CodeQL Analysis
         uses: github/codeql-action/analyze@v2
+
+  build-prqlc:
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - os: ubuntu-latest
+            target: x86_64-unknown-linux-musl
+          - os: ubuntu-latest
+            target: aarch64-unknown-linux-musl
+          - os: macos-latest
+            target: x86_64-apple-darwin
+          - os: macos-latest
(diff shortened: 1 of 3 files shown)


max-sixty (approved):
Great! Thank you!

>  Doing this with every push was clearly excessive.

Should we reduce this, then? I don't think this PR changes the prqlc build on PRs.

FWIW — I originally thought it was a bit excessive, but also appreciate your perspective and am happy to have it run frequently if you / others use them.

(We could also add them to `test-all.yaml`, then they run on each merge, or on request, if someone wants to try out `prqlc`. I leave the decision to you!)

eitsupi:
Thanks for your review!

> I don't think this PR changes the prqlc build on PRs.

I mean #2778 already changed that.

eitsupi:
> (We could also add them to `test-all.yaml`, then they run on each merge, or on request, if someone wants to try out `prqlc`. I leave the decision to you!)

My concern is these are numerous and take a long time to build, as you say, so every time I run it, we have to wait quite a bit if we are creating multiple PRs at the same time.

If we want to try a release build we can use the `pr-nightly` tag...... (great design!)

max-sixty:
> I mean #2778 already changed that.

Ah great, I see what you mean by `release` — yes this is a great balance — we get the fast debug builds, but also the tests for `release` before we do the actual releases!

eitsupi:
But thinking about it again, it may be better to call `release.yaml` using `workflow_call` than to copy the contents.

PR #2893, opened 2023-06-21
docs: Add pull request template

khoa165:
## Description
- Add a pull request template

## Motivation
- Quick to fill out the next time someone makes a Pull Request. I realized that when I made my first PR in this repo, I wasn't sure what to put in the PR description since there was no template to fill in.

## Type
- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [X] Documentation
- [ ] Improvements

## Tests
- Looked at preview of Markdown to make sure it looks right

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,0 +1,20 @@
+## Description
+
+- List your changes here
+
+## Motivation
+
+- [e.g. issue #1234](Link to issue here)
+- Other motivations for this change
+
+## Type
+
+- [ ] Bug fix
+- [ ] New feature
+- [ ] Breaking change
+- [ ] Documentation
+- [ ] Improvements
+
+## Tests
+
+- List how you verified your changes


khoa165:
Hi @max-sixty I wonder if when I make a new PR, do I need to tag someone specifically? Normally, at companies I work for, I can request reviewer(s) directly but since I don't have write permissions here, I suppose the only way is to tag someone to notify.

max-sixty:
I'm open to a template, but I'm not sure the current proposal adds much that is missing from most PRs. We get the type of PR from the semantic commit message (`docs:`), and the remainder is rarely missing.

I'd rather have more, lower quality PRs! I can't recall a time we ever had a net-negative PR...

Does anyone @PRQL/prql-team have ideas for things they'd like to see?

aljazerzen:
We don't get enough PRs for this to make sense. As you say max, we already enforce `docs:` or `feat:` and even `feat!:` for breaking changes.

Thanks for taking the time to write this down!

max-sixty:
OK. I'll close. I appreciate the effort @khoa165 (and lmk if you think we're missing something, maybe there's a smaller version of this that would make sense)

PR #2896, opened 2023-06-21
build: Upgrade duckdb to `0.8.1`

max-sixty:
Previously reverted in #2892. We can attempt to fix here.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -160,9 +160,9 @@ checksum = "8da52d66c7071e2e3fa2a1e5c6d088fec47b593032b254f5e980de8ea54454d6"
 
 [[package]]
 name = "arrow"
-version = "39.0.0"
+version = "41.0.0"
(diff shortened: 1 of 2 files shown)


max-sixty:
~This looks like it's working...~

~@eitsupi do you happen to know about the differences in the toolchains? Is it OK to test in CI on `-windows-gnu`? Or is the default `-windows-msvc` and it might break on local machines?~

Edit: no actually... :)

eitsupi:
`windows-gnu` target is for something like MinGW, right?
Probably not very common. (Although used in the R package)
https://github.com/eitsupi/prqlr/blob/944b1964f6c9bdc85ccae308f3db3b204a8f5770/src/Makevars.win#L1

max-sixty:
Yes, to confirm, the goal here is to get duckdb 0.8.1 compiling, not to use `-gnu` — that was just an idea based on the recommendation in https://stackoverflow.com/a/68596301

PR #2894, opened 2023-06-21
test: Add Stack-Overflowing query

max-sixty:
#2857

PR #2899, opened 2023-06-21
build: Use a feature for integration tests

max-sixty:
As discussed in https://github.com/PRQL/prql/pull/2894#issuecomment-1601395315


max-sixty:
~This is still a default feature at the moment, so to skip it requires `--no-default-features`. That's easy to change if we decide we don't want it to be default.~

PR #2912, opened 2023-06-22
feat(prql-python): add type hints

philpep:
Following documentation in https://pyo3.rs/v0.18.1/python_typing_hints#if-you-do-not-have-other-python-files


max-sixty:
Great @philpep , thank you!

Do we also need to add an empty `py.typed` file? https://github.com/PyO3/maturin/blob/084cfaced651b28616aeea1f818bdc933a536bfe/guide/src/project_layout.md#adding-python-type-information

max-sixty:
If we can test this, that would be great — I think just adding `-> None` to the tests in `test_all.py`, then I _think_ pre-commit will pick that up so that the test is run on every commit.

(That said — if you don't have the patience then we can do this later, I'm fairly confident the current code is correct...)

philpep:
Quoting the documentation mentioned above: _In a pure Rust project, add type stubs in a <module_name>.pyi file in the project root. Maturin will automatically include this file along with the required py.typed file for you._

I forgot to add type for `__version__`, I'll fix this.

Yes running mypy is a good but prql_python might not be installed when running pre-commit. So I propose to add a nox environment for this.

philpep:
CI run : https://github.com/PRQL/prql/actions/runs/5356393251/jobs/9715943140?pr=2912#step:8:56

```
nox > mypy 
Success: no issues found in 3 source files
```

bindings/prql-python/mypy.ini
eitsupi, line 1: Can we use pyproject.toml instead of mypy.ini?

philpep: Sure, done.

eitsupi: Thanks!

max-sixty:
> Quoting the documentation mentioned above: _In a pure Rust project, add type stubs in a <module_name>.pyi file in the project root. Maturin will automatically include this file along with the required py.typed file for you._

Sorry, mea culpa

bindings/prql-python/pyproject.toml
```diff
@@ -29,3 +29,13 @@ ignore = [
   # No lambdas — too strict
   "E731",
 ]
+
+[tool.mypy]
+files = "."
+show_error_codes = true
+warn_unused_ignores = true
+strict = true
```
max-sixty, line 37: Great!

max-sixty:
Thank you @philpep . If you're up for adding a note to the Changelog, please go ahead in a mini-PR!

PR #2918, opened 2023-06-25
docs: Update playground queries

richb-hanover:
Update the `introduction.prql` query in the Playground to describe (most) PRQL features in a comment.

I would be glad for others to chime in either making this a more sensible query (it's kinda silly), to correct any descriptions, or to add important features. Thanks.


web/playground/src/examples.js
```diff
@@ -1,28 +1,31 @@
...
-take 10               # Limit to top 10 spenders
-join c=customers (==customer_id)
-derive name = f"{c.last_name}, {c.first_name}"
-select {              # Select only these columns
-  c.customer_id, name, sum_income
-}
-derive db_version = s"version()" # S-string, escape hatch to SQL
+join c=customers (==customer_id)     # join on "customer_id" from both tables
+derive name = f"{c.last_name}, {c.first_name}" # F-strings like Python
+derive db_version = s"version()"     # S-string give escape hatch to SQL
```
max-sixty:
```suggestion
derive db_version = s"version()"     # S-string offers escape hatch to SQL
```
(or `gives`)

web/playground/src/examples.js
```diff
@@ -1,28 +1,31 @@
...
+select {                             # Select certain columns
```
max-sixty:
```suggestion
select {                             # "select" takes only these columns
```

web/playground/src/examples.js
```diff
@@ -1,28 +1,31 @@
...
-    ct = count sum_income,
+    `from invoices                        # PRQL query begins with a table
+derive {                             # "derive" adds columns
+  transaction_fees = 0.8,            # "=" sets a column name
+  income = total - transaction_fees  # Calculations can use other column names
+}                                    # "#" starts a comment - even at start of line
+filter income > 1                    # "filter" replaces both of SQL's WHERE & HAVING
+filter invoice_date >= @1970-01-16   # Clear date syntax
+group customer_id (                  # "group" performs the pipeline in (...) on each group
+  aggregate {                        # "aggregate" reduces each group to a single value
```
max-sixty:
```suggestion
  aggregate {                        # "aggregate" reduces each group to a single row
```

web/playground/src/examples.js
```diff
@@ -1,28 +1,31 @@
...
-group customer_id (   # Use a nested pipeline on each group
-  aggregate {         # Aggregate each group to a single row
-    average total,
-    sum_income = sum income,
-    ct = count sum_income,
+    `from invoices                        # PRQL query begins with a table
+derive {                             # "derive" adds columns
+  transaction_fees = 0.8,            # "=" sets a column name
+  income = total - transaction_fees  # Calculations can use other column names
+}                                    # "#" starts a comment - even at start of line
```
max-sixty:
```suggestion
}                                    # "#" starts a comment
```
Seems implicit?

web/playground/src/examples.js
```diff
@@ -1,28 +1,31 @@
...
+filter income > 1                    # "filter" replaces both of SQL's WHERE & HAVING
+filter invoice_date >= @1970-01-16   # Clear date syntax
+group customer_id (                  # "group" performs the pipeline in (...) on each group
+  aggregate {                        # "aggregate" reduces each group to a single value
+    average total,                   # get the average of the total for each customer_id
+    sum_income = sum income,         # ... and the total of income
+    ct = count customer_id,          # ... and the count of items
```
max-sixty: Possibly these don't need comments?

web/playground/src/examples.js
```diff
@@ -1,28 +1,31 @@
...
-}
-filter income > 1     # Transforms can be repeated.
-group customer_id (   # Use a nested pipeline on each group
-  aggregate {         # Aggregate each group to a single row
-    average total,
-    sum_income = sum income,
-    ct = count sum_income,
+    `from invoices                        # PRQL query begins with a table
+derive {                             # "derive" adds columns
+  transaction_fees = 0.8,            # "=" sets a column name
```
max-sixty:
```suggestion
  transaction_fee = 0.8,             # "=" sets a column name
```
singular column names?

web/playground/src/examples.js
```diff
@@ -1,28 +1,31 @@
...
+  income = total - transaction_fees  # Calculations can use other column names
```
max-sixty:
```suggestion
  income = total - transaction_fee   # Calculations can use other column names
```

web/playground/src/examples.js
```diff
@@ -1,28 +1,31 @@
...
+derive name = f"{c.last_name}, {c.first_name}" # F-strings like Python
+derive db_version = s"version()"     # S-string give escape hatch to SQL
+select {                             # Select certain columns
+  c.customer_id, name, sum_income, ct, db_version,
+}                                    # trailing comma ignored
+sort {-sum_income}                   # "sort" sorts the result; "-" is decreasing order
+take 1..10                           # Limit to a range - could also be "take 10"
+#
+# "output.sql" tab at right shows the SQL generated from this PRQL query
+# "output.arrow" tab shows the result of the query
```
max-sixty: Great!

web/playground/src/examples.js
```diff
@@ -1,28 +1,31 @@
...
-select {              # Select only these columns
-  c.customer_id, name, sum_income
-}
-derive db_version = s"version()" # S-string, escape hatch to SQL
+join c=customers (==customer_id)     # join on "customer_id" from both tables
+derive name = f"{c.last_name}, {c.first_name}" # F-strings like Python
+derive db_version = s"version()"     # S-string give escape hatch to SQL
+select {                             # Select certain columns
+  c.customer_id, name, sum_income, ct, db_version,
+}                                    # trailing comma ignored
```
max-sixty: Should we line break these and have the comment at the end? At the moment the comment is on a different line (no travesty though!)

max-sixty (approved):
This is excellent! 

I think this could be the canonical example query we use everywhere, including the website & readme.

I added a few tiny refinements, lmk your thoughts @richb-hanover . 

Thank you v much!

web/playground/src/examples.js
richb-hanover: Good point. See my next commit.

web/playground/src/examples.js
richb-hanover: Yes, that's implicit.

But I realize what I want to say is that you can comment out any line/statement and retain a valid PRQL query... See my next commit

web/playground/src/examples.js
richb-hanover: see the newest commit...

web/playground/src/examples.js
richb-hanover: I think I like an example of select { ... } all on one line.

richb-hanover:
> I think this could be the canonical example query we use everywhere, including the website & readme.

It certainly could. But it wouldn't hurt my feelings if we came up with a more compelling query that demonstrates the same stuff... This one's kinda dopey/nonsensical - who cares about `(count customer_id)`? For example:

- It'd be nice to show a join on two columns
- It'd be nice to sort on two columns
- Or a `group` on two columns
- Maybe more didactic stuff. 
- All the while keeping it understandable to a newcomer

(Although I am kind of proud of the note that commenting out a line preserves a valid query, then making it possible to comment out the two following `filter` commands. :-)

max-sixty:
> (Although I am kind of proud of the note that commenting out a line preserves a valid query, then making it possible to comment out the two following `filter` commands. :-)

Yes!

---

There's some more stuff we could add — , `=~` for regex, `append`, `window`, `let`. The more things we add, the more likely someone thinks it looks complicated and moves on. Maybe it's OK to get progressively more complicated as we go down the query?

max-sixty:
I would also be up for replacing the Readme query with this (either programmatically or copy&paste); we can see if others agree...

Issue #819, opened 2022-07-17
Syntax to break an expression over multiple lines?


ajfriend:
The PRQL syntax currently usually (with the exception of lists?) treats newlines as terminating characters for statements. (And please do correct me if I'm wrong about this!)

For long expressions (and perhaps also statements?), it would be nice to have a syntax that would allow users to break that expression up over multiple lines.

This idea is motivated by the discussion in https://github.com/prql/prql/issues/788.

Is there a current way to do this? If not, would parentheses be the natural way to express this?

max-sixty:
This is a good point.

I don't think we should need it _that_ often — one of the benefits of PRQL should be to split up logic into modular pieces, represented by transforms on separate lines. 

But it would be presumptuous to assume we can always do this, and #788 is a good case — even if the "split up into multiple filters" suggestion is reasonable, it doesn't cover the case of a long list of `or`-ed conditions.

Newlines have an important semantic meaning, so we've "used up" that syntax. So a couple of options:
- Use `\` to allow line breaks
- Find the cases that force long lines and offer a solution to them (e.g. an `any` function to take a list of conditions and string them together with `or`s)

Any thoughts?

ajfriend:
My comment around parentheses was wondering if you could do something like what Python does to "free up" the newline syntax inside of parentheses:
<img width="835" alt="Screen Shot 2022-07-18 at 10 50 55 AM" src="https://user-images.githubusercontent.com/984039/179572488-e95a3bbc-61c2-4760-9055-cc38590486e9.png">

mklopets:
`\` also seems like a decent solution to me

max-sixty:
Yes, interesting. So we could have a rule that finishing the line on an operator (e.g. `+` / `or` / etc) causes a line continuation.

Are there any instances where we want to finish on an operator?

One concern is that while editing a line in the middle of a query, the error messages might be quite confusing:

```elm
from foo
derive bar = baz or<cursor>
aggregate (
  max fuz
)
```

...would this have a confusing error message saying "`baz or aggregate` is invalid"?

ajfriend:
> Yes, interesting. So we could have a rule that finishing the line on an operator (e.g. `+` / `or` / etc) causes a line continuation.

But that's maybe valid *only* inside a set of parens? (I'm not sure what the full Python parsing logic is.) Additionally, it might also be valid if the next line *starts* with an operator, which is that last Python example (inside parens).

ajfriend:
[...]

PR #2937, opened 2023-07-03
feat: multiline function definition syntax

aljazerzen:


max-sixty:
I would have preferred #819 with a preceding `\` over something specific to functions. Do you think if we do that, then we can remove this? Or you like this per se?

aljazerzen:
This would be hard to do in general, as new lines are parsed slightly differently in different cases.

Maybe it's doable, but I wanted to add back the `func` for the function definition since there are some cases where it is hard to parse an expression if you are not sure if it is a function definition or not. Unfortunately, I don't remember the example.

PR #2938, opened 2023-07-03
fix: unsupported window frames

### LINKED ISSUE #1247: OVER clause should not include frame, when function does not support it

### LINKED ISSUE #2771: ORDER BY incorrectly inserted in WINDOW function expression
### What happened?

The `OVER` clause for column `d` should not contain `ORDER BY a ROWS BETWEEN UNBOUNDED PRECEDING AND UNBOUNDED FOLLOWING`.

### PRQL input

```elm
from_text format:csv "i,a\n1,3\n2,1\n3,2\n4,3\n5,1"
select [i, a, b=lag 1 a]
window (
  sort a
  derive [c = lag 1 a]
  )
window (
  derive [d = lag 1 a]
  )
sort i
```

### SQL output

```SQL
WITH table_0 AS (
  SELECT
    '1' AS i,
    '3' AS a
  UNION
  ALL
  SELECT
    '2' AS i,
    '1' AS a
  UNION
  ALL
  SELECT
    '3' AS i,
    '2' AS a
  UNION
  ALL
  SELECT
    '4' AS i,
    '3' AS a
  UNION
  ALL
  SELECT
    '5' AS i,
    '1' AS a
)
SELECT
  i,
  a,
  LAG(a, 1) OVER () AS b,
  LAG(a, 1) OVER (
    ORDER BY
      a ROWS BETWEEN UNBOUNDED PRECEDING AND UNBOUNDED FOLLOWING
  ) AS c,
  LAG(a, 1) OVER (
    ORDER BY
      a ROWS BETWEEN UNBOUNDED PRECEDING AND UNBOUNDED FOLLOWING
  ) AS d
FROM
  table_0 AS table_1
ORDER BY
  i

-- Generated by PRQL compiler version:0.8.1 (https://prql-lang.org)
```

### Expected SQL output

```SQL
WITH table_0 AS (
  SELECT
    '1' AS i,
    '3' AS a
  UNION
  ALL
  SELECT
    '2' AS i,
    '1' AS a
  UNION
  ALL
  SELECT
    '3' AS i,
    '2' AS a
  UNION
  ALL
  SELECT
    '4' AS i,
    '3' AS a
  UNION
  ALL
  SELECT
    '5' AS i,
    '1' AS a
)
SELECT
  i,
  a,
  LAG(a, 1) OVER () AS b,
  LAG(a, 1) OVER (
    ORDER BY
      a ROWS BETWEEN UNBOUNDED PRECEDING AND UNBOUNDED FOLLOWING
  ) AS c,
  LAG(a, 1) OVER () AS d
FROM
  table_0 AS table_1
ORDER BY
  i

-- Generated by PRQL compiler version:0.8.1 (https://prql-lang.org)
```

### MVCE confirmation

- [X] Minimal example
- [X] New issue

### Anything else?

_No response_

aljazerzen:
Closes https://github.com/PRQL/prql/issues/1247
Closes https://github.com/PRQL/prql/issues/2771

Some SQL window functions like ROW_NUMBER do not support
window frames (ROW & RANGE).
This PR removes the frames from OVER clauses of those functions.


prql-compiler/tests/integration/main.rs
```diff
@@ -318,35 +320,45 @@ fn test_rdbms() {
...
+        assert_snapshot!("results", first_result, &prql);
+
+        // Then check every other result against the first result
+        results.iter().for_each(|(dialect, result)| {
+            assert_eq!(
+                *first_result, *result,
+                "{} == {}: {test_name}",
+                first_dialect, dialect
+            );
+        })
```
max-sixty, line 361: You weren't a fan of my attempt to just snapshot test each result against the same snapshot? :)

max-sixty (commented):
Great improvement, thank you!

prql-compiler/tests/integration/main.rs
aljazerzen: Yeah, it was not reporting which two dialects are different from each other, so the debugging got quite confusing.

Instead of investigating, I reverted the commit.

max-sixty: > Instead of investigating, I reverted the commit.

Support!

PR #2940, opened 2023-07-03
feat: more aggregation functions, enforce ret type

### LINKED ISSUE #985: Fixed points of aggregation functions
From Postgres docs:

> It should be noted that except for count, these functions return a null value when no rows are selected. In particular, sum of no rows returns null, not zero as one might expect, and array_agg returns null rather than an empty array when there are no input rows. The coalesce function can be used to substitute zero or an empty array for null when necessary.

This is in SQL standard:
```
sqlite> SELECT count(a) FROM (SELECT 5 as a WHERE false) t;
+----------+
| count(a) |
+----------+
| 0        |
+----------+
sqlite> SELECT sum(a) FROM (SELECT 5 as a WHERE false) t;
+--------+
| sum(a) |
+--------+
|        |
+--------+
```

I believe that aggregation functions with no input rows should have more reasonable defaults. In some cases this is should be [zero-objects](https://en.wikipedia.org/wiki/Initial_and_terminal_objects) (the object which doesn't have effect if it's reapplied). For `sum` that is 0, for `string_agg` it is an empty string.

So I propose that:
```
from employees
select (sum salary)
```
... compiles to ...
```sql
SELECT COALESCE(SUM(salary), 0) FROM employees
```

aljazerzen:
Closes #985

Add aggregation functions `every`, `any` and `concat_array`.
Remove `avg` in favor of `average`.

Enforce that `sum`, `every`, `any` and `concat_array` return non-null values
since they have a reasonable and expected fixed value.


vanillajonathan:
Note I think what here is called "every" is called "all" in LINQ in .NET.

https://learn.microsoft.com/en-us/dotnet/api/system.linq.enumerable.all?view=net-7.0

vanillajonathan:
Python also has function named `all`.
https://docs.python.org/3/library/functions.html#all

aljazerzen:
Rust also `all`. https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.all

Javascript has `every`. https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/every

Postgres has `every`. https://www.postgresql.org/docs/9.5/functions-aggregate.html

I'm -0.5 on changing `every` to `all`. If we get another +1, we can change. @max-sixty @snth @eitsupi

max-sixty:
~Python (& pandas & numpy) are also `all`, I'd favor `all`. ~

~`all` also seems more consistent with `any` (whereas `every` possibly seems some of the way to `each`?)~

~I'm probably +0.75, but also OK to downweight my vote given my recent contributions! (now back from my very European vacation...)~

I misunderstood this at first — I thought this was the aggregation function over an array (i.e. a column). Instead, this is a function over a tuple, equivalent to `BOOL_AND`.

In that case, what do we think about `and` & `or`?

And then `any` & `all` can be used for arrays. (Possibly when we can infer types these maaaybe could be unified, but not yet).

prql-compiler/src/tests/test.rs
```diff
@@ -853,7 +853,7 @@ fn test_window_functions_02() {
         TO_CHAR(co.order_date, '%Y-%m-%d') AS order_day,
         COUNT(DISTINCT co.order_id) AS num_orders,
         COUNT(*) AS num_books,
-        SUM(ol.price) AS total_price
+        COALESCE(SUM(ol.price), 0) AS total_price
```
max-sixty, line 856: My initial reaction was "is it worth the SQL noise", but actually this does seem quite reasonable.

PR #2946, opened 2023-07-04
fix: annotation parsing

### LINKED ISSUE #2738: Comments break annotations
### What happened?

Currently comments between an annotation and a function breaks the annotation. PRQL's semantics should be transparent to comment additions.

### PRQL input

```elm
@{binding_strength=1}
# comment
let add = x y -> s"x + y"

from tracks
```

### SQL output

```SQL
Error:
   ╭─[Root.prql:5:1]
   │
 5 │ let add = x y -> s"x + y"
   │ ─┬─
   │  ╰─── pipeline expected one of !, (, +, -, ->, ==, [, an identifier, keyword case or {, but found keyword let
───╯
```

### Expected SQL output

```SQL
-- Without the comment:

SELECT
  *
FROM
  tracks
```

### MVCE confirmation

- [X] Minimal example
- [X] New issue

### Anything else?

Currently these are only internal, so not a big priority

aljazerzen:
Closes #2738

### DIFF
--- a/prql-compiler/src/parser/mod.rs
+++ b/prql-compiler/src/parser/mod.rs
@@ -2681,9 +2681,9 @@ join s=salaries (==id)
     #[test]
     fn test_annotation() {
         assert_yaml_snapshot!(parse_single(r#"
-      @{binding_strength=1}
-      let add = a b -> a + b
-      "#).unwrap(), @r###"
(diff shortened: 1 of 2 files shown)


max-sixty:
Thanks!

PR #2961, opened 2023-07-09
build: Bump duckdb-rs to 0.8.2

max-sixty:


max-sixty:
This has been tagged but not released, can reopen when it gets released

PR #2966, opened 2023-07-11
fix: expand set of quoted SQL keywords

### LINKED ISSUE #1966: Expand set of quoted keywords to include sqlite keywords

aljazerzen:
Closes #1966

### DIFF
--- a/prql-compiler/src/sql/gen_expr.rs
+++ b/prql-compiler/src/sql/gen_expr.rs
@@ -2,16 +2,11 @@
 
 use anyhow::{bail, Result};
 use itertools::Itertools;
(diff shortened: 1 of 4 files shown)


max-sixty:
Thanks!

PR #2971, opened 2023-07-14
fix: prqlc compile_help test

not-my-profile:
The test introduced in 0d6481633a0be823ed7880af85e16de067031df1 never
worked because a pre-commit hook had automatically stripped the trailing
whitespace. This wasn't spotted in the CI because cargo-insta 1.30.0 has
a bug that it returns exit code 0 even if there's a test failure.[1]

This commit restores the necessary trailing whitespace and disables the
trailing-whitespace pre-commit hook (we already run `cargo fmt` anyway,
which is intelligent enough to preserve trailing whitespace in strings).

[1]: https://github.com/mitsuhiko/insta/issues/392

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -2,7 +2,6 @@ repos:
   - repo: https://github.com/pre-commit/pre-commit-hooks
     rev: v4.4.0
     hooks:
-      - id: trailing-whitespace
       - id: end-of-file-fixer
         exclude: '(.*\.snap|.*render-link.html|head.hbs)'
       - id: check-yaml
--- a/prql-compiler/prqlc/tests/test.rs
+++ b/prql-compiler/prqlc/tests/test.rs
@@ -112,13 +112,13 @@ fn compile_help() {
 
     -t, --target <TARGET>
             Target to compile to
-
+            
             [env: PRQLC_TARGET=]
             [default: sql.any]
 
         --color <WHEN>
             Controls when to use color
-
+            
             [default: auto]
             [possible values: auto, always, never]
 


aljazerzen:
Great debugging and explanation of what's happening! I've noticed that the test is failing locally, but didn't bother to investigate.

I'm a bit hesitant to disable the precommit rule for removing whitespace, as it also acts on non-rust files. On the other hand we do also have prettier, which should handle a big majority of those files. And removing whitespace in the Rust string changes semantics of the code, which should never happen with a formatter.

not-my-profile:
Thanks for the quick merge. Just a small heads-up that I prefer my commit messages to be preserved on merge. Especially in this cases like this one where the context isn't obvious.

.pre-commit-config.yaml
```diff
@@ -2,7 +2,6 @@ repos:
   - repo: https://github.com/pre-commit/pre-commit-hooks
     rev: v4.4.0
     hooks:
-      - id: trailing-whitespace
```
max-sixty, line 5: This did cover files other than rust files, so rustfmt isn't a perfect replacement.

But I agree the case is a tough one; not sure there's a good solution. Let's see if things creep in...

not-my-profile: I am pretty sure `prettier` gets rid of trailing whitespace in most other file formats.

eitsupi: Can we exclude only Rust files with `exclude`?
https://pre-commit.com/#adding-pre-commit-plugins-to-your-project

max-sixty: Good idea @eitsupi ; done!

PR #2972, opened 2023-07-14
refactor: move website examples into data files

not-my-profile:
The examples showcased on the website were previously all defined in the the YAML front matter of _index.md. This commit moves them to separate data files[1], which are easier to edit and test.

[1]: https://gohugo.io/templates/data-templates/

### DIFF
--- a/web/book/tests/documentation/website.rs
+++ b/web/book/tests/documentation/website.rs
@@ -1,3 +1,5 @@
+use std::fs::read_dir;
+
 use regex::Regex;
 use serde_yaml::Value;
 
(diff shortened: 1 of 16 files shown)


aljazerzen:
Thank you for the contribution!

Do you think that it is worth splitting the new .yaml files into separate .prql and .sql files, so we get syntax highlighting when editing them? We do something similar with integration tests and book tests.

not-my-profile:
Currently the example metadata also includes a `label`, so I think it makes sense to have everything together (i.e. label, prql and sql) in one file.

aljazerzen:
Could we put the label into `_index.md`?

not-my-profile:
Yeah we could but then the data for one example would be split up across three different files, which I think would make editing examples more cumbersome.

aljazerzen:
I'm -0.5 on merging this, as the examples equality accessible as before as since they are still in YAML, but spread over multiple file.

@max-sixty I'll defer this one to you.

not-my-profile:
I want to note that the refactor did enable the test code to be simplified since it no longer has to extract the tests from `_index.md` or special case the hero example.

max-sixty:
Oh I'm +1 on this, test code is much simpler.

I would be even keener if they were `.prql` files we could snapshot; then we get the benefits of editing PRQL files directly, and zero test code. But that requires running the compilation as part of building the website (unless we instead extract from the snapshot files https://github.com/mitsuhiko/insta/issues/353)

PR #2973, opened 2023-07-14
refactor: move website testimonials into data file

not-my-profile:
Having the testimonials in a dedicated file
enables a dedicated git history for them.


aljazerzen:
We already get git history with them in the `_index.md` and they are not likely to change. But this is a bit more organized than a single big file, so let's merge.

PR #2726, opened 2023-06-04
docs: fix the link to example tests

eitsupi:
Follow up #2719

Fix the broken link https://github.com/PRQL/prql/actions/runs/5167542767/jobs/9308485299#step:7:1466

PR #2975, opened 2023-07-14
fix(website): download profile images

not-my-profile:
The website embeds images from Twitter for the testimonials. The Firefox tracking protection on strict mode apparently blocks these images.[1] This commit adapts the Hugo template to download the images on site build as a workaround.

[1]: https://bugzilla.mozilla.org/show_bug.cgi?id=1458915


aljazerzen:
Nice work!

not-my-profile:
CC @max-sixty do you have to manually trigger the website build? Because prql-lang.org hasn't updated for me.

max-sixty:
> CC @max-sixty do you have to manually trigger the website build? Because prql-lang.org hasn't updated for me.

I added the label to backport it. We currently have the website on a branch off the previous release (though much discussion on what the best approach is!)

not-my-profile:
Ah, I see thanks :) It seems like you'll also need to backport #2726, since that broken link is currently preventing the deploy job from being started.

max-sixty:
Ah, good point. 

I _think_ that's been removed in `main` , so a broken link won't stop a deployment.

I'll backport that. If that doesn't work, then the release is only days away...

aljazerzen:
@max-sixty FYI: I think this is a classic example of a PR that should not be backported. It's a new "feature" in the terms of expanded functionality of the website and not a critical bug/bad wording.

Minimization of things to backport would ease the toll of maintaining the web branch...

---

A bit unrelated: the broken link checker is a PITA, because we have a lot of links pointing to http://prql-lang.org so when you move the page, it will check the currently-deployed site, not the site that is about-to-be-deployed.

So it's quite helpful that this action is not a requirement for merging.

max-sixty:
> @max-sixty FYI: I think this is a classic example of a PR that should not be backported. It's a new "feature" in the terms of expanded functionality of the website and not a critical bug/bad wording.

OK. Agree it's not a bug. Deploying sooner does let us iterate quicker, if we need to see the result. But no strong view on my end.

---

> So it's quite helpful that this action is not a requirement for merging.

Agree!

PR #2977, opened 2023-07-14
refactor: impl Display for Reason

not-my-profile:


aljazerzen:
Thanks, this is an improvement.

PR #2980, opened 2023-07-14
refactor: change range in rq::Take to Range<i64>

not-my-profile:


aljazerzen:
Woah, nice work! You really cleaned-up a few places.

Changing `Range<Expr>` into `Range<i64>` will prevent support for other things in Take, such as parameters. As you have noticed in #2981, it would make sense to support that. Essentially, `rq::Take` should be able to contain this `rq::ExprKind` variant:

https://github.com/PRQL/prql/blob/main/prql-compiler/src/ast/rq/expr.rs#L35-L36

not-my-profile:
No it won't prevent supporting parameters. The code currently assumes that the range contains integers ... the type we use should reflect the current assumption ... because otherwise we just have needless type matching everywhere.

Supporting parameters in take will probably entail introducing something like:

```rs
enum TakeRange {
    Known(Range<i64>),
    Unknown(Range<i64OrParameter>),
}
```

(and should be done in a followup PR)

aljazerzen:
Oh, yes we do assume that the range contains literals only:

https://github.com/prql/prql/blob/main/prql-compiler/src/sql/gen_expr.rs#L498-L500

That's because we do some clever range combining: 

https://github.com/prql/prql/blob/main/prql-compiler/src/sql/gen_expr.rs#L469-L495

If we wanted to support params in take, this would have to be updated too. Maybe emitting SQL that does this computation, or just splitting query into multiple chained SELECTs.

----

I do agree that the code should strive toward having the tightest possible types, since this minimizes memory consumption and provides cleaner code. But in this case, I do think we would be doing something that would have to be reverted in the future.

Imagine that we wanted to support binary expressions too:

```
from x
take $param..(3 + 2)
```

I'd say that the best option for Take under these requirements would be to contain `Range<Expr>`.

Also, `Expr` contains `Span` too, which is needed to produce nice error messages.

not-my-profile:
Ah yeah, that makes sense. Closing this for now ... will probably open a similar PR in the future that also accounts for the other take constructs we want to support.

PR #2982, opened 2023-07-14
fix(book): fix illogical sentence

not-my-profile:
If there is **no** such function in SQL an s-string won't help you.


aljazerzen:
Hmmm, that's really strange wording. I wonder how it evolved into this. In any case, thanks.

PR #2987, opened 2023-07-14
fix: panic on empty string intepolations

### LINKED ISSUE #2985: `{}` in f-strings leads to compiler panic
E.g.

```prql
from invoices
derive [
   foo = f"{}"
]
```

panics with:
```
thread 'main' panicked at 'called `Option::unwrap()` on a `None` value', .../prql-compiler/src/ast/pl/ident.rs:23:31
```

aljazerzen:
Closes #2985

### DIFF
--- a/prql-compiler/src/parser/expr.rs
+++ b/prql-compiler/src/parser/expr.rs
@@ -97,7 +97,8 @@ pub fn expr() -> impl Parser<Token, Expr, Error = PError> + Clone {
                     finish(vec![])
                 }
             }
-        });
+        })
+        .labelled("interpolated string");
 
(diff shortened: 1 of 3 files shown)


not-my-profile:
Should this fix get a changelog entry?

aljazerzen:
yes. I've forgot to add most of the fixes since 8.1, so this should be done before releasing.

max-sixty:
Our Changelog is "best effort" — I do a pass before releasing of any really important changes, but many don't make it in. 

I'd be up for having more things on our changelog, but probably not if it reduces our development velocity...

PR #2988, opened 2023-07-16
docs: work around mdbook bug & rust problem

richb-hanover:
**Update:** _I have opened a new issue (#2991) to discuss the initial reason for creating this issue. Let's use this issue to talk about solving the `mdbook` bug for PRQL documentation and the rust problem that causes this PR to fail._

See #2991 for the remainder of the original question.


web/book/src/queries/pipelines.md
```diff
@@ -43,5 +43,5 @@ select {e.name, gross_salary, d.name}
 
 ## See also
 
-- [Transforms](../transforms/README.md) - PRQL Transforms
-- [Syntax](../syntax/README.md) - Notation for PRQL queries
+- [Transforms](../transforms/index.html) - PRQL Transforms
+- [Syntax](../syntax/index.html) - Notation for PRQL queries
```
not-my-profile: Good catch! You may want to add a comment:

```
<!-- we're linking the README.md files as index.html to work around https://github.com/rust-lang/mdBook/issues/984 -->
```

richb-hanover: Yes, but...

The checkin actions run `markdown-link-check` which grumbles about this - there are no files named _index.html_ in the source tree.

I was about to write this up as another issue, but I'll say it here: It's intriguing that this only bites us when there's a directory in the path. Normal links to _./foo.md_ work fine.

I suspect it's something to do with `mdbook` converting a _README.md_ into _index.html_ for unclear reasons (well, I know why this could be useful - every directory since 1993 has needed an _index.html_ file - but that isn't necessarily helpful here...) I'm hopeful that we can turn this off.

richb-hanover:
I updated _book.toml_ to turn off the README.md -> index.html conversion in `mdbook`. `test-rust` is failing now. I don't know enough to address this. 

1) I would still appreciate a review of the "Variables & Let" page in the Book
2) How can we fix the rust tests?

Thanks.

web/book/src/queries/pipelines.md
not-my-profile: Yeah the `markdown-link-check` doesn't know about the mdbook bug but it can be disabled for these links via a comment.

> I suspect it's something to do with mdbook converting a README.md into index.html for unclear reasons 

Yes it's a bug in mdbook ... see the GitHub issue I linked in the comment.

I have opened #2990 to fix these two links, so you can drop these changes from this PR and focus on the content changes :)

web/book/src/language-features/README.md
```diff
@@ -3,16 +3,20 @@
...
+| Coalesce                                              | [Handle potentially NULL values](./coalesce.md)                         |
+| Dates & times                                         | [Handle dates and times](./dates-and-times.md)                          |
+| Distinct                                              | [Equivalent of SQL `DISTINCT`](./distinct.md)                           |
+| Relation Literals                                     | [Define a table](./relation-literals.md)                                |
+| Null handling                                         | [Handle `NULL` values](./null.md)                                       |
+| Ranges                                                | [Syntax for all forms of ranges](./ranges.md)                           |
+| Regex                                                 | [Handle regular expressions](./regex.md)                                |
+| <!-- markdown-link-check-disable --> Standard Library | [Common functions in SQL](./standard-library/index.html)                |
+| <!-- markdown-link-check-enable--> Strings            | [Rules for creating strings](./strings.md)                              |
```
not-my-profile: Having these comments in the table is rather ugly. I think it should be possible to put the link definition after the table i.e. having just `[Common functions in SQL]` in the table and then after the table declaring:

```
[Common functions in SQL]: ./standard-library/index.html
```

web/book/src/language-features/README.md
```diff
@@ -3,16 +3,20 @@
...
-| Distinct         | [Equivalent of SQL `DISTINCT`](./distinct.md)                           |
-| Null handling    | [Handle `NULL` values](./null.md)                                       |
-| Ranges           | [Syntax for all forms of ranges](./ranges.md)                           |
-| Regex            | [Handle regular expressions](./regex.md)                                |
-| Strings          | [Rules for creating strings](./strings.md)                              |
-| S-strings        | [Insert SQL directly into a query with an S-string](./s-strings.md)     |
-| F-strings        | [Combine several column's data with F-strings](./f-strings.md)          |
-| Case             | [Create a new column based on the contents of other columns](./case.md) |
-| Target & Version | [Specify a target SQL engine and PRQL version](./target.md)             |
+<!-- we're linking to README.md files as index.html to work around https://github.com/rust-lang/mdBook/issues/984  by using "markdown_link_check_disable"-->
```
not-my-profile: I'd remove the `by using "markdown_link_check_disable` from that comment since it's not actually part of the workaround but just to make the link checker happy.

web/book/src/language-features/README.md
richb-hanover: I am not fond of the comments either, but I'm not sure how to eliminate them without removing the link-check for the entire table. 

Or am I missing something? I would entertain a different suggestion. Thanks.

web/book/src/language-features/README.md
richb-hanover: I think the comment should be rewritten to say:

```
<!-- 
We use "markdown_link_check_disable" to work around https://github.com/rust-lang/mdBook/issues/984  
so we can link to "index.html" files without complaints from `markdown-link-check`
-->

File: web/book/src/language-features/README.md (hunk 2 above)
Comment (not-my-profile) [reply web/book/src/language-features/README.md]: Yes ... you're missing what I said. Markdown links can be broken up like:

```
[test]

[test]: foo.md
```

File: web/book/src/language-features/README.md (hunk 3 above)
Comment (not-my-profile) [reply web/book/src/language-features/README.md]: that's still wrong ... markdown_link_check_disable does not work around the mdbook bug ... the workaround is the change of the extension in the link

File: web/book/src/language-features/README.md (hunk 2 above)
Comment (richb-hanover) [reply web/book/src/language-features/README.md]: Hah! You're right. Thanks for reminding me that Markdown has this facility.

File: web/book/src/language-features/README.md (hunk 3 above)
Comment (richb-hanover) [reply web/book/src/language-features/README.md]: Right again! (Actually, making the link checker happy is _another_ workaround...) How about...
```
<!-- 
To work around https://github.com/rust-lang/mdBook/issues/984 we link to "index.html" instead of "README.md"
files. To avoid complaints from `markdown-link-check` in the CI, we use "markdown-link-check-disable" 
-->
```

File: web/book/src/language-features/README.md [hunk 4]
```diff
@@ -3,16 +3,27 @@
...
-| Regex            | [Handle regular expressions](./regex.md)                                |
-| Strings          | [Rules for creating strings](./strings.md)                              |
-| S-strings        | [Insert SQL directly into a query with an S-string](./s-strings.md)     |
-| F-strings        | [Combine several column's data with F-strings](./f-strings.md)          |
-| Case             | [Create a new column based on the contents of other columns](./case.md) |
-| Target & Version | [Specify a target SQL engine and PRQL version](./target.md)             |
+<!--
+To work around https://github.com/rust-lang/mdBook/issues/984 we link to "index.html" instead of "README.md".
+To avoid complaints from `markdown-link-check` in the CI, we use "markdown-link-check-disable"
+-->
```
not-my-profile, line 9: I think the comment should be moved down next to the link-check-disable comment. And I don't think elaborating on the markdown-link-check-disable comment is necessary.

web/book/src/queries/variables.md
```diff
@@ -34,7 +60,12 @@ let grouping = s"""
 from grouping
 ```
 
-## `into`
+```admonish info
+**Why introduce `into`?** It seems that `let = ...` completely covers the use case without introducing a new language feature.
+Does `into` work with scalars and functions?
+```
```
aljazerzen, line 66: Because it fits nicely with the concept of a pipeline:

```elm
from x
derive
filter
into y

from y
filter
join y
```

---

Yes, `into` works with everything that `let` does.

web/book/src/queries/variables.md
```diff
@@ -49,15 +80,20 @@ from top_50      # Starts a new pipeline
 ```
 
 ```admonish info
+**Is this advice still relevant?**
 In PRQL variables are far less common than CTEs are in SQL, since a linear series
 of CTEs can be represented with a single pipeline.
 ```
```
aljazerzen, line 86: No, this can all go.

web/book/src/queries/variables.md
```diff
@@ -49,15 +80,20 @@ from top_50      # Starts a new pipeline
...
 
+```admonish info
+**NO LONGER TRUE?**
 Currently defining variables is restricted to relations. We'd like to extend
 this to expressions that evaluate to scalars.
+```
```
aljazerzen, line 92: Yup, no longer true, can be removed.

web/book/src/queries/variables.md
```diff
@@ -71,4 +107,4 @@ table recursive_example = (s"""
 
 from recursive_example
 
--->
+```
```
aljazerzen, line 110: Now this could be implemented using `loop`, but this is not the place for it.

This example can be removed.

web/book/src/queries/variables.md
```diff
@@ -1,11 +1,37 @@
-# Variables
+# Variables and `let`
 
-We can define a relation — similar to a CTE in SQL — with two approaches — a
-prefix `let` or a postfix `into.
+`let` assigns a scalar value, a function, an array, or a pipeline to a named
+variable.
```
aljazerzen, line 4: `let` can be also assigned a tuple, function call or a range.

I think we can just say "any expression, including functions".

aljazerzen (commented):
Nice cleanup. This could be merged individually, or I can cherry-pick it into the "docs-restructure" branch, where I'm working on splitting the documentation into "reference" and "tutorial".

richb-hanover:
> Nice cleanup. This could be merged individually, or I can cherry-pick it into the "docs-restructure" branch, where I'm working on splitting the documentation into "reference" and "tutorial".

Thanks. I am completely content to have you cherry pick the best parts into your new docs.  I'll close this.

PS I opened #2991 to bring out more of my questions. Thanks again

PR #2990, opened 2023-07-16
fix(book): fix two broken links

not-my-profile:


max-sixty:
Great spot, thanks @not-my-profile

not-my-profile:
Credit goes to @richb-hanover for spotting it :)

PR #2992, opened 2023-07-17
refactor!: clean up the public API

not-my-profile:
These are some refactor commits to clean up the public API ... they are breaking changes.

This PR contains 8 commits, you'll want to review the changes commit by commit.
For the reasoning about the `Expr::from` to `Expr::new` change please refer to the last commit message.

I want these commits to be merged *as they are*, that is I **do not** want this PR to be squashed. Let me know if that's ok with you, then I can mark the PR as ready.

### DIFF
--- a/prql-compiler/src/ast/pl/expr.rs
+++ b/prql-compiler/src/ast/pl/expr.rs
@@ -1,6 +1,6 @@
 use std::collections::HashMap;
 
-use anyhow::{anyhow, Result};
+use anyhow::Result;
 use enum_as_inner::EnumAsInner;
 
 use serde::{Deserialize, Serialize};
@@ -497,13 +497,5 @@ impl<T> Default for WindowFrame<T> {
             kind: WindowKind::Rows,
             range: Range::unbounded(),
         }
-    }
-}
-
-impl From<TransformKind> for anyhow::Error {
-    // https://github.com/bluejekyll/enum-as-inner/issues/84
-    #[allow(unreachable_code)]
-    fn from(kind: TransformKind) -> Self {
-        anyhow!("Failed to convert `{kind:?}`")
     }
 }
--- a/prql-compiler/src/ast/pl/expr.rs
+++ b/prql-compiler/src/ast/pl/expr.rs
@@ -500,14 +500,6 @@ impl<T> Default for WindowFrame<T> {
     }
 }
 
-impl From<ExprKind> for anyhow::Error {
-    // https://github.com/bluejekyll/enum-as-inner/issues/84
-    #[allow(unreachable_code)]
-    fn from(kind: ExprKind) -> Self {
-        anyhow!("Failed to convert `{}`", Expr::from(kind))
-    }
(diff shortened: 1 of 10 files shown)


prql-compiler/src/ast/pl/expr.rs
```diff
@@ -446,15 +446,15 @@ impl Expr {
...
             })
             .with_span(self.span)
         })
     }
 }
 
-impl From<ExprKind> for Expr {
-    fn from(kind: ExprKind) -> Self {
+impl Expr {
+    pub fn new(kind: ExprKind) -> Self {
```
aljazerzen, line 457: Makes sense. 

We used to use `.into()` and it does make sense because `kind` is the most important field and easily adding all other default fields was on-par with the idea of the from trait. But it was not obvious what `.into()` was doing so I started using `Expr::from` a lot. This PR is essentially saying "don't use `into` at all".

prql-compiler/src/semantic/resolver.rs
```diff
@@ -1015,7 +1015,11 @@ pub(super) mod test {
             .kind
             .into_transform_call()
             .unwrap_or_else(|e| panic!("Failed to convert `{}`", Expr::from(e)));
-        let exprs = derive.kind.into_derive()?;
+        let exprs = derive
+            .kind
+            .into_derive()
+            .unwrap_or_else(|e| panic!("Failed to convert `{e:?}`"));
```
aljazerzen, line 1021: This is a test, we can just `.unwrap()`

prql-compiler/src/semantic/resolver.rs
```diff
@@ -1011,7 +1011,10 @@ pub(super) mod test {
 
     fn resolve_derive(query: &str) -> Result<Vec<Expr>> {
         let expr = parse_and_resolve(query)?;
-        let derive = expr.kind.into_transform_call()?;
+        let derive = expr
+            .kind
+            .into_transform_call()
+            .unwrap_or_else(|e| panic!("Failed to convert `{}`", Expr::from(e)));
```
aljazerzen: This is a test, we can just `.unwrap()`.

aljazerzen:
Nice cleanup. We can enable "rebase and merge" for this commit, so it would not get squashed.

prql-compiler/src/semantic/resolver.rs
not-my-profile: Right ... I think the intention is to have a nice error message if the `unwrap` fails.

not-my-profile:
Thanks :) I'd like you to merge all of my recent refactor PRs via rebasing.

max-sixty:
> Thanks :) I'd like you to merge all of my recent refactor PRs via rebasing.

Just saw this, ignore my question on the other PR!

max-sixty:
> Thanks :) I'd like you to merge all of my recent refactor PRs via rebasing.

Just saw this, ignore my question on the other PR!

PR #2993, opened 2023-07-17
refactor: move parser tests to parser and `Span` to ast

not-my-profile:
This is the followup PR for #2992. The first 8 commits are from that other PR. This PR will be rebased once the other PR has been merged so that only the last ~6~ 2 commits remain. (this is a workaround since GitHub unfortunately doesn't support stacked PRs)

### DIFF
--- a/prql-compiler/src/ast/mod.rs
+++ b/prql-compiler/src/ast/mod.rs
@@ -2,3 +2,6 @@
 //!
 pub mod pl;
 pub mod rq;
+mod span;
+
+pub use span::Span;
--- a/prql-compiler/src/ast/span.rs
+++ b/prql-compiler/src/ast/span.rs
@@ -1,0 +1,90 @@
+use serde::de::Visitor;
+use serde::{Deserialize, Serialize};
+use std::fmt::{self, Debug, Formatter};
+use std::ops::Range;
+
+#[derive(Clone, PartialEq, Eq, Copy)]
+pub struct Span {
+    pub start: usize,
+    pub end: usize,
+
+    pub source_id: u16,
+}
+
+impl From<Span> for Range<usize> {
+    fn from(a: Span) -> Self {
+        a.start..a.end
+    }
+}
+
+impl Debug for Span {
+    fn fmt(&self, f: &mut Formatter<'_>) -> fmt::Result {
+        write!(f, "{}:{}-{}", self.source_id, self.start, self.end)
+    }
+}
+
+impl Serialize for Span {
+    fn serialize<S>(&self, serializer: S) -> std::result::Result<S::Ok, S::Error>
+    where
+        S: serde::Serializer,
+    {
+        let str = format!("{self:?}");
+        serializer.serialize_str(&str)
+    }
+}
+
+impl<'de> Deserialize<'de> for Span {
+    fn deserialize<D>(deserializer: D) -> std::result::Result<Self, D::Error>
+    where
+        D: serde::Deserializer<'de>,
+    {
+        struct SpanVisitor {}
+
+        impl<'de> Visitor<'de> for SpanVisitor {
+            type Value = Span;
+
+            fn expecting(&self, f: &mut fmt::Formatter) -> fmt::Result {
+                write!(f, "A span string of form `file_id:x-y`")
+            }
+
+            fn visit_str<E>(self, v: &str) -> std::result::Result<Self::Value, E>
+            where
+                E: serde::de::Error,
+            {
+                use serde::de;
+
+                if let Some((file_id, char_span)) = v.split_once(':') {
+                    let file_id = file_id
+                        .parse::<u16>()
+                        .map_err(|e| de::Error::custom(e.to_string()))?;
+
+                    if let Some((start, end)) = char_span.split_once('-') {
+                        let start = start
+                            .parse::<usize>()
+                            .map_err(|e| de::Error::custom(e.to_string()))?;
+                        let end = end
+                            .parse::<usize>()
+                            .map_err(|e| de::Error::custom(e.to_string()))?;
+
+                        return Ok(Span {
+                            start,
+                            end,
+                            source_id: file_id,
+                        });
+                    }
+                }
+
+                Err(de::Error::custom("malformed span"))
+            }
+
+            fn visit_string<E>(self, v: String) -> std::result::Result<Self::Value, E>
+            where
+                E: serde::de::Error,
+            {
+                self.visit_str(&v)
+            }
+        }
+
+        deserializer.deserialize_string(SpanVisitor {})
+    }
+}
--- a/prql-compiler/src/error.rs
+++ b/prql-compiler/src/error.rs
@@ -2,8 +2,7 @@ use anstream::adapter::strip_str;
 pub use anyhow::Result;
 
 use ariadne::{Cache, Config, Label, Report, ReportKind, Source};
-use serde::de::Visitor;
-use serde::{Deserialize, Serialize};
+use serde::Serialize;
 
 use std::error::Error as StdError;
 use std::fmt::{self, Debug, Display, Formatter};
@@ -13,13 +12,7 @@ use std::{collections::HashMap, io::stderr};
 
 use crate::SourceTree;
 
-#[derive(Clone, PartialEq, Eq, Copy)]
-pub struct Span {
-    pub start: usize,
-    pub end: usize,
-
-    pub source_id: u16,
-}
+pub use crate::ast::Span;
 
 #[derive(Debug, Clone)]
 pub struct Error {
@@ -399,12 +392,6 @@ impl Display for Reason {
     }
 }
 
-impl From<Span> for Range<usize> {
-    fn from(a: Span) -> Self {
-        a.start..a.end
-    }
-}
-
 impl Add<usize> for Span {
     type Output = Span;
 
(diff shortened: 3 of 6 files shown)


prql-compiler/src/span.rs
```diff
@@ -0,0 +1,128 @@
...
+    }
+
+    fn start(&self) -> Self::Offset {
+        self.start
+    }
+
+    fn end(&self) -> Self::Offset {
+        self.end
+    }
+}
```
aljazerzen: Let's keep this in the parser module, as it is very parser specific. If we didn't want to be implementing traits so far from the type declaration, we could create a newtype for the parser span.

prql-compiler/src/ast/pl/ident.rs
```diff
@@ -153,6 +155,21 @@ pub fn display_ident_part(f: &mut std::fmt::Formatter, s: &str) -> Result<(), st
...
 
+pub static VALID_IDENT: Lazy<Regex> = Lazy::new(|| {
+    // One of:
+    // - `*`
+    // - An ident starting with `a-z_\$` and containing other characters `a-z0-9_\$`
+    //
+    // We could replace this with pomsky (regex<>pomsky : sql<>prql)
+    // ^ ('*' | [ascii_lower '_$'] [ascii_lower ascii_digit '_$']* ) $
+    Regex::new(r"^((\*)|(^[a-z_\$][a-z0-9_\$]*))$").unwrap()
+});
```
aljazerzen: I'm confused. Are we using the same regex for PRQL idents and SQL idents? It's quite unlikely that the two have the same lexing rules.

aljazerzen: This is not a concern of this PR, but just something we should investigate.

aljazerzen: Although this PR is essentially saying this regex is dealing with the PRQL ident.

aljazerzen (approved):
Looks good to me. The changes make sense and are nicely separated, which makes them easy to review.

And I feel the pain of stacked PRs on GitHub. How is this not a thing?

prql-compiler/src/span.rs
not-my-profile: Very good point! I just moved the whole new `span` module into the `parser` module :)

prql-compiler/src/ast/pl/ident.rs
not-my-profile: Yes ... to be frank I was also a bit confused by this. I agree that we should investigate it.

The main reason I moved this regex into the `ident` module is that the `Display` impl of `Ident` indirectly depends on this regex, and I want to move `Ident` to a new `prql_ast` crate in #2994. Currently much of the codebase depends on this Display impl ... I'm planning on removing the dependencies on the Display impl in the future, which will let us move this regex elsewhere.

aljazerzen: Ah yes.

Think about this, I see that this essentially makes codegen a dependency of the AST, which is a bit strange. We could remove the Display impl to sever this dependency and require explicit call to some function for codegen.

not-my-profile: Yes I agree that codegen obviously shouldn't be a dependency of the AST ... however this is only temporarily until I fix that ... because I also believe that errors shouldn't be stringly typed ... the `Reason` enum currently just contains a bunch of strings, which I want to change so that it contains the actual values ... then the formatting happens via a `Display` impl on the Reason type rather than everywhere where the errors are constructed ... and because of this I'll be touching all the places where errors are constructed anyway, and I'd rather only have to touch them once (removing the format string and specifying a proper enum variant) rather than having to touch them twice (firstly using some explicit function call to avoid the Display impl for Expr only to then remove it in the followup change). Does that make sense to you?

tl;dr: Yeah but I want to fix it in a followup PR^^

not-my-profile:
Ready to be merged via rebase :)

prql-compiler/src/span.rs
not-my-profile: I have now actually moved the `span` module back into the root of the crate where I had it originally and just kept the `chumsky::Span` impl in the parser module as you suggested. The reasoning being that in the followup PR I'll make the parser generic over the used Span type so it wouldn't make sense to have the concrete type definition residing in the parser module.

not-my-profile: Nevermind ... I looked into it and making the parser generic over `chumsky::Span` is actually quite cumbersome since a bunch of code in the parser currently does some arithmetic on the `usize`s. So in the followup PR I'll just restrict the associated type of the trait to the `Span` from the parser module.

prql-compiler/src/ast/pl/ident.rs
aljazerzen: I see, this makes sense.

---

Before you start working on a major refactor of the errors, I'd want us to share the vision of what they are supposed to do.

prql-compiler and PRQL-the-language are work-in-progress, which means that requirements, architecture and some code snippets change quite quickly. Because of that, I would really want that it is easy to construct new errors. By easy, I mean that it is not much harder than just panicking.

The string-contents of the `Error` are quite a nice fit for that, and a text message is exactly what's needed to display the error in an IDE or terminal. Obviously, this is very different if you wanted to integrate the compiler and display the errors with some alternative formatting/UI. But currently, this is not the priority of the development.

I don't want to discourage you - I'm just making sure that my work will not more tedious overnight due to premature optimization. If you think that the refactor can keep the errors easy to construct, you have my blessing to go ahead.

---

PS: We have something that aids "programmatic consumption of errors": error codes, which are used by the vscode extension to match instead of relying on the message. This way, code is part of the "stable public API" and the message is not (i.e. we reserve the freedom to change it anytime).

not-my-profile: Thanks, ... yes of course that makes sense ... after these PRs are resolved I'll open a draft PR that just converts a few error variants to showcase what I am thinking of :)

prql-compiler/src/span.rs
aljazerzen: > make the parser generic over the used Span

Wait, why? Can we not make an assumption that we are parsing text files, where span is a range of chars?

> So in the followup PR I'll just restrict the associated type of the trait to the Span from the parser module.

What do you mean here?

not-my-profile: It's not really important. It just would be nice if you could use e.g. `()` as the span type if you don't care about spans at all but yeah that's more of a micro optimization ... not important.

>What do you mean here?

If you look at the parser files in #2994 now you'll see a bunch of `<T: Extension<Span = Span>>` for every method, which says if you want to use the parser the Span has to be the span from the parser crate. But yeah I think that's actually fine ... the important improvement is that the Span stays with the parser and is no longer moved to the new prql_ast crate as I did initially (which also means there's no awkward optional dependency from prql_ast to chumsky).

aljazerzen: Ok, yes, `<T: Extension<Span = Span>>` just means that there is a dependency from our parser to chumsky. Which is completely ok.

not-my-profile: I mean it of course depends on chumsky since it's implemented with chumsky ... a chumsky parser could in theory be implemented to work with an arbitrary span type that implements `chumsk::Span` ... `<T: Extension<Span = Span>>` constraints the span type supported by our parser to our own span type ... but yeah that constraint already exists anyway.

aljazerzen: Wait, so why is span now under `parser` module? That should be under AST, as the AST uses the span?

not-my-profile: The prql_ast crate introduced by the followup PR is generic over the span. We'll want to split off the `parser` module from `prql-compiler` in the future into a separate crate at which point `prql-compiler` will depend on `prql-parser` but the `Span` type will be part of `prql-parser` since you cannot use the parser without it.

So the answer is because I am already planning on the `parser` module becoming its own crate.

aljazerzen: Let's keep the span under root module or ast module for now, so this can be merged.

not-my-profile: I can do that but I am curious on why you prefer that. I think it's clear that it has the tightest dependency with the parser and extracting the parser into its own crate will require the Span to be moved along.

prql-compiler/src/ast/pl/ident.rs
not-my-profile: I have now dropped these regex related changes from this PR since if we don't go with the generic AST approach then we don't have the probem with the Display impl of Expr in the ast crate requiring the regex.

prql-compiler/src/span.rs
aljazerzen: I see the span as a part of AST, and the parser is just something that produces the span. This is not the only interpretation, but to me it makes the most sense.

aljazerzen: When we extract the parser it would depend on ast crate the same way as it depends on ast crate for other AST nodes.

not-my-profile: Ok yeah that makes sense.

not-my-profile:
Haha ok this should now finally be ready for merging. I split off the `RelationLiteral` move to #2998 so that only two commits remain in this PR (so still please merge via rebase).

The second commit now only moves the `Span` struct, the Debug impl and the serde impls ... it intentionally doesn't move the Add<usize> and chumsky::Span impls since they are only used by the parser. (I have also explained this in the commit message :)

max-sixty:
> Ready to be merged via rebase :)

Would you prefer rebase vs. squash? Generally we've used squash, but we can enable rebase merging if you prefer.

(We're also generally quick to do reviews, and happy for folks to self-merge if they're confident in small changes; which mitigates the disadvantage of squashing, for the future)

not-my-profile:
I think whether to rebase or to squash depends on the work style of the contributor.

If a PR contains a dozen messy commits, it's probably best to squash them ... however I in general like to take care of splitting up my commits and adding respective commit messages, and I of course want this work I put in to be preserved on merge.

not-my-profile:
Closing this as these two commits are now part of #3000 ... (which contains 27 commits) ... it doesn't make much sense to split off these 2 commits in particular into their own PR.

max-sixty:
> Closing this as these two commits are now part of #3000 ... (which contains 27 commits) ... it doesn't make much sense to split off these 2 commits in particular into their own PR.

Sorry I hadn't merged before, I'm not sure why I hadn't; possibly me & Aljaz were waiting for each other.

FWIW I do find it much easier to work in smaller PRs — there is almost no PR that's too small. (Ofc that relies on someone reviewing quickly...)

Issue #2857, opened 2023-06-18
Stack overflow on Windows


eitsupi:
See https://github.com/PRQL/prql/pull/2700#issuecomment-1575206792 and #2713 and https://github.com/PRQL/prql/issues/2781#issuecomment-1595272822

----

I tried a few simple examples with the latest prqlc.exe built from main (8f9b77802da49971b94ebeb4baee7ccc725675bf), and I suspect that this has nothing to do with `count` and everything to do with the number of `filter` and `take` and `sort`.

## Working

```elm
from employees
filter gross_cost > 0
group {title} (
  aggregate {
    ct = count s"*",
  }
)
sort ct
filter ct > 200
```

```elm
from employees
filter gross_cost > 0
take 20
sort ct
filter ct > 200
sort ct
take 20
```

```elm
from employees
filter gross_cost > 0
take 20
sort ct
filter ct > 200
sort ct
take 20
filter ct > 200
sort ct
filter ct > 200
sort ct
```

## Not working

```elm
from employees
filter gross_cost > 0
group {title} (
  aggregate {
    ct = count s"*",
  }
)
sort ct
filter ct > 200
take 20
sort ct
filter ct > 200
take 20
sort ct
filter ct > 200
take 20
sort ct
filter ct > 200
take 20
```

```elm
from employees
filter gross_cost > 0
sort ct
filter ct > 200
take 20
sort ct
filter ct > 200
take 20
sort ct
filter ct > 200
take 20
sort ct
filter ct > 200
take 20
```

```elm
from employees
filter gross_cost > 0
take 20
sort ct
filter ct > 200
sort ct
take 20
filter ct > 200
sort ct
filter ct > 200
sort ct
filter ct > 200
```

_Originally posted by @eitsupi in https://github.com/PRQL/prql/issues/2717#issuecomment-1596099365_

max-sixty:
Ah, great discovery!

Does `prqlc parse` fail too? I would have thought it were the parser...

eitsupi:
> Does `prqlc parse` fail too? I would have thought it were the parser...

No, it seems working.

```powershell
❯ ./prqlc.exe parse
Enter PRQL, then press ctrl-z to compile:

from employees
filter gross_cost > 0
take 20
sort ct
filter ct > 200
sort ct
take 20
filter ct > 200
sort ct
filter ct > 200
sort ct
filter ct > 200

sources:
  '':
  - name: main
    VarDef:
      value:
        Pipeline:
          exprs:
[...]

PR #2994, opened 2023-07-17
refactor: extract prql_ast crate

not-my-profile:
There are a bunch of reasons why it makes sense to have the actual
abstract syntax tree (AST) in a dedicated crate, e.g:

* the code formatter just needs the AST, it doesn't need type
  information (and the previous mixing up of syntax elements
  with inferred info was very confusing ... case in point the
  "codegen" formatter even formatted inferred args for
  Pipeline syntax nodes).

* Currently there's just the prql-compiler which does PRQL -> SQL,
  but the other direction SQL -> PRQL is also interesting,
  and you should be able to implement mappings to PRQL syntax elements
  without depending on the compiler or the parser.

* Some crates may even want to interpret PRQL queries
  directly without SQL being involved at all
  (e.g. using PRQL as a simple filtering language or
  perhaps even linting PRQL queries.)

So this PR decouples a prql_ast crate by making the Expr struct
generic over a new trait with a bunch of associated types.

This is the followup PR for #2993. It will be rebased once that PR has been merged so that only the last 12 commits remain.

For more detailed reasoning please refer to the individual commit messages.

Todos:

* [ ] figure out why test-rust is failing on windows
* [x] don't move Span to the new crate, instead introduce a new `Span` associated type on the trait


not-my-profile:
@aljazerzen do you have an idea what's up with the Windows test failure?

```
error: failed to run custom build command for `compile-files v0.8.1 (D:\a\prql\prql\prql-compiler\examples\compile-files)`

Caused by:
  process didn't exit successfully: `D:\a\prql\prql\target\debug\build\compile-files-4feec7a1a22f2d69\build-script-build` (exit code: 0xc00000fd, STATUS_STACK_OVERFLOW)
  --- stderr

  thread 'main' has overflowed its stack
```

the example builds fine for me on linux :shrug:

aljazerzen:
The parser is behaving a bit strange on Windows, but I'm not willing to investigate. https://github.com/PRQL/prql/issues/2857

not-my-profile:
So I assume this is not a blocker? It would be nice to see the CI passing on the other targets ... should we perhaps reconfigure the CI to still run the other targets even if windows-latest fails?

aljazerzen:
> So I assume this is not a blocker?

I want to say it is not, but ask @max-sixty

aljazerzen:
I don't have the bandwidth to review this PR now, but from the first glance this this major refactor seems a bit premature. The amount of new lines added to the `semantic` module, for example, does not seem to outweigh the benefits of having two separate crates.

If we wanted better separation between "the syntactic AST" and "the inferred AST", we could look into producing a Concrete Syntax Tree with the parser and compiling that to AST as the first step of resolving. 

Edit: again, I didn't look through the whole PR, I might be missing something.

not-my-profile:
I'd like to use the PRQL AST and parser in a project without adding a bunch of build dependencies I don't need (I don't need the PRQL -> SQL step). And besides that I think the benefits on the project overall are clear ... e.g. `fmt` would be easier if there was a clear separation between syntactic AST and inferred AST.

I am not sure what you mean by "amount of new lines added to the semantic module" the last commit doesn't add much:

```
 prql-compiler/src/semantic/context.rs         |  12 ++-
 prql-compiler/src/semantic/eval/mod.rs        |  43 ++++++-----
 prql-compiler/src/semantic/lowering.rs        |  92 +++++++++++-----------
 prql-compiler/src/semantic/reporting.rs       |  20 +++--
 prql-compiler/src/semantic/resolver.rs        | 209 +++++++++++++++++++++++++++++---------------------
 prql-compiler/src/semantic/static_analysis.rs |   8 +-
 prql-compiler/src/semantic/transforms.rs      |  92 +++++++++++++---------
 prql-compiler/src/semantic/type_resolver.rs   |  25 +++---
 8 files changed, 280 insertions(+), 221 deletions(-)
```

I think +60 lines in `semantic` is a very small price to pay for enabling users to depend on the PRQL AST without pulling in dozens of unneeded build dependencies. So I'd personally rather consider this factorization as overdue rather than premature.^^

And I don't really see the appeal of a Concrete Syntax Tree, I think ASTs are much easier to work with.

aljazerzen:
Having a use-case for the change helps a lot.

New lines is not a good measure, but there is quite a lot of added friction with `.extra` and `ExprKindExtra`. It's cognitive overhead that's unnecessary when working on an-already-very-complex-system.

I would insist this adds too much complexity:

https://github.com/PRQL/prql/pull/2994/files#diff-d37c9885d3249f53321b812dd527d9db975bf613e3aa3f8e9b1006620e8c8f38R9-R15

---

Can you consider the "early syntactic" AST idea I floated before? Call it Concrete Syntax Tree or something else, we could have something like this:

```
Syntactic AST       PL AST

Ident           ->  Ident
Binary          ->  Binary
                    RqOperator
```

It could also be generic over span (which is not something we'd need for the PL) and would not have all of the fields required for resolving.

There would be a bit of code duplication and an additional tree pass to map the early nodes into PL nodes. I do think this would be the more readable approach, as opposed to generics.

not-my-profile:
Yeah that makes sense. Needing to convert the syntactic AST to the PL AST will come with some runtime overhead but I guess that's better than mental overhead.

PR #2997, opened 2023-07-18
fix(test): skip compile_help test on Windows

not-my-profile:
This caused the test failure in  #2996.


max-sixty:
Thanks a lot @not-my-profile.

PR #2998, opened 2023-07-18
refactor: move `RelationLiteral` and generic structs out of `pl`

not-my-profile:
(as usual please merge via rebase)

### DIFF
--- a/prql-compiler/src/semantic/lowering.rs
+++ b/prql-compiler/src/semantic/lowering.rs
@@ -8,8 +8,8 @@ use itertools::Itertools;
 
 use crate::ast::pl::fold::AstFold;
 use crate::ast::pl::{
-    self, BinaryExpr, Expr, ExprKind, Ident, InterpolateItem, Lineage, LineageColumn, QueryDef,
-    Range, SwitchCase, TupleField, UnaryExpr, WindowFrame,
+    self, BinaryExpr, Ident, InterpolateItem, Lineage, LineageColumn, QueryDef, Range, SwitchCase,
+    TupleField, UnaryExpr, WindowFrame,
 };
 use crate::ast::rq::{
     self, CId, Query, RelationColumn, RelationLiteral, TId, TableDecl, Transform,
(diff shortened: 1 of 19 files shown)


prql-compiler/src/semantic/lowering.rs
```diff
@@ -552,18 +555,18 @@ impl Lowerer {
...
-    fn lower_range(&mut self, range: pl::Range<Box<pl::Expr>>) -> Result<Range<rq::Expr>> {
+    fn lower_range(&mut self, range: Range<Box<pl::Expr>>) -> Result<Range<rq::Expr>> {
         Ok(Range {
             start: range.start.map(|x| self.lower_expr(*x)).transpose()?,
             end: range.end.map(|x| self.lower_expr(*x)).transpose()?,
         })
     }
 
-    fn lower_sorts(&mut self, by: Vec<pl::ColumnSort>) -> Result<Vec<pl::ColumnSort<CId>>> {
+    fn lower_sorts(&mut self, by: Vec<ColumnSort<Box<pl::Expr>>>) -> Result<Vec<ColumnSort<CId>>> {
```
not-my-profile, line 565: I think this nicely shows how the code becomes more readable when you don't have implicit type parameter defaults.

prql-compiler/src/semantic/lowering.rs
```diff
@@ -849,7 +852,7 @@ impl Lowerer {
 
     fn lower_interpolations(
         &mut self,
-        items: Vec<InterpolateItem>,
+        items: Vec<InterpolateItem<pl::Expr>>,
     ) -> Result<Vec<InterpolateItem<rq::Expr>>> {
```
not-my-profile, line 856: Same here ... just more readable.

max-sixty:
Thanks @not-my-profile , these are indeed nice improvements

For transparency, @aljazerzen knows the code better than I do — he has written most of it! — so I'm reviewing what I can but biased to just merge, given it seems you more than know what you're doing... If there's something you think requires more review, it's worth raising that in case I miss it...

PR #3000, opened 2023-07-19
refactor: extract `prql_ast` and `prql_parser` crates

not-my-profile:
Splitting the AST and the parser up into separate crates allows other crates that don't need the PRQL->SQL compilation step to parse PRQL without having to add unnecessary build dependencies. It additionally speeds up compile times and cleans up the code since it enforces boundaries (e.g. the parser previously depended on `semantic` and `codegen` ... which it really shouldn't).

As usual please merge via rebasing. You'll want to review this commit by commit. Let me know if you have any questions :) If it helps with reviewing, I can also split these commits across several PRs.

Planned followups:
* move `prql-compiler`, `prql-compiler/prql-compiler-macros` and `prql-compiler/prqlc` to `crates`
* move the documentation tests to the new `misc_tests` crate (it doesn't make much sense that the tests in book also test the top-level README and the website examples)
* improve the API documentation of the new crates


misc_tests/Cargo.toml
max-sixty, line 1: Can I ask why this needs to be in a separate crate? Additional test binaries slow down tests a bit (check out our development docs for more details) — can it go in `prql-compiler`?

misc_tests/tests/ast_code_matches.rs
```diff
@@ -0,0 +1,150 @@
...
+        /// This comment was changed
+        Bar,
+
+        Baz,
+
+        /// This comment was added
+        Fiz,
+
+        /// This variant was added
+        Buz,
```
max-sixty, line 108: Wouldn't we want `Buz` in the diff?

crates/prql_ast/Cargo.toml
max-sixty, line 1: Do you think if we have a `crates` path, we should flatten the crates within `prql-compiler` and put them here? I've seen advice to have a `crates` path, am supportive.

We can do it later to stop this PR growing too big though.

(`bindings` could still be separate?)

max-sixty: It would be consistent if we inherit the values from the workspace in these, like we do in https://github.com/prql/prql/blob/41cf29d315e9e350962e7e773386093eb81b2933/prql-compiler/Cargo.toml#L5-L11, is that reasonable?

max-sixty (approved):
Overall it looks great.

@aljazerzen has more context on some of this, so might want to voice a view (but is away, and wouldn't us to wait)

I added a couple of small comments on the broad shape. The additional code looks good!

(I would be keener on smaller PRs if you're up for it, and will reciprocate by endeavoring to respond quickly...)

misc_tests/Cargo.toml
not-my-profile: When you publish a crate the the source files get put in the `.crate` archive file. I think it would be best if the tests still work within that archive file which means they cannot rely on files outside of the crate (unless we include them in the build process). And this test compares files of two different crates.

misc_tests/tests/ast_code_matches.rs
not-my-profile: The PL AST adds a bunch of additional variants and fields to the parser AST. This test is only meant to check that the shared fields have the same doc comments ... listing all additional variants and fields in the diff wouldn't actually add anything and would just make editing them more cumbersome since you'd always have to update the snapshot.

misc_tests/Cargo.toml
max-sixty: I still don't quite get the disadvantage of having the test in `prql-compiler`... Surely lots of `prql-compiler` tests require `prql-compiler` and its dependency? What's the principal for when to create a new crate?

misc_tests/tests/ast_code_matches.rs
max-sixty: Ah I see, that makes sense, thanks. Possibly we could rename `diff_code` to something more explicit about the doc comments?

misc_tests/Cargo.toml
not-my-profile: There's a difference between using an API from a dependency defined in `Cargo.toml` (which is sound) and directly reading a source file from another crate and thereby relying on it being present in the parent directory of the current crate (which will break when trying to run the tests in the extracted crate).

misc_tests/tests/ast_code_matches.rs
not-my-profile: I have updated the comment in the test to clarify why `Buz` won't show up in the diff.

crates/prql_ast/Cargo.toml
not-my-profile: Yes we should indeed flatten the crates within `prql-compiler` and put them there (as I have already written under followups of the PR description). And yes we should do that in a followup PR. I also agree that it's best to keep `bindings` separate.

I just updated the two new `Cargo.toml` files to set:

```
edition.workspace = true
license.workspace = true
repository.workspace = true
rust-version.workspace = true
```

I did not add `version.workspace = true` since we'll want these crates to have independent versions.

I also didn't add `metadata.msrv = "1.65.0"` since I am hoping that foresterre/cargo-msrv#590 will be resolved soon.

not-my-profile:
> aljazerzen has more context on some of this, so might want to voice a view (but is away, and wouldn't us to wait)

There is no particular hurry. If you'd rather wait for more feedback this can wait.

> I would be keener on smaller PRs if you're up for it, ...

Right ... that makes sense. Splitting up larger changes like these across multiple PRs is kind of a pain with GitHub since it doesn't support stacked PRs ... however I don't think I have any changes planned for PRQL that are this big, so my future PRs should be significantly smaller :)

I did just address your feedback but now the CI is failing because it suddenly cannot download the crates for some reason:

```
Downloading crates ...
warning: spurious network error (2 tries remaining): [6] Couldn't resolve host name (Could not resolve host: static.crates.iocrates)
warning: spurious network error (2 tries remaining): [6] Couldn't resolve host name (Could not resolve host: static.crates.iocrates)
warning: spurious network error (1 tries remaining): [6] Couldn't resolve host name (Could not resolve host: static.crates.iocrates)
warning: spurious network error (1 tries remaining): [6] Couldn't resolve host name (Could not resolve host: static.crates.iocrates)
error: failed to download from `[https://crates.io/api/v1/crates/regex/1.9.1/download`](https://crates.io/api/v1/crates/regex/1.9.1/download%60)

Caused by:
  [6] Couldn't resolve host name (Could not resolve host: static.crates.iocrates)
Error: The process '/home/runner/.cargo/bin/cargo' failed with exit code 101
```

crates/prql_ast/Cargo.toml
max-sixty: Super, thank you!

> I did not add `version.workspace = true` since we'll want these crates to have independent versions.

FWIW, so far we've been successful with versioning everything as one, which also makes aligning git tags with cargo versions very easy. But open-minded and we can discuss...

> (as I have already written under followups of the PR description).

Apologies, not sure how I missed this

misc_tests/Cargo.toml
max-sixty: Let's merge because I don't want my slow understanding to prevent progress. 

But I still don't get the advantage — I get that depending on the exact file structure is temperamental — but don't both approaches have this disadvantage — the test here also calls `read_to_string` on a file in another crate? (BTW would `include_str!` be better here, since we can do this at compile time?)

Would it be reasonable for me to open an issue to get others' thoughts? Or are you very confident and don't think a discussion is productive?

not-my-profile: Ah right the point to note here is that `misc_tests` is only an internal crate we will never publish it to crates.io. For crates that we intend to publish on crates.io like `prql_ast` and `prql-compiler` I think we should take care that their tests still work in the `.crate` archive files we publish. I'll add a README to the `misc_tests` crate to explain this :)

Regarding the choice between `read_to_string` and `include_str!` ... I don't think `include_str!` comes with any advantage here ... it probably just means that the `misc_tests` crate has to be (rather needlessly) recompiled when these files change.

crates/prql_ast/Cargo.toml
not-my-profile: No worries this is a long page ... it's easy to miss something :)

Yeah versioning all crates in a workspace as one seems to be quite common practice ... however I personally don't really like the idea of needlessly version bumping crates e.g. why should the version number of `prql_ast` be bumped if there haven't been any changes to it or any of its dependencies? Ad git tag & cargo version alignment: `cargo publish` does actually record the git commit in the published crate in a `.cargo_vcs_info.json` file ... unfortunately this doesn't currently really seem to be surfaced by any tooling :/

misc_tests/Cargo.toml
max-sixty: Thanks for the explanation re the `.crate` archive files.

The advantage of `include_str!` is that it runs on compilation, so it _would_ presumably work from any compiled file that's distributed, and I think would satisfy that concern?

max-sixty: (I'm keen to keep merging things and not slow you down! I would vote to not move _more_ code into `misc_tests` unless you're really confident that this is the right path, until we're all more in sync; hope that's reasonable!)

not-my-profile: I see that you're confused about `.crate` files ... they don't contain the compiled files ... they contain the source code to build the compiled files. So any file included via `include_str!` would still be need to be present in the `.crate` file.

crates/prql_ast/Cargo.toml
max-sixty: > however I personally don't really like the idea of needlessly version bumping crates e.g. why should the version number of `prql_ast` be bumped if there haven't been any changes to it or any of its dependencies?

I empathize with this. But it is much simpler to just say "version X" and have that represent the version of everything. I'm not sure what we get for the heterogeneous versioning...

(I realize we're talking about quite narrow things lots, sorry if that's because of me, I don't mean to nerd-snipe us away from the important work. Possibly we open an issue for each of these and see what others think...)

misc_tests/Cargo.toml
max-sixty: Ah, sorry, I see. I think we're at least in sync on the facts now!

I'm still not sure how much I would weigh the `.crate` of `prql-compiler` passing tests independently, relative to a smaller number of crates & test binaries, though. I'm not sure how to balance our respective intuitions — is there anything out there (blogs etc) that promotes prioritizing this?

not-my-profile: I am not aware of any blog posts on this ... however getting this right doesn't really pose much additional challenge for us ... I'd consider not randomly relying on files from the parent directory to be good practice in any case.

aljazerzen:
Woah, this is a big effort. Too big to review and comment on in the PR alone. Please open multiple PRs for such refactors in the future.

I have objections on a few decisions, but nothing major, so I can be updated in my followup PRs. I haven't read all the new code, so this will get done when I find them.

In any case, now I have a lot of work to do to rebase my Work-In-Progress branch, which I did not merge before this because it's large effort, which touches quite a few files :sweat_smile:

not-my-profile:
I already offered to split the PR if it helps with reviewing ... I'll just do so in the future :)

>  Too big to review and comment on in the PR alone.

You can step through the individual commits and comment on them this way.

I am very curious about the decisions you object to, since I don't see anything objectionable ... if you follow up on this PR please tag me.

As I said in #3024 I have a bunch more changes to the semantic module coming up ... I am happy to rebase your branch for you ... as long as it compiles and passes the tests ;)

Cargo.toml
```diff
@@ -1,5 +1,7 @@
 [workspace]
 members = [
+  "crates/*",
+
```
aljazerzen: This repo structure feels a bit off. The AST, parser, resolver and SQL backend and cli are all parts of prql-compiler effort, so having `crates/` and then `prql_compiler` inside seems backwards.

To me, it would make sense that we'd have:

```
web/
  ...
prql-compiler/
  prql_ast/
  prql_parser/
  prql_compiler/ <-- the crate that is exposed as the main library
  prql_compiler_macros/
  prqlc/

  prql_frontend/ <-- split this off someday
  prql_backend_sql/ <-- split this off someday
  prql_backend_substrait/ <-- someday
  prql_backend_polars/ <-- someday
```

And also `bindings/` would be under `prql-compiler/`. And packages would be under `prqlc/`.

Also, we might want to omit the `prql_` prefix.

@max-sixty Any thoughts?

Maybe I'm just overthinking things.

prql-compiler/src/parser.rs
```diff
@@ -0,0 +1,274 @@
...
+        }),
+        // We want a friendlier message than "found end of input"...
+        None => Error::new(Reason::Simple(format!(
+            "Expected {expected}, but didn't find anything before the end."
+        ))),
+    }
+    .with_span(Some(*span))
+}
+
+struct DisplayToken<'a>(&'a Token);
```
aljazerzen, line 141: Why do we need `DisplayToken`? Because `impl Display for Token` needs `impl Display for Literal` and that's in prql_compiler?

Technically this is codegen, but this would get much simpler if `impl Display for Literal` was in `prql_ast` crate.

crates/prql_parser/src/lib.rs
```diff
@@ -1,46 +1,42 @@
 mod expr;
 mod interpolation;
-mod lexer;
+pub mod lexer;
```
aljazerzen, line 3: lexer should be private, right?

But we need it public because we are mapping the chumsky error (and `lexer::Token`) into our error in the `prql_compiler` crate? If `prql_parser` is to be used elsewhere, we should probably do the conversion into our error in the `prql_parser` and not expose chumsky errors.

But this would require moving our error out of `prql_compiler`.

crates/prql_ast/Cargo.toml
max-sixty: Opened #3047

Cargo.toml
max-sixty: I find https://matklad.github.io/2021/08/22/large-rust-workspaces.html a good structure, which broadly is consistent with the version in this PR.

Putting `bindings` separately is different from that advice. Though I can see the split making sense — I net to +/-0...

not-my-profile: A `crates` directory containing directories named after the crates is a well-established convention that helps new contributors quickly find the source code of crates. Putting the AST and the parser under compiler doesn't make any sense to me since they can very well be used on their own.

Anyway please open new issues for new discussions rather than making comments on old PRs.

PR #3005, opened 2023-07-20
chore: move crates in prql-compiler to crates/

not-my-profile:


max-sixty:
Looks great! Thanks for fixing all the links etc.

The link checker will fail for one commit but then will start passing immediately, so we can merge as soon as the other tests pass

not-my-profile:
It seems like everything is passing except link-check ... which as you said just doesn't know that the paths are changed by this PR.

PR #3015, opened 2023-07-22
ci: check internal links in mdbook after build

not-my-profile:
We already run `markdown-link-check` in our CI, however by virtue of it running against `.md` files it cannot detect links that are broken due to mdbook-specifics (such as forgetting to list the .md file in SUMMARY.md or [the README.md bug](https://github.com/rust-lang/mdBook/issues/984)). I already fixed 5 such broken links in #3017 and #2990 ... we should really detect these broken links in the CI.

This PR therefore adds a new `link-check-book` job to our CI that builds the mdbook and then checks the generated HTML for broken links with the [hyperlink](https://crates.io/crates/hyperlink) link checker.

### DIFF
--- a/.github/workflows/check-links-book.yaml
+++ b/.github/workflows/check-links-book.yaml
@@ -5,10 +5,6 @@
 name: check-links-book
 
 on:
-  pull_request:
-    paths:
-      - "web/book/**"
-      - ".github/workflows/check-links-book.yaml"
   workflow_call:
 
 env:
@@ -20,7 +16,25 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
+  changes:
+    runs-on: ubuntu-latest
+    permissions:
+      pull-requests: read
+    outputs:
+      src: ${{ steps.filter.outputs.src }}
+    steps:
+      - uses: dorny/paths-filter@v2
+        id: filter
+        with:
+          filters: |
+            src:
+              - '.github/workflows/check-links-book.yaml'
+              - 'web/book/**'
+
   check:
+    needs: changes
+    if: ${{ needs.changes.outputs.src == 'true' }}
+
     runs-on: ubuntu-latest
 
     steps:
--- a/.github/workflows/check-links-book.yaml
+++ b/.github/workflows/check-links-book.yaml
@@ -1,0 +1,53 @@
+# We also have a check-links-markdown job, however it will not spot mdbook
+# mistakes such as forgetting to list an .md file in SUMMARY.md.
+# Running a link checker on the generated HTML is more reliable.
+
+name: check-links-book
+
+on:
+  pull_request:
+    paths:
+      - "web/book/**"
+      - ".github/workflows/check-links-book.yaml"
+  workflow_call:
+
+env:
+  RUSTFLAGS: "-C debuginfo=0"
+
+concurrency:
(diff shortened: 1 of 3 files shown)


.github/workflows/pull-request.yaml
```diff
@@ -92,7 +92,7 @@ jobs:
     # Check out `test-all.yaml` for more ideas on organizing these.
     if: always()
     needs:
```
not-my-profile, line 97: I think it would be nice to include the new `check-links-book` job under `check-ok-to-merge.needs` ... however I guess then we could no longer use `on.pull_request.paths` in the new workflow file? And would probably have to do something like [what's described in this blog post](https://www.meziantou.net/executing-github-actions-jobs-or-steps-only-when-specific-files-change.htm)?

max-sixty (approved):
Great idea!

.github/workflows/check-links-book.yaml
```diff
@@ -0,0 +1,53 @@
...
+          save-if: ${{ github.ref == 'refs/heads/main' }}
+
+      - uses: arduino/setup-task@v1
+        with:
+          repo-token: "${{ secrets.GITHUB_TOKEN }}"
+
+      - name: Build the mdbook
+        run: mdbook build web/book/
+
+      - name: Check links
```
max-sixty, line 66: One option would be to add this to the existing workflow, or to call the existing workflow as part of this. But no strong view.

max-sixty (approved):

.github/workflows/pull-request.yaml
max-sixty: Yes, unfortunately I'm not sure it's easy to do this...

.github/workflows/check-links-book.yaml
max-sixty: Though possibly moving the _job_ into `build-web.yaml` would be a reasonable consolidation — it would run slightly too often, but not by much.

not-my-profile: I am planning on linking relevant pages of the book from doc comments in `prql_ast` in the future ... at which point I want to expand this CI job to check these links as well ... so I think it would be best to keep these separate since they have different path requirements and do largely different things.

.github/workflows/check-links-book.yaml
```diff
@@ -0,0 +1,67 @@
...
+
+      # the link checker
+      - uses: baptiste0928/cargo-install@v2
+        with:
+          crate: hyperlink
+
+      - name: Cache
+        uses: Swatinem/rust-cache@v2
+        with:
+          prefix-key: 0.8.1
```
eitsupi, line 56: FYI, prefix-key should be updated by the Cargo.toml.

https://github.com/PRQL/prql/blob/0224b6efb140b8aec4d8c85bd62967cebdf7b2e2/crates/prql_compiler/Cargo.toml#L78-L85

max-sixty: Ah, missed that, thanks @eitsupi

.github/workflows/pull-request.yaml
max-sixty: Just saw what you did here — nice! I had seen this action before, but hadn't tried it out yet. Looks good; we could think about extending to more places if it turns out well.

PR #3016, opened 2023-07-22
fix(docs): fix two links broken by crate move

not-my-profile:
This commit fixes two links I forgot to update in
ef89bba0ca3f35c5b1cc1aa48cf814fe6ed5d4a4.


max-sixty:
Thanks for fixing all these

PR #3019, opened 2023-07-22
refactor: make Main a StmtKind in prql_ast

not-my-profile:
The types in prql_ast should model the language as closely as possible. You can never specify a name or a type for a main statement, so it should not use the VarDef struct in prql_ast. Moving the Main variant from VarDefKind to StmtKind also lets us make name in VarDef required.


max-sixty:
Makes sense to me (albeit I'm not the authority here...)

PR #3023, opened 2023-07-23
refactor: privatize Resolver struct 

not-my-profile:
Previously you had to use the Resolver as follows:

    let mut resolver = Resolver::new(context, options);
    resolver.current_module_path = path;
    resolver.fold_statements(stmts)?;
    let new_context = resolver.context;

This API is problematic since you could forget to set the current module
path and it's not at all obvious that you're supposed to access .context
after calling fold_statements. In this and the following four commits
we will simplify the API to:

    resolve(&mut context, path, stmts &options)?;

### DIFF
--- a/crates/prql_compiler/src/semantic/context.rs
+++ b/crates/prql_compiler/src/semantic/context.rs
@@ -6,6 +6,7 @@ use std::{collections::HashMap, fmt::Debug};
 use super::*;
 use crate::ast::pl::*;
 use crate::error::Span;
+use crate::utils::IdGenerator;
 
 /// Context of the pipeline.
 #[derive(Default, Clone)]
@@ -14,6 +15,8 @@ pub struct Context {
     pub(crate) root_mod: Module,
 
     pub(crate) span_map: HashMap<usize, Span>,
+
+    pub(crate) id: IdGenerator<usize>,
 }
 
 /// A struct containing information about a single declaration.
--- a/crates/prql_compiler/src/semantic/context.rs
+++ b/crates/prql_compiler/src/semantic/context.rs
@@ -8,7 +8,7 @@ use crate::ast::pl::*;
 use crate::error::Span;
 
 /// Context of the pipeline.
-#[derive(Default, Serialize, Deserialize, Clone)]
+#[derive(Default, Clone)]
 pub struct Context {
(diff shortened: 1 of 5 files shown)


crates/prql_compiler/src/semantic/context.rs
```diff
@@ -6,14 +6,17 @@ use std::{collections::HashMap, fmt::Debug};
...
 /// Context of the pipeline.
-#[derive(Default, Serialize, Deserialize, Clone)]
+#[derive(Default, Clone)]
 pub struct Context {
     /// Map of all accessible names (for each namespace)
     pub(crate) root_mod: Module,
 
     pub(crate) span_map: HashMap<usize, Span>,
+
+    pub(crate) id: IdGenerator<usize>,
```
aljazerzen, line 19: Why should the id generator be part of the context?

Conceptually, `Context` has semantic value, as the environment within which expressions are resolved. Id generator would be a part of the thing that resolves, not the environment itself. Also, the generator is not meant to be serialized.

crates/prql_compiler/src/semantic/context.rs
```diff
@@ -6,14 +6,17 @@ use std::{collections::HashMap, fmt::Debug};
 use super::*;
 use crate::ast::pl::*;
 use crate::error::Span;
+use crate::utils::IdGenerator;
 
 /// Context of the pipeline.
-#[derive(Default, Serialize, Deserialize, Clone)]
```
aljazerzen, line 11: Context is meant be serializable, so it can be persisted to disk. This is not used right now, but there are plans to do it.

aljazerzen (approved):
Looks good, the interface is clearer now.

crates/prql_compiler/src/semantic/context.rs
not-my-profile: The Context currently also contains the `span_map` which doesn't have semantic value.

aljazerzen: True, but that's something that should be kept close to declarations for things like "goto definition" in LSP.

not-my-profile:
Closing this as, as Aljaz pointed out, the IdGenerator doesn't belong in the Context ... which should really be called ResolvedQuery ... I'll take another stab at this in a new PR.

PR #3024, opened 2023-07-23
refactor: avoid .unwrap() by narrowing param types 

not-my-profile:


aljazerzen (commented):
Which unwraps does this remove? To me, this looks like just unpacking params earlier?

crates/prql_compiler/src/semantic/resolver/transforms.rs
```diff
@@ -18,41 +19,44 @@ use super::{Context, Lineage};
 use super::{NS_PARAM, NS_THIS};
 
 /// try to convert function call with enough args into transform
-pub fn cast_transform(resolver: &mut Resolver, closure: Func) -> Result<Expr> {
-    let internal_name = closure.body.kind.as_internal().unwrap();
```
not-my-profile, line 22: This one.

aljazerzen: I see. As this is marginal, could I merge my branch before we do a lot of changes to semantic module?

I cannot guarantee the timeline, so I know it's a lot to ask. Such refactors really complicate merging-in the new changes.

not-my-profile: I am afraid I have a bunch of changes upcoming for the semantic module :sweat_smile:. I am only starting with the trivial stuff they are building up to more high-value changes.

What are your changes about?

aljazerzen: Removing Lineage and moving inference of that into Ty.

This removes Expr::target_id, Expr::target_ids, ExprKind::All and introduces Expr::TupleFields and Expr::TupleExclude. It's a cleanup I wanted to do for sometime, because right now, we are essentially doing type inference in two places.

not-my-profile: Ah that sounds lovely! I am currently working on getting rid of `ExprKind::Internal` since leaking compiler internals in the AST and the parser is just poor design ... I am like 90% done and have like 20 commits that can be merged already but just have to be split up across multiple PRs to be easily reviewable.

I am very well versed with rebasing branches after they have become out of date, so I could rebase your branch for you after I'm done with my planned refactors ... this way I wouldn't be blocked and you could continue your work at your convenience ... how does that sound?

aljazerzen: Wait, I did check and this specific PR does not conflict. My branch is `aljaz` on the main repo. It does not pass tests right now (maybe it doesn't even compile).

---

But regarding this PR, I don't feel like this is an improvement. The `.unwrap` would never get hit and new code is quite more verbose. The choice for passing in a closure that has been checked to contain an `Internal` node was conscious. It could be documented though.

aljazerzen: Rebasing my branch sounds great, go ahead!

aljazerzen: How are you planning to get rid of `ExprKind::Internal`? If you want, we could talk about stuff like this on Discord, which we use for faster, low-importance comms.

not-my-profile: I am not gonna rebase a branch that doesn't pass all tests. I think it's very important that all tests pass before you start rebasing because otherwise you cannot easily check that everything is still ok during rebasing.

> But regarding this PR, I don't feel like this is an improvement. The .unwrap would never get hit and new code is quite more verbose.

I strongly disagree. Every `.unwrap()` is a liability ... a runtime panic waiting to happen if the API is accidentally misused ... if we can avoid unwraps just by refining types we should *always* do so. Refined types also just make the code more readable. You immediately see with which types you are allowed to call the method ... otherwise we'd need to document that and doc comments can easily get out of date ... types are checked ... types are great ... we should use types. Turning runtime errors into comp-time errors is very much what makes Rust code idiomatic.

> How are you planning to get rid of ExprKind::Internal?

By simply post processing the std lib PL AST rather than doing hacks like the following:

https://github.com/PRQL/prql/blob/44457ebe48919147790080199cbb52f5d4f46022/crates/prql_compiler/src/semantic/resolver/mod.rs#L93-L97

I'd be happy to talk about stuff like this but I don't use Discord. If you want to we chat via [Matrix](https://matrix.org/).

not-my-profile: Ok your branch did compile ... I just rebased it and pushed it to `type-infer-2.0`.

aljazerzen: Great, thanks for rebasing. It is much easier if the tests pass, but if they did, I would have merged it sooner.

----

> Every .unwrap() is a liability

That is generally true, but not always. Sometimes it is obvious that something cannot panic, and in those cases it does not make sense increase code complexity for the sake of chasing ideomatic code. Rust already has slower development speed compared to some other languages and I don't want to add to the problem by sticking to guidelines that work against that.

The idea is that we can start quickly with code that works, has low complexity and is easy to experiment on. Only when the project is deemed feature-complete, we transition into optimizations and idiomatic code.

I know that this might feel like working against all of the Rust principles, but it a way of how not to get bogged down by making the code perfect.

max-sixty:
@not-my-profile requested a review from me. I don't have a strong view on the specifics; this comment is unlikely to be a great revelation to anyone...

On the general point, I agree that minimizing `.unwrap`s is good. "Make invalid states unrepresentable" is an excellent principal.

But I also think that it's possible to take that too far, particularly while the codebase is still going through changes. If we have an AST that requires unwraps everywhere and we're hitting them regularly, then we don't have a great AST. But if there are corner-cases where we have the odd unwrap which we never hit, adding complexity to the code is _probably_ not a net improvement, at least without a plan to make the overall codebase simpler.

I'm keen to be a welcoming and accepting project — I think we're all quite low-ego here — and we're committed to making the codebase better regardless of who wrote the previous code.

Anything else? Can we be better at signaling the refactorings which we'd prioritize, to avoid wasted work?

PR #3031, opened 2023-07-24
refactor!: rename prql_compiler::ast to prql_compiler::ir

aljazerzen:
Now that we have a proper crate for AST, we can call IRs
what they are: intermediate representations.


aljazerzen:
Just to recap all the representations:

- AST - plain AST for general use
- IR: PL - Pipelined Language (AST + all of the resolver stuff) 
- IR: RQ - Relational Query (expressive power of an SQL SELECT)
- IR: SRQ - SQL Relational Query (RQ, but closer to SQL)

crates/prql_compiler/src/ir/pl/mod.rs
```diff
@@ -8,21 +8,23 @@
...
-pub mod extra;
-pub mod fold;
+mod expr;
+mod extra;
+mod fold;
 mod from_ast;
-pub mod lineage;
-pub mod stmt;
-pub mod types;
-pub mod utils;
```
aljazerzen, line 18: Should I do the same to `prql_ast` crate?

So all of the stuff is exposed on the top-level only and imports are simpler?

aljazerzen: @not-my-profile

not-my-profile: I'd rather not expose everything at the top level in prql_ast, I find the rustdoc to be more readable when the types are grouped by module.

not-my-profile (approved):
Rebasing my local commits will be a pain with all the changed imports and the renaming of `AstFold` but the changes make sense.

crates/prql_compiler/src/ir/pl/mod.rs
aljazerzen: Oh, rustdoc likes this better, yup, agree.

aljazerzen:
I know the pain :D We should probably stick to small and frequent PRs to avoid conflicts.

PR #3032, opened 2023-07-24
refactor!: emit AST instead of PL from prql-compiler parser

aljazerzen:
Based on #3023

The new AST before PL gave me ideas on how to simplify resolver code, by removing a few `ExprKind`, which are just syntactic sugar for some other basic exprs. For example, `ExprKind::Pipeline` is just a bunch of `ExprKind::FuncCall`, semantically.

So this PR prepares ground for that: it converts `from_ast` module into a proper AST pass (similar to what `lowering` module does). It also returns `prql_ast::Stmt` instead of `pl::Stmt` from the parser, since that is the AST people would expect from the compiler, instead of the de-sugared PL version.

This has the knock-on effect of codegen having to consume AST instead of PL. And this has a knock-on effect of semantic module not being able to codegen from PL nodes. So I also had to add reverse mapping from PL back to AST.


not-my-profile:
I am confused ... what do you mean by "proper AST pass"? I don't see the point of changing from_ast to expand_expr ... what's the advantage? This PR doesn't actually remove any ExprKinds does it?

> semantic module not being able to codegen from PL nodes

Why does the semantic module need to codegen in the first place? (other than displaying nodes within error messages?)

aljazerzen:
> what do you mean by "proper AST pass"?

An explicit one. This is not conversion between two roughly equivalent data structures, but will be expanded to do de-sugaring as well.

>  This PR doesn't actually remove any ExprKinds does it?

No.

> Why does the semantic module need to codegen in the first place? (other than displaying nodes within error messages?)

Just for error (and other) messages.

crates/prql_compiler/src/ir/pl/from_ast.rs
```diff
@@ -1,165 +0,0 @@
...
-fn map_vec_into<I: Into<O>, O>(exprs: Vec<I>) -> Vec<O> {
-    exprs.into_iter().map(Into::into).collect::<Vec<_>>()
-}
-
-#[allow(clippy::boxed_local)]
-fn map_box_into<I: Into<O>, O>(expr: Box<I>) -> Box<O> {
-    Box::new((*expr).into())
-}
-
-impl From<prql_ast::expr::Expr> for pl::Expr {
```
not-my-profile, line 17: I consider a `From<>` impl to be more idiomatic than an `expand_expr` function.

crates/prql_compiler/src/lib.rs
```diff
@@ -305,12 +307,12 @@ pub mod json {
...
-    pub fn from_pl(pl: Vec<ir::pl::Stmt>) -> Result<String, ErrorMessages> {
+    pub fn from_pl(pl: Vec<prql_ast::stmt::Stmt>) -> Result<String, ErrorMessages> {
         serde_json::to_string(&pl).map_err(|e| anyhow::anyhow!(e).into())
     }
 
     /// JSON deserialization
-    pub fn to_pl(json: &str) -> Result<Vec<ir::pl::Stmt>, ErrorMessages> {
+    pub fn to_pl(json: &str) -> Result<Vec<prql_ast::stmt::Stmt>, ErrorMessages> {
         serde_json::from_str(json).map_err(|e| anyhow::anyhow!(e).into())
     }
```
not-my-profile, line 317: I think we should actually get rid of these functions so that we can drop the `serde_json` dependency.

max-sixty: IIRC, bindings use these. As does the playground...

crates/prql_compiler/src/ir/pl/from_ast.rs
aljazerzen: I do too, under the condition that the mapping is simple, just a conversion between two roughly equivalent types. This PR prepares ground for that not to be so.

crates/prqlc/Cargo.toml
```diff
@@ -23,7 +23,8 @@ colorchoice-clap = "1.0.0"
 env_logger = {version = "0.10.0", features = ["color"]}
 itertools = "0.11.0"
 notify = "^6.0.0"
-prql-compiler = {path = '../prql_compiler', version = "0.9.0", features = ["serde_yaml"]}
```
max-sixty, line 26: FYI this was required for releasing!

aljazerzen: Ah yes, of course. Sorry.

PR #3033, opened 2023-07-24
refactor!: rename `rq::Query` and `semantic::Context`

not-my-profile:
While we're at making widespread renames, I think we might as well make this change.


crates/prql_compiler/src/semantic/context.rs
```diff
@@ -9,7 +9,7 @@ use crate::ir::pl::*;
 
 /// Context of the pipeline.
 #[derive(Default, Serialize, Deserialize, Clone)]
-pub struct Context {
+pub struct ResolvedQuery {
```
aljazerzen: Yeah, name `Context` is too generic.

Although this does not include just a query, but all declarations: functions, types, modules.

Maybe we could call this `Environment`?

crates/prql_compiler/src/ir/rq/mod.rs
```diff
@@ -22,7 +22,7 @@ use super::pl::Ident;
 use super::pl::{Literal, QueryDef};
 
 #[derive(Debug, PartialEq, Clone, Serialize, Deserialize)]
-pub struct Query {
+pub struct RelationalQuery {
```
aljazerzen: Yup, much clearer.

aljazerzen (changes requested):

crates/prql_compiler/src/semantic/context.rs
not-my-profile: Environment isn't much better ... I think I actually prefer Context ... I think the main thing that bothered me was that lower_to_iq took just a `Context` ... but this can be resolved by introducing a new `ResolvedQuery` type like in the commit I just pushed ... please take a look.

crates/prql_compiler/src/semantic/lowering.rs
```diff
@@ -12,44 +12,37 @@ use crate::ir::generic::{InterpolateItem, Range, SwitchCase};
...
-pub fn lower_to_ir(context: Context, main_path: &[String]) -> Result<(Query, Context)> {
-    // find main
-    log::debug!("lookup for main pipeline in {main_path:?}");
-    let (_, main_ident) = context.find_main_rel(main_path).map_err(|hint| {
-        Error::new_simple("Missing main pipeline")
-            .with_code("E0001")
-            .with_hints(hint)
-    })?;
-
+pub fn lower_to_ir(query: ResolvedQuery) -> Result<(RelationalQuery, Context)> {
```
not-my-profile: I think this signature has become much clearer.

crates/prql_compiler/src/semantic/context.rs
aljazerzen: That's the key insight: lowering takes a set of resolved declarations, picks the one pointed to by `main_ident` and lowers that to `RelationalQuery`.

The `ResolvedQuery` struct you've created now also contains the `main_ident`, so more appropriate name would be `LoweringRequest`? But this is weird.

What about renaming `Context` to `ResolvedDeclarations`? A bit long...

not-my-profile: I'd say the lowering requires a main_ident it cannot take place when there isn't one.

aljazerzen: I'd agree, yes

not-my-profile:
@aljazerzen I have renamed `Context` to `RootModule` ... what do you think? Yes this currently makes accessing the inner module a bit awkward since we'd have e.g. `Resolver.root_mod.root_mod` however I think that can be solved in a followup PR at some point in the future by implementing `Deref<Target=Module> for RootModule` (as well as `DerefMut`).

aljazerzen (approved):
`RootModule` makes the most sense of everything we came up with :+1:

PR #3038, opened 2023-07-24
docs: Update aggregation.md

richb-hanover:


max-sixty:
(would have been v happy to comment here, might have even been easier to merge, but instead commented on the latest one — which is great — at https://github.com/PRQL/prql/pull/3088)

PR #3058, opened 2023-07-25
feat: func type

aljazerzen:
Add a new type that contains all functions, regardless of the number of
paramaters and the return type. It can referenced by using
`` std.`func` `` or `<func>` keyword.

This is needed for internal implementation.


eitsupi:
Perhaps the syntax file needs to be updated?
I believe `func` was excluded from the keywords in #2840.

PR #3060, opened 2023-07-25
refactor: desugar pipeline

aljazerzen:
Remove `pl::ExprKind::Pipeline`.
Desugar `ast::ExprKind::Pipeline` into `FuncCall` in `ast_expand`.

### DIFF
--- a/crates/prql-compiler/src/semantic/resolver/mod.rs
+++ b/crates/prql-compiler/src/semantic/resolver/mod.rs
@@ -12,7 +12,6 @@ use crate::utils::IdGenerator;
 
 use super::context::{Context, Decl, DeclKind, TableExpr};
 use super::module::Module;
-use super::reporting::debug_call_tree;
 use super::{write_pl, NS_DEFAULT_DB, NS_INFER, NS_STD, NS_THAT, NS_THIS};
 use flatten::Flattener;
 use transforms::coerce_into_tuple_and_flatten;
@@ -370,11 +369,6 @@ impl PlFold for Resolver {
                 self.fold_function(func, span)?
             }
 
-            ExprKind::Pipeline(pipeline) => {
-                self.default_namespace = None;
-                self.resolve_pipeline(pipeline)?
-            }
-
             ExprKind::Func(closure) => self.fold_function(*closure, span)?,
 
             ExprKind::Unary(UnaryExpr {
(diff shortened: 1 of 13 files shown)


crates/prql_compiler/src/semantic/ast_expand.rs
```diff
@@ -86,7 +64,47 @@ fn expand_expr_kind(value: ExprKind) -> pl::ExprKind {
...
+fn expand_exprs(exprs: Vec<prql_ast::expr::Expr>) -> Vec<pl::Expr> {
+    exprs.into_iter().map(expand_expr).collect::<Vec<_>>()
+}
+
+#[allow(clippy::boxed_local)]
+fn expand_expr_box(expr: Box<prql_ast::expr::Expr>) -> Box<pl::Expr> {
+    Box::new(expand_expr(*expr))
+}
+
+fn desugar_pipeline(mut pipeline: prql_ast::expr::Pipeline) -> pl::Expr {
```
not-my-profile: This function deserves some comment explaining what it does ... what the logic behind it is.

not-my-profile: Also the signature should probably be changed to return `pl::FuncCall` to make it clear what it gets desugared to.

aljazerzen: It cannot return `pl::FuncCall` because it needs to set the span and alias. I added a comment about it.

crates/prql-compiler/src/semantic/resolver/mod.rs
```diff
@@ -982,16 +958,89 @@ impl Resolver {
...
+    //             Func {
+    //                 params: [a, b],
+    //                 args: [a, b],
+    //                 body: arg_body
+    //             },
+    //             z
+    //         ],
+    //         body: parent_body
+    //     }
+    // }
```
not-my-profile: This should be a snapshot test rather than a comment.

aljazerzen: This is not something that compiles, it a note about what's happening. Quite verbose one.

aljazerzen: (this was part of the previous commit, that this one was based on)

not-my-profile: Huh I'm slightly confused why this showed up in the diff ... oh yeah this should probably say "Input (simplified)" and "Output (simplified)" ...  but yeah that's off-topic for this PR.

PR #3061, opened 2023-07-25
refactor: desugar BinaryExpr and UnaryExpr

aljazerzen:


not-my-profile:
Makes sense. Could this negatively effect our error messages? It might confuse users if some error speaks of `std.add` when they've used `+`.

aljazerzen:
> It might confuse users if some error speaks of std.add when they've used +.

True. The same goes for the pipelines. When we encounter this, I can add rules for converting `std.add` into `+` in `restrict_expr`.

PR #3065, opened 2023-07-25
build: Make packages releasable

max-sixty:
- Add back versions into `Cargo.toml`s, which are required for releasing
- Revert some hyphen / underscore changes (I should have reviewed these more closely, mea culpa)


max-sixty:
> Revert some hyphen / underscore changes (I should have reviewed these more closely, mea culpa)

I don't think making this change was a good tradeoff @not-my-profile — this is 100% on me for not reviewing well enough. I'm already reflecting on my failures of the past few days. 

For future changes — small aesthetic things that a) aren't reversible and b) create an inconsistency — crates would have used hyphens vs. underscores differently in their names — I would strongly vote to not change, at least without significant consensus.

(I hope this comes across appropriately, we're all excited you're helping so much, and you've already made the project better. And that this got merged was on me, not you. Lmk if this is a reasonable message...)

PR #3071, opened 2023-07-25
web: Try removing slash for playground data

max-sixty:
As discussed on Discord -- could this be causing the playground failing?


max-sixty:
image

image


(Maybe 20% probability, seems unlikely)

PR #3085, opened 2023-07-25
web: Fix links

max-sixty:


max-sixty:
@aljazerzen we probably should have done a more thorough search of these before merging!

eitsupi:
I think the lack of a redirect feature in mdBook is also causing links to break frequently.
In that regard I see merit in moving to something like Docusaurus.

max-sixty:
Yes — though mdbook does support them — but they have to be manually specified. 
Maybe docusaurus can do that automatically? 

(though for transparency — still concerned about a big change unless someone really wants to own the whole change...)

eitsupi:
> Yes — though mdbook does support them — but they have to be manually specified.
> Maybe docusaurus can do that automatically?

Oh, sorry. My lack of research. Perhaps all static site generators require us to set it up manually; if mdBook supports it, I think using it would alleviate problems like PRQL/prql-vscode#272.

aljazerzen:
Thanks for cleaning this up. Many links were broken just because the restructure has not been published yet and we have full URLs committed, so after a while I stopped searching.

Link redirection would help, yes.

PR #3086, opened 2023-07-25
ci: not require VCRUNTIME140.dll for to execute release version of prqlc.exe

### LINKED ISSUE #3080: prqlc.exe 0.9.0 requires VCRUNTIME140.dll on Windows
https://github.com/microsoft/winget-pkgs/pull/113133#issuecomment-1650410476

Maybe this issue is helpful. rust-lang/rust#100874

eitsupi:
Fix #3080


max-sixty:
Great! Thanks for finding this.

eitsupi:
It seems working.
I tested on Windows sandbox.

prqlc.exe 0.9.3

image


From this PR

image

eitsupi:
My concern is that adding the flag will increase the time it takes to build without the cache being used.
Do we need to limit the flag setting to release time only?

max-sixty:
> My concern is that adding the flag will increase the time it takes to build without the cache being used. Do we need to limit the flag setting to release time only?

Is it only required with `--release` mode? Otherwise why the difference between release an non-release?

One lesson from today is to minimize differences when releasing, since they don't get tested often, and it's a bad time to break things :)

eitsupi:
I am simply concerned with whether it is worth reducing the execution time of frequently running jobs.
We have already switched profiles at the time of release and at each PR run to make effective use of the cache. (#2639)

If there is no possibility that the addition of this flag will increase build time, then it is always good to add the flag.

max-sixty:
> My concern is that adding the flag will increase the time it takes to build without the cache being used.

I don't have context for the whole problem, so feel free to discount this — but one option is to put it in _all_ windows builds, so then the GHA build will be cached with it (but slow without it)

max-sixty:
> If there is no possibility that the addition of this flag will increase build time, then it is always good to add the flag.

Great (we both commented at the same time :) )

eitsupi:
> one option is to put it in _all_ windows builds

That case, we can add the following to the `.cargo/config.toml`. (https://github.com/rust-lang/rust/issues/100874#issuecomment-1222405985)

```toml
[target.'cfg(all(windows, target_env = "msvc"))']
rustflags = ["-C", "target-feature=+crt-static"]
```

However, since I am not sure how doing this would affect other builds, I thought it would be better to set the flag only on prqlc builds for now.

eitsupi:
I think it makes sense to set the flag only on release builds for now, so I merge this.
(The nightly CI failure has nothing to do with this change)

max-sixty:
> That case, we can add the following to the `.cargo/config.toml`. ([rust-lang/rust#100874 (comment)](https://github.com/rust-lang/rust/issues/100874#issuecomment-1222405985))

I had some trouble with doing this for Mac & Elixir — it means specifying the target explicitly with `--target=...` causes the cache to miss (super annoying!). 

But we could do for all Windows in GHA. Though that has a downside that we then have a different env in GHA vs local.

No perfect solution!!

PR #3088, opened 2023-07-26
docs: First cut draft of revised Tutorial - Relations, Filtering, Aggregation

richb-hanover:
This time for sure. 

* I finally have the `docs-tutorial` back in my own Github repo, so I can use my normal workflow.
* Changed all examples to use `invoices`, not `albums` because math makes more sense on prices than on `artist_id`.
* I used the term "passing through" to describe what a transform does. Is this the best terminology?
* I _think_ I know how this stuff works, but would be pleased to have someone correct this.

### DIFF
--- a/web/book/src/tutorial/aggregation.md
+++ b/web/book/src/tutorial/aggregation.md
@@ -1,95 +1,98 @@
 # Aggregation
 
 A key feature of analytics is reducing many values down to some summary. This
-act is called aggregation and always includes an aggregation function, for
-example `average` or `sum`:
+act is called "aggregation" and always includes a function &mdash; for
+example, `average` or `sum` &mdash; that reduces the table to a single row.
 
-<!-- aggregating ids does not make much sense, should we switch to some other table? -->
-
-```
-from albums
-aggregate {sum_of_ids = sum album_id}
-```
-
-`aggregate` is quite similar to `derive` and `select` as it is creating a new
-column, with the option to assign it a name. Just as select, it will discard all
-columns that are not present in the tuple, but with a key difference: it will
-always produce a single row.
-
-Just as `select` (and actually `derive` too), `aggregate` takes a tuple of
-aggregation expressions, which means that you can produce multiple summaries at
-once:
+**aggregate:** The `aggregate` transform takes a tuple to create one or many new columns
+that "distill down" data from all the rows.
 
 ```
-from albums
+from invoices
+aggregate { sum_of_orders = sum total }
+```
+In the example above, the `total` column of all rows of the `invoices` table
+is summed to produce a single value.
+
+`aggregate` can produce multiple summaries at once when
+one or more aggregation expressions are contained in a tuple.
+`aggregate` discards all columns that are not present in the tuple.
+
+```
+from invoices
 aggregate {
-    sum_of_ids = sum album_id,
-    mean_id = average album_id,
+    num_orders = count this,
+    sum_of_orders = sum total,
 }
 ```
 
-Here we've also split the tuple over multiple lines to make it more readable.
-Obviously, you can do this with `select` and `derive` too.
+In the example above, the result is a single row with two columns.
+The `count` function displays the number of rows in the table that was passed in;
+the `sum` function adds up the values of the `total` column of all rows.
 
 ## Grouping
 
-Say that now we want to produce summaries of albums _for each artist
-separately_. We cannot do this, as `aggregate` will always produce a single row.
+Suppose we want to produce summaries of invoices _for each city_
+in the table. 
+`aggregate` cannot do this because it will always produce a single row.
 
-What we can do, is separate the relation into groups corresponding to individual
-artists, and apply `aggregate` to each one of them:
-
-```
-from albums
-filter artist_id == 1
-aggregate {sum_of_ids = sum album_id}
-```
+We could separate the relation into groups corresponding to individual
+cities, and apply `aggregate` to each one of them:
 
 ```
 from albums
-filter artist_id == 2
-aggregate {sum_of_ids = sum album_id}
+filter billing_city == "Oslo"
+aggregate { sum_of_orders = sum total }
 ```
 
 ```
 from albums
-filter artist_id == 3
-aggregate {sum_of_ids = sum album_id}
+filter billing_city == "Frankfurt"
+aggregate { sum_of_orders = sum total }
 ```
-
-... and so on. Of course this is not the way to go, because it is way too
-repetitive and we need to type all of the artist ids by hand. Luckily, `group`
-will do exactly what we want: separate relation into groups, apply a function to
-each of the groups and combine the results back together:
 
 ```
 from albums
-group artist_id (
-    aggregate {sum_of_ids = sum album_id}
+filter billing_city == "London"
+aggregate { sum_of_orders = sum total }
+```
+
+... and so on. Of course this is repetitive (and boring) and error prone (because we would need to type each `billing_city` by hand). Moreover, we would need to create the `billing_city` list before we started.
+
+**group:** The `group` transform separates the table into groups (say, those having the same city)
+using information that's already in the table.
+It then applies a transform to each group, and combines the results back together:
+
+```
+from invoices
+group billing_city (
+    aggregate {
+        num_orders = count this,
+        sum_of_orders = sum total,
+    }
 )
 ```
 
-People who know how this is done in SQL would probably noticed that we just
-decoupled aggregation from grouping. These two very connected operations benefit
-immensely from each being a standalone function. Firstly, this allows each of
-them to have some invariants that the query engine can leverage to produce more
-efficient query planes. Additionally, they can be used with other transform
-functions:
+People who know how this is done in SQL will probably have noticed that we just
+decoupled aggregation from grouping. These two very connected operations (in SQL) benefit
+immensely from each being a standalone function. 
+
+Firstly, each can have invariants that the query engine can 
+leverage to produce more efficient queries. 
+Additionally, they can be used with other transform functions, such as:
 
 ```
-from albums
-group artist_id (
+from invoices
+group billing_city (
     take 2
 )
 ```
 
-We can derive what this code chunk is doing using the same logic use used
-before: `group` splits the relation into chunks each of which is corresponding
-to some `artist_id`. Then it applies function `take 2`, which will take first
-two rows of the chunk, and at the end it combines the rows back together.
+This code collects the first two rows for each city's `group`. 
 
-SQL needed to replicate this behavior includes window functions and multiple
-sub-queries. Some dialects (PostgreSQL, DuckDB, Google BigQuery) have a special
-syntax to improve performance and reduce the query complexity with
-`DISTINCT ON`. This approach only works when we have `take 1` instead of
-`take 2`, so it's not a general case solution.
+The SQL needed to replicate this behavior might include window functions and
+sub-queries. PRQL handles all the complexity.
+Some dialects (PostgreSQL, DuckDB, Google BigQuery) have a special
+syntax to improve performance and reduce the query complexity
+(for example ,`DISTINCT ON` if the query uses `take 1`).
+But it's not a general solution for the case of `take 2`.
--- a/web/book/src/tutorial/aggregation.md
+++ b/web/book/src/tutorial/aggregation.md
@@ -33,7 +33,7 @@ the `sum` function adds up the values of the `total` column of all rows.
 ## Grouping
 
 Suppose we want to produce summaries of invoices _for each city_
-in the table. 
+in the table.
 `aggregate` cannot do this because it will always produce a single row.
 
 We could separate the relation into groups corresponding to individual
@@ -75,10 +75,10 @@ group billing_city (
 
 People who know how this is done in SQL will probably have noticed that we just
 decoupled aggregation from grouping. These two very connected operations (in SQL) benefit
-immensely from each being a standalone function. 
+immensely from each being a standalone function.
 
-Firstly, each can have invariants that the query engine can 
-leverage to produce more efficient queries. 
+Firstly, each can have invariants that the query engine can
+leverage to produce more efficient queries.
 Additionally, they can be used with other transform functions, such as:
 
 ```
@@ -88,7 +88,7 @@ group billing_city (
 )
 ```
 
-This code collects the first two rows for each city's `group`. 
+This code collects the first two rows for each city's `group`.
 
 The SQL needed to replicate this behavior might include window functions and
 sub-queries. PRQL handles all the complexity.
--- a/web/book/src/tutorial/aggregation.md
+++ b/web/book/src/tutorial/aggregation.md
@@ -4,15 +4,16 @@ A key feature of analytics is reducing many values down to some summary. This
 act is called "aggregation" and always includes a function &mdash; for
 example, `average` or `sum` &mdash; that reduces the table to a single row.
 
-**aggregate:** The `aggregate` transform takes a tuple to create one or more new columns
+### `aggregate` transform
+
+The `aggregate` transform takes a tuple to create one or more new columns
 that "distill down" data from all the rows.
 
 ```
 from invoices
 aggregate { sum_of_orders = sum total }
 ```
-In the example above, the `total` column of all rows of the `invoices` table
-is summed to produce a single value.
+The query above computes the sum of the `total` column of all rows of the `invoices` table to produce a single value.
 
 `aggregate` can produce multiple summaries at once when
 one or more aggregation expressions are contained in a tuple.
@@ -34,10 +35,7 @@ the `sum` function adds up the values of the `total` column of all rows.
 
 Suppose we want to produce summaries of invoices _for each city_
 in the table.
-`aggregate` cannot do this because it will always produce a single row.
-
-We could separate the relation into groups corresponding to individual
-cities, and apply `aggregate` to each one of them:
+We could create a query for each city, and aggregate its rows:
 
 ```
 from albums
@@ -45,21 +43,11 @@ filter billing_city == "Oslo"
 aggregate { sum_of_orders = sum total }
 ```
 
-```
-from albums
-filter billing_city == "Frankfurt"
-aggregate { sum_of_orders = sum total }
-```
+But we would need to do it for each city: `London`, `Frankfurt`, etc. Of course this is repetitive (and boring) and error prone (because we would need to type each `billing_city` by hand). Moreover, we would need to create a list of each `billing_city` before we started.
 
-```
-from albums
-filter billing_city == "London"
-aggregate { sum_of_orders = sum total }
-```
+### `group` transform
 
-... and so on. Of course this is repetitive (and boring) and error prone (because we would need to type each `billing_city` by hand). Moreover, we would need to create the `billing_city` list before we started.
-
-**group:** The `group` transform separates the table into groups (say, those having the same city)
+The `group` transform separates the table into groups (say, those having the same city)
 using information that's already in the table.
 It then applies a transform to each group, and combines the results back together:
 
@@ -74,7 +62,8 @@ group billing_city (
 ```
 
 Those familiar with SQL have probably noticed that we just
-decoupled aggregation from grouping. These connected operations (in SQL) benefit
+decoupled aggregation from grouping.
+These connected operations (in SQL) benefit
 from each being a standalone function.
 
 Firstly, each can have invariants that the query engine can
@@ -89,6 +78,10 @@ group billing_city (
 ```
 
 This code collects the first two rows for each city's `group`.
+
+### Move to Reference?
+
+_The following items might be better moved to the Reference section_
 
 The SQL needed to replicate this behavior might include window functions and
 sub-queries. PRQL handles all the complexity.
--- a/web/book/src/tutorial/aggregation.md
+++ b/web/book/src/tutorial/aggregation.md
@@ -4,7 +4,7 @@ A key feature of analytics is reducing many values down to some summary. This
 act is called "aggregation" and always includes a function &mdash; for
 example, `average` or `sum` &mdash; that reduces the table to a single row.
 
-**aggregate:** The `aggregate` transform takes a tuple to create one or many new columns
+**aggregate:** The `aggregate` transform takes a tuple to create one or more new columns
 that "distill down" data from all the rows.
 
 ```
--- a/web/book/src/tutorial/aggregation.md
+++ b/web/book/src/tutorial/aggregation.md
@@ -75,7 +75,7 @@ group billing_city (
 
 People who know how this is done in SQL will probably have noticed that we just
 decoupled aggregation from grouping. These two very connected operations (in SQL) benefit
-immensely from each being a standalone function.
+from each being a standalone function.
 
 Firstly, each can have invariants that the query engine can
 leverage to produce more efficient queries.
--- a/web/book/src/tutorial/aggregation.md
+++ b/web/book/src/tutorial/aggregation.md
@@ -73,8 +73,8 @@ group billing_city (
 )
 ```
 
-People who know how this is done in SQL will probably have noticed that we just
-decoupled aggregation from grouping. These two very connected operations (in SQL) benefit
+Those familiar with SQL have probably noticed that we just
+decoupled aggregation from grouping. These connected operations (in SQL) benefit
 from each being a standalone function.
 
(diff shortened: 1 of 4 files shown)


web/book/src/tutorial/aggregation.md
```diff
@@ -1,95 +1,98 @@
...
 
-<!-- aggregating ids does not make much sense, should we switch to some other table? -->
+**aggregate:** The `aggregate` transform takes a tuple to create one or many new columns
+that "distill down" data from all the rows.
 
 ```
-from albums
-aggregate {sum_of_ids = sum album_id}
+from invoices
+aggregate { sum_of_orders = sum total }
```
max-sixty, line 14: Much better!

web/book/src/tutorial/aggregation.md
```diff
@@ -1,95 +1,98 @@
 # Aggregation
 
 A key feature of analytics is reducing many values down to some summary. This
-act is called aggregation and always includes an aggregation function, for
-example `average` or `sum`:
+act is called "aggregation" and always includes a function &mdash; for
+example, `average` or `sum` &mdash; that reduces the table to a single row.
 
-<!-- aggregating ids does not make much sense, should we switch to some other table? -->
+**aggregate:** The `aggregate` transform takes a tuple to create one or many new columns
```
max-sixty:
```suggestion
**aggregate:** The `aggregate` transform takes a tuple to create one or more new columns
```

web/book/src/tutorial/aggregation.md
```diff
@@ -1,95 +1,98 @@
 # Aggregation
 
 A key feature of analytics is reducing many values down to some summary. This
-act is called aggregation and always includes an aggregation function, for
-example `average` or `sum`:
+act is called "aggregation" and always includes a function &mdash; for
```
max-sixty, line 4: Can we just use a dash in markdown?

```suggestion
act is called "aggregation" and always includes a function — for
```

web/book/src/tutorial/aggregation.md
max-sixty, line 14:
```suggestion
aggregate { sum_of_orders = sum amount }
```

using `total` is not the clearest name for the thing being summed...

web/book/src/tutorial/aggregation.md
```diff
@@ -1,95 +1,98 @@
...
+The `count` function displays the number of rows in the table that was passed in;
+the `sum` function adds up the values of the `total` column of all rows.
 
 ## Grouping
 
-Say that now we want to produce summaries of albums _for each artist
-separately_. We cannot do this, as `aggregate` will always produce a single row.
+Suppose we want to produce summaries of invoices _for each city_
+in the table.
+`aggregate` cannot do this because it will always produce a single row.
```
max-sixty: This could be rephrased away from the negative (particularly since it's not exactly correct — it can do it — it just requires a `group`

web/book/src/tutorial/aggregation.md
```diff
@@ -1,95 +1,98 @@
...
-aggregate {sum_of_ids = sum album_id}
+filter billing_city == "London"
+aggregate { sum_of_orders = sum total }
 ```
 
-... and so on. Of course this is not the way to go, because it is way too
-repetitive and we need to type all of the artist ids by hand. Luckily, `group`
-will do exactly what we want: separate relation into groups, apply a function to
-each of the groups and combine the results back together:
+... and so on. Of course this is repetitive (and boring) and error prone (because we would need to type each `billing_city` by hand). Moreover, we would need to create the `billing_city` list before we started.
```
max-sixty: Not a verdict — feel free to ignore — but I think this is very laborious given the average reader — they probably know about group-by from SQL, or at least the concept of a pivot table in Excel.

So I would skip this explanation and just jump into the concept of grouping...

web/book/src/tutorial/aggregation.md
```diff
@@ -1,95 +1,98 @@
...
 
-People who know how this is done in SQL would probably noticed that we just
-decoupled aggregation from grouping. These two very connected operations benefit
-immensely from each being a standalone function. Firstly, this allows each of
-them to have some invariants that the query engine can leverage to produce more
-efficient query planes. Additionally, they can be used with other transform
-functions:
+People who know how this is done in SQL will probably have noticed that we just
+decoupled aggregation from grouping. These two very connected operations (in SQL) benefit
+immensely from each being a standalone function.
```
max-sixty:
```suggestion
from each being a standalone function.
```

("show, don't tell")

web/book/src/tutorial/aggregation.md
```diff
@@ -1,95 +1,98 @@
...
 ```
 
-People who know how this is done in SQL would probably noticed that we just
-decoupled aggregation from grouping. These two very connected operations benefit
-immensely from each being a standalone function. Firstly, this allows each of
-them to have some invariants that the query engine can leverage to produce more
-efficient query planes. Additionally, they can be used with other transform
-functions:
+People who know how this is done in SQL will probably have noticed that we just
+decoupled aggregation from grouping. These two very connected operations (in SQL) benefit
```
max-sixty:
```suggestion
Those familiar with SQL have probably noticed that we just
decoupled aggregation from grouping. These connected operations (in SQL) benefit
```

(brevity)

web/book/src/tutorial/aggregation.md
```diff
@@ -1,95 +1,98 @@
...
+group billing_city (
     take 2
 )
 ```
 
-We can derive what this code chunk is doing using the same logic use used
-before: `group` splits the relation into chunks each of which is corresponding
-to some `artist_id`. Then it applies function `take 2`, which will take first
-two rows of the chunk, and at the end it combines the rows back together.
+This code collects the first two rows for each city's `group`.
```
max-sixty, line 80:
```suggestion
This code collects the first two rows for each `billing_city`.
```

Is that still right?

web/book/src/tutorial/aggregation.md
```diff
@@ -1,95 +1,98 @@
...
-sub-queries. Some dialects (PostgreSQL, DuckDB, Google BigQuery) have a special
-syntax to improve performance and reduce the query complexity with
-`DISTINCT ON`. This approach only works when we have `take 1` instead of
-`take 2`, so it's not a general case solution.
+The SQL needed to replicate this behavior might include window functions and
+sub-queries. PRQL handles all the complexity.
+Some dialects (PostgreSQL, DuckDB, Google BigQuery) have a special
+syntax to improve performance and reduce the query complexity
+(for example ,`DISTINCT ON` if the query uses `take 1`).
+But it's not a general solution for the case of `take 2`.
```
max-sixty, line 91: I think this is great but probably belongs in the reference — I liked the guide that @aljazerzen originally posted (which I now can't find...) about the relative uses of each type of guide

web/book/src/tutorial/filtering.md
```diff
@@ -1,74 +1,73 @@
...
-> orthogonality". It's goal is to keep transform functions composable by
-> minimizing interference of their effects.
->
-> In SQL, you can see this lack of invariant when an aggregation function is
-> used in the `SELECT` clause. Before, the number of rows was kept constant, but
-> introduction of an aggregation function caused the whole statement to produce
-> only one row (per group).
-
-Filter pick rows based on their values:
+**filter:** The `filter` transform picks (passes through) rows based on their values:
```
max-sixty:
```suggestion
**filter:** The `filter` transform picks rows based on their values:
```

I'm not sure `picks` is great either, but `passes through` seems not great, given its role is to not pass things through? "selectively passes through"??

web/book/src/tutorial/filtering.md
```diff
@@ -1,74 +1,73 @@
...
-artist with id 2.
+The resulting table contains all the rows that came from Berlin.
 
-In SQL, `filter` would be expressed with either `WHERE` or `HAVING`, depending
-on where in the pipeline it is used.
+PRQL converts the single `filter` transform to use the SQL `WHERE` or `HAVING` command,
+depending on where it appears in the pipeline it is used.
 
----
+**take:** The `take` transform passes through rows based on their position within the relation (table).
```
max-sixty:
```suggestion
**take:** The `take` transform selects rows based on their position within the relation (table).
```
?

web/book/src/tutorial/filtering.md
```diff
@@ -1,74 +1,73 @@
...
 
-# skip first 10 rows and take only the next 10
+# skip first 10 rows and take the next 10
 take 11..20
 
-# take only first 5 rows
-take 5
+# take only first 3 rows of *that* result
+take 3
 ```
```
max-sixty, line 55: Nice!

web/book/src/tutorial/filtering.md
```diff
@@ -1,74 +1,73 @@
...
-transform `take 11..15`. When compiled to SQL this two ways of expressing the
-same thing actually produce an identical query. This is a nice example of how
-PRQL allows fast data exploration, optimizing your queries so you have the
+We did something a bit odd at the end: first we took rows `11..20` and
+then took the first 3 rows from that result.
+Although this could have been done in a single
+transform `take 11..13`, this is a nice example of how
+PRQL allows fast data exploration.
+
+You have the
```
max-sixty: I think we should try and keep a consistent voice, and one way I've found to do that is to never use "You", only use a descriptive voice...

web/book/src/tutorial/filtering.md
```diff
@@ -1,74 +1,73 @@
...
+> Side note: each of the transform functions has at least some invariant: `select`
+> and `derive` will not change the number of rows, `filter` and `take` will not change
+> the number of columns. We call this separation of effects "transform
+> orthogonality". Its goal is to keep transform functions composable by
+> minimizing interference of their effects.
+>
+> In SQL, you can see this lack of invariant when an aggregation function is
+> used in the `SELECT` clause. Before, the number of rows was kept constant, but
+> introduction of an aggregation function caused the whole statement to produce
+> only one row (per group). _I don't understand the importance of this paragraph -richb-hanover_
```
max-sixty: I kinda agree with @richb-hanover — is this good for the tutorial?

web/book/src/tutorial/filtering.md
```diff
@@ -1,74 +1,73 @@
...
 freedom to stack transforms on top of each other, without worrying about
 interactions of a new transform with the previous query.
+When PRQL compiles these two statements/transforms
+(`take 11..20` and `take 3`) to SQL, it produces the same result as `take 11..13`.
+
+> Side note: each of the transform functions has at least some invariant: `select`
+> and `derive` will not change the number of rows, `filter` and `take` will not change
+> the number of columns. We call this separation of effects "transform
+> orthogonality". Its goal is to keep transform functions composable by
+> minimizing interference of their effects.
```
max-sixty, line 77: If we do want to keep this (I might move to reference), then we can put in an admonish block

web/book/src/tutorial/relations.md
```diff
@@ -1,78 +1,110 @@
...
-> Side note: most definitions use unordered relations, which cannot have
-> duplicate rows. PRQL defines relations to have an order and to be able to
-> contain duplicate rows.
-
-PRQL is designed to query relations such as the `albums` table above. The most
-basic query would be:
+PRQL is designed on top of _relational algebra_, which is the established data
+model used by modern SQL databases.
+A _relation_ has a rigid mathematical definition,
+which can be simplified to "a table of data".
```
max-sixty, line 6: Could we reduce the level of technical terms for the tutorial?

I get that we like the term "relation" and everyone else thinks it's "table" — if we do want to say "relation" rather than "table", would just saying "think of a relation as a table" be enough?

web/book/src/tutorial/relations.md
```diff
@@ -1,78 +1,110 @@
...
+| 4        | 14 | Edmonton  | ...         | 8.91 |
+| 5        | 23 | Boston    | ...         | 13.86 |
+| 6        | 37 | Frankfurt | ...         | 0.99 |
+
+A table (relation) is composed of columns, each of which has an unique name and a designated data type.
+Every table has zero to many rows, each containing the same set of columns.
+The table above has `invoice_id` and `customer_id` columns with a data type of "integer number",
+a `billing_city` column with a data type of "text",
+a number of other columns, and
+and a `total` column that contains floating-point number. [^2]
```
max-sixty:
```suggestion
and a `total` column that contains floating-point numbers. [^2]
```

web/book/src/tutorial/relations.md
```diff
@@ -1,78 +1,110 @@
...
+PRQL is designed to build queries that combine and select data from relations such as the `invoices` table above. Here is the most basic query:
 
 ```
-from albums
+from invoices
 ```
 
-This is not impressive, it is just the same table as we had before. But we can
-add on to it:
+_Note: You can try each of these examples here in the [Playground.](https://prql-lang.org/playground/)
```
max-sixty: Can we remove "You"?

web/book/src/tutorial/relations.md
```diff
@@ -1,78 +1,110 @@
...
+Enter the query on the left-hand side,
+and click **output.arrow** in the right-hand side to see the result._
+
+The result of the query above is not terribly interesting, it's just the same table as before.
+But PRQL can do more...
+
```
max-sixty, line 40:
```suggestion
```

web/book/src/tutorial/relations.md
```diff
@@ -1,78 +1,110 @@
...
+or can be computed from the value of an existing column.
+Note that the name of the new column is the expression that defined it.
+To give that column an interesting name,
+assign the expression to a name `value_added_tax`.
 
 ```
-from albums
-derive my_column = artist_id + 1
+from invoices
+derive { value_added_tax = total * 0.19 }
```
max-sixty: I think we can just immediately give it a name; no need to have different examples?

web/book/src/tutorial/relations.md
```diff
@@ -1,78 +1,110 @@
...
 ```
 
-Now let's say that we don't need the original `artist_id` column and would want
-to remove it. We can use select:
+**select:** PRQL can also _remove_ columns from a table.
```
max-sixty:
```suggestion
**select:** PRQL can also select specific columns from a table.
```

I don't really see `select` as removing, it's more like you're narrowing to the things you want. But I realize `select` is in the definition, which isn't great

web/book/src/tutorial/relations.md
```diff
@@ -1,78 +1,110 @@
...
+The `select` statement passes along only the columns named in a list
+and discards all others.
+Formally, that list is a _tuple_ of comma-separated expressions wrapped in `{ ... }`.
+You can write the items in the tuple on one or several lines:
+trailing commas are ignored.
+Finally, you can assign any of the expressions to a _variable_
+that becomes the name of the resulting column in the SQL output.
+
+Suppose you only need the `order_id`, `total`, and `value_added_tax`.
```
max-sixty:
```suggestion
Suppose we only need the `order_id`, `total`, and `value_added_tax`.
```

web/book/src/tutorial/relations.md
```diff
@@ -1,78 +1,110 @@
...
+select {
+  OrderID = invoice_id,
+  Total = total,
+  VAT = value_added_tax,
+}
 ```
 
-`select` will take a tuple of columns and discard all others. Note that it can
-also contain expressions, similar to derive:
+Once you `select` certain columns, subsequent transforms will have access only to those columns named in the tuple.
```
max-sixty:
```suggestion
Once we `select` certain columns, subsequent transforms will have access only to those columns named in the tuple.
```

web/book/src/tutorial/relations.md
```diff
@@ -1,78 +1,110 @@
...
-> duplicate rows. PRQL defines relations to have an order and to be able to
-> contain duplicate rows.
-
-PRQL is designed to query relations such as the `albums` table above. The most
-basic query would be:
+PRQL is designed on top of _relational algebra_, which is the established data
+model used by modern SQL databases.
+A _relation_ has a rigid mathematical definition,
+which can be simplified to "a table of data".
+For example, the `invoices` table from the Chinook database[^1] looks like this:
```
max-sixty: I don't think the book supports these kinds of footnotes...

web/book/src/tutorial/aggregation.md
max-sixty: Could we replace these inline emphases with more minor headings? Or you like the current style?

max-sixty (approved):
Overall this is a big improvement, thank you @richb-hanover 

I made lots of comments, I hope that's OK.

Could we link to the pages in the Reference (e.g. https://prql-lang.org/book/reference/stdlib/transforms/select.html)

I also think we could merge the definitions of each transform between the tutorial and the reference — e.g. `select` has "Picks and computes columns.". I find that pretty good? But if there's something better, we could update in both places?

web/book/src/tutorial/aggregation.md
aljazerzen: `total` is a column in `chinook/invoices`

Before the tutorial merges, I'd like to have actual tables with chinook data in the book, which would add a lot to understanding of PRQL.

web/book/src/tutorial/aggregation.md
aljazerzen: https://github.com/PRQL/prql/blob/db9ba055721c39dcc73e050736ccfbd9ad718f0e/web/book/src/overview.md

web/book/src/tutorial/filtering.md
aljazerzen: It's half way between design justification and something that it is nice to know so you can rely on it.

Maybe it should be made more of the latter.

And yes, admonish.

web/book/src/tutorial/filtering.md
aljazerzen: No, it's a bash at SQL :D And definitely a design justification -> to the reference.

web/book/src/tutorial/relations.md
aljazerzen: Yeah that's a good idea

web/book/src/tutorial/aggregation.md
richb-hanover: I don't think Markdown will emit an em dash (which is typographically more correct) for a bare "-" in the source. I'm OK either way - we can add the decision to our style guide. Let's leave &mdash; there as a reminder

web/book/src/tutorial/aggregation.md
richb-hanover: @aljazerzen is correct - the `invoices` table has a `total` column. Not my favorite name, either, but I wanted the examples to work in the playground.

web/book/src/tutorial/relations.md
richb-hanover: I'm not really fond of "we" or "you" in any of this... In another pass over it, I'll look at what might make it better. Accepting the change to "we"

web/book/src/tutorial/aggregation.md
richb-hanover: I agree. I'll rephrase as "We could have a series of aggregates - one for each city..."

web/book/src/tutorial/aggregation.md
richb-hanover: For newcomers, I think it's important to motivate the notion of grouping. (The experts will only be reading the Tutorial to see if they're smarter than we are :-) 

That said, I can trim it down to one example PRQL query and then describe London and Oslo, etc.

web/book/src/tutorial/aggregation.md
richb-hanover: Yes, a lot of the side notes may not belong in the Tutorial.  We can evolve their location as we iterate on this.

web/book/src/tutorial/filtering.md
richb-hanover: I'll change it to say, "... picks the rows to pass through..." in this first description, then use "picks" thereafter. Same for `take`

web/book/src/tutorial/filtering.md
richb-hanover: same comment as above

web/book/src/tutorial/filtering.md
richb-hanover: Tricky using the data from the `invoices` table. The largest number of rows for any city is 14. So `take 11..20` really only gets 14 rows...

web/book/src/tutorial/filtering.md
richb-hanover: Busted! :-) Thanks.

web/book/src/tutorial/filtering.md
richb-hanover: OK by me :-)

web/book/src/tutorial/filtering.md
richb-hanover: Yes, let's move to the reference section

web/book/src/tutorial/relations.md
richb-hanover: This needs to go into our Style Guide (or some place where we agree on this stuff.) What do we call these things: Table? Relation? Is there any difference? What do practitioners say?

It'll be easy enough to go through the docs harmonizing the choice once we decide.

web/book/src/tutorial/relations.md
richb-hanover: Yes

web/book/src/tutorial/relations.md
richb-hanover: I wanted to motivate the notions that a) "just type it in the Playground" and b) a bare `from ....` transform generates a sensible, human-readable result.

aljazerzen:
@richb-hanover when you're done merge into docs-tutorial so I can squash it. The merges there are getting out of hand :D

web/book/src/tutorial/relations.md
richb-hanover: I'll need to see whether this precedes the _tuple_ definition or not... (That explains about giving results a name)

web/book/src/tutorial/relations.md
richb-hanover: I'll try describing it as "picking the columns to pass through" (just as filter picks rows...)

web/book/src/tutorial/relations.md
richb-hanover: Dang. I'll write it out. (and many of the other footnotes may migrate to the reference)

web/book/src/tutorial/aggregation.md
richb-hanover: I went back and forth. I'll switch all to a minor heading to see how they look

richb-hanover:
@aljazerzen @max-sixty  - Thanks for all these comments.

> @richb-hanover when you're done merge into docs-tutorial so I can squash it. The merges there are getting out of hand :D

That's nothing. You should have seen the commit history on my local machine!

web/book/src/tutorial/aggregation.md
richb-hanover: I think so - that's how it looks in the Playground

web/book/src/tutorial/aggregation.md
max-sixty: FYI on Mac this is just `alt+shift+-`: —

web/book/src/tutorial/aggregation.md
max-sixty: Sorry, my comment was unclear. I just meant: we can "for each city" rather than "for each city's group"

max-sixty:
This is really great!

My only tweak would be to not use `total` as the column being summed — it means "it's the total" means several things. But possibly there's not that much we can do with the sample data.

richb-hanover:
> My only tweak would be to not use total as the column being summed

Yeah, but there's nothing for it. `invoices` is the only table in the Chinook database with "interesting" data, and the `total` column is the only one that really makes sense to aggregate. I, too, wish they had chosen a better name for that column. 

I think the current description and variable name (_sum\_of\_orders_) are good enough. Let's commit what we have so others can give it a good going over. 

PS And what's the deal with the _pre-commit.ci_ failing with Prettier...???

max-sixty:
> Yeah, but there's nothing for it. `invoices` is the only table in the Chinook database with "interesting" data, and the `total` column is the only one that really makes sense to aggregate. I, too, wish they had chosen a better name for that column.

OK, good point



> PS And what's the deal with the _pre-commit.ci_ failing with Prettier...???

This branch needs to merge in `main`...

max-sixty:
I'm hitting the button! Thank you @richb-hanover !

richb-hanover:
Thanks!

PR #3091, opened 2023-07-26
ci: update iFaxity/wait-on-action to use node.js 16

eitsupi:
v1 tag seems outdated (iFaxity/wait-on-action#10)

### DIFF
--- a/.github/workflows/test-rust.yaml
+++ b/.github/workflows/test-rust.yaml
@@ -86,7 +86,7 @@ jobs:
         working-directory: ./crates/prql-compiler/tests/integration
         if: ${{ inputs.target == 'x86_64-unknown-linux-gnu' }}
       - name: Wait for database
-        uses: ifaxity/wait-on-action@v1
+        uses: ifaxity/wait-on-action@v1.1.0
         with:
           resource: "tcp:1433 tcp:3306 tcp:5432 tcp:9004"
           timeout: 60000


.github/workflows/test-rust.yaml
```diff
@@ -86,7 +86,7 @@ jobs:
         working-directory: ./crates/prql-compiler/tests/integration
         if: ${{ inputs.target == 'x86_64-unknown-linux-gnu' }}
       - name: Wait for database
-        uses: ifaxity/wait-on-action@v1
+        uses: ifaxity/wait-on-action@v1.1.0
```
max-sixty, line 89: Doesn't `v1` resolve to the latest within `v1`?

eitsupi: Not resolved.
This actually refers to a tag in the Git repository, and the v1 and v1.1.0 tags are tied to different commits because the v1 tag has not been updated.

max-sixty: Ah interesting, TIL.

I wonder why dependabot didn't upgrade, in that case...

max-sixty: Interestingly GH recommends that actions change their `v1` tag to the latest `v1` release: https://github.com/actions/toolkit/blob/main/docs/action-versioning.md...

eitsupi: Good point.
I think Dependabot has judged `v1` is greater than `v1.1.0`...

max-sixty: > I think Dependabot has judged `v1` is greater than `v1.1.0`...

😬

max-sixty: Oh, and actually that's what the issue you linked says...

PR #3093, opened 2023-07-26
docs: Add project update

max-sixty:
I let this slide for too many months (others should feel free to update / ping me to update!)

Is this a good update? Is this what we're focusing on? We (/ I) have let the focus on tools to _use_ PRQL drop, which we had planned to prioritize. But we've also done a great job on closing bugs -- we actually have zero Priority bugs now (thanks to @aljazerzen ...)


eitsupi:
Is it worth to mention further breaking changes maybe done in near future? #2723

max-sixty:
> Is it worth to mention further breaking changes maybe done in near future? #2723

I think if it's an area of focus for the project, it's interesting to include, so we could include this — @aljazerzen ? 

I don't think we need to forecast breaking changes; this is more like "so what are you working on?"

aljazerzen:
Jup, I agree, that's interesting for people to see.

README.md
```diff
@@ -62,25 +62,30 @@ For more on the language, more examples & comparisons with SQL, visit
...
-## Current Status - April 2023
+## Current Status - August 2023
 
 PRQL is being actively developed by a growing community. It's ready to use by
 the intrepid, either with our supported integrations, or within your own tools,
 using one of our supported language bindings.
 
-PRQL still has some minor bugs and some missing features, and probably is only
-ready to be rolled out to non-technical teams for fairly simple queries.
+PRQL still has some infrequent minor bugs and some missing features, and
```
aljazerzen: > infrequent minor bug

Sounds a bit defensive and also not entirely true - some bugs are quite major :D

PR #3102, opened 2023-07-26
fix: playground csv loading, take 2

aljazerzen:
Followup for #3076 

I'm surprised that yesterdays fix even worked in any way - I wasn't able to test if fully because prql-js was having some wasm problems.

This one is tested.


max-sixty:
Thanks a lot 😀

PR #3106, opened 2023-07-26
feat: cli no-format

### LINKED ISSUE #3101: `hide_signature_comment` doesn't seem to work in some circumstances
### What's up?

Here's an example: https://github.com/PRQL/prql/blob/9a835b66c62c1856c21aec561433f6a9dfe344f9/crates/prqlc/src/cli.rs#L590-L624

This causes bumping the version numbers to fail tests

aljazerzen:
Also closes #3101

`Compile::hide_signature_comment`  was badly named, as it
has inverted logic applied.


max-sixty:
Very much agree on negation naming!

PR #3107, opened 2023-07-26
ci: get rid of pre-release-replacements for GitHub workflows

not-my-profile:
This gets rid of the pitfall of forgetting to add a pre-release-replacements when adding new workflows or actions. (ref 53ada61b29941a2ff3800390aacc466ad47b978e)


max-sixty:
Ok yes this is better; good idea

max-sixty (approved):

Issue #3096, opened 2023-07-26
Unify GHA linux cache


max-sixty:
### What's up?

Currently `prqlc` is built on every PR with `-musl`: https://github.com/PRQL/prql/actions/runs/5663989436/job/15346734099?pr=3093#step:3:42

But the linux GHA cache is build with `-gnu`: https://github.com/PRQL/prql/actions/runs/5663989436/job/15346734529?pr=3093#step:7:4

So the prqlc cache misses every time.

Is it possible for us to use the same one for both?

(Tagging @eitsupi because they seem to know this area, hope that's OK)

eitsupi:
Isn't the problem that we are setting `save-if: false`?

https://github.com/PRQL/prql/blob/975ffa505dcc1c0f1ddb94a876c201b156d7dd10/.github/actions/build-prqlc/action.yaml#L22-L27

eitsupi:
As an aside, I don't know what the `prefix-key` does for us.
If I understand correctly, the cache is currently only updated from the main branch, so there is no problem if this does not exist.

max-sixty:
> Isn't the problem that we are setting `save-if: false`?

We only have one cache per OS, because we're out of cache space.

If we enabled saving from the `prqlc` job, we'd need a different `shared-key` from the `test-rust` job — or we'd get a race condition the OSs where they're the same — and we'd overflow our cache space.

eitsupi:
> If we enabled saving from the `prqlc` job, we'd need a different `shared-key` from the `test-rust` job — or we'd get a race condition the OSs where they're the same — and we'd overflow our cache space.

In other words, if we only enable the cache for `inputs.target == 'rust-x86_64-unknown-linux-musl'`, which is not tested, the problem would be solved.

max-sixty:
> As an aside, I don't know what the `prefix-key` does for us.

The cache isn't perfect at detecting what is / isn't required, and so after a while it can accrue some files which we don't use. We were hitting 3GB caches at one point.

Cycling a cache key occasionally starts afresh, solving this problem

eitsupi:
> The cache isn't perfect at detecting what is / isn't required, and so after a while it can accrue some files which we don't use. We were hitting 3GB caches at one point.

Oh, I see. Thanks.

max-sixty:
> In other words, if we only enable the cache for `inputs.target == 'rust-x86_64-unknown-linux-musl'`, which is not tested, the problem would be solved.

[...]

PR #3117, opened 2023-07-26
ci: Build workspace in `prqlc` build

max-sixty:
As noted inline, I think this will be much faster, ironically


max-sixty:
Unfortunately hit the target issue in https://github.com/PRQL/prql/issues/3096...

PR #3118, opened 2023-07-26
ci: Remove quotes from cache key

max-sixty:


.github/set_version.sh
```diff
@@ -2,4 +2,4 @@
 # since the caches seem to accumulate cruft over time;
 # ref https://github.com/PRQL/prql/pull/2407
 
-grep '^version =' Cargo.toml | tr -d ' ' >> $GITHUB_ENV
+grep '^version =' Cargo.toml | cut -d\" -f2
```
max-sixty:
```suggestion
grep '^version =' Cargo.toml | cut -d\" -f2  >> $GITHUB_ENV
```

max-sixty:
No luck with this — though works locally — is there another way we can remove the quotes @not-my-profile ?

not-my-profile:
How did you observe that the quotes are in the cache key? They actually shouldn't be since in `foo="bar"` the quotation marks are part of the shell syntax.

not-my-profile:
Ah, nvm, I see it.

max-sixty:
The caches are at https://github.com/PRQL/prql/actions/caches

not-my-profile:
Closed in favor of #3122.

PR #3119, opened 2023-07-26
ci: Fix cache version on Windows

max-sixty:
The existing code doesn't seem to be working correctly -- maybe windows isn't running the script correctly? https://github.com/PRQL/prql/actions/runs/5673273615/job/15374368488#step:8:60

Let's try this.

### DIFF
--- a/.github/workflows/test-all.yaml
+++ b/.github/workflows/test-all.yaml
@@ -103,6 +103,7 @@ jobs:
         with:
           crate: cargo-tarpaulin
       - run: ./.github/set_version.sh
+        shell: bash
       - name: 💰 Cache
         uses: Swatinem/rust-cache@v2
         with:
--- a/.github/workflows/test-rust.yaml
+++ b/.github/workflows/test-rust.yaml
@@ -37,6 +37,7 @@ jobs:
         with:
           crate: cargo-insta
       - run: ./.github/set_version.sh
+        shell: bash
       - name: 💰 Cache
         uses: Swatinem/rust-cache@v2
         with:


max-sixty:
No luck unfortunately — https://github.com/PRQL/prql/actions/runs/5673550570/job/15375269508?pr=3119

Happy if we can resolve a different way @not-my-profile . The old approach was worse but worked, so if you prefer to revert, that's fine too.

eitsupi:
Perhaps it has something to do with the lack of shebang?

.github/workflows/test-all.yaml
```diff
@@ -103,6 +103,7 @@ jobs:
         with:
           crate: cargo-tarpaulin
       - run: ./.github/set_version.sh
+        shell: bash
```
not-my-profile, line 106: Oh ... yeah I think we have to add it in `test-rust.yaml` as well.

not-my-profile:
> Perhaps it has something to do with the lack of shebang?

I don't think Windows supports shebangs ... and the [GitHub workflow doc](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstepsrun) also doesn't mention them.

eitsupi:
> I don't think Windows supports shebangs ... and the [GitHub workflow doc](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstepsrun) also doesn't mention them.

If we specify shell as bash on Windows on GHA, it is bash on MinGW or something and shebangs will be used, I believe.

not-my-profile:
Yes but at that point it's already running in bash so we don't need a shebang anymore.

.github/workflows/test-all.yaml
max-sixty: Ok, to the extent you'd be up for owning this all the way through, that would be great

not-my-profile (approved):
[LGTM](https://github.com/PRQL/prql/actions/runs/5683865097/job/15405345029?pr=3119#step:8:23)

PR #3122, opened 2023-07-27
ci: remove " quotes from cache key

not-my-profile:


.github/workflows/test-rust.yaml
```diff
@@ -37,6 +37,7 @@ jobs:
         with:
           crate: cargo-insta
       - run: ./.github/set_version.sh
+        shell: bash
```
not-my-profile: Are there other workflows that we run on Windows where we should set this as well?

eitsupi:
I checked and yq (which should be installed in the GitHub Actions runner) seems to support toml reading (https://github.com/mikefarah/yq/issues/1364#issuecomment-1483957510), could this be used?

.github/workflows/test-rust.yaml
not-my-profile: I'll leave this to you to do in #3119 and reduce this PR to the quote fix.

not-my-profile:
We wouldn't want to repeat the `yq` command in every workflow file so I think we'd still have to use a script ... so we'd still have to address #3119.

eitsupi:
> We wouldn't want to repeat the `yq` command in every workflow file so I think we'd still have to use a script ... so we'd still have to address #3119.

My suggestion was to use yq within a shell script. I don't think we need to worry about spaces and quotes if we parse toml correctly.

not-my-profile:
The `tr` command works fine and is part of POSIX ... I don't think proper TOML parsing is necessary here.

> yq (which should be installed in the GitHub Actions runner)

I don't know.

eitsupi:
> I don't know.

Please check the document. https://github.com/actions/runner-images/blob/ubuntu22/20230724.1/images/linux/Ubuntu2204-Readme.md#tools

not-my-profile:
Oh yeah that should indeed work ... still we might want to switch to a different docker image in the future ... e.g. switching to alpine would let contributors run the workflows locally with [act](https://github.com/nektos/act) without having to pull in all of ubuntu :shrug:. I don't know ... if you think using yq would be an improvement you're welcome to make a PR.

Issue #3108, opened 2023-07-26
Stop leaking chumsky error types from prql_parser


aljazerzen:
Follow up for https://github.com/PRQL/prql/pull/3000#discussion_r1272323992

Currently, `prql_parser` crate emits a special error type:

https://github.com/prql/prql/blob/e0fd9202d61390a452cef331fbe685a7d02e9562/crates/prql-parser/src/lib.rs#L20-L24

In the `prql_compiler::parser` we then map the error into the `prql_compiler::Error`.

I propose we emit `prql_compiler::Error` from `prql_parser` instead. This will:
- make chumsky an internal detail of the parser,
- standardize a single error type over all of the compiler crates,
- require `prql_compiler::Error` to be moved to a new `prql_common` crate or maybe `prql_ast`?
- make lexer an internal detail of the parser,

@not-my-profile

not-my-profile:
If we want to make chumsky an implementation detail we should create thin wrappers around its error types within the parser crate ... however we certainly shouldn't emit `prql_compiler::Error` from `prql_parser` ... a parser cannot emit compiler errors ... it doesn't do compilation.

aljazerzen:
Well then it would be `prql_common::Error` :D

You don't seem to treat the parser as a part of the compiler? It is producing the AST specific to this compiler.

aljazerzen:
Alternatively, we could create a new `prql_parser::Error` that is a copy-paste of the `prql_compiler::Error`.

not-my-profile:
> It is producing the AST specific to this compiler.

`prql_ast` should contain nothing compiler specific. Yes I know it currently contains `ExprKind::Internal` but I'll remove that.

And no `prql_compiler::Error` is a really poor error type since it just stringly types errors ... we certainly don't want to duplicate it anywhere.

aljazerzen:
Actually, thinking about it, maybe we should simplify the `prql_compiler::Error` and embrace what it is:

```rust
struct Error {
  message: String,
  span: Option<Span>,
  code: Option<String>,
}
```

aljazerzen:
I still find the idea that parser is not the part of compiler strange. I'd say that "parsing" is part of "compiling". Technically, it may be not, but in the same way as PRQL codegen or SQL codegen is not. And after that you are left with just a few IR mappings?

[...]

PR #3126, opened 2023-07-27
refactor: cleanup literal Display trait

aljazerzen:
Moves display of literal to prql_ast crate.
This code is quite inherent to Literal, so it can reside in prql_ast.
Especially because it simplifies other parts of the codebase.


not-my-profile:
I think it makes more sense to have that logic be part of the PRQL codegen/formatter.

max-sixty:
> I think it makes more sense to have that logic be part of the PRQL codegen/formatter.

Are the orphaning rules an issue here? i.e. that we can't implement `Display` on structs from another crate?

not-my-profile:
Yes of course they are but that can be solved via a newtype, as I did in 2017621e8dea0ed662ec0a88a760293de9181086, which this PR essentially seeks to revert.

To clarify what I had in mind: I think we should get rid of this Display impl entirely and rather implement it via the codegen trait ... (which should also be adjusted since it currently requires needless string allocations rather than operating on e.g. `std::fmt::Write`).

max-sixty:
> Yes of course they are but that can be solved via a newtype, as I did in [2017621](https://github.com/PRQL/prql/commit/2017621e8dea0ed662ec0a88a760293de9181086), which this PR essentially seeks to revert.

Yes, I guess this is a case of the tradeoff in https://github.com/PRQL/prql/issues/3108#issuecomment-1652202443

---

BTW @not-my-profile — just to emphasize — I think we're all v happy you're here and have already made such an impact. 

It's possible that we don't immediately end up exactly where you want, as we compromise on some of these tradeoffs. I hope you're OK with that. Nothing is fixed, and we can revisit in the future.

not-my-profile:
I personally think that keeping the codegen together and out of prql_ast is much more preferable than slightly simplifying the code by getting rid of a newtype ... especially when the newtype can be gotten rid of by just properly implementing (a better) codegen trait for it. You're welcome to any tradeoffs you deem reasonable ... I just disagree with this one.

aljazerzen:
Motivation here is that `prql_parser::lexer::Token` needs a display, so it can produce hints and error messages as "expected let, but found an identifier". The problem is that is contains `prql_ast::Expr::Literal`. Because `Token` is in parser and `Literal` is in ast, they cannot depend on codegen, which is in the main crate.

So this PR essentially says "tokens and literals are so simple constructs that they can be converted to strings on-the-fly. They don't require the whole codegen machinery, which deals with precedence and indentation".

not-my-profile:
I think that dependency problem can very simply be resolved with just a3d61a8737424cfe76a5f8853d097be24bbc88a3 (the first commit from #3141). We already have the spans, so I don't think repeating the literal in the error message really adds anything.

aljazerzen:
We can do without this refactor and avoid merge conflicts.

Issue #3105, opened 2023-07-26
PRQL project structure


aljazerzen:
_Followup for https://github.com/PRQL/prql/pull/3000#discussion_r1272280056_

I propose that we restructure the repo, so it reflects the whole project not just the prql-compiler.

This more of a issue about the structure of the PRQL project, and not the files and directories.

I see the project structured like this:

```
PRQL/
  website & documentation/
  prql-compiler/
  pq/
  prql-vscode/
```

Our main repo does correspond to the prql-compiler, but the language as a whole, so having `crates/` in the root tells me that these crates are essential to the language. Which is not true - they are just a reference implementation of the compiler, which could be implemented some other way.

If we wanted to follow advice in https://matklad.github.io/2021/08/22/large-rust-workspaces.html, we'd have something like this:

```
prql/
  web/
  prql-compiler/
    crates/
      ast/
      parser/
      frontend/
      backend-sql/
```

not-my-profile:
To repeat what I said in the other thread:

> A `crates` directory containing directories named after the crates is a well-established convention that helps new contributors quickly find the source code of crates. Putting the AST and the parser under compiler doesn't make any sense to me since they can very well be used on their own.

But I appreciate that you now explained your motivation for suggesting this change:

> Our main repo does correspond to the prql-compiler, but the language as a whole, so having crates/ in the root tells me that these crates are essential to the language. Which is not true - they are just a reference implementation of the compiler, which could be implemented some other way.

I think you're missing a "not" in there. Anyway I think your concern that people migth conflate the language with the reference implementation is a reasonable concern ... however I think the proper way to address this would be to move the specification of the language to a separate repository ... perhaps `prql-spec`. This would also enable us to strictly separate language design issues from implementation-specific issues. And with GitHub issue templates we could link the "new issue" page for issues about the language to the other issue tracker ... just an idea.

max-sixty:
I agree with the structure — but how would that work in practice? We'd have `prql-compiler` with exactly one path, `crates`?

Where would the external `compile` function live?

Not to pull in the whole of https://slatestarcodex.com/2014/11/21/the-categories-were-made-for-man-not-man-for-the-categories/ — but no representation is perfect, we're just trying to group approximate clusters of things that require more shared context...

aljazerzen:
> I think you're missing a "not" in there.

I do, sorry:

Our main repo does **not** correspond to the prql-compiler, ...

---

[...]

PR #3131, opened 2023-07-27
refactor!: parser

### LINKED ISSUE #3108: Stop leaking chumsky error types from prql_parser

aljazerzen:
- refactor: move Error into prql_ast
- refactor: move generic
- refactor!: stop leaking chumsky prql_parser
- refactor: move parser tests into own file
- fix: report syntax error from all files

Closes #3108


not-my-profile:
No moving the Error to prql_ast doesn't make any sense. #3108 should be fixed by introducing thin newtypes around the chumsky types ... not by having the parser emit the Error type from the compiler that we actually want to get rid of.

max-sixty:
@aljazerzen what should we do with this — I worry that #3265 is going to cause a conflict.

As I said in #3105, I think you should feel free to make adjustments after the big refactor even if there isn't unanimity, and then we can make bigger refactors as we gain consensus...

aljazerzen:
We can do without this refactor and avoid merge conflicts.

PR #3141, opened 2023-07-27
refactor!: hide chumsky from public prql_parser API

### LINKED ISSUE #3108: Stop leaking chumsky error types from prql_parser

not-my-profile:
Resolves #3108 (and unlike #3131 keeps the compiler-specific error type in prql-compiler).

### DIFF
--- a/crates/prql-compiler/src/parser.rs
+++ b/crates/prql-compiler/src/parser.rs
@@ -3,13 +3,9 @@ use std::collections::HashMap;
 use anyhow::Result;
 use prql_ast::stmt::Stmt;
 
-use crate::error::Errors;
+use crate::error::{Errors, WithErrorInfo};
 use crate::utils::IdGenerator;
-use crate::SourceTree;
-
-use self::error::{convert_lexer_error, convert_parser_error};
-
-mod error;
+use crate::{Error, Reason, SourceTree};
 
 pub fn parse(file_tree: &SourceTree<String>) -> Result<SourceTree<Vec<Stmt>>> {
     let mut res = SourceTree::default();
@@ -35,9 +31,11 @@ fn parse_source(source: &str, source_id: u16) -> Result<Vec<prql_ast::stmt::Stmt
         Errors(
             errors
                 .into_iter()
-                .map(|err| match err {
-                    prql_parser::Error::Lexer(err) => convert_lexer_error(source, err, source_id),
-                    prql_parser::Error::Parser(err) => convert_parser_error(err),
+                .map(|err| {
+                    // TODO: we actually want to to avoid this stringification
+                    // here but it's currently just there because Reason
+                    // currently must implement Clone but we don't necessarily want prql_parser::Error to implement Clone.
+                    Error::new(Reason::Simple(err.kind.to_string())).with_span(Some(err.span))
                 })
                 .collect(),
         )
@@ -78,9 +76,9 @@ mod tests {
                     span: Some(
                         0:22-23,
                     ),
-                    reason: Unexpected {
-                        found: "’",
-                    },
+                    reason: Simple(
+                        "unexpected ’",
+                    ),
                     hints: [],
                     code: None,
                 },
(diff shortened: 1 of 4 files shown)


crates/prql-parser/src/error.rs
```diff
@@ -0,0 +1,196 @@
...
+        span: *span,
+        kind: ErrorKind::Parser(ParserError(match e.found() {
+            Some(found) => format!(
+                "{who}expected {expected}, but found {found}",
+                who = e.label().map(|l| format!("{l} ")).unwrap_or_default(),
+                found = DisplayToken(found)
+            ),
+
+            // We want a friendlier message than "found end of input"...
+            None => format!("Expected {expected}, but didn't find anything before the end."),
```
max-sixty, line 145: Are we losing anything with this stringification? Or once we leave `prql-parser`, we're can't do anything with this structured info anyway?

crates/prql-compiler/src/parser.rs
```diff
@@ -34,9 +31,11 @@ fn parse_source(source: &str, source_id: u16) -> Result<Vec<prql_ast::stmt::Stmt
         Errors(
             errors
                 .into_iter()
-                .map(|err| match err {
-                    prql_parser::Error::Lexer(err) => convert_lexer_error(source, err, source_id),
-                    prql_parser::Error::Parser(err) => convert_parser_error(err),
+                .map(|err| {
+                    // TODO: we actually want to to avoid this stringification
+                    // here but it's currently just there because Reason
+                    // currently must implement Clone but we don't necessarily want prql_parser::Error to implement Clone.
```
max-sixty, line 37: Thanks for the code comment. 

Would there be a downside of `prql_parser::Error` implementing `Clone`? 

If it did, how would we avoid the stringification? Wouldn't `reason` still need to be a flat string?

not-my-profile: I think the only reason why the compiler error type implements `Clone` is because of the downcasting that `WithErrorInfo` does ... we should get rid of this downcasting anyway, at which point all the Clone impls would be rather redundant. But adding it now and removing it later would be a breaking change, so I'd rather not add it at all.

The idea for the future is to rename `Reason` to `ErrorKind` and have a variant like `Parser(prql_parser::Error)`, so the stringification would only happen on `Display`.

crates/prql-parser/src/error.rs
not-my-profile: I don't think we're loosing anything. If we at some point in the future want to provide more structured parser errors we can just change the `ParserError` struct to be an enum instead.

crates/prql-compiler/src/parser.rs
max-sixty: What's the downside of implementing `Clone` though?

crates/prql-parser/src/error.rs
aljazerzen: Right now, we are not. All errors eventually get converted into `ErrorMessage`, which contains just `message: String, code: String` (and a few others).

crates/prql-compiler/src/parser.rs
not-my-profile: Derive impls increase the compilation time ... for this particular instance probably rather negligible but I like to in general avoid derive impls that aren't actually needed.

max-sixty: I see, thanks

aljazerzen:
LGTM

PR #3142, opened 2023-07-27
docs: Fix Changelog example

max-sixty:
We changed the `filter` syntax after merging the regex change


max-sixty:
Thanks @matthias-Q for spotting!

---

Possibly we _should_ evaluate the code blocks in Changelog? At least until we release the next version? Or we should link to docs which are tested and current?

PR #3147, opened 2023-07-28
docs: Update join.md

richb-hanover:
Add a little more detail for the `join` parameters

### DIFF
--- a/web/book/src/reference/stdlib/transforms/join.md
+++ b/web/book/src/reference/stdlib/transforms/join.md
@@ -9,7 +9,8 @@ join side:{inner|left|right|full} table (condition)
 ## Parameters
 
 - `side` decides which rows to include, defaulting to `inner`.
-- `table` - a reference to a relation, possibly including an assignment, e.g. `var= ...`
+- `table` - a reference to a relation, possibly including an assignment, e.g.
+  `var= ...`
 - `condition` - a boolean condition
   - If the condition evaluates to True, the rows will be joined
   - If name is the same from both tables, it can be expressed with only `==col`.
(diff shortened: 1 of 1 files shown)


web/book/src/reference/stdlib/transforms/join.md
```diff
@@ -9,8 +9,10 @@ join side:{inner|left|right|full} table (condition)
 ## Parameters
 
 - `side` decides which rows to include, defaulting to `inner`.
-- Relation reference
-- A condition
+- `table` - a reference to a relation, possibly including an assignment, e.g.
+  `var= ...`
+- `condition` - a boolean condition
```
max-sixty:
```suggestion
- a reference to a relation, possibly including an assignment, e.g.
  `var= ...`
- a boolean condition
```

Only `side` is a named param, so removing the others as quoted names — is that reasonable?

max-sixty (approved):
Thanks!

web/book/src/reference/stdlib/transforms/join.md
richb-hanover: Ahah! I was aiming for visual consistency to call out/mention each of the parameters. Try this...

max-sixty: But surely we do want to visually discriminate between `side` and the others?

web/book/src/reference/stdlib/transforms/join.md
```diff
@@ -8,9 +8,11 @@ join side:{inner|left|right|full} table (condition)
 
 ## Parameters
 
-- `side` decides which rows to include, defaulting to `inner`.
-- Relation reference
-- A condition
+- _side_ decides which rows to include, defaulting to `inner`.
```
max-sixty:
```suggestion
- `side` specifies which rows to include, defaulting to `inner`.
```

How about this?

max-sixty:
Thanks @richb-hanover !

Issue #3124, opened 2023-07-27
Lack of a page of "this" on the book


eitsupi:
#2935 and #2936

like

```elm
let games = (
  from tab
  filter s"COLUMNS(*) IS NOT NULL"
  derive date_new = (s"strptime(date, '%a, %b %d, %Y')" | as date)
  select !{this.date}
  sort date_new
  derive game_id = (row_number this)
)

from games
take 5
```

PR #3148, opened 2023-07-28
docs: Add docs on `this` & `that`

max-sixty:
Not my finest literature, but fills the gap


max-sixty:
Closes #3124

eitsupi:
Thanks!
Example including something like `this.time` may helpful.

max-sixty:
> Example including something like `this.time` may helpful.

Good idea: https://github.com/PRQL/prql/pull/3152

Issue #3098, opened 2023-07-26
`cargo build -p prqlc` isn't really cached after `cargo build`


max-sixty:
### What's up?

```
cargo clean
cargo build # takes 30 seconds
cargo build -p prqlc # takes 17 seconds — *not 0 seconds!*
```

```
cargo clean
cargo build -p prqlc # takes 22 seconds.
```

So the really curious thing is why running `cargo build` hardly helps with `cargo build -p prqlc` given it presumably builds prqlc as part of its build.

This hits us in the CI runtime of building prqlc.

Is it something to do with the dependencies being slightly different — e.g. features being different?

max-sixty:
I tried aligning the dependency features etc, but it didn't seem to help.

```diff
diff --git a/crates/prql-compiler/Cargo.toml b/crates/prql-compiler/Cargo.toml
index 4b6745e2..e0748a90 100644
--- a/crates/prql-compiler/Cargo.toml
+++ b/crates/prql-compiler/Cargo.toml
@@ -11,15 +11,15 @@ version.workspace = true
 metadata.msrv = "1.65.0"
 
 [features]
-default = []
+default = ["serde_yaml"]
 # Technically tokio could be limited to external tests, but its types are in
 # signatures which would require lots of conditional compilation.
 test-dbs = ["duckdb", "rusqlite", "tokio"]
 test-dbs-external = ["chrono", "duckdb", "mysql", "pg_bigdecimal", "postgres", "rusqlite", "tiberius", "tokio", "tokio-util"]
 
 [dependencies]
-prql-ast = {path = "../prql-ast", version = "0.9.2" }
-prql-parser = {path = "../prql-parser", version = "0.9.2" }
+prql-ast = {path = "../prql-ast", version = "0.9.2"}
+prql-parser = {path = "../prql-parser", version = "0.9.2"}
 
 anstream = {version = "0.3.2", features = ["auto"]}
 anyhow = {version = "1.0.57", features = ["backtrace"]}
@@ -39,7 +39,7 @@ sqlparser = {version = "0.36.0", features = ["serde"]}
 strum = {version = "0.25.0", features = ["std", "derive"]}
 strum_macros = "0.25.0"
 
-serde_yaml = {version = "0.9", optional = true}
+serde_yaml = {version = "0.9.1", optional = true}
 
 [target.'cfg(not(target_family="wasm"))'.dependencies]
 # For integration tests. These are not listed as dev-dependencies because
[...]

PR #3157, opened 2023-07-28
ci: Disable duckdb tests while caching has issues

max-sixty:
ref https://github.com/duckdb/duckdb-rs/issues/178; related to https://github.com/PRQL/prql/issues/3098


max-sixty:
Let's try — we may need to revert if caching doesn't go as well as expected

PR #3163, opened 2023-07-28
refactor: Should `prql-lib` use `char`?

max-sixty:
I'm less strong on this stuff — I saw a warning and made this change — is it good?


aljazerzen:
Yeah, probably. I doesn't really matter - I know cases where people just use `void *` for all pointers - but still, char gives more info that these are strings.

vanillajonathan:
Keep in mind that the `libprql_lib.h` header file was autogenerated with [cbindgen](https://github.com/eqrion/cbindgen). That there is also a header file for C++ called `libprql_lib.hpp` which you did not touch.

max-sixty:
> Keep in mind that the `libprql_lib.h` header file was autogenerated with [cbindgen](https://github.com/eqrion/cbindgen). That there is also a header file for C++ called `libprql_lib.hpp` which you did not touch.

Ah, thanks for spotting @vanillajonathan — you're right, it says it at the top of the file 🤦 .

Do you happen to know whether we can change it to generate `char`? Otherwise I guess this is going to be overwritten next time we autogen...

vanillajonathan:
> Do you happen to know whether we can change it to generate `char`? Otherwise I guess this is going to be overwritten next time we autogen...

I generated a new header file and saw that it generated `char`, I don't know why.

max-sixty:
Ah interesting. 

If we can add the command to generate a new header to `task build-lib` (or `task build-c`), that would be a good reference, and ensure it keeps up to date...

PR #3166, opened 2023-07-29
ci: make version extraction more reliable

eitsupi:
https://github.com/PRQL/prql/pull/3122#issuecomment-1653053214

There is no need to remove spaces or quotation marks because `cargo metadata` correctly parses the TOML and extracts the version number.

### DIFF
--- a/.github/set_version.sh
+++ b/.github/set_version.sh
@@ -2,4 +2,4 @@
 # since the caches seem to accumulate cruft over time;
 # ref https://github.com/PRQL/prql/pull/2407
 
-grep '^version =' Cargo.toml | tr -d ' "' >> $GITHUB_ENV
+echo "version=$(cargo metadata --format-version=1 --no-deps | jq -r '.packages[0].version')" >>"$GITHUB_ENV"
--- a/.github/set_version.sh
+++ b/.github/set_version.sh
@@ -1,3 +1,5 @@
+#!/bin/bash
+
 # We set prefix-key to the version from Cargo.toml for Swatinem/rust-cache@v2
 # since the caches seem to accumulate cruft over time;
 # ref https://github.com/PRQL/prql/pull/2407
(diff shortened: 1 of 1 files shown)


eitsupi:
It does not seem working on Windows

```log
  ./.github/set_version.sh
  shell: C:\Program Files\Git\bin\bash.EXE --noprofile --norc -e -o pipefail {0}
  env:
    CARGO_TERM_COLOR: always
    CLICOLOR_FORCE: 1
    RUSTFLAGS: -C debuginfo=0
./.github/set_version.sh: line 5: yq: command not found
```

I thought yq was available on GitHub Actions, but it doesn't seem to be included on Windows at this time. (actions/runner-images#7443)
https://github.com/actions/runner-images/blob/win22/20230724.1/images/win/Windows2022-Readme.md

max-sixty:
FYI `cargo metadata` & `jq` is probably the "optimizer" (not necessarily good!) approach...

eitsupi:
> FYI `cargo metadata` & `jq` is probably the "optimizer" (not necessarily good!) approach...

Good idea! Thanks.

.github/set_version.sh
```diff
@@ -1,5 +1,7 @@
+#!/bin/bash
+
 # We set prefix-key to the version from Cargo.toml for Swatinem/rust-cache@v2
 # since the caches seem to accumulate cruft over time;
 # ref https://github.com/PRQL/prql/pull/2407
 
-grep '^version =' Cargo.toml | tr -d ' "' >> $GITHUB_ENV
+echo "version=$(cargo metadata --format-version=1 --no-deps | jq -r '.packages[0].version')" >>"$GITHUB_ENV"
```
not-my-profile: I don't think we should rely on the first package having the version number we want ... we should rather filter for the name to be "prql-compiler".

not-my-profile: And I think `--raw-output` would be more readable than `-r`.

max-sixty (approved):
Thanks!

Issue #3150, opened 2023-07-28
confusing errors for incorrect function calls in a `filter`


vthriller:
(Yes, I know, it should be `(lower f"...")` or `(f"..." | lower)` instead. Still, reporting because it might hint to other problems with the parser.)

```
from f = foo
join b = bar (==id)
filter lower(f"{b.x}/{b.y}") == f.z
select {
  lower(f"{b.x}/{b.y}") == f.z
}
```

yields:

```
Error: 
   ╭─[:3:16]
   │
 3 │ filter lower(f"{b.x}/{b.y}") == f.z
   │                ──┬──  
   │                  ╰──── Unknown name
───╯
```

Commenting out `filter` actually produces the output:

```sql
SELECT
  LOWER(CONCAT(b.x, '/', b.y) = f.z)
FROM
  foo AS f
  JOIN bar AS b ON f.id = b.id
```

And `filter (lower(f"{b.x}/{b.y}") == f.z)` throws error that actually makes sense:

```
Error: 
   ╭─[:3:9]
   │
 3 │ filter (lower(f"{b.x}/{b.y}") == f.z)
   │         ──────────────┬─────────────  
   │                       ╰─────────────── function std.filter, param `condition` expected type `bool`, but found type `text`
   │ 
   │ Help: Type `bool` expands to `bool`
───╯
```

max-sixty:
Good example!

Because there are so many error messages, we track them in `bad_error_messages.rs` — not sure whether you're set up / technical enough to do that; otherwisei or someone else can. Thanks!

vthriller:
I've reduced this example a bit and now I'm not sure if it belongs to `bad_error_messages.rs`. I mean, this particular example does:

```
from foo
select z == lower f"{x}/{y}"
```

because `select {z == lower f"{x}/{y}"}` is also invalid, but

```
from foo
select lower f"{x}/{y}"
```

throws the exact same error, even though `select { lower f"{x}/{y}" }` and `select (lower f"{x}/{y}")` are parsed as expected. So this now looks more like a parser issue rather than misleading error. I guess I should rewrite this issue instead.

vthriller:
...then again, `select lower "mooo"` is also invalid, so this actually belongs to `bad_error_messages` all along.

```
Error: 
   ╭─[:2:14]
   │
 2 │ select lower "mooo"
   │              ───┬──  
   │                 ╰──── function std.select, param `tbl` expected type `relation`, but found type `text`
   │ 
[...]

PR #3171, opened 2023-07-29
test: another confusing error message (via #3150)

vthriller:


max-sixty (approved):
Yes this is a great one, thanks

Issue #3127, opened 2023-07-27
standard functions sometimes can't be used after the `as` function


eitsupi:
### What happened?

From https://github.com/eitsupi/querying-with-prql/blob/4acff8fa395a9a2c33cad15ed61866dabac25157/docs/method_chaining.qmd#L128-L158

I rewrote the example that worked in PRQL 0.8 with the 0.9 syntax and an error occurred.

```elm
let time_to_datetime = string -> (FlightDate | as timestamp) + s"""
  TRY_CAST(regexp_replace({string}, '^2400$', '0000').substr(1, 2).concat(' hours') AS INTERVAL) +
  TRY_CAST(regexp_replace({string}, '^2400$', '0000').substr(3, 2).concat(' minutes') AS INTERVAL)
"""

from tab
select {
  FlightDate,
  DepTimeOld = DepTime
}
derive {
  DepTime = (time_to_datetime DepTimeOld)
}
take 5
```

For example, the following minimal example will result in the following error:

```
Error: 
   ╭─[:2:23]
   │
 2 │ select col1 = (col1 | as int) + 1
   │                       ───┬──  
   │                          ╰──── function std.add, param `left` expected type `int || float || timestamp || date`, but found type `scalar`
───╯
```

### PRQL input

```elm
from tab
select col1 = (col1 | as int) + 1
```

### SQL output

```SQL
NA
```

### Expected SQL output

```SQL
SELECT CAST(col1 AS int) + 1 AS col1
FROM tab
```

### MVCE confirmation

- [X] Minimal example
- [X] New issue

vthriller:
Interestingly,

```
from tab
select col1 = (col1 | as int) * 2
```

works. It's `std.add`/`std.sub` that throws this error.

max-sixty:
Good find.

And to note, adding parentheses doesn't help:

```elm
from tab
select col1 = ((col1 | as int) + 1)
```

```
Error: 
   ╭─[:2:24]
   │
 2 │ select col1 = ((col1 | as int) + 1)
   │                        ───┬──  
   │                           ╰──── function std.add, param `left` expected type `int || float || timestamp || date`, but found type `scalar`
───╯
```

eitsupi:
It seems that only `add` and `sub` are specifing types.
Perhaps `scalar` should also be allowed?

[...]

PR #3174, opened 2023-07-29
test: cases for both valid and confusing error messages re: types

vthriller:


vthriller:
(Initially as part of independent discovery of #3127.)

crates/prql-compiler/src/tests/test_error_messages.rs
```diff
@@ -189,6 +189,26 @@ fn test_basic_type_checking() {
     "###);
 }
 
+// See also test_bad_error_messages::misplaced_type_error
```
max-sixty, line 192:
```suggestion
// See also test_bad_error_messages::misplaced_type_error
// Note that the ``` Help: Type `bool` expands to `bool` ``` is not that useful
```

max-sixty (approved):
Good one!

PR #3177, opened 2023-07-30
tweak: Wordsmith line in book

max-sixty:


web/book/src/project/contributing/development.md
```diff
@@ -49,8 +49,8 @@ website, we have two options:
 ### Option 1: Use the project's `task`
 
 ```admonish note
-This is tested on MacOS, should work on Linux, but won't work on Windows.
-Because of using the `brew` command inside of `Taskfile.yml`.
+This is tested on MacOS, should work on Linux, but won't work on Windows,
```
eitsupi: Homebrew only supports amd64 Linux, I think.

```suggestion
This is tested on macOS, should work on amd64 Linux, but won't work on others (include Windows),
```

PR #3181, opened 2023-07-30
ci: test release dry run in test-all workflow

### LINKED ISSUE #3179: Improve release workflow to prevent release error
Releases 0.9.0 and 0.9.1 failed to publish prql-compiler to crates.io. (#3051)

To prevent this, it is necessary to test whether it is possible to publish by executing a command like `cargo release` in advance.

```sh
$ cargo release
error: tag `0.9.2` already exists (for `prql-compiler-macros`)
error: prql-ast 0.9.2 is already published
error: prql-parser 0.9.2 is already published
error: prql-compiler 0.9.2 is already published
error: prql-compiler-macros 0.9.2 is already published
error: prqlc 0.9.2 is already published
warning: updating prql-compiler-macros to 0.9.2 despite no changes made since tag 0.9.2
warning: updating prql-java to 0.9.2 despite no changes made since tag 0.9.2
  Publishing prql-ast
    Updating crates.io index
   Packaging prql-ast v0.9.2 (/workspaces/prql/crates/prql-ast)
    Packaged 10 files, 17.1KiB (5.1KiB compressed)
   Uploading prql-ast v0.9.2 (/workspaces/prql/crates/prql-ast)
warning: aborting upload due to dry run
  Publishing prql-parser
    Updating crates.io index
   Packaging prql-parser v0.9.2 (/workspaces/prql/crates/prql-parser)
    Packaged 9 files, 101.9KiB (16.4KiB compressed)
   Uploading prql-parser v0.9.2 (/workspaces/prql/crates/prql-parser)
warning: aborting upload due to dry run
  Publishing prql-compiler
    Updating crates.io index
   Packaging prql-compiler v0.9.2 (/workspaces/prql/crates/prql-compiler)
    Packaged 157 files, 1.1MiB (298.4KiB compressed)
   Uploading prql-compiler v0.9.2 (/workspaces/prql/crates/prql-compiler)
warning: aborting upload due to dry run
  Publishing prql-compiler-macros
    Updating crates.io index
   Packaging prql-compiler-macros v0.9.2 (/workspaces/prql/crates/prql-compiler-macros)
    Packaged 3 files, 2.5KiB (1.3KiB compressed)
   Uploading prql-compiler-macros v0.9.2 (/workspaces/prql/crates/prql-compiler-macros)
warning: aborting upload due to dry run
  Publishing prqlc
    Updating crates.io index
   Packaging prqlc v0.9.2 (/workspaces/prql/crates/prqlc)
warning: package `hermit-abi v0.3.1` in Cargo.lock is yanked in registry `crates-io`, consider updating to a version that is not yanked
    Packaged 15 files, 157.0KiB (27.7KiB compressed)
   Uploading prqlc v0.9.2 (/workspaces/prql/crates/prqlc)
warning: aborting upload due to dry run
     Pushing Pushing 0.9.2, compile-files-v0.9.2, main, misc-v0.9.2, prql-ast-v0.9.2, prql-compiler-v0.9.2, prql-parser-v0.9.2, prql-v0.9.2, prqlc-v0.9.2 to origin
error: dry-run failed, resolve the above errors and try again.
```

This check will fail if the same version has already been published on crates.io.

Therefore, the local version must be upgraded before the test is run.
Our current workflow is to version up just before release, but is it worth versioning up right after release?
Or would it be better to configure GitHub Actions to only run tests if the version of crates.io and the local version differ?

https://github.com/PRQL/prql/blob/1f9e8c59e5889f31a605b870ea5b1e8fb3323135/web/book/src/project/contributing/development.md?plain=1#L383-L435

eitsupi:
Close #3179

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -6,7 +6,6 @@ on:
     inputs:
       dry-run:
         description: "Whether to do a dry run"
-        required: true
         default: true
         type: boolean
   workflow_dispatch:
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -73,7 +73,7 @@ jobs:
     # TODO: currently this doesn't publish it as an artifact; would be a welcome contribution
     # to add that (and eventually as a release asset)
     runs-on: ubuntu-latest
-    if: false
+    if: ${{ !inputs.dry-run }}
     steps:
       - name: 📂 Checkout code
         uses: actions/checkout@v3
@@ -101,7 +101,7 @@ jobs:
   build-rpm-package:
     # TODO: This doesn't publish the rpm yet, that would be a welcome follow-up (even as a CI artifact)
     runs-on: ubuntu-latest
-    if: false
+    if: ${{ !inputs.dry-run }}
     steps:
       - name: 📂 Checkout code
         uses: actions/checkout@v3
@@ -131,7 +131,7 @@ jobs:
 
   build-and-publish-snap:
     runs-on: ubuntu-latest
-    if: false
+    if: ${{ !inputs.dry-run }}
     steps:
       - name: 📂 Checkout code
         uses: actions/checkout@v3
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -6,14 +6,14 @@ on:
     inputs:
       dry-run:
         description: "Whether to do a dry run"
-        required: false
+        required: true
         default: true
         type: boolean
   workflow_dispatch:
     inputs:
       dry-run:
         description: "Whether to do a dry run"
-        required: false
+        required: true
         default: true
         type: boolean
 
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -2,12 +2,26 @@ name: release
 on:
   release:
     types: [released]
+  workflow_call:
+    inputs:
+      dry-run:
+        description: "Whether to do a dry run"
+        required: false
+        default: true
+        type: boolean
   workflow_dispatch:
+    inputs:
+      dry-run:
+        description: "Whether to do a dry run"
+        required: false
+        default: true
+        type: boolean
 
 jobs:
   brew-dispatcher:
     name: Release on homebrew-prql
     runs-on: ubuntu-latest
+    if: github.event.inputs.dry-run != 'true'
     steps:
       - uses: actions/github-script@v6
         with:
@@ -26,6 +40,7 @@ jobs:
 
   build-prqlc:
     runs-on: ${{ matrix.os }}
+    if: github.event.inputs.dry-run != 'true'
     strategy:
       fail-fast: false
       matrix:
@@ -49,6 +64,7 @@ jobs:
           target: ${{ matrix.target }}
           profile: release
       - name: Upload release artifact
+        if: github.event_name == 'release'
         uses: softprops/action-gh-release@v1
         with:
           append_body: true
@@ -58,6 +74,7 @@ jobs:
     # TODO: currently this doesn't publish it as an artifact; would be a welcome contribution
     # to add that (and eventually as a release asset)
     runs-on: ubuntu-latest
+    if: false
     steps:
       - name: 📂 Checkout code
         uses: actions/checkout@v3
@@ -85,6 +102,7 @@ jobs:
   build-rpm-package:
     # TODO: This doesn't publish the rpm yet, that would be a welcome follow-up (even as a CI artifact)
     runs-on: ubuntu-latest
+    if: false
     steps:
       - name: 📂 Checkout code
         uses: actions/checkout@v3
@@ -114,10 +132,7 @@ jobs:
 
   build-and-publish-snap:
     runs-on: ubuntu-latest
-
-    # Skip running workflow on forks
-    if: github.repository_owner == 'prql'
-
+    if: false
     steps:
       - name: 📂 Checkout code
         uses: actions/checkout@v3
@@ -137,6 +152,7 @@ jobs:
 
   build-python-wheels:
     runs-on: ${{ matrix.os }}
+    if: github.event.inputs.dry-run != 'true'
     strategy:
       fail-fast: true
       matrix:
@@ -186,7 +202,9 @@ jobs:
           registry-url: "https://registry.npmjs.org"
 
       - name: Publish package on npm
-        run: npm publish
+        run:
+          npm publish ${{ github.event.inputs.dry-run == 'true' && '--dry-run'
+          || '' }}
         working-directory: bindings/prql-js/
         env:
           NODE_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }}
@@ -202,7 +220,9 @@ jobs:
       - uses: richb-hanover/cargo@v1.1.0
         with:
           command: release
-          args: publish --no-confirm -x
+          args:
+            publish --no-confirm ${{ github.event.inputs.dry-run == 'true' && ''
+            || '--execute' }}
         env:
           CARGO_REGISTRY_TOKEN: ${{ secrets.CARGO_REGISTRY_TOKEN }}
 
@@ -227,10 +247,12 @@ jobs:
 
   push-web-branch:
     runs-on: ubuntu-latest
+    if: github.event.inputs.dry-run != 'true'
     steps:
       - name: 📂 Checkout code
         uses: actions/checkout@v3
       - run: git push origin HEAD:web --force
 
   push-devcontainer-base-image:
+    if: github.event.inputs.dry-run != 'true'
     uses: ./.github/workflows/build-devcontainer.yaml
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -215,6 +215,7 @@ jobs:
           crate: cargo-release
       # Currently, we can only check prql-ast which is not dependent other local crates with --dry-run.
       # https://github.com/crate-ci/cargo-release/issues/691
+      # --no-verify is required to prevent build.
       - run:
           cargo release publish --no-confirm ${{ inputs.dry-run && '--no-verify
           --package prql-ast' || '--execute' }}
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -202,9 +202,7 @@ jobs:
           registry-url: "https://registry.npmjs.org"
 
       - name: Publish package on npm
-        run:
-          npm publish ${{ github.event.inputs.dry-run == 'true' && '--dry-run'
-          || '' }}
+        run: npm publish ${{ github.event.inputs.dry-run && '--dry-run' || '' }}
         working-directory: bindings/prql-js/
         env:
           NODE_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }}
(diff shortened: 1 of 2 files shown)


max-sixty:
An alternative would be to run the whole `release.yaml` workflow with a dry run option — which would avoid rewriting the jobs / keeping them aligned. We could make any steps which don't have dry run options into no-ops. 

But this is fine to start!

.github/workflows/test-all.yaml
```diff
@@ -153,3 +153,21 @@ jobs:
...
         run: cargo msrv verify
+
+  test-publish-to-cargo:
+    runs-on: ubuntu-latest
+    steps:
+      - name: 📂 Checkout code
+        uses: actions/checkout@v3
+      - run: ./.github/set_version.sh
+      - run: |
+          published_version=$(curl -sL https://crates.io/api/v1/crates/prql-compiler/versions | jq -r '.versions[].num' | sort -V | tail -n 1)
```
max-sixty, line 171: Why have this? Aren't they only equal for the actual release commit?

eitsupi: > Why have this? Aren't they only equal for the actual release commit?

My understanding is that we can only find out which version has been published to crates.io by contacting crates.io.
I don't think we can check the Git repository to see if it has been successfully published to crates.io or not.

max-sixty: Oh, I see, because otherwise it'll fail...

eitsupi:
It seems to fail because the version of the dependency in the workspace is not published on crates.io. (crate-ci/cargo-release#691)

max-sixty:
Awesome! Thanks a lot for doing the full version @eitsupi , this could be great!

Hmmm re `cargo release publish`, that's a shame. Let me have a look and see whether there's a workaround

.github/workflows/release.yaml
```diff
@@ -2,12 +2,26 @@ name: release
...
+        description: "Whether to do a dry run"
+        required: false
+        default: true
+        type: boolean
 
 jobs:
   brew-dispatcher:
     name: Release on homebrew-prql
     runs-on: ubuntu-latest
+    if: github.event.inputs.dry-run != 'true'
```
max-sixty: nit: does this work?

```suggestion
    if: ! github.event.inputs.dry-run
```

.github/workflows/release.yaml
```diff
@@ -186,7 +202,9 @@ jobs:
           registry-url: "https://registry.npmjs.org"
 
       - name: Publish package on npm
-        run: npm publish
+        run:
+          npm publish ${{ github.event.inputs.dry-run == 'true' && '--dry-run'
```
max-sixty: (but here it looks like `== 'true'` *is* required, since inside the braces, everything is a string...

.github/workflows/release.yaml
```diff
@@ -202,7 +220,9 @@ jobs:
       - uses: richb-hanover/cargo@v1.1.0
         with:
           command: release
```
max-sixty: FYI in many of these (maybe all? at least anything not dependent on toolchains) we don't actually need the `cargo` action, we can just use normal `run:`

```suggestion
```

.github/workflows/release.yaml
```diff
@@ -202,7 +220,9 @@ jobs:
...
-          args: publish --no-confirm -x
+          args:
+            publish --no-confirm ${{ github.event.inputs.dry-run == 'true' && ''
```
max-sixty: FYI I _think_ the `''` will evaluate to false and `'--execute'` would always be returned!

GHA syntax is kinda dangerous there...

.github/workflows/release.yaml
```diff
@@ -202,7 +220,9 @@ jobs:
...
+            || '--execute' }}
```
```suggestion
          # In a dry-run, we only attempt to publish `prql-ast`, because of
          # https://github.com/crate-ci/cargo-release/issues/691
          run:
            cargo release publish --no-confirm
            ${{ github.event.inputs.dry-run == 'true' 
            && '--execute' 
            || '--dry-run -p prql-ast' }}
```

max-sixty:
> Hmmm re `cargo release publish`, that's a shame. Let me have a look and see whether there's a workaround

I think we can try limiting to `prql-ast`, I added a suggestion, WDYT?

.github/workflows/release.yaml
max-sixty: > GHA syntax is kinda dangerous there...

(but given the number of jobs, I think this is the correct design, relative to a `task release` (which only would work in GHA anyway given the credentials)

.github/workflows/release.yaml
eitsupi: I wasn't sure it would work, but as you pointed out in your comment (https://github.com/PRQL/prql/pull/3181#discussion_r1278603944), GitHub Actions seems to evaluate an empty string as false, so it could work.
I'll see.

eitsupi: ~~It does not seem working.~~

`!github.event.inputs.dry-run` does not work, but `!inputs.dry-run` works...

.github/workflows/release.yaml
eitsupi: ~~That's true. `== 'true'` seems needed here.~~

When using `inputs.dry-run` instead of `github.event.inputs.dry-run`, `== 'true'` is not needed...

eitsupi:
ref: https://github.blog/changelog/2022-06-10-github-actions-inputs-unified-across-manual-and-reusable-workflows/

.github/workflows/release.yaml
```diff
@@ -58,6 +74,7 @@ jobs:
     # TODO: currently this doesn't publish it as an artifact; would be a welcome contribution
     # to add that (and eventually as a release asset)
     runs-on: ubuntu-latest
+    if: false
```
max-sixty: Why do we disable this & the rpm? Even if we're not publishing it yet, it's still useful to build, and is an intermediate step to publishing, no?

.github/workflows/release.yaml
```diff
@@ -199,10 +214,11 @@ jobs:
...
         with:
           crate: cargo-release
-      - uses: richb-hanover/cargo@v1.1.0
-        with:
-          command: release
-          args: publish --no-confirm -x
+      # Currently, we can only check prql-ast which is not dependent other local crates with --dry-run.
+      # https://github.com/crate-ci/cargo-release/issues/691
+      - run:
+          cargo release publish --no-confirm ${{ inputs.dry-run && '--no-verify
```
max-sixty, line 220: (I had thought that we wouldn't need the `--no-verify` if we were only doing `prql-ast`, but no big deal even if that is correct)

.github/workflows/test-all.yaml
```diff
@@ -153,3 +153,30 @@ jobs:
       - name: Verify minimum rust version — prqlc
         working-directory: crates/prqlc
         run: cargo msrv verify
+
+  check-published:
```
max-sixty, line 161:
```suggestion
  # One possible alternative to having this workflow is to automatically
  # bump the version after a commit, auto-merge the PR, and then there
  # should be no time where the version in the repo has been
  # published.
  check-published:
```

.github/workflows/release.yaml
eitsupi: > (I had thought that we wouldn't need the `--no-verify` if we were only doing `prql-ast`, but no big deal even if that is correct)

If this is not added, prql-ast will be built, which will take time.

.github/workflows/release.yaml
eitsupi: > Why do we disable this & the rpm? Even if we're not publishing it yet, it's still useful to build, and is an intermediate step to publishing, no?

In my opinion these jobs need to be rewritten as the package needs to be created based on the uploaded prqlc.
Currently they are building prqlc from scratch, which takes extra time.

.github/workflows/release.yaml
max-sixty: OK — I do think that's worth noting in a comment, so the next person knows why we have it. (Or is it fairly obvious and I was missing it? No offense taken if so, we can't explain every line of code... :) )

.github/workflows/release.yaml
max-sixty: But then why not run on releases only?

By turning off tests, it'll decay, and we'll be further from getting it publishing. Generally stuff that isn't tested breaks! And then contributors are less likely to contribute smaller changes, since their work won't be kept current.

Computing time is so much cheaper than contributor time! So anything we can do to make contributing easier is likely to be worthwhile...

(A future version could run nightly rather than on every merge and then run these, but that's another project)

.github/workflows/release.yaml
eitsupi: I added a comment.

.github/workflows/release.yaml
eitsupi: I enabled when not dry run.

.github/workflows/release.yaml
```diff
@@ -114,10 +131,7 @@ jobs:
 
   build-and-publish-snap:
     runs-on: ubuntu-latest
-
-    # Skip running workflow on forks
-    if: github.repository_owner == 'prql'
```
max-sixty, line 119: Sorry to keep adding comments. I do think we need this — otherwise the build fails on a fork...

max-sixty:
Looking great! I added one comment, then we can merge.

Thanks, this is much better than it was!

.github/workflows/release.yaml
eitsupi: I dont know why you care about it. Other release jobs basically fail too, is that a problem?

max-sixty: Hmmm, good point. 

I tried to look at the history of this; I think it inherited from https://github.com/PRQL/prql/blob/40ddf29b588f70b8dda3bf9aed57d0e017e7e973/.github/workflows/build-web.yaml#L26 — where it is needed, since that workflow regularly runs on forks.

Agree re removing, we can see if something breaks...

.github/workflows/release.yaml
max-sixty: OK — just to confirm, we're sure that `${{ !inputs.dry-run }}` will evaluate as `true` when `dry-run` is false?

To quote the [docs](https://docs.github.com/en/actions/learn-github-actions/expressions): 

> Using the ${{ }} expression syntax turns the contents into a string, and strings are truthy. For example, if: true && ${{ false }} will evaluate to true.

I had previously thought that `!inputs.dry-run` would evaluate as `true` when `dry-run` is `false`, because it's not inside the braces (but all of our are inside braces)

max-sixty (approved):

.github/workflows/release.yaml
max-sixty: Happy to see whether something breaks, don't want to slow you down. But on thinking about it again — surely every push to a fork *is* going to fail — should we have `github.repository_owner == 'prql'` as a condition on https://github.com/PRQL/prql/pull/3181/files#diff-e3cb659af0d95ff75113e6a8678a6ac089b653b083d316c3f44b4da8086b59d2R183 ?

eitsupi: if `dry-run: true`, the workflow will not fail on forks as no secrets are used.
So we don't need to check the repository owner.

image

max-sixty:
FYI this worked great. 

Sorry for doubting you at https://github.com/PRQL/prql/pull/3181#discussion_r1278850594 ! 😄

PR #3182, opened 2023-07-30
ci: Share CI cache between book & link checks

max-sixty:


max-sixty:
This was merged as part of #3183

PR #3183, opened 2023-07-30
ci: Consolidate check-links-book into pull-request.yaml

max-sixty:
Stacked on https://github.com/PRQL/prql/pull/3182


not-my-profile:
What's the reasoning for this change?

max-sixty:
> What's the reasoning for this change?

Less indirection — is there a reason to have them separate if the child is only called from a single parent?

max-sixty:
I hadn't realized it at the time, but I guess it'll also solve the issue in https://github.com/PRQL/prql/pull/3185 re too many workflows

eitsupi (approved):
I think it makes sense and I want to merge this to fix https://github.com/PRQL/prql/pull/3185#issuecomment-1658345218

PR #3185, opened 2023-07-31
ci: call release workflow from nightly workflow

eitsupi:
Previously, the contents of `nightly.yaml` were copied to `release.yaml` to build prqlc and prql-python with the same options as in the release, but this has been changed to call `release.yaml` from `nightly.yaml` to reduce code duplication.


eitsupi:
I tested on my fork and seems working fine.

eitsupi:
Seems limited by this https://github.com/orgs/community/discussions/39394
https://github.com/PRQL/prql/actions/runs/5714228549?pr=3185

max-sixty:
Looks good! 

I wonder whether we could run the same steps for both nightly and `test-all`, and so collapse the possible configs. But the downside of that is the build time, as we discussed on the previous PR. (Unless we think running nightly is sufficient?)

~For the GitHub limitation, would commenting out the snap / rpm workflows, until the limit is raised, get us under 20?~ edit: fixed with a different approach

---

Edit: `dry-run` now more means more like "fast run" — while "not being on a tag" makes it a dry run. (no great stress, this is a minor point, I hope it's reasonable to make it while being positive on the change)

max-sixty (approved):

max-sixty:
> For the GitHub limitation, would commenting out the snap / rpm workflows, until the limit is raised, get us under 20?

Very worst case, we could have `pull-request-1.yaml` & `pull-request-2.yaml`...!

max-sixty:
(CI error fixed by #3197)

eitsupi:
Thanks!

max-sixty:
Thank you!

PR #3203, opened 2023-08-01
ci: Skip rust tests on non-rust-changes

max-sixty:
Since we have the `changes` job, this is now possible


max-sixty:
(thanks @not-my-profile for adding the `changes` job!)

PR #3209, opened 2023-08-02
revert: #3208

max-sixty:
One dependency in `test-dbs` requires 1.70: https://github.com/PRQL/prql/actions/runs/5733147399/job/15537358810#step:9:114

...which means that technically our build isn't valid for all `Cargo.toml` specs. But it's only on our test feature, so prob fine — and not sure how to fix while remaining available on debian

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4,9 +4,9 @@ version = 3
 
 [[package]]
 name = "addr2line"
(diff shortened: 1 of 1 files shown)


max-sixty:
ref https://github.com/starkat99/half-rs/issues/97

PR #3216, opened 2023-08-02
refactor: chinook to tables

### LINKED ISSUE #3201: Rename "chinook" section in playground
The sidebar in the playground currently has the queries for the sample tables under the "chinook" heading.

Users might be confused what a "chinook" is ... I certainly was. "sample tables" or something like that would probably be more clear.

The key would apparently have to be changed in this object:

https://github.com/PRQL/prql/blob/97cdc93b317d19712fba7ae0414d4eeb8b50c4a6/web/playground/src/app/App.js#L27-L32

postmeback:
Fixes: #3201 

I saw that internally "chinook" word is used multiple times. So, rather than changing it entirely, I used an alias.

### DIFF
--- a/web/playground/src/app/App.js
+++ b/web/playground/src/app/App.js
@@ -26,7 +26,7 @@ class App extends React.Component {
   state = {
     library: {
       examples,
-      chinook,
+      tables: chinook,
       book,
       "local storage": loadLocalStorage(),
     },


max-sixty:
Thanks @postmeback ! 

Scores highly on impact per line :)

PR #3225, opened 2023-08-02
build: Add C++ minimal example

vanillajonathan:
A minimal C++ based upon the C example.

Includes the `libprql_lib.hpp` C++ header file, has a `using namespace` statement for the prql namespace and calls the `compile` function.

@aljazerzen can you peek at this?

I don't know if it possible to spruce it up with `std::string`, `std::vector`, for-each loop for messages, or something.

### DIFF
--- a/.github/workflows/test-lib.yaml
+++ b/.github/workflows/test-lib.yaml
@@ -35,3 +35,6 @@ jobs:
       - name: Run basic C example
         working-directory: bindings/prql-lib/examples/minimal-c
         run: make run
+      - name: Run basic C++ example
+        working-directory: bindings/prql-lib/examples/minimal-cpp
+        run: make run
--- a/bindings/prql-lib/examples/minimal-cpp/Makefile
+++ b/bindings/prql-lib/examples/minimal-cpp/Makefile
@@ -1,0 +1,17 @@
+PRQL_PROJECT=../../../..
+
+build-prql:
+	cargo build -p prql-lib --release
+
+# TODO: would be helpful to allow running with a debug build too.
+build: main.cpp build-prql
+	g++ main.cpp -o main.out \
+		-I${PRQL_PROJECT}/bindings/prql-lib \
+		-L${PRQL_PROJECT}/target/release \
+		${PRQL_PROJECT}/target/release/libprql_lib.a
+
+run: build
+	./main.out
+
+valgrind: build
+	valgrind ./main.out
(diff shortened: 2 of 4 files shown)


max-sixty:
Looks great, thanks @vanillajonathan !

max-sixty:
Could we add this to the `test-lib.yaml` job? https://github.com/prql/prql/blob/4d6b8b80d52ac91acdbbc199d34552c1484c539e/.github/workflows/test-lib.yaml#L36

---

(one other thing — though not in this PR — is to put these documented commands, such as `make run`, and the build commands in the Readme — https://github.com/prql/prql/blob/8ecfdebedb031042b78d9ecfbb83910a5db1494a/bindings/prql-lib/README.md#L99 — as tasks in the taskfile — then they'll be tested and easy to reference. Even better than docs is running code!)

eitsupi:
Please check the MegaLinter's warnings.
https://github.com/PRQL/prql/actions/runs/5743682548/job/15568484487#step:4:768

max-sixty:
> Please check the MegaLinter's warnings. https://github.com/PRQL/prql/actions/runs/5743682548/job/15568484487#step:4:768

Some of these are superfluous — e.g. 
- `  bindings/prql-lib/libprql_lib.h:0:  No copyright message found.  You should have a line: "Copyright [year] <Copyright Owner>"  [legal/copyright] [5]` — I don't think we should add this (my understanding is that it's not necessary to retain copyright, even if that were something we cared about)

I don't know C well enough to know whether these are legit or not: 

-   `bindings/prql-lib/examples/minimal-cpp/main.cpp:4:  Include the directory when naming header files  [build/include_subdir] [4]`
-   `bindings/prql-lib/examples/minimal-cpp/main.cpp:6:  Do not use namespace using-directives.  Use using-declarations instead.  [build/namespaces] [5]`

Re the formatting: 
- if there's an autoformatter for these files, great, let's use that. If there isn't, we shouldn't let formatting slow us down.

So overall — @vanillajonathan if you want to make refinements for these, great, but we don't *need* to fix in order to merge... (unless others disagree?)

vanillajonathan:
Okay, I added it as a task to the `test-lib.yaml` file.

As for the "no copyright message" lint, the [`cbindgen.toml`](/PRQL/prql/blob/main/bindings/prql-lib/cbindgen.toml) configuration file could be modified to include a copyright header if we would like to.

As for the two warnings in the `main.hpp` I don't know C++ well enough. I would like to leave the warnings for now, maybe someone with C++ expertise can improve the example.

aljazerzen (approved):

max-sixty:
Thanks a lot @vanillajonathan . Very nice, reasonably-sized contribution.

Issue #3223, opened 2023-08-02
Run post-release steps as part of workflow


max-sixty:
### What's up?

I think we can probably make an action to run the post-release steps as part of the release workflow, and then PR the commit in:

https://github.com/PRQL/prql/blob/49783393503bb237f7a7c6d8abd3bc3866b0beac/web/book/src/project/contributing/development.md?plain=1#L413-L416

Even though it's not a _huge_ time-saver, it makes doing releases easier, lets more folks do them, and removes the chance of mistakes, which can be painful (https://github.com/PRQL/prql/issues/3051#issuecomment-1650654463)

---

As an optional extra step, once this is done, we _might_ be able to remove the [`check-published`](https://github.com/prql/prql/blob/bec5168978ea0955fce8e49bd5a011b1efcfb906/.github/workflows/test-all.yaml#L162) step, since the version will reliably change from the published version immediately. (there would be a few minutes of overlap, but very short, and we could even skip the required CI jobs so it's immediate)

PR #3226, opened 2023-08-02
ci: Fix `check-published` job

max-sixty:
This wasn't working in https://github.com/PRQL/prql/actions/runs/5743166103/job/15570503983?pr=3222#step:9:19

(Though https://github.com/PRQL/prql/issues/3223 might supersede this...)


eitsupi (approved):
Thanks for fixing this!

PR #2885, opened 2023-06-20
chore: bump clio from 0.3.1 to 0.3.3

aj-bagwell:
This bumps clio to version 0.3.2

It needs a rust version bump as clio now use [`PathBuf::as_mut_os_string`](https://doc.rust-lang.org/std/path/struct.PathBuf.html#method.as_mut_os_string) which is only stablized in 1.70.0.

Clio 0.3.2 added `is_tty()` which means that `atty` goes to being a transative dependency.

PR #3240, opened 2023-08-03
fix: Detect tty for prqlc prompt

### LINKED ISSUE #3228: Can the `prqlc` prompt only show when waiting?
### What's up?

We have a prompt which shows when running `prqlc compile`, to avoid appearing to hang while waiting for an input — "Enter PRQL, then press ctrl-d to compile:`" 

I think that's great, I quite dislike the confusion in CLIs where it's unclear if they're thinking or waiting. IIRC, doing this is in a bunch of "CLI best practices" guides

But it also shows `Enter PRQL, then press ctrl-d to compile:` when something is piped in:

```sh
$ echo "from artists" | prqlc compile -
Enter PRQL, then press ctrl-d to compile:

SELECT
  *
FROM
  artists

-- Generated by PRQL compiler version:0.9.2 (https://prql-lang.org)
```

Weirdly I can't remember that happening previously; do others?

Is there some way we can only show when it's waiting, and not when piping text in?

max-sixty:
Closes https://github.com/PRQL/prql/issues/3228

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,9 @@
 
 **Fixes**:
 
+- `prolc` no longer displays a prompt when piping a query into its stdin
+  (@max-sixty, #3248).
+
 **Documentation**:
 
 **Web**:
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -283,7 +283,7 @@ version = "41.0.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "2b71d8d68d0bc2e648e4e395896dc518be8b90c5f0f763c59083187c3d46184b"
 dependencies = [
- "bitflags 2.3.1",
+ "bitflags 2.3.3",
 ]
 
 [[package]]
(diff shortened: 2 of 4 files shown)


max-sixty:
FYI @aj-bagwell — not sure if this is a helpful @ or not — feel free to ignore if not. 

The approach in https://github.com/PRQL/prql/pull/2885/files#diff-5042ecb23fbc2e8750e6452e008e9e96007bc06be4dff012653aedeaede02865R410 didn't seem to detect a `tty` correctly (or at least how I'm thinking about a TTY, where someone is typing into stdin, rather than piping text in). 

More details in #3228. I spent 20 minutes looking through dependencies — it seems like `clio` is calling the same code, though the code changes slightly by versions, so I'm likely missing something.

We're v happy to stick with this for a bit while upgrades percolate through the ecosystem.

eitsupi (commented):
Thanks for working on this!
Could we add a test for this?

max-sixty:
> Thanks for working on this! Could we add a test for this?

For sure we should have one — but it should be in our tests already! e.g. https://github.com/PRQL/prql/blob/9743446298ce4e72424239c234288a6c54173d74/crates/prqlc/tests/test.rs#L65

I think possibly it isn't picking it up because it's not a terminal, whereas the error was happening while we were in a terminal but not a tty? (I'm not tight on these concepts...)

eitsupi:
Thank you for pointing that out. It seems difficult to test this....

PR #3246, opened 2023-08-04
ci: Automerge pre-commit-ci updates & backports

max-sixty:


max-sixty:
40% chance this will work first time, but fixes should be easy!

PR #3263, opened 2023-08-06
feat: Limit quotes to an odd number of quotes

max-sixty:
With an even number of quotes, it's ambiguous whether it represents an empty string or the start of a new string — i.e. `""""` could be:
- `""` followed by `""` 
- or an opening `""""`.


aljazerzen:
I do think this is a step in right direction, but maybe we should rethink the "multiple quotes" strategy to quoting more broadly.

aljazerzen (approved):
In any case, we could use more tests for string parsing.

Issue #2989, opened 2023-07-16
r-strings are not mentioned in the documentation


vanillajonathan:
### What's up?

The documentation mentions f-strings and s-strings, but there is no mention of r-strings.

m-span:
addressed in #3281

PR #3281, opened 2023-08-08
docs: Re-write string docs

max-sixty:
I'm not sure we actually should have done the `r` strings — we're compiling for another language, not the final result, so `\t\t` would have been pretty reasonable to keep as `\t\t` rather than two tab characters.

I'm not in a rush to revert though — folks can always just use the r-strings to get the same thing.


max-sixty:
Closes #2989

Issue #3290, opened 2023-08-08
Cutting time from CI


max-sixty:
### What's up?

We're getting non-trivial CI queues. I'm trying to ensure CI is fast, since I worry that once queues start getting long, they don't recover — folks do fewer bigger PRs, care less about instant feedback, we feel less pressure to keep CI times short, we get longer queues — somewhat as described in https://matklad.github.io/2021/09/04/fast-rust-builds.html.

We also have an annoying effect in our pipeline around merging lots of PRs quickly; for more details:

<details>

- For a PR to merge the `check-ok-to-merge` job needs to run. It waits for most jobs within that PR's pipeline before running.
- When there's a queue, we want to merge things as soon as possible, because merging new code causes the old CI pipeline to be replaced by the new one, saving the wasted work of running tests on non-current code
- Because the `check-ok-to-merge` job waiting, the job is then queued _behind_ tests from later PRs, which can run immediately
- So CI waits for later PRs to finish their pipelines before running `check-ok-to-merge` on a previous PR and so before merging that PR
- As a result, we often can't merge quicker than the time it takes to run a full `test-all` pipeline, which is ~15 minutes atm, even we're otherwise unconstrained.

</details>

Here's a snapshot of where `test-all` spends its time: https://github.com/PRQL/prql/actions/runs/5798841411/usage

### Some ideas

- [x] Don't run bindings' tests on multiple platforms in `test-all`, at least if there are no changes. We can run them on nightly.
  - This would cut a lot — ~2/3 python jobs~, 1/2 java jobs, 1/2 elixir jobs, ~2/3 js jobs~ (edit with the ones done)
- [ ] Don't run `test-release-dry-run` in `test-all` — move to nightly. Bugs here are very rare (though very annoying when they do happen, so it's good we have a test here). 
  - We could also unify a "dry-run" (which is really a "quick-run") and running when not on a tag, which is really a dry-run. (this would also obviate `check-published`...)
- [x] Don't run `test-all` twice on some PRs — because `test-all` can be triggered in two ways, it can sometimes run twice.
- [x] Add back caching for `test-js`; it could probably share a cache with `build-web`
- ~[ ] Don't run `build-prqlc` on a standard PR. By reducing `test-all`, we can also make adding a `pr-test-all` label more palatable in order to generate a `prqlc` binary.~ Decided not to do this

eitsupi:
[...]

PR #3291, opened 2023-08-08
ci: Consolidate `test-js`, restrict to ubuntu outside nightly 

max-sixty:
One step on #3290


max-sixty:
This turned out much much more complicated than I expected. Going to pause and reflect if this is worthwhile

max-sixty:
OK, by using JSON, it's now better. Still too complicated, will think about whether it's worth extending

Issue #3271, opened 2023-08-06
Table name and transform name collisions


FinnRG:
### What happened?

I have a table named `group`, which I tried to query. I tried using backticks (as suggested in the docs: https://prql-lang.org/book/reference/stdlib/transforms/from.html), which doesn't seem to work.

### PRQL input

```elm
from `group`
filter id == $1
```

### SQL output

```SQL
No output
```

### MVCE confirmation

- [X] Minimal example
- [X] New issue

### Anything else?

Exact error:
```Error: 
   ╭─[:2:1]
   │
 2 │ filter id == $1
   │ ───────┬───────  
   │        ╰───────── main expected type `relation`, but found type `infer -> infer`
   │ 
   │ Help: Have you forgotten an argument to function main?
   │ 
   │ Note: Type `relation` expands to `[tuple_of_scalars]`
───╯
```

max-sixty:
Thanks @FinnRG . Yes, repro-ed. And a particularly bad error message...

As a workaround in the meantime, this works:

```prql
from s"SELECT * from group"
```

```sql
WITH table_0 AS (
  SELECT
    *
  from
    group
)
SELECT
  *
FROM
  table_0

-- Generated by PRQL compiler version:0.9.3 (https://prql-lang.org)
```

aljazerzen:
This is expected behavior. An unfortunate one, but a compound of how we do name resolution and table inference.

So the problem is that identifier `group` matches `std.group` (the group function) instead of `default_db.group` (the table declaration).

The solution is this:

```
from default_db.group
filter id == $1
```
... except that now, this will compile to `SELECT * FROM default_db.group ...`, because this will infer that we are referring to `default_db.default_db.group`.

The real solution is:

```
default_db.group
filter id == $1
```
... which bypasses `from` function and its behavior of defaulting to `default_db` module.

max-sixty:
Completely abstract of the current implementation, I do think that's a bit unfriendly / inconsistent.

I recognize we do need to make some tradeoffs because of PRQL's use of "bare words" — there'll always be some ambiguity. But (thinking on the fly here, not super confident):
- there would be a robust way which machines could write — e.g. in column selection, we can always have `this.foo`, then `this.time` will resolve to a column even when `time` resolves to a type.
- the robust way would follow from the less robust way. My objection to the current implementation is that if a machine wanted to generate PRQL, it would generate `default_db.foo` rather than `from foo`, which then doesn't look much like PRQL.

---

[...]

PR #3292, opened 2023-08-08
docs: Update `from` docs

FinnRG:
Related to #3271.


max-sixty:
Great! Do you have an env set up to run the tests and write the resulting SQL? Otherwise I'm happy to

FinnRG:
@max-sixty I've updated the snapshots

max-sixty:
Thanks @FinnRG ! Happy to have you as a contributor.

(FYI for transparency I'm going to add a note to this so folks don't think this is a good state)

PR #3304, opened 2023-08-09
ci: Create issue when nightly workflow fails

max-sixty:


max-sixty:
I expect this won't be great the first time but we can iterate

Issue #3168, opened 2023-07-29
`from` isn't a transform


not-my-profile:
The book refers to `from` as a transform ... which is wrong ... a `from` doesn't transform anything.

https://github.com/PRQL/prql/blob/35e511d9328fdb323f8c1914b942045ecbd42d35/web/book/src/reference/stdlib/transforms/README.md?plain=1#L12-L16

I think this conflation is confusing since transforms can be at any point in the pipeline while `from` cannot ... since it's not a transform.

max-sixty:
~Yes, we could possibly split that section into Transforms and Keywords?~ Actually don't think this a great idea on reflection

richb-hanover:
_Contrary viewpoint warning:_ I suppose `from` is not truly a transform. Although it _does_ turn "nothing" into a table/relation, it cannot be present in the middle of a pipeline, so probably doesn't meet the true definition of "transform".

But from a documentation standpoint, there is real value in retaining it in the current [Transforms](https://prql-lang.org/book/reference/stdlib/transforms/index.html) section. That page is the handy-dandy reference for all PRQL statements, and sending the reader to some other page for this one statement would be awkward.

The [from](https://prql-lang.org/book/reference/stdlib/transforms/from.html) page could state these limitations (not really a transform, must begin a pipeline, can't be in the middle) and I don't think anyone will be confused or led astray. Thanks.

not-my-profile:
I did not suggest it to be removed from the transforms page but the page should be renamed to Functions if it is to list `from`. Or alternatively the From page should be moved out of the section but it could still be linked from the page.

> I don't think anyone will be confused or led astray

I disagree. Listing `from` under "transforms" is just misleading.

max-sixty:
I do think there's value in having a "PRQL primitives" grouping, basically answering "what are my tools for transforming data", which includes how to pull in data.

We've called this "Transforms", because they're mostly transforms. Possibly `from` isn't technically a transform, though we could define it as transforming nothing into something...

What are other ways of grouping this?

<img width="117" alt="image" src="https://github.com/PRQL/prql/assets/5635139/496062e3-7734-44b8-8eea-24fc693bc444">

One approach would be to call it something like "Primitives", add in `loop`? Or even just flatten all of those to the Standard Library?

---

[...]

PR #3312, opened 2023-08-11
docs(book): Document `loop` under transforms

not-my-profile:
See https://github.com/PRQL/prql/issues/3168#issuecomment-1661959577.


max-sixty (approved):
👍  from me

PR #3322, opened 2023-08-11
ci: Pass features correctly in prqlc build

max-sixty:


.github/actions/build-prqlc/action.yaml
```diff
@@ -43,15 +43,15 @@ runs:
     - if: runner.os == 'Windows' && inputs.profile == 'release'
       shell: bash
       run: |
-        echo 'RUSTFLAGS=-Ctarget-feature=+crt-static' >>"$GITHUB_ENV"
+        echo 'RUSTFLAGS=-Ctarget-feature=+crt-static' >> "$GITHUB_ENV"
```
eitsupi, line 46: Why was this change made?
The style before the change was consistent with the formatting by shfmt, and I don't see why it should be changed here. I think it will only make the history harder to understand.

max-sixty: Ah sorry, I hadn't realized the existing one was the standard approach — we can revert...

eitsupi: I see. Never mind. It doesn't make much difference.

max-sixty: May as well! I put in #3326...

PR #3326, opened 2023-08-12
style: Use shfmt for expressions in CI

max-sixty:
If there's some way to get `shfmt` for all expressions here, that would be nice...


max-sixty:
CC @eitsupi , thanks for pointing it out...

Issue #3145, opened 2023-07-28
Unify compilation across CI jobs


max-sixty:
### What's up?

One frustration with the [0.9.0-2 release polava](https://github.com/PRQL/prql/issues/3051) was that builds took ages. While we were trying to fight fires — I made mistakes on things, the playground broke, etc — it was difficult to get fast feedback from CI, or get the updates out. 

Because our builds are very parallel, it's not so much of a problem generally. But when we have lots of PRs, and when some of those call `test-all`, then the queue gets long. Building `prqlc` is a big share of the`pul-request` CI time (though not much of the `test-all` time)

https://github.com/PRQL/prql/pull/3117 has made compilation faster, but building `prqlc` is still doing lots of recompilation because:
- prql's workspace builds with different dependency features than `prqlc` does — https://github.com/PRQL/prql/issues/3098
- prql's features are different across the two CI jobs because of https://github.com/prql/prql/blob/a9db1e26fe8380626f2fe0075837dc0d44e72cb5/.github/workflows/test-rust.yaml#L82-L85

A couple options:
- Cache them separately. But we're completely out of cache space — we use all the 10GB on a single cache version — so we can't do this without dropping something else from the cache, and I'm not sure what we'd drop
- Have a separate action which compiles rust, encapsulating the configs for that target. It would basically run this https://github.com/prql/prql/blob/a9db1e26fe8380626f2fe0075837dc0d44e72cb5/.github/workflows/test-rust.yaml#L82-L85. Both `test-rust.yaml` and `build-prqlc` would call it
- Spend more time unifying duplicate dependency features, e.g. https://github.com/PRQL/prql/pull/3144. (But this doesn't seem to scale well — how do we ensure things don't creep back in??)

Other quick ideas for speeding up CI:
- ~Drop `prqlc` from the PR build; run only on `pr-test-all` label~ Edit: decided not to atm
- Reduce the size of `test-all`, possibly along the lines of https://github.com/prql/prql/blob/a9db1e26fe8380626f2fe0075837dc0d44e72cb5/.github/workflows/test-all.yaml#L11-L16. It's the huge parallel job, so it's the thing causing queues to form. (edit: moved to #3290)

max-sixty:
In the end we build the cache in a single place, and try and ensure other jobs which use it use the same flags; e.g. `--features` and `--all-targets` & env. 

It's a bit of a whack-a-mole — #3331 was one example where a seemingly innocuous flag made a huge difference.

PR #3331, opened 2023-08-12
ci: Another attempt at prqlc caching

max-sixty:


max-sixty:
It works! Down to 1.5mins from 9mins, for two jobs we run on every commit.

Enough to close #3096, close #3145 & contributes to #3290

PR #3332, opened 2023-08-13
docs(book): remove outdated R example

eitsupi:
The latest prqlr supports PRQL 0.9.

Perhaps it would be better to just link to the prqlr documentation instead of writing an untested code block here.


max-sixty (approved):
Good idea...

Issue #3149, opened 2023-07-28
How can I run `pre-commit.ci` before pushing to my repo?


richb-hanover:
### What's up?

`pre-commit.ci` makes a bunch of (useful) changes after I push to my repo. This dirties up the commit history.

How can I run it prior to pushing? Is there a `task run...` that will do this? Thanks.

max-sixty:
(4) [here](https://pre-commit.com/)!

richb-hanover:
To flesh out this answer for the Dev Container: use `pre-commit run --all-files` To set this up:

- I installed `pre-commit` within the Dev Container with `pip install pre-commit`
- (Optional) I updated pip with `python -m pip install --upgrade pip`
- I then ran `pre-commit run --all-files` to check the files prior to committing them.

PR #3333, opened 2023-08-14
devops: Install pre-commit into devcontainer

max-sixty:
Does https://github.com/PRQL/prql/issues/3149#issuecomment-1676590241 automatically

### DIFF
--- a/.devcontainer/base-image/Dockerfile
+++ b/.devcontainer/base-image/Dockerfile
@@ -12,6 +12,8 @@ RUN apt-get update \
 COPY Taskfile.yml /tmp/Taskfile.yml
 USER vscode
 
-RUN task -t /tmp/Taskfile.yml install-cargo-tools-binstall
+RUN task -t /tmp/Taskfile.yml install-cargo-tools-binstall \
+    && task install-pre-commit \
+    && pre-commit install-hooks
 
 USER root


eitsupi (changes requested):
I think the installation command is not optimized. There is a cache issue.

And it is definitely more appropriate to do it in postCreateCommand instead of on the Dockerfile.

.devcontainer/base-image/Dockerfile
```diff
@@ -12,6 +12,8 @@ RUN apt-get update \
 COPY Taskfile.yml /tmp/Taskfile.yml
 USER vscode
 
-RUN task -t /tmp/Taskfile.yml install-cargo-tools-binstall
+RUN task -t /tmp/Taskfile.yml install-cargo-tools-binstall \
+    && task install-pre-commit \
+    && pre-commit install-hooks
```
eitsupi, line 17: Does this work?

eitsupi (changes requested):

.devcontainer/base-image/Dockerfile
eitsupi: `pre-commit install-hooks` would make the commit fail when committing outside the container.

max-sixty: > `pre-commit install-hooks` would make the commit fail when committing outside the container.

Hmm, why?

NB: this is different from `pre-commit install` — it just installs the current hooks, it doesn't have them run on each commit...

max-sixty:
> There is a cache issue. And it is definitely more appropriate to do it in postCreateCommand instead of on the Dockerfile.

I was thinking it would be better for the cache to do them here, so that it's cached? 

Or the versions will be out of date and so the hooks will often need to be reinstalled?

(tbc you know this better, I defer to you...)

.devcontainer/base-image/Dockerfile
eitsupi: > NB: this is different from `pre-commit install` — it just installs the current hooks, it doesn't have them run on each commit...

Ah, my mistake. Thanks for pointing that out.

eitsupi:
The installation of Rust tools takes so long that we have no choice but to install them on the Dockerfile, but I think it is far more flexible for users to install things that take an extremely long time to install, rather than including them in the pre-build image.

> I was thinking it would be better for the cache to do them here, so that it's cached?

Sorry I couldn't explain it adequately from my smart phone. What I meant was that the `pip install` command bakes the cache file into the container. 

https://github.com/PRQL/prql/blob/34507c322da0bd69bc5983c56b156d8ad4f83310/Taskfile.yml#L144-L145

Here it is common to use the `--no-cache-dir` option.

eitsupi:
I looked at it again and thought that I should really rewrite the contents of the Taskfile.....

We should not be forced to install something as huge as a language runtime via Homebrew....

eitsupi:
Closing in favor of #3344

.devcontainer/base-image/Dockerfile
eitsupi: As an addendum, I think `task install-pre-commit` will fail because it cannot find the Taskfile.

max-sixty:
> Sorry I couldn't explain it adequately from my smart phone. What I meant was that the `pip install` command bakes the cache file into the container.

Ah yes, very good point...

> We should not be forced to install something as huge as a language runtime via Homebrew....

I agree it's not great. What's the alternative? Maybe we stop & print some instructions if python isn't installed, in the same way we do for VS Code extensions?

max-sixty:
> The installation of Rust tools takes so long that we have no choice but to install them on the Dockerfile, but I think it is far more flexible for users to install things that take an extremely long time to install, rather than including them in the pre-build image.

To confirm, do you mean "for users to install things that _don't_ take an extremely long time to install"?

eitsupi:
> To confirm, do you mean "for users to install things that _don't_ take an extremely long time to install"?

I means that the installation is done via user side by `postCreateCommand` etc. using the definition files that are included in the repository at the time.

eitsupi:
> I agree it's not great. What's the alternative? Maybe we stop & print some instructions if python isn't installed, in the same way we do for VS Code extensions?

I think that would be better.

For example, I think the current command forces Python3 installation via brew, even for those who prefer to build and install Python from source. This is quite surprising.
It is far better to fail to build this project than to be forced to install something.

PR #3343, opened 2023-08-14
ci: Add `web` cache to `js` test

max-sixty:


max-sixty:
This does make it a bit faster — down to 4min from 7min on ubuntu.

PR #3346, opened 2023-08-15
devops: Don't auto-install brew deps

max-sixty:
As discussed in https://github.com/PRQL/prql/pull/3333#issuecomment-1678374434

(need to have another look in the morning)

### DIFF
--- a/Taskfile.yml
+++ b/Taskfile.yml
@@ -48,6 +48,8 @@ tasks:
     preconditions:
       - sh: which clang
         msg: |
+
+
           🔴 Can't find `clang`, which is required to install a full development environment (we use `duckdb` in our tests, which requires it).
 
           Please install it. On MacOS, that's
@@ -169,6 +171,7 @@ tasks:
     cmds:
       - cmd: |
           echo "
+
           🟡 It looks like at least one brew dependency is missing from:
 
           {{ .brew_dependencies }}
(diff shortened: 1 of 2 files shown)


Taskfile.yml
```diff
@@ -26,6 +26,12 @@ vars:
     bacon cargo-audit cargo-insta cargo-release default-target mdbook
     mdbook-admonish mdbook-footnote mdbook-toc wasm-bindgen-cli wasm-opt
     wasm-pack
+  brew_dependencies: |
+    brew
+    hugo
+    npm
+    python3
+    elixir
```
eitsupi: I'm wondering if it's worth including elixir here.
I don't think prql-elixir has been substantially edited in the last few months.

max-sixty: Yes OK

max-sixty:
Having some trouble with the devcontainer...

This last passed two weeks ago.

I know you're keen on not running this too often, but I do wonder whether running this nightly would help us understand the cause when we get a problem like this — is it the parent image? Is it `cargo binstall`? I'm not sure the best way to figure this out...

---

I added a fallback to installing from source [here](https://github.com/PRQL/prql/pull/3346/files#diff-cd2d359855d0301ce190f1ec3b4c572ea690c83747f6df61c9340720e3d2425eR109), but now the binaries aren't installed to the `PATH`: https://github.com/PRQL/prql/actions/runs/5870585733/job/15917918066?pr=3346#step:6:1016

Taskfile.yml
```diff
@@ -22,17 +22,34 @@ vars:
...
+    cargo-release
+    default-target
+    mdbook
+    mdbook-admonish
+    mdbook-footnote
+    mdbook-toc
+    wasm-bindgen-cli
+    wasm-opt
     wasm-pack
+  # Excluding `elixir` atm given it's not enabled on Mac and not yes supported
```
```suggestion
  # Excluding `elixir` atm given it's not enabled on Mac and currently unsupported
```

Taskfile.yml
```diff
@@ -22,17 +22,34 @@ vars:
...
+    python3
 
 tasks:
   # main installer is "setup-dev" which calls other tasks
   setup-dev:
     preconditions:
       - sh: which clang
         msg: |
+
+
```
```suggestion
```

max-sixty:
Merging this despite the devcontainer failure — it's unrelated.

max-sixty:
> is it the parent image? Is it `cargo binstall`? I'm not sure the best way to figure this out...

Looks like that issue is related to https://github.com/cargo-bins/cargo-binstall/issues/1292

PR #3353, opened 2023-08-18
ci: change Winget Releaser job to `ubuntu-latest`

sitiom:
Winget Releaser supports non-Windows runners, and `ubuntu-latest` is generally faster.

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -70,7 +70,7 @@ jobs:
           files: ${{ steps.build-artifact.outputs.artifact-name }}
 
   winget-release:
-    runs-on: windows-latest
+    runs-on: ubuntu-latest
     needs: build-prqlc
     if: github.ref_type == 'tag' && !inputs.dry-run
     steps:


max-sixty:
Thanks @sitiom , good to have you as a contributor.

LGTM; @eitsupi knows this better so I'm tagging them, if that's OK.

eitsupi (commented):
Thanks, but Komac seems having a bug on non-windows platforms? (russellbanks/Komac#110)

Since @sitiom looks the author of the issue, I think we can merge if @sitiom is sure it's okay to use Linux here.

sitiom:
> Thanks, but Komac seems having a bug on non-windows platforms? ([russellbanks/Komac#110](https://github.com/russellbanks/Komac/issues/110))

The only issue now is that it's not saving the GitHub token and you would have to enter it every time when using Linux or Mac. That should not affect this action since the token is passed via the environment variables set by the action inputs.

eitsupi (approved):

eitsupi:
> The only issue now is that it's not saving the GitHub token and you would have to enter it every time when using Linux or Mac. That should not affect this action since the token is passed via the environment variables set by the action inputs.

Thanks, let's merge this now.

PR #3357, opened 2023-08-19
ci: Try getting devcontainer to build on arm64

max-sixty:


richb-hanover:
@max-sixty Slightly unrelated, but since you're editing _Taskfile.yml_, could you add a "pre-commit" task that invokes `pre-commit run --all-files`? That'd decrease the cognitive load for ~~people using the Dev Container~~ me to force reformatting of files prior to pushing to Github... Thanks.

max-sixty:
> add a "pre-commit" task that invokes `pre-commit run --all-files`? That'd decrease the cognitive load for ~people using the Dev Container~ me to force reformatting of files prior to pushing to Github... Thanks.

Would this be easier than `pre-commit run -a` though?

max-sixty:
This seems to work, so merging. It'll run every night at the moment, but won't push. 

Open to feedback on it running every night — it does seem quite expensive, but it breaking is also quite annoying, and confusing if it breaks after not having run in a while.

richb-hanover:
> Would this be easier than pre-commit run -a though?

@max-sixty Yes - it's easier because it decreases cognitive load for everyone who follows by embodying all the knowledge about that action into one "name". 

This means I can simply type `task` in the Dev Container and be reminded of the name of what I want to do. (My current alternative is a Joplin (Evernote replacement) file that holds a bunch of secret invocations...)

Issue #3345, opened 2023-08-15
Release plan for 0.9.4


eitsupi:
It has been about 2 weeks since the last release and I am wondering if I should make the next release.
Would it be worthwhile to do a release since several bug fixes have been made?

eitsupi:
@max-sixty I was surprised to read the documentation, are you writing the "New Contributors" section manually?
How do you find new contributors?

(In general, I imagine such sections are automatically generated.)

max-sixty:
Would be great to do a release if you'd be up for it! Thank you! Happy to review anything.

If you do, let me know what you think of the process / how it can be improved. I've tried to make it fairly automated. Most of the work is updating the Changelog, but in the current state that's not really necessary — only when we have a big release like 0.9.0...

---


> are you writing the "New Contributors" section manually?

New contributors are generated by GitHub, see https://github.com/PRQL/prql/blob/49f8bd9e12b43eeb286b6cf993ca8889896ebf75/web/book/src/project/contributing/development.md#L384 (maybe not that clearly described by me...)

eitsupi:
I opened a PR for to update the changelog #3379

eitsupi:
0.9.4 was released.

PR #3359, opened 2023-08-19
docs: Document command to reset snapshot tests

### LINKED ISSUE #3358: How can I rebuild all the snapshots?
### What's up?

Using the Dev Container, what is the minimal set of commands that will rebuild all the SQL snapshots after modifying a PRQL query in the book?

I think `task test-all` will do it, but that takes forever (~20-25 minutes from a new Dev Container because it re-built hundreds of packages a couple times).

max-sixty:
Closes #3358, thanks for the question


max-sixty:
Leaving until #3345

PR #3363, opened 2023-08-20
ci: Remove `test-all` workflow

### LINKED ISSUE #3290: Cutting time from CI

max-sixty:
Closes #3290, as discussed in https://github.com/PRQL/prql/issues/3290#issuecomment-1675647441

`test-all` is now completely consolidated into `pull-request`. 

This consolidates the _testing_ of `release.yaml` into nightly builds, which run on schedule and on very occasional PRs. Within this, it consolidates "dry-run" with "not running on a tag". Given we run it less frequently, I think it's simpler to only have two possible modes, but very open to feedback.

I said I didn't want to step on toes, so won't merge this without consensus. And again, the notion of a "test release" is excellent and this doesn't reduce the utility we get from it -- this change is a consolidation, hope that comes across...


eitsupi (approved):
I think this makes sense, thanks!

PR #3372, opened 2023-08-22
build: Add Zig minimal example

vanillajonathan:
This is a Zig example which uses the `libprql_lib.h` C header file and loads the PRQL library using dynamic linking.

### DIFF
--- a/.github/workflows/test-lib.yaml
+++ b/.github/workflows/test-lib.yaml
@@ -37,6 +37,21 @@ jobs:
   test-zig:
     runs-on: ubuntu-latest
     steps:
+      - name: 📂 Checkout code
+        uses: actions/checkout@v3
+      - run: ./.github/set_version.sh
+      - name: 💰 Cache
+        uses: Swatinem/rust-cache@v2
+        with:
+          prefix-key: ${{ env.version }}
+          save-if: ${{ github.ref == 'refs/heads/main' }}
+          shared-key: lib
+      - name: Build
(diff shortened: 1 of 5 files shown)


max-sixty:
Cool!

Could we add a small test? Otherwise it's likely to decay.

(an alternative would be posting it to a gist / a blog, if you prefer to just publish some code quickly)

vanillajonathan:
I have now added a unit test.

max-sixty:
Yes I also mean to _run_ the test in CI...

I think that could be as small as just adding a job to https://github.com/PRQL/prql/blob/f9a983d1bc0d139b94246aa240d2c59edc35d709/.github/workflows/test-lib.yaml, which is easier that making a new workflow...

.github/workflows/test-lib.yaml
```diff
@@ -34,3 +34,11 @@ jobs:
       - name: Run basic C++ example
         working-directory: bindings/prql-lib/examples/minimal-cpp
         run: make run
+      - name: 🔧 Setup Zig
```
max-sixty, line 55: Let's do a separate job within this workflow though, so it's clearly delineated, particularly when there's a failure. You can copy & paste most things up to this line.

vanillajonathan: Okay, I've done that now.

max-sixty:
Looks great, thanks @vanillajonathan !

max-sixty:
Thanks @vanillajonathan !

PR #3376, opened 2023-08-23
build: Update Snap from core20 to core22

vanillajonathan:
Update to core22 which is the new recommended base.


max-sixty:
Thanks a lot!

PR #3378, opened 2023-08-23
ci: Fix job skips

max-sixty:
Based on https://github.com/re-actors/alls-green/issues/23

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -465,6 +465,7 @@ jobs:
       - name: Decide whether the needed jobs succeeded or failed
         uses: re-actors/alls-green@release/v1
         with:
+          jobs: ${{ toJSON(needs) }}
           # We don't include `check-links-markdown`, since occasionally we'll want to merge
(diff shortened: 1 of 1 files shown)


.github/workflows/tests.yaml
```diff
@@ -478,9 +478,29 @@ jobs:
...
-          # Easier than writing out every job — and it's our own rules deciding
-          # whether jobs are skipped or not.
-          allowed-skips: ${{ toJSON(needs) }}
+          # We skip jobs deliberately, so we are OK if any are skipped.
+          #
+          # Copy-pasted from `needs`, since it needs to be a json list, so `${{
+          # toJSON(needs) }}` (which is a map) doesn't work.
+          # https://github.com/re-actors/alls-green/issues/23
+          allowed-skips: |
+            [
```
webknjaz, line 487: FYI, you don't have to wrap this as a JSON list, a comma-separated list works too:
https://github.com/python/cpython/blob/809ea7c/.github/workflows/build.yml#L550-L577

PR #3394, opened 2023-08-24
ci: Also disable pushing cache in devcontainer build

max-sixty:


eitsupi:
Why?

Is making changes without any written reason an encouraged practice in this organization?

max-sixty:
Sorry, I should have referenced the previous issue: https://github.com/PRQL/prql/pull/3389

eitsupi:
> Sorry, I should have referenced the previous issue: #3389

Thanks, but I think inline cache is nothing to do with gha cache.

max-sixty:
I can open a PR to re-enable this, but it does seem to break the build...

Edit: possibly I'm mis-attributing the failure — I'll try re-enabling the inline cache. Sorry if I moved too quickly earlier, I wanted to fix the build and lots of things were failing

eitsupi:
Where is the log of failed builds due to inline cache?

max-sixty:
> Where is the log of failed builds due to inline cache?

(made an edit above, I may have mis-attributed this. Let's see if #3398 passes)

eitsupi:
I think you should have looked more into buildx if you didn't want to play around with this and wait for someone to review it.

max-sixty:
> I think you should have looked more into buildx if you didn't want to play around with this and wait for someone to review it.

OK, thanks for the feedback, I appreciate it.

eitsupi:
I don't know what is going on, but looking at the container registry, it seems that several useless pushes have been made in the last few hours anyway. Clearly, this is a waste of energy and storage, and the workflow needs to be fundamentally rethought....

max-sixty:
Sounds like you're a bit frustrated — sorry for my part in that. I have been doing lots of work on trying to get CI into a better place, very possibly I haven't given the devcontainer part of it enough attention.

Let me try and write down where it's at. I think it's probably not far from where it should be, but let's see. 

The current state is:
- `build-devcontainer` runs when 
    1. there are changes to the devcontainer files 
    2. or when there are changes to the workflow
- When it runs on a push to main, it also pushes its image
- It also runs every night as a test

It probably shouldn't push when there are only changes to the workflow. Today I tried lots of times to fix the workflow, and so made lots of changes, and so that resulted in lots of pushes.

It does seem to me (and maybe we disagree, happy to discuss) that the existing model of only ever running when the devcontainer files changed meant that it became quite difficult to understand why something broke.  I made that comment [here](https://github.com/PRQL/prql/pull/3346#issuecomment-1679427273):

> Having some trouble with the devcontainer...
> 
> This last passed two weeks ago.
> 
> I know you're keen on not running this too often, but I do wonder whether running this nightly would help us understand the cause when we get a problem like this — is it the parent image? Is it `cargo binstall`? I'm not sure the best way to figure this out...

(I also felt like this fell on me, while CI was broken, and I was spending a decent amount of time trying to fix it — though maybe because I'm doing a bad job at it!)

I can try and make two triggers — one for running a test, and one for pushing, so we don't push more than we need to. Would that fix the issues you're seeing? Do you disagree re running tests fairly often?

eitsupi:
@max-sixty I still have things like https://github.com/PRQL/prql/pull/2880#issuecomment-1598430007 and #2898 on my to-do list, but I'm sorry I haven't spent more time on them.

I think the division of jobs makes sense.

max-sixty:
OK thank you! 

Sorry again for changing some of the code you wrote & possibly making it worse. Hopefully it being worse was temporary, and it's now improved...

Thanks a lot for expressing yourself above, I really appreciate the feedback. Please continue to be frank when I make mistakes.

PR #3395, opened 2023-08-24
ci: Actually disable devcontainer cache

max-sixty:
Sorry, third time lucky, probably should have been more conscientious on the first two PRs


eitsupi:
Why disable inline cache?

max-sixty:
Sorry, I should have referenced the previous issue: https://github.com/PRQL/prql/pull/3389

PR #3398, opened 2023-08-24
ci: Re-enable inline docker cache

max-sixty:
While keeping GHA docker cache disabled.

Ref #3394


max-sixty:
So far so good @eitsupi — you were right...

max-sixty:
I'll merge since we don't see the failure we previously saw

PR #3402, opened 2023-08-25
build: Bump toolchain version to 1.72.0

max-sixty:
(this does not affect the MSRV)


max-sixty:
Should we do this now? `rust-analyzer` only supports latest, but it's been working fine for me. So fine to wait until `n-1` or we see any issues before merging

eitsupi (approved):
Thanks.

I too considered reverting ba0c7c74313acc1a2ceb047838c061e84b4c40c1 when Rust 1.72 was released yesterday, but decided against it when I saw the latest -1 set by CI.
For me, either merge now or when Rust 1.73 is released is fine.

max-sixty:
OK, I'll update on my recent experience being over-enthusiastic on merging CI changes and wait for a bit.

If anyone else has a view, feel free to say (or hit the button...)

max-sixty:
I am getting errors in rust-analyzer now, so I will merge. If this causes any issues, fine to immediately revert (as ever)

PR #3403, opened 2023-08-25
ci: Add a couple of cargo files to changes for nightly

max-sixty:


.github/workflows/tests.yaml
```diff
@@ -92,6 +92,8 @@ jobs:
             nightly:
               - .github/workflows/nightly.yaml
               - Cargo.lock
+              - rust-toolchain.toml
+              - .cargo
```
max-sixty:
```suggestion
              - .cargo/**
```

PR #3404, opened 2023-08-25
ci: Don't publish devcontainer on workflow changes

max-sixty:
Trying to address https://github.com/PRQL/prql/pull/3394#issuecomment-1692604266

(apologies again for not giving this enough thought prior)


.github/workflows/release.yaml
```diff
@@ -259,6 +259,8 @@ jobs:
         uses: actions/checkout@v3
       - run: git push origin HEAD:web --force
 
-  push-devcontainer-base-image:
+  publish-devcontainer:
```
eitsupi: I don't know if the word "publish" is appropriate here; I think "push" is more common.

.github/workflows/build-devcontainer.yaml
```diff
@@ -2,7 +2,15 @@ name: build-devcontainer
 
 on:
   workflow_call:
+    inputs:
+      publish:
+        type: boolean
+        required: true
```
eitsupi: Is `required: true` appropriate here? Or `default: false`?

.github/workflows/build-devcontainer.yaml
```diff
@@ -2,7 +2,15 @@ name: build-devcontainer
 
 on:
   workflow_call:
+    inputs:
+      publish:
```
eitsupi: Why not `push`?

.github/workflows/build-devcontainer.yaml
```diff
@@ -2,7 +2,15 @@ name: build-devcontainer
...
+        type: boolean
+        required: true
   workflow_dispatch:
+    inputs:
+      publish:
+        type: boolean
+        required: true
```
eitsupi: Ditto.

.github/workflows/build-devcontainer.yaml
```diff
@@ -41,17 +49,11 @@ jobs:
...
-          push:
-            # If it's a push to main, or a tag, or explicitly called, push the image to the registry.
-            # Don't if it's on a schedule.
-            ${{ (github.ref == 'refs/heads/main' && github.event_name == 'push')
-            || startsWith(github.ref, 'refs/tags/') || github.event_name ==
-            'workflow_dispatch' }}
+          push: ${{ inputs.publish == 'true' }}
           cache-from: |
             ${{ steps.meta.outputs.tags }}
             type=gha
```
eitsupi, line 56: This line can be removed.

eitsupi (changes requested):

.github/workflows/release.yaml
max-sixty: Yeah, I thought it was consistent with the rest of the workflow, but agree re docker conventions, will change

.github/workflows/build-devcontainer.yaml
max-sixty: Done!

.github/workflows/build-devcontainer.yaml
max-sixty: I thought it would be easier to revert the `cache-to` change by keeping this, and the comment below somewhat explains the current incongruity between them.

Will add a comment (but feel free to remove this if I'm not making sense :) )

eitsupi: Hmmm, given the fact that communicating with gha can cause problems, wouldn't leaving gha here cause a CI failure?

.github/workflows/build-devcontainer.yaml
```diff
@@ -41,17 +49,12 @@ jobs:
           file: .devcontainer/base-image/Dockerfile
           tags: ${{ steps.meta.outputs.tags }}
           platforms: linux/amd64, linux/arm64
-          push:
-            # If it's a push to main, or a tag, or explicitly called, push the image to the registry.
-            # Don't if it's on a schedule.
-            ${{ (github.ref == 'refs/heads/main' && github.event_name == 'push')
-            || startsWith(github.ref, 'refs/tags/') || github.event_name ==
-            'workflow_dispatch' }}
+          push: ${{ inputs.push == 'true' }}
```
eitsupi, line 52: This can be?

```suggestion
          push: inputs.push
```

max-sixty: I'm not sure — I get confused by these. Within an `if`, we don't need the brackets. But outside, it's possible that `'false'` is interpreted as truthy...

If you're confident, then commit — I've resorted to almost always doing the verbose `== 'true'` thing

.github/workflows/build-devcontainer.yaml
max-sixty: It seems to only raise an error on the push (this run was passing until it was canceled...)

.github/workflows/build-devcontainer.yaml
eitsupi: I see, thanks. Let's keep as is now.

eitsupi (approved):
LGTM, thanks!

.github/workflows/tests.yaml
```diff
@@ -70,10 +71,11 @@ jobs:
               - bindings/prql-dotnet/**
               - bindings/prql-lib/**
               - .github/workflows/test-dotnet.yaml
-            devcontainer:
+            devcontainer-push:
               - .devcontainer/**/*Dockerfile
-              - .github/workflows/build-devcontainer.yaml
               - Taskfile.yml
+            devcontainer-test:
+              - .github/workflows/build-devcontainer.yaml
```
eitsupi, line 78: Sorry, I missed the need to include the Dockerfile and Taskfile here. I will fix it later (After #3423 merged).

max-sixty: Sorry! Good spot. Lmk if you prefer I fix this...

eitsupi: Thank you. I'm glad you are doing this as I am about to go to work and can't do this right away (just note that it takes 3+ hours to get through the CI as the cache doesn't exist until after it is pushed from the main branch)

PR #3405, opened 2023-08-25
fix: Potentially fix logic error

max-sixty:
Am I missing something or are `all` & `any` the same at the moment? Is this change correct?

### DIFF
--- a/crates/prql-compiler/src/utils/mod.rs
+++ b/crates/prql-compiler/src/utils/mod.rs
@@ -101,7 +101,7 @@ where
 
     fn any_true(self) -> Option<bool> {
         self.cloned()
-            .fold(Some(true), |a, x| a.zip(x).map(|(a, b)| a && b))
+            .fold(Some(true), |a, x| a.zip(x).map(|(a, b)| a || b))
     }
 }
 


crates/prql-compiler/src/utils/mod.rs
```diff
@@ -101,7 +101,7 @@ where
 
     fn any_true(self) -> Option<bool> {
         self.cloned()
-            .fold(Some(true), |a, x| a.zip(x).map(|(a, b)| a && b))
+            .fold(Some(true), |a, x| a.zip(x).map(|(a, b)| a || b))
```
aljazerzen, line 104: Yes this is backwards.

And it should also start with `Some(false)`.

aljazerzen: Also this comment is a bit confusing: https://github.com/PRQL/prql/pull/3405/files#diff-e59d2f5d0883e7d1ee71dfa13faef8218e2ef13dab795774895f4573cc86ad69R87-R89

If this is not used anywhere, maybe it should be removed because there are a few different interpretations of what it is supposed to do.

max-sixty: OK, replacing this with #3413

PR #3408, opened 2023-08-25
feat: Add a line continuation character

### LINKED ISSUE #819: Syntax to break an expression over multiple lines?

### LINKED ISSUE #3380: Line continuation inside parentheses
### What happened?

I tried my hand at converting the "heinous SQL CASE statement" at https://github.com/mattermost/mattermost-data-warehouse/blob/577c4e85adbe03e50bc989612a65120efda62823/transform/sql/snowflake/automated_nps_feedback_category_update.sql to PRQL.

There are a couple of problems with what I've done so far. For a start I think my string quoting with the r-strings isn't correct but let's set that aside for the moment (because I just started refactoring this and copied the strings "as-is" from the original SQL).

What I'm concerned about is the error below that I'm seeing in the playground. In other languages it is usual to be able to continue to current line without a line continuation character if you are inside an expression as would be after a `&&` or `||` or when inside parentheses `()` as is the case here.

These expressions are already very hard to follow, forcing them onto a single line would make them really unreadable.

### PRQL input

```elm
let prefix  = r"""(^| |\\.|\;|,|:|\n|\t|\-|\"|\'|\/)"""
let suffix  = r"""(^|\\.|\;|,| |\/|\n|\t|\-|\"|\'|\\?|\!|$)"""
let suffix2 = r"""(s|^|\\.|\;|,| |\/|\n|\t|\-|\"|\'|\\?|\!|$)"""

let matches = func term _prefix:f"{prefix}" _suffix:f"{suffix}" input -> ((lower input) ~= f"{_prefix}{term}{_suffix}") != null

from n1=analytics.mattermost.nps_user_daily_score
join n2=analytics.mattermost.nps_feedback_classification (n1.user_id == n2.user_id && n1.server_id == n2.server_id && n1.last_feedback_date == n2.last_feedback_date)
derive subcategory = case {
  (n1.feedback | matches "oaut(h|h2|hs|hentication)")  => 'Account Authorization',
  (n1.feedback | matches "password" _suffix:f"{suffix2}") => 'Account Authorization',
  (n1.feedback | matches "authenticat(e|ion|ed|es|ed|ing|or|er)")  => 'Account Authorization',
  (n1.feedback | matches "(sso|single sign on|single-sign-on|single signon|single sign-on|single-sign on)")  => 'Account Authorization',
  (n1.feedback | matches "(quit|stop) (bugging|annoying|spamming|surveying) me" _suffix:f"{suffix2}") => 'Invalid',
  ( (n1.feedback | matches "mobile" _suffix:f"{suffix2}") ||
    (n1.feedback | matches "(i-p|ip|p)hone" _suffix:f"{suffix2}") ||
    (n1.feedback | matches "ios" _suffix:f"{suffix2}") ||
    (n1.feedback | matches "android" _suffix:f"{suffix2}")
  ) => 'Mobile Stability',
}
```

### SQL output

image

### Expected SQL output

```SQL
SELECT
  n1.*,
  n2.*,
  CASE
    WHEN REGEXP(
      LOWER(n1.feedback),
      CONCAT(
        '(^| |\\.|\;|,|:|\n|\t|\-|\"|\'|\/)',
        'oaut(h|h2|hs|hentication)',
        '(^|\\.|\;|,| |\/|\n|\t|\-|\"|\'|\\?|\!|$)'
      )
    ) IS NOT NULL THEN 'Account Authorization'
    WHEN REGEXP(
      LOWER(n1.feedback),
      CONCAT(
        '(^| |\\.|\;|,|:|\n|\t|\-|\"|\'|\/)',
        'password',
        '(s|^|\\.|\;|,| |\/|\n|\t|\-|\"|\'|\\?|\!|$)'
      )
    ) IS NOT NULL THEN 'Account Authorization'
    WHEN REGEXP(
      LOWER(n1.feedback),
      CONCAT(
        '(^| |\\.|\;|,|:|\n|\t|\-|\"|\'|\/)',
        'authenticat(e|ion|ed|es|ed|ing|or|er)',
        '(^|\\.|\;|,| |\/|\n|\t|\-|\"|\'|\\?|\!|$)'
      )
    ) IS NOT NULL THEN 'Account Authorization'
    WHEN REGEXP(
      LOWER(n1.feedback),
      CONCAT(
        '(^| |\\.|\;|,|:|\n|\t|\-|\"|\'|\/)',
        '(sso|single sign on|single-sign-on|single signon|single sign-on|single-sign on)',
        '(^|\\.|\;|,| |\/|\n|\t|\-|\"|\'|\\?|\!|$)'
      )
    ) IS NOT NULL THEN 'Account Authorization'
    WHEN REGEXP(
      LOWER(n1.feedback),
      CONCAT(
        '(^| |\\.|\;|,|:|\n|\t|\-|\"|\'|\/)',
        '(quit|stop) (bugging|annoying|spamming|surveying) me',
        '(s|^|\\.|\;|,| |\/|\n|\t|\-|\"|\'|\\?|\!|$)'
      )
    ) IS NOT NULL THEN 'Invalid'
    WHEN REGEXP(
      LOWER(n1.feedback),
      CONCAT(
        '(^| |\\.|\;|,|:|\n|\t|\-|\"|\'|\/)',
        'mobile',
        '(s|^|\\.|\;|,| |\/|\n|\t|\-|\"|\'|\\?|\!|$)'
      )
    ) IS NOT NULL
    OR REGEXP(
      LOWER(n1.feedback),
      CONCAT(
        '(^| |\\.|\;|,|:|\n|\t|\-|\"|\'|\/)',
        '(i-p|ip|p)hone',
        '(s|^|\\.|\;|,| |\/|\n|\t|\-|\"|\'|\\?|\!|$)'
      )
    ) IS NOT NULL
    OR REGEXP(
      LOWER(n1.feedback),
      CONCAT(
        '(^| |\\.|\;|,|:|\n|\t|\-|\"|\'|\/)',
        'ios',
        '(s|^|\\.|\;|,| |\/|\n|\t|\-|\"|\'|\\?|\!|$)'
      )
    ) IS NOT NULL
    OR REGEXP(
      LOWER(n1.feedback),
      CONCAT(
        '(^| |\\.|\;|,|:|\n|\t|\-|\"|\'|\/)',
        'android',
        '(s|^|\\.|\;|,| |\/|\n|\t|\-|\"|\'|\\?|\!|$)'
      )
    ) IS NOT NULL THEN 'Notifications Reliability'
    ELSE NULL
  END AS subcategory
FROM
  analytics.mattermost.nps_user_daily_score AS n1
  JOIN analytics.mattermost.nps_feedback_classification AS n2 ON n1.user_id = n2.user_id
  AND n1.server_id = n2.server_id
  AND n1.last_feedback_date = n2.last_feedback_date

-- Generated by PRQL compiler version:0.9.3 (https://prql-lang.org)
```

### MVCE confirmation

- [X] Minimal example
- [X] New issue

### Anything else?

Also, where did we settle on a line continuation character? I recall there being a debate but not what the outcome was.

max-sixty:
Would close #3380, close #819

This kinda works but we need to decide how to handle comments. It's described inline — but basically:

```elm
derive x = 1 +
# \ 2 +
\ 3
```

...doesn't work, because the middle line is commented out, so is interpreted as a new line.

The big advantage of the line continuation character at the start of the line vs. the end is that it's easy to comment out parts of an expression, which this doesn't support.

---

One idea — have a line continuation character join to _the previous line with content_. So these would all be equal:

> (the elm formatting makes this look confusing, it'll be better in PRQL formatting, I think?)

```elm
derive x = 1 + 3

derive x = 1 +
\ 3

derive x = 1 +
# \ 2 +
\ 3

derive x = 1 +
# hello

# \ 2 +
\ 3

```

A refinement of that idea would be to only allow comments, not blank lines, to be joined to.

What do folks think?


crates/prql-parser/src/lexer.rs
```diff
@@ -140,6 +127,38 @@ pub fn ident_part() -> impl Parser<char, String, Error = Cheap<char>> {
...
+        .repeated()
+        .ignored()
+}
+
+fn line_continuation() -> impl Parser<char, (), Error = Cheap<char>> {
+    just('\n')
+        .then(whitespace().repeated().or_not())
+        .then(just('\\'))
+        .ignored()
+}
```
max-sixty: I split some parsers out to test the new function individually; on reflection maybe not worthwhile, am happy to lump them all back into the big func if folks prefer that

eitsupi:
This looks great!

When I look at the discussion in #819, I see that the similarity to the comma at the end of the line is an issue, so perhaps the following example would be more appropriate?

```elm
derive x = 1
# \ + 2
\ + 3
```

aljazerzen:
Nice work!

Looking at this I had quite a controversial idea: the fact that a new line has semantic meaning is quite strange for a language. How much worse would PRQL look like if we required the `|` for pipes?

```elm
from albums
| select {album_id, title}
| group title (
   sort album_id
   | take 1
)
```

max-sixty:
> How much worse would PRQL look like if we required the `|` for pipes?

I thought about this a while ago. I think one of the beautiful features of PRQL is that it's completely based around pipelines, and so can design the syntax around that. And that means that we don't need special pipe characters...

This does seem mostly opinion / aesthetic. One more objective measure is comparing the number of times we need a line-continuation character with the number of times we need a pipe — the latter is 1-2 orders of magnitude greater.

aljazerzen:
That's true, continuation character is much less common than a pipeline. Cool, I concur.

crates/prql-parser/src/lexer.rs
max-sixty: I put them back, which speeds up compilation (not just parsing or even lexing) by 1.5%...

max-sixty:
I implemented an approach to "skip over" lines with comments, which I think is the best approach.

One caveat — it does mean that when lexing PRQL, we always need to find the next non-comment, non-empty line to know when to finish lexing that "logical" line. I think that's probably OK given our performance requirements are lower than most...

eitsupi:
One of my questions is that the meaning of line breaks changes (line breaks are ignored) in tuples `{...}` and arrays `[...]`.
Outside of these the line break is equivalent to a pipe, but only within these it does not seem to be so?

max-sixty:
> Outside of these the line break is equivalent to a pipe, but only within these it does not seem to be so?

Exactly correct!

https://github.com/PRQL/prql/blob/b612f093adff8ad0020a841a046cbbfcc15b71ef/web/book/src/reference/syntax/pipes.md#L25

Lmk if we could make that clearer

max-sixty:
This is ready to go, let me know any concerns?

One reason not to merge is that we think it makes lexing less performant, and prevents us from optimizing in the future. It's basically doing exactly what `@matlkad` suggests _not_ doing in https://matklad.github.io/2023/08/01/on-modularity-of-lexical-analysis.html ...

But otherwise I'll merge...

eitsupi:
> > Outside of these the line break is equivalent to a pipe, but only within these it does not seem to be so?
> 
> Exactly correct!
> 
> https://github.com/PRQL/prql/blob/b612f093adff8ad0020a841a046cbbfcc15b71ef/web/book/src/reference/syntax/pipes.md#L25
> 
> Lmk if we could make that clearer

Oh, I did not check that excellent document. Thanks for pointing that.
I think this is enough to clear up my question.

max-sixty:
Let me cross link that existing doc & this new documentation

max-sixty:
FYI I renamed this in the docs to "line wrapping", from "line continuation"...

web/book/src/reference/syntax/pipes.md
```diff
@@ -1,10 +1,10 @@
...
 
-Pipes — the connection between [transforms](../stdlib/transforms/) that make up
-a pipeline — can be either line breaks or a pipe character (`|`).
+Pipes — the connection between [transforms](../stdlib/transforms/) that make up
+a pipeline — can be either linebreaks or a pipe character (`|`).
 
-In almost all situations, line-breaks pipe the result of a line's transform into
-the transform on the following line. For example, the `filter` transform
+In almost all situations, a line break pipe the result of a line's transform
+into the transform on the following line. For example, the `filter` transform
```
snth: This doesn't read right. I think the original text was ok or otherwise maybe something like the following:

> In almost all situations, a linebreak (\n) is equivalent to inserting a pipe character (|), which means the relation produced by the current transform is used as the input (last argument) for the transform on the following line.
> 
> For example, the `filter` transform

web/book/src/reference/syntax/pipes.md
```diff
@@ -15,19 +15,20 @@ filter department == "Product"
...
 ```
 
-A line-break doesn't create a pipeline in a couple of cases:
+A line break doesn't create a pipeline in a few cases:
 
 - within a tuple
 - within an array
 - when the following line is a new statement, which starts with a keyword of
-  `func`, `let` or `from`.
+  `func`, `let` or `from`
```
snth, line 31: Is the reference to `func` still correct here or is this outdated since the introduction of `let`?

IIRC, a function definition including the optional `func` would need to be written as

```elm
let my_func = func x -> ...
```

So a line would never start with `func` anymore.

snth:
> Looking at this I had quite a controversial idea: the fact that a new line has semantic meaning is quite strange for a language. 

I don't think it is that uncommon. bash, Python, and Julia all have `\n` generally equivalent to `;` to name just a few that come to mind.

> How much worse would PRQL look like if we required the | for pipes?

I see you've already moved on from this but I just wanted to add that I think it would be substantially worse.

snth:
> > One of my questions is that the meaning of line breaks changes (line breaks are ignored) in tuples {...} and arrays [...].
> > Outside of these the line break is equivalent to a pipe, but only within these it does not seem to be so?
>
> Exactly correct!

Could we not add parentheses `()` to this list? That would address most, if not all, of my use cases raised in #3380 .

It's also quite a common pattern, for example in Python for things like the following:

```python
s = ("this string "
     "reads better when "
     "written over multiple lines")
```
EDIT: I guess one counterpoint is that you would need to know whether you are inside a "column expression" or a parenthesis for a pipeline like after a `group` so I understand if that's not possible with our setup.

Similarly, many languages seem to be able to cope with unfinished expressions where a `\n` comes after an infix operator as in
```elm
derive x = 1 +
   3
```
IDK enough about compilers to know what makes that possible for them and what the performance impact is. I do see Max's point about the impact of `\` on the Lexer.

snth:
Thank you very much for implementing this @max-sixty ! 🙏

web/book/src/reference/syntax/pipes.md
max-sixty: I was actually just about to remove this! I was going to completely remove `func`...

web/book/src/reference/syntax/pipes.md
max-sixty: Sorry for the grammar, your proposal is great, I'm adding (you can always add as a suggestion btw!)

max-sixty:
> EDIT: I guess one counterpoint is that you would need to know whether you are inside a "column expression" or a parenthesis for a pipeline like after a `group` so I understand if that's not possible with our setup.

Yeah re pipelines. 

And I also think that grammar being different inside vs. outside of parentheses is less consistent, and so worse, all-else-equal

web/book/src/reference/syntax/pipes.md
max-sixty: I actually restructured it slightly...

eitsupi:
> And I also think that grammar being different inside vs. outside of parentheses is less consistent, and so worse, all-else-equal

I agree. I hate using parentheses in Python method chains just to bring the dot to the beginning of the line...... (Of course I don't have to do it myself because the wonderful Black handles it automatically.)

PR #3412, opened 2023-08-25
ci: Fix failing Snap build

vanillajonathan:


max-sixty:
Thanks @vanillajonathan !

vanillajonathan:
Can you get 0.9.4 out on the Snap store?

max-sixty:
I actually not sure that's easy to do, since working code to release in not in the place as the rust code for 0.9.4. If we've had issues with something like `prql-js`, I've released locally or done another release.

Are we OK to wait until 0.9.5? The changes are small from 0.9.3...

vanillajonathan:
Sure, the reason I wanted it published is to download it from Snap and verify that the core22 works as expected, so that I can go ahead and create a new PR that changes the `grade` property from `devel` to `stable`.

max-sixty:
Can we run a dry-run? That won't let you download it, but it might test that it's working. We now have a good system for testing publishing in the workflow, by setting it to only actually publish when the `publish` input is true...

vanillajonathan:
We can wait, I want to download it from the Snap store anyway.

PR #3414, opened 2023-08-25
ci: Pin `alls-green` to bugfix commit

max-sixty:
Per https://github.com/re-actors/alls-green/issues/23

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -452,10 +452,7 @@ jobs:
       - check-links-book
       - check-links-markdown
       - lint-megalinter
-      # Excluding until https://github.com/re-actors/alls-green/issues/23 is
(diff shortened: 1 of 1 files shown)


.github/workflows/tests.yaml
```diff
@@ -469,7 +466,8 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Decide whether the needed jobs succeeded or failed
-        uses: re-actors/alls-green@release/v1
+        # https://github.com/re-actors/alls-green/issues/23
+        uses: re-actors/alls-green@release/cf9edfcf932a0ed6b431433fa183829c68b30e3f
```
max-sixty:
```suggestion
        uses: re-actors/alls-green@cf9edfcf932a0ed6b431433fa183829c68b30e3f
```

PR #3420, opened 2023-08-26
devops: Reduce number of cargo tools installed

### LINKED ISSUE #3419: Nightly tests failed — slow devcontainer build
Nightly tests [failed on 2023-08-26](https://github.com/PRQL/prql/actions/runs/5984368557)

max-sixty:
Trying to get the devcontainer build faster, which will fix #3419. This is mostly a QEMU issue but are the constraints we have...

### DIFF
--- a/Taskfile.yml
+++ b/Taskfile.yml
@@ -23,11 +23,7 @@ vars:
     prql-lang.prql-vscode
     rust-lang.rust-analyzer
   cargo_crates: |
-    bacon
-    cargo-audit
     cargo-insta
-    cargo-release
-    default-target
     mdbook
     mdbook-admonish
     mdbook-footnote
@@ -35,6 +31,12 @@ vars:
     wasm-bindgen-cli
     wasm-opt
     wasm-pack
+  # Crates for administering the repo
+  cargo_crates_admin: |
+    bacon
+    cargo-audit
+    cargo-release
+
   # Excluding `elixir` atm given it's not enabled on Mac and currently unsupported
   brew_dependencies: |
     hugo
@@ -76,6 +78,12 @@ tasks:
           🟢 Setup complete! ✅🚀"
         silent: true
 
+  setup-dev-admin:
+    desc: Install tools for PRQL development & admin.
+    cmds:
+      - task: setup-dev
+      - task: install-cargo-tools-admin
+
   install-cargo-tools:
     desc: Install cargo tools for PRQL development.
     cmds:
@@ -96,6 +104,11 @@ tasks:
       - cmd: |
           [ "$(which cargo-insta)" ] || echo "🔴 Can't find a binary that cargo just installed. Is the cargo bin path (generally at `~/.cargo/bin`) on the \$PATH?"
         silent: true
+
+  install-cargo-tools-admin:
+    desc: Install cargo tools for PRQL development & admin.
+    cmds:
+      - cargo install --locked {{.cargo_crates_admin | splitLines | join " " }}
 
   install-cargo-tools-source:
     cmds:


Taskfile.yml
```diff
@@ -23,18 +23,20 @@ vars:
...
     mdbook-footnote
     mdbook-toc
     wasm-bindgen-cli
     wasm-opt
     wasm-pack
+  # Crates for administering the repo
+  cargo_crates_admin: |
+    bacon
+    cargo-audit
+    cargo-release
```
eitsupi, line 38: There is a question about the selection of tools. Why did you choose these?

Taskfile.yml
```diff
@@ -97,6 +105,11 @@ tasks:
           [ "$(which cargo-insta)" ] || echo "🔴 Can't find a binary that cargo just installed. Is the cargo bin path (generally at `~/.cargo/bin`) on the \$PATH?"
         silent: true
 
+  install-cargo-tools-admin:
+    desc: Install cargo tools for PRQL development & admin.
+    cmds:
+      - cargo install --locked {{.cargo_crates_admin | splitLines | join " " }}
```
eitsupi, line 111: Why don't use bintall?

eitsupi (commented):
I don't know if this is the right solution.

Taskfile.yml
max-sixty: The others are required for building & testing the repo — or are there any that I misallocated? 

I can add a comment with the principle

Taskfile.yml
max-sixty: We only use `binstall` for CI, because it would write hundreds of `429` messages to the terminal when running locally.

Maybe that's improved? Though it installing the wrong version of `cargo admonish` doesn't give me confidence we should expand our use of it...

max-sixty:
Are there other simple solutions? Very open to them!

eitsupi:
> Are there other simple solutions? Very open to them!

Why did it recently become impossible to fit within the 3,600 minute limit to begin with?

My guess is that it is because of the source installation of cargo-binstall?
Since the latest cargo-binstall seems to have solved this problem, I suspect that simply doing a binary install of cargo-binstall will solve the problem.
https://github.com/PRQL/prql/blob/28b9653b4d0a52427dde2269d2d5826d09a695f3/Taskfile.yml#L120-L127

Taskfile.yml
eitsupi: > because it would write hundreds of 429 messages to the terminal when running locally.

I don't understand exactly what happens, but wouldn't it be solved by simply suppressing the output to the terminal?

max-sixty: But then there's no output for 10 minutes?

max-sixty:
> My guess is that it is because of the source installation of cargo-binstall?

Yes, good idea. Let's try that. I opened #3421

max-sixty:
@eitsupi please can we not cancel the CI jobs? It's means I go and check why it failed, and our badge says our tests are failing...

max-sixty:
> @eitsupi please can we not cancel the CI jobs? It's means I go and check why it failed, and our badge says our tests are failing...

Remembering a couple of other discussions — possibly we weigh "wasted compute" vs. "developer easiness / simplicity" differently? 

Happy to come up with a compromise!

eitsupi:
> @eitsupi please can we not cancel the CI jobs? It's means I go and check why it failed, and our badge says our tests are failing...

I'm sorry, but don't you think it doesn't make sense to run all the CI's until multiple fixes are split into multiple PR's and they are all merged?
Of course, they will spend several hours each time building and pushing Docker images for useless purposes until #3404 is merged.

max-sixty:
I think we sent messages sent at similar times!

I agree it's very wasteful of compute. (CI is generally quite wasteful, and I would love to see a usable incremental testing framework for open-source projects... — It's kinda crazy that someone adds a comment and we need to run basically all the tests again). Our taskfile tests & devcontainer build are especially wasteful.

I forgot that it was still pushing, so fine re canceling that — it sounds like that's an inconvenience to _users_ of the devcontainer too.

In general though, can you see my perspective too? That we save something in the order of magnitude of $0.20 (5 hours * 2 CPUs * $0.02c/cpu/hour) — but I go and check why the build failed, and we have a badge which states it's failing? (we could change the latter)

Is a reasonable compromise that when we get this working — which I've been spending lots of time on (maybe inefficiently?) — then we don't cancel them? Or what would you suggest?

Taskfile.yml
eitsupi: Ok, it certainly may not be a good idea to suppress everything.
But given the frequency with which this command is executed, I'm not sure if the large amount of output to the console is a compelling reason not to use binstall.
At least for me, I prefer a lot of logging to spending dozens of minutes compiling.

eitsupi:
One suggestion is that I did not want this merged until I saw this. As I have already pointed out, I do not believe this is the best way to handle this. Perhaps #3421 works better.

Another idea is to stop referencing Taskfile from within the Dockerfile.
Of course we should avoid writing the same thing in two places, but it is a complete waste to refer to the whole Taskfile and discard layers for a small change to the Taskfile, when in fact we just want a list of cargo tools. (It is also useless to install Task twice, once when building the base image and once when building the Dev Container)
I haven't come up with a good way to do this, but after a bit of thought I've come up with one way and will try it later.

> In general though, can you see my perspective too? That we save something in the order of magnitude of $0.20 (5 hours * 2 CPUs * $0.02c/cpu/hour) — but I go and check why the build failed, and we have a badge which states it's failing? (we could change the latter)

I apologize for wasting your time. I should have commented that I stopped the job.

eitsupi:
I cancelled the test workflow triggered by 28b9653b4d0a52427dde2269d2d5826d09a695f3

max-sixty:
> One suggestion is that I did not want this merged until I saw this. As I have already pointed out, I do not believe this is the best way to handle this. Perhaps #3421 works better.
> 
> Another idea is to stop referencing Taskfile from within the Dockerfile. Of course we should avoid writing the same thing in two places, but it is a complete waste to refer to the whole Taskfile and discard layers for a small change to the Taskfile, when in fact we just want a list of cargo tools. (It is also useless to install Task twice, once when building the base image and once when building the Dev Container)

I very much agree!

> I apologize for wasting your time. I should have commented that I stopped the job.

No apology necessary — my motivation is to find a good way of working together... 

(I appreciate the gesture of saying you stopped it, though then you're taking time to do it, still not ideal...)

Hopefully your new design will be a good solution for all

PR #3423, opened 2023-08-27
ci(devcontainer-base-image): Simplify the Dockerfile to avoid coping Taskfile into Docker image

eitsupi:
Since the only item in the Taskfile that must be referenced from within the Dockerfile is a string in a single field (cargo_crates), instead of copying the Taskfile into the Dockerfile, the string is extracted from the Taskfile by `yq` and referenced by the build argument of the `docker build` command.

This should prevent minor changes in the Taskfile from modifying the Docker image layers.

Since the Taskfile no longer needs to be copied and we could change the build context, the `.dockerignore` file in the repository root can also be remove.


max-sixty:
Very clever! Good idea!

.devcontainer/base-image/Dockerfile
```diff
@@ -1,17 +1,20 @@
-# Some of this is shared with /Dockerfile
+# syntax=docker/dockerfile:1.4
 
 FROM mcr.microsoft.com/devcontainers/rust:1-1-bullseye
 
 RUN apt-get update \
     && apt-get install -y --no-install-recommends \
     cmake \
-    && rm -rf /var/lib/apt/lists/* \
-    && sh -c "$(curl --location https://taskfile.dev/install.sh)" -- -d -b /usr/local/bin
```
max-sixty, line 9: I do think we need this in the image though, or none of our `task` command will work in the devcontainer itself — i.e. it's not just for the build that we want `task`...

eitsupi: Good point.
The latest Task will be installed on the user side via the Dev Container Feature ([that I mantain](https://github.com/eitsupi/devcontainer-features/blob/main/src/go-task/README.md)), the advantage of installing via the Feature is that the Task's VSCode extension and shell completion are also configured.
https://github.com/PRQL/prql/blob/28b9653b4d0a52427dde2269d2d5826d09a695f3/.devcontainer/devcontainer.json#L8

In other words, Tasks previously installed on the Dockerfile were used only temporarily.

max-sixty: OK great!

max-sixty (dismissed):

.github/workflows/build-devcontainer.yaml
```diff
@@ -34,11 +37,17 @@ jobs:
       - name: Set up Docker Buildx
         uses: docker/setup-buildx-action@v2
 
+      - name: Prep args
+        run: |
+          echo "cargo_crates=$(yq -r '.vars.cargo_crates' Taskfile.yml)" >>"$GITHUB_ENV"
```
max-sixty, line 42: One slight disadvantage of this is that it's not easy to run the build locally. I think that's fine though, especially since most of the existing build instructions are in the GHA format, not in a `docker buildx build` CLI format.

An alternative would be a `task` which ran this, which we also called from CI (we could also avoid `yq` in that case).

(But the current state is completely fine tbc)

.devcontainer/devcontainer.json
```diff
@@ -6,9 +6,12 @@
     "ghcr.io/devcontainers/features/python:1": {},
     "ghcr.io/devcontainers/features/node:1": {},
     "ghcr.io/eitsupi/devcontainer-features/go-task:1": {},
+    "ghcr.io/eitsupi/devcontainer-features/jq-likes:1": {
+      "yqVersion": "latest"
```
max-sixty, line 10: No problem for including this — presumably it's not strictly required?

max-sixty (approved):

max-sixty:
Thanks @eitsupi !

.github/workflows/build-devcontainer.yaml
eitsupi: Good point.
However, this is essentially a release command, and I believe the other release commands are not defined on the Taskfile, but on the workflow files.

.devcontainer/devcontainer.json
eitsupi: Yeah, I just wanted to use yq to complete this PR. (I just know that yq installs quickly, so even with this, the build time would only increase by a few seconds)

PR #3425, opened 2023-08-27
ci: Use `github.event_name` for releases

### LINKED ISSUE #3382: Release workflow does not seem to be working as expected when creating a release
~~I have created a release on 18b9cba60e70ee740e506c8868d2d351cc8846e6, but nothing appears to have been published.~~
~~I probably should have checked first on my fork to make sure the workflow was working correctly (I think I checked on the changes I made to my release workflow, but didn't keep track of the other changes).~~

~~I removed the release and tag as it seemed to run as a dry run.~~

~~@max-sixty Any thoughts?~~

Sorry, I did not understand the difference between creating a release as a pre-release and then making changes to the release versus creating it as a release from the beginning.
The workflow seems to work correctly only in the former case.

49 is the latter and 50 is the former in this picture.

image

max-sixty:
Conceptually closer to our goal.

Inspired by, and possibly fixes https://github.com/PRQL/prql/issues/3382

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -16,7 +16,7 @@ jobs:
   brew-dispatcher:
     name: Release on homebrew-prql
     runs-on: ubuntu-latest
-    if: github.ref_type == 'tag'
+    if: github.event_name == 'release'
     steps:
       - uses: actions/github-script@v6
         with:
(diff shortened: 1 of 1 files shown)


eitsupi (approved):
LGTM, thanks!

PR #3431, opened 2023-08-27
ci: Refine devcontainer push rules

max-sixty:
Follow up to #3404


max-sixty:
I'm going to merge this now, because I think it might otherwise push too often — happy to make any refinements on post-merge review

PR #3439, opened 2023-08-28
revert: #3436

max-sixty:
MSRV issue after merging

https://github.com/PRQL/prql/actions/runs/6005715274/job/16288962369#step:5:14


max-sixty:
#3440 fixes that it could merge

PR #3468, opened 2023-09-04
web: Refine #3445

max-sixty:
- Combine task which wraps another
- Rename `install-` to `build-` for things which create artifacts internally (vs. installing something to the system)

@richb-hanover for your review


richb-hanover:
This is good -  fire away. Thanks

PR #3472, opened 2023-09-05
fix: Rewrite precedence, fix corner-case

### LINKED ISSUE #3469: wrong expression precedence
### What happened?

wrong "/" and "*" expression precedence。

### PRQL input

```elm
from foo
select a / (b * 0.7)
```

### SQL output

```SQL
SELECT
  (a / b * 0.7)
FROM
  foo
```

### Expected SQL output

```SQL
SELECT
  (a / (b * 0.7))
FROM
  foo
```

### MVCE confirmation

- [ ] Minimal example
- [ ] New issue

### Anything else?

_No response_

max-sixty:
This fixes #3469, by re-writing how we do precedence. I think these are pretty robust rules now. It was surprisingly difficult — I definitely didn't expect to be thinking so much about precedence!

(weirdly I couldn't find a source so mostly came up with these myself, though surely every language should have some implementation of this??)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,9 @@
 
 - Fix panic when unresolved lineage appears in group or window (@davidot, #3266)
 
+- Fix a corner-case in handling precedence, and remove unneeded parentheses in
+  some outputs (@max-sixty, #3472)
+
 **Documentation**:
 
 **Web**:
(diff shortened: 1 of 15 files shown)


aljazerzen (approved):
This is great! From the examples I think the precedence is now spot on - no unnecessary parenthesis.

PR #3492, opened 2023-09-10
docs(prqlc): minor update of sample commands

eitsupi:
Related to #3483

I think we should encourage users to use single quotes here.


max-sixty (approved):
Good idea

PR #3493, opened 2023-09-11
docs: Add Ace JavaScript editor

vanillajonathan:
**Ace** is an embeddable code editor written in JavaScript. It includes syntax highlighting for PRQL.

PR: https://github.com/ajaxorg/ace/pull/5307

PR #3512, opened 2023-09-15
docs: Create ace-demo.html

vanillajonathan:
This is a demo of using [Ace](https://ace.c9.io/) with the PRQL syntax highlighting mode.


max-sixty:
OK cool. 

Would be nice to have some sort of test / check that it's still working correctly, but I guess that's difficult with HTML.

Shall we link to this from the grammars readme? Otherwise it's totally orphaned.

Separately, we could add the link from https://github.com/PRQL/prql/pull/3493#issuecomment-1721627217 too.

max-sixty:
If you want to spend more time with grammars, adding tests for our existing grammars would be really helpful.

And one *huge* win would be allowing the PRQL lezer grammar to work in Jupyter notebooks. The grammar is our best grammar, and is the format (CodeMirror) required for Jupyter notebooks, but the integration hasn't been writen. That would be a difficult project but would immensely valuable.

grammars/README.md
```diff
@@ -9,6 +9,8 @@ an index.
 
 - [Ace](https://ace.c9.io/) — supported. The grammar is upstream
   ([prql_highlight_rules.js](https://github.com/ajaxorg/ace/blob/master/src/mode/prql_highlight_rules.js)).
+  See the
+  [demo](https://htmlpreview.github.io/?https://github.com/PRQL/prql/blob/main/grammars/ace-demo.html).
```
max-sixty: Does this work? I'm not sure I'm doing it right but if I try https://htmlpreview.github.io/?https://github.com/PRQL/prql/blob/0bcb3d530aa9eb411a9c09380738732f5be1407f//grammars/ace-demo.html, I just get some text.

vanillajonathan: Ugh, I had hoped so, but I guess not. I can move to `/web/website/content/` though.

max-sixty: OK, let's try that

max-sixty: (need to update link too :) )

vanillajonathan: I hope this works!

grammars/README.md
```diff
@@ -9,6 +9,7 @@ an index.
 
 - [Ace](https://ace.c9.io/) — supported. The grammar is upstream
   ([prql_highlight_rules.js](https://github.com/ajaxorg/ace/blob/master/src/mode/prql_highlight_rules.js)).
+  See the [demo](https://prql-lang.org/ace-demo).
```
max-sixty: I didn't realize this was a top-level page. Could we put it down a level to indicate that it's not a sibling of the key pages? Maybe a `content/demos/ace-demo.html` page?

vanillajonathan: Sounds good.

max-sixty:
We're doing a release v soon anyway, so it'll publish then

PR #3513, opened 2023-09-15
feat: Add r-string to prql.grammar

vanillajonathan:
This adds `r"strings"` to the grammar.


max-sixty:
Super, thank you!

(Another thing that would be nice is to be able to test some of these grammars against our examples; there are some other projects that do this successfully...)

PR #3514, opened 2023-09-15
feat(Lezer grammar): Support Unicode in identifiers in prql.grammar

vanillajonathan:
This adds support for Unicode characters for identifiers to the Lezer grammar.

### DIFF
--- a/grammars/prql-lezer/src/prql.grammar
+++ b/grammars/prql-lezer/src/prql.grammar
@@ -58,7 +58,8 @@ Op_bin { op_bin_only | !term op_unary }
   // TODO: does excluding spaces here work? It seems to make a difference, but
   // then I think the `Number` rule doesn't allow spaces...
   Date { "@" ![ ] (( @digit | "-" ) ![ ] )+ }
-  ident_part { @asciiLetter (@asciiLetter | "_" | @digit )* }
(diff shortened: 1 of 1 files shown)


grammars/prql-lezer/src/prql.grammar
```diff
@@ -58,7 +58,8 @@ Op_bin { op_bin_only | !term op_unary }
   // TODO: does excluding spaces here work? It seems to make a difference, but
   // then I think the `Number` rule doesn't allow spaces...
   Date { "@" ![ ] (( @digit | "-" ) ![ ] )+ }
-  ident_part { @asciiLetter (@asciiLetter | "_" | @digit )* }
+  identifier_char { @asciiLetter | $[_\u{a1}-\u{10ffff}] }
+  ident_part { identifier_char (@asciiLetter | "_" | @digit )* }
```
max-sixty: I don't think this is quite right — this means the unicode character can only be at the start of the `ident_part`?

I do think tests would be very very helpful here. There's a nice example in https://lezer.codemirror.net/examples/test/ if you'd be up for doing that separately...

vanillajonathan: You're not, it's not quite right, and I had a feeling it wasn't but it worked, but now I see that it produces multiple `Expr`.

max-sixty:
Great! Will be cool to add a test for this when #3529 merges!

Issue #3477, opened 2023-09-06
Release plan for 0.9.5


eitsupi:
Might it be worth doing the next release because of the interesting new feature #3408?

Since some of the issues under discussion (#3462) would be breaking changes, I believe now might be a good time for a release. (And I am personally curious if the winget release will work properly #3387)

max-sixty:
Yes v happy to do one!

If you'd like to @eitsupi , you're of course welcome. If it's my turn then I can do it... :)

eitsupi:
Thanks.
Perhaps it would be a good idea to release it once #3472 is merged?
(And #3408 hasn't been added to the changelog yet.)

I'll be a little busy until Sunday, so I'd appreciate it if you could work on the release.

max-sixty:
Ah great, I'll add #3408, thanks for pointing that out.

Cool — I'll do the release!

snth:
#3472 and #3408 are two great new features/fixes coming. Thanks to everyone involved!

max-sixty:
I would like to add a link to the talks coming up, I've only got a link to the QCon one. @snth if you're online, do you have yours? Can we add in https://github.com/PRQL/prql/pull/3501? Or paste here and I'll do the md.

Otherwise will release regardless at the weekend.

snth:
Hi, one sec.

max-sixty:
Done! Sorry to take a few more days than I should have

eitsupi:
Wonderful! Thanks @max-sixty.

Winget releaser seems working fine. microsoft/winget-pkgs#120030
I think you will be asked to respond to CLA later, so please respond at that time.

max-sixty:
> Winget releaser seems working fine. [microsoft/winget-pkgs#120030](https://github.com/microsoft/winget-pkgs/pull/120030)

Great — nice work!

PR #3517, opened 2023-09-16
fix!: Update case syntax to use brackets over braces

### LINKED ISSUE #3462: `case` should use `[]` instead of `{}`
### What's up?

Currently, this is syntax for case:

```prql
from employees
derive distance = case {
  city == "Calgary" => 0,
  city == "Edmonton" => 300,
}
```

In the last dev call, @snth made a point that `case` syntax should be more similar to arrays than tuples, since the elements of the case block:
- must have the same type,
- are not named.

This is the proposed syntax:

```prql
from employees
derive distance = case [
  city == "Calgary" => 0,
  city == "Edmonton" => 300,
]
```

AaronMoat:
BREAKING CHANGE: This is a breaking change to case syntax.

Resolves #3462

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,10 +4,9 @@
 
 **Language**:
 
-- _Breaking:_ Case syntax now uses brackets `[]` rather than braces `{}`.
-
-  To convert previous PRQL queries to this new syntax simply change
-  `case { ... }` to `case [ ... ]`.
+- _Breaking:_ Case syntax now uses brackets `[]` rather than braces `{}`. To
+  convert previous PRQL queries to this new syntax simply change `case { ... }`
+  to `case [ ... ]`. (@AaronMoat, #3517)
 
 **Features**:
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,11 @@
 
 **Language**:
 
+- _Breaking:_ Case syntax now uses brackets `[]` rather than braces `{}`.
+
+  To convert previous PRQL queries to this new syntax simply change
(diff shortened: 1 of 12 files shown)


eitsupi:
Hi, thanks for working on this!
We can't merge this into the current version (See #3477), so I'll turn it into a draft to prevent merging.

CHANGELOG.md
```diff
@@ -4,6 +4,11 @@
 
 **Language**:
 
+- _Breaking:_ Case syntax now uses brackets `[]` rather than braces `{}`.
+
+  To convert previous PRQL queries to this new syntax simply change
+  `case { ... }` to `case [ ... ]`.
+
```
eitsupi: This should move to 0.10.0.

eitsupi (changes requested):

CHANGELOG.md
AaronMoat: I've moved to a new unreleased section, though will probably get conflicts that need fixing when #3477 is in.

max-sixty:
Great work! Thanks @AaronMoat .

Let's merge as soon as we release 0.9.5?

CHANGELOG.md
```diff
@@ -4,6 +4,32 @@
...
+**Features**:
+
+**Fixes**:
+
+- Update case syntax to brackets `case [ ... ]` over braces `case { ... }`
+  (@AaronMoat, #3517)
```
max-sixty: I think we can consolidate this into the changelog above, so we only have one entry?

max-sixty:
We can merge now!

eitsupi (approved):
Thanks!

PR #3521, opened 2023-09-17
chore(prqlc): bump MSRV to 1.70.0

eitsupi:
prqlc can now be binary installed on each platform.

I think the some of current problem with MSRV in dependencies is the prqlc dependency (#3520), so I think it makes sense to only update prqlc's MSRV.

### DIFF
--- a/crates/prqlc/Cargo.toml
+++ b/crates/prqlc/Cargo.toml
@@ -5,10 +5,8 @@ name = "prqlc"
 edition.workspace = true
 license.workspace = true
 repository.workspace = true
-rust-version.workspace = true
+rust-version = "1.70.0"
 version.workspace = true
-
-metadata.msrv = "1.65.0"
 
 [target.'cfg(not(target_family="wasm"))'.dependencies]
 anstream = {version = "0.3.2", features = ["auto"]}
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,8 @@
 
 **Integrations**:
 
+- Bump `prqlc`'s MSRV to 1.70.0 (@eitsupi, #3521)
+
 **Internal changes**:
 
 **New Contributors**:


max-sixty:
Ah, brilliant! 

Because we don't mind about Debian for prqlc, only for prqlr, and prqlr only requires `prql-compiler`?

This will make managing dependencies much easier...

eitsupi:
> Because we don't mind about Debian for prqlc, only for prqlr, and prqlr only requires `prql-compiler`?

Of course I think there is value in keeping MSRV small, but I thought it was worth thinking about this since clap's MSRV is now 1.70.

- prql-compiler needs to worry about downstream projects, but prqlc doesn't need to worry about downstream projects, so I think the impact of MSRV is small.
- prqlc has binaries in Homebrew and the official GitHub release, so I think many people install it from there.

eitsupi:
By the way, Debian testing is now on the migration from rustc 1.67 to 1.69, and 1.70 is now experimental, so there is a possibility that it will be updated to 1.70 soon.
<https://tracker.debian.org/pkg/rustc>

Ubuntu LTSs are 1.66 now.
<https://launchpad.net/ubuntu/+source/rustc>

crates/prqlc/Cargo.toml
```diff
@@ -5,11 +5,9 @@ name = "prqlc"
 edition.workspace = true
 license.workspace = true
 repository.workspace = true
-rust-version.workspace = true
+rust-version = "1.70.0"
 version.workspace = true
 
-metadata.msrv = "1.65.0"
```
max-sixty, line 11: I think this changes the wrong thing:
- The rust version is the standard version when developing etc. There's not much reason to have that be anything but latest / latest-1. We definitely want it the same throughout the workspace
- The MSRV is the version we guarantee we can still compile against

Because compilers are always backward-compatible, it's always fine to use a more recent version of the compiler if it's available.

So we should only adjust the MSRV...

max-sixty:
> Of course I think there is value in keeping MSRV small, but I thought it was worth thinking about this since clap's MSRV is now 1.70.

Yes very much agree.

To confirm though — is the main reason we stay on a lower version because of `prqlr` on debian? And not `prqlc`?

crates/prqlc/Cargo.toml
eitsupi: > The rust version is the standard version when developing etc. There's not much reason to have that be anything but latest / latest-1. We definitely want it the same throughout the workspace

Isn't that `rust-toolchain`?
https://github.com/PRQL/prql/blob/33b71715136380255482e902c8afe92bd5a8fe10/rust-toolchain.toml#L1-L4

<https://doc.rust-lang.org/cargo/reference/manifest.html#the-rust-version-field> says:

> The rust-version field is an optional key that tells cargo what version of the Rust language and compiler your package can be compiled with. If the currently selected version of the Rust compiler is older than the stated version, cargo will exit with an error, telling the user what version is required.

eitsupi: I think the reason we're using `metadata.msrv` instead of `rust-version` here is because cargo-msrv doesn't support inheritance from the workspace. foresterre/cargo-msrv#590

eitsupi:
> To confirm though — is the main reason we stay on a lower version because of `prqlr` on debian? And not `prqlc`?

At least that's true for me.
Of course I'm monitoring Debian's rustc version, so as soon as they reach 1.70 I'll make changes here as well.
(I think setting it to 1.70 makes sense since many crates are starting to set MSRV to 1.70.)

crates/prqlc/Cargo.toml
max-sixty: Sorry sorry, you're completely correct...

max-sixty:
> > 
> 
> At least that's true for me. Of course I'm monitoring Debian's rustc version, so as soon as they reach 1.70 I'll make changes here as well. (I think setting it to 1.70 makes sense since many crates are starting to set MSRV to 1.70.)

Perfect, thanks for the clarification and agree.

max-sixty (approved):

PR #3529, opened 2023-09-17
feat(Lezer grammar): Add tests

vanillajonathan:
This adds tests to the PRQL lezer grammar.


max-sixty:
This is excellent! Thank you very much @vanillajonathan .

The only issue is that the tests don't pass yet — do you get the same result? If you want to reduce the number of tests for the moment, that's OK, and then we can add them back / add more later. Let's ensure that the tests we have are passing, though.

Ideally we would run these in CI, though if you'd prefer to leave that for another PR, that's completely fine, this is a great improvement already.

If you know — do the tests require writing out the result manually? Or they snapshot the result? I'm not familiar with the JS test ecosystem. IIRC there's a library called `jest` from Meta which will snapshot the result, similar to `insta` in our rust tests. We probably don't want to be writing out these results manually.

---

So — to get this merged, let's at least limit it to the tests that pass, and then we can iterate on it in future PRs.

---

```
npm run test

> prql-lezer@1.0.0 test
> mocha test/test-*.js



  numbers
    ✔ Integer
    ✔ Integer with underscore
    ✔ Integer with underscores
    ✔ Decimal

  operators
    1) == Equals
    2) != Not equals
    3) >= Greater than
    4) <= Less than
    5) ~= Regex match
    6) && And
    7) || Or

  strings
    ✔ Single-quoted string
    ✔ Double-quoted string
    ✔ Single-quoted f-string
    ✔ Double-quoted f-string
    ✔ Single-quoted s-string
    ✔ Double-quoted s-string


  10 passing (10ms)
  7 failing
```

vanillajonathan:
Oh my bad, the tests worked locally because I had an old version of the grammar file, after getting the latest grammar, building and testing the tests failed, so I fixed them. Now they work.

I also added a test for Unicode characters in identifiers. We need more tests though so it can be more comprehensive.

Also we need make sure the expressions defined in the test files are those we want so that tests which pass actually pass because they're correct.

max-sixty:
I now get the unicode test failing. Can we add these to CI? It's difficult to manage code additions without CI if the failure rate is non-negligible.

I think we also want to either commit or ignore the `package-lock.json`. (I think we generally commit them...)

vanillajonathan:
Oh, the Unicode fails because I foolishly added them to this PR after I saw the other PR got merged, so now this PR got an outdated grammar, without support for Unicode.

Yeah, I am looking into CI.

vanillajonathan:
I got tests running on CI now, but maybe you know something about this measure-code-cov / tarpaulin thing?

max-sixty:
> maybe you know something about this measure-code-cov / tarpaulin thing?

Yes, an upstream dependency changed, let me fix!

max-sixty:
I also pushed a change here for the CI — we've changed how we organize those a bit

vanillajonathan:
> I also pushed a change here for the CI — we've changed how we organize those a bit

Alright. I had it in one separate file as a separate job that only ran when the files in the /grammars/prql-lezer/ directory were modified as I didn't want the job to always run even when files that were not related were modified.

max-sixty:
> Alright. I had it in one separate file as a separate job that only ran when the files in the /grammars/prql-lezer/ directory were modified as I didn't want the job to always run even when files that were not related were modified.

Yes, great. We have similar functionality as that, but with more fine-grained rules, and can gate merges based on those rules. And don't run lots of things twice when we merge like we used to. So my changes keep your functionality.

max-sixty:
Merging!

Great work @vanillajonathan . If we want to extend this, we could add some of the canonical examples — e.g. the examples from the playground. They are big, so we wouldn't want to type out what we expect — but hopefully the test framework can generate the result.

---

Separately — me trying to be helpful — I also notice that you might be creating branches & uploading files from GitHub's web interface. If so — it's impressive you've made so much progress doing it that way! If you have access to a more native way of using git — either from the CLI or from Codespaces, you might find it much easier to make changes & test code that way. Let me know if I can be helpful!

Issue #3260, opened 2023-08-05
Special function to return PRQL version


eitsupi:
Integrations such as the prql dialect on ClickHouse (which does not display post-compilation SQL) do not easily know the version of the PRQL language that can be used.
I wonder if it might be worth implementing a special function (?) that returns the PRQL compiler version as a string to know the PRQL version in such cases.

Something like:

```elm
from prql_version
```

Compiles to

```sql
SELECT '0.9.4'
```

This is also useful:

```elm
from foo
select bar = prql_version
```

Compiles to

```sql
SELECT
  '0.9.4' AS bar
FROM
  foo
```

max-sixty:
Yes — maybe a `prql` namespace, so like `prql.version`?

eitsupi:
> Yes — maybe a `prql` namespace, so like `prql.version`?

I like this.
It makes sense to reserve the `prql` namespace.

vanillajonathan:
Now imagine if `prql.version` was a struct or a object, so you could do `prql.version.major` and `prql.version.minor`.

hulxv:
I want to take this issue. How can I start from the codebase?

max-sixty:
> I want to take this issue. How can I start from the codebase?

Great.

We can model this on `std.floor`. Here's an example of its use: https://github.com/PRQL/prql/blob/0d342efd4482580244f702565c97cc779d692c42/crates/prql-compiler/src/semantic/eval/mod.rs#L474

Here's where it's defined: https://github.com/PRQL/prql/blob/0d342efd4482580244f702565c97cc779d692c42/crates/prql-compiler/src/semantic/eval/mod.rs#L157-L167

And here's where we can get the version from: https://github.com/PRQL/prql/blob/9c6ea2a2fa3eea2597b8596c4231f46036a30f4b/crates/prql-compiler/src/lib.rs#L106-L108

Does that make sense? Feel free to ask any questions!

vanillajonathan:
But could it be a struct on `prql` though so you could do `prql.version.major` and `prql.version.minor`?

max-sixty:
> But could it be a struct on `prql` though so you could do `prql.version.major` and `prql.version.minor`?

Maybe later, let's keep it simple at first

hulxv:
Thanks, @max-sixty  for help. but I have another question, please.

> Yes — maybe a `prql` namespace, so like `prql.version`?

[...]

PR #3533, opened 2023-09-18
feat: Special function to return PRQL version

hulxv:
Add a new function to std library to return PRQL version (see #3260)

**Simple example** 
```elm
from updates
filter version = prql_version
```
```sql
Select 
  *
from 
  updates
WHERE
  version = "0.9.5"
```

### DIFF
--- a/crates/prql-compiler/src/semantic/eval/mod.rs
+++ b/crates/prql-compiler/src/semantic/eval/mod.rs
@@ -277,11 +277,6 @@ impl Evaluator {
                 ExprKind::Array(array)
             }
 
-            "std.prql_version" => {
-                let ver = COMPILER_VERSION.to_string();
-                ExprKind::Literal(String(ver))
-            }
-
             _ => {
                 return Err(Error::new_simple(format!("unknown function {func_name}"))
                     .with_span(span)
@@ -404,7 +399,6 @@ fn std_module() -> Expr {
             new_func("columnar", &["closure", "relation"]),
             new_func("sum", &["x"]),
             new_func("lag", &["x"]),
-            new_func("prql_version", &[]),
         ]
         .to_vec(),
     ))
@@ -542,21 +536,6 @@ mod test {
             std.columnar {g = std.lag b}
         ").unwrap(),
             @"[{g = null}, {g = 4}, {g = 5}]"
-        );
-    }
-
-    #[test]
-    fn prql_version() {
-        println!("{:?}", eval("std.prql_version"));
-        println!(
-            "{:?}",
-            crate::compile(
-                r"
-            from updates
-            filter version = prql_version
-        ",
-                &crate::Options::default()
-            )
         );
     }
 }
--- a/crates/prql-compiler/src/semantic/eval/mod.rs
+++ b/crates/prql-compiler/src/semantic/eval/mod.rs
@@ -6,7 +6,6 @@ use itertools::Itertools;
 use super::ast_expand;
 use crate::error::{Error, Span, WithErrorInfo};
 use crate::ir::pl::{Expr, ExprKind, Func, FuncParam, Ident, Literal, PlFold};
-use crate::COMPILER_VERSION;
 
 pub fn eval(expr: prql_ast::expr::Expr) -> Result<Expr> {
     let expr = ast_expand::expand_expr(expr)?;
@@ -105,7 +104,6 @@ fn lookup(base: Option<&Expr>, name: &str) -> Result<Expr> {
     if name == "std" {
         return Ok(std_module());
     }
-
     Err(Error::new_simple(format!("cannot find `{}` in {:?}", name, base)).into())
 }
 
--- a/crates/prql-compiler/src/semantic/eval/mod.rs
+++ b/crates/prql-compiler/src/semantic/eval/mod.rs
@@ -3,10 +3,10 @@ use std::iter::zip;
 use anyhow::Result;
 use itertools::Itertools;
 
+use super::ast_expand;
 use crate::error::{Error, Span, WithErrorInfo};
 use crate::ir::pl::{Expr, ExprKind, Func, FuncParam, Ident, Literal, PlFold};
-
-use super::ast_expand;
+use crate::COMPILER_VERSION;
 
 pub fn eval(expr: prql_ast::expr::Expr) -> Result<Expr> {
     let expr = ast_expand::expand_expr(expr)?;
@@ -277,6 +277,11 @@ impl Evaluator {
                 ExprKind::Array(array)
             }
 
+            "std.prql_version" => {
+                let ver = COMPILER_VERSION.to_string();
+                ExprKind::Literal(String(ver))
+            }
+
             _ => {
                 return Err(Error::new_simple(format!("unknown function {func_name}"))
                     .with_span(span)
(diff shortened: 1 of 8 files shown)


crates/prql-compiler/src/tests/test.rs
```diff
@@ -3402,6 +3402,16 @@ fn test_header_target_error() {
     "###);
 }
 
+#[test]
+fn prql_version() {
```
max-sixty, line 3406: @Hulxv this is the test I added

crates/prql-compiler/src/semantic/resolver/snapshots/prql_compiler__semantic__resolver__test__frames_and_names-2.snap
```diff
@@ -10,12 +10,12 @@ columns:
       input_name: customers
       except: []
 inputs:
-  - id: 185
+  - id: 186
```
max-sixty, line 13: FYI ignore these changes; it's an inconvenience that these change when we add an item to the stdlib...

max-sixty:
OK, I spent 20 minutes on this — it's actually harder than I expect. `std.floor` doesn't work all the way through the compiler yet, so isn't a good example. The functions that do work all the way through are mostly operators.

Let me have another look tomorrow. Sorry this isn't as easy as it should be — we haven't really been adding functions so far, so this is new territory.

If @aljazerzen is online, he'd likely know quite quickly...

crates/prql-compiler/src/semantic/eval/mod.rs
```diff
@@ -277,6 +277,11 @@ impl Evaluator {
                 ExprKind::Array(array)
             }
 
+            "std.prql_version" => {
+                let ver = COMPILER_VERSION.to_string();
+                ExprKind::Literal(String(ver))
+            }
```
aljazerzen: Eval is a separate branch of the compiler, which is used only for `prqlc debug eval` command. It is used (probably only by me) for quick experimentation with language design and is not part of the general compilation process.

So this chunk is not needed, really.

crates/prql-compiler/src/semantic/eval/mod.rs
```diff
@@ -399,6 +404,7 @@ fn std_module() -> Expr {
             new_func("columnar", &["closure", "relation"]),
             new_func("sum", &["x"]),
             new_func("lag", &["x"]),
+            new_func("prql_version", &[]),
```
aljazerzen: Same as above.

aljazerzen (commented):
So this function can be statically evaluated during resolving, as it does not depend on run-time data. This would translate the function call to a string literal in the first stage of the compiler, so this function call never hits the RQ IR.

To implement that, you can add a new case in `transforms.rs`, something similar to [`std.in`](https://github.com/PRQL/prql/blob/main/crates/prql-compiler/src/semantic/resolver/transforms.rs#L203-L232).

crates/prql-compiler/src/semantic/std.prql
```diff
@@ -195,3 +195,8 @@ let upper = column -> <text> internal std.upper
 ## File-reading functions, primarily for DuckDB
 let read_parquet = source<text> -> <relation> internal std.read_parquet
 let read_csv = source<text> -> <relation> internal std.read_csv
+
+
+## PRQL compiler functions
+
+let prql_version = -> <string> internal std.prql_version
```
max-sixty:
```suggestion
let prql_version = -> <text> internal std.prql_version
```

max-sixty:
> To implement that, you can add a new case in `transforms.rs`, something similar to [`std.in`](https://github.com/PRQL/prql/blob/main/crates/prql-compiler/src/semantic/resolver/transforms.rs#L203-L232).

Excellent, thanks. So @Hulxv — we should use `std.in` as the thing to copy, not `std.floor`. Sorry for giving the wrong info before, but now this should be quite reasonable from here.

hulxv:
What's the right way to test this function?
I tried this unit:
```rust
    fn test_prql_version() {
        println!(
            "{:?}",
            crate::compile(
                r"
        from updates
        derive version = prql_version
    ",
                &crate::Options::default()
            )
        )
    }
```
but now, I got this:
```
running 1 test
thread 'semantic::resolver::transforms::tests::test_prql_version' panicked at 'called `Option::unwrap()` on a `None` value', crates/prql-compiler/src/sql/operators.rs:49:53
test semantic::resolver::transforms::tests::test_prql_version ... FAILED

failures:

failures:
    semantic::resolver::transforms::tests::test_prql_version

test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 160 filtered out; finished in 0.59s
```

crates/prql-compiler/src/semantic/std.prql
```diff
@@ -195,3 +195,8 @@ let upper = column -> <text> internal std.upper
 ## File-reading functions, primarily for DuckDB
 let read_parquet = source<text> -> <relation> internal std.read_parquet
 let read_csv = source<text> -> <relation> internal std.read_csv
+
+
+## PRQL compiler functions
+
+let prql_version = -> <text> internal prql_version
```
max-sixty, line 202: @Hulxv I tried removing the `std.`, and it worked. 

I think that's the distinction between some of these that require going later in the pipeline vs. those that don't. But it's not that clear, I think we can use this effort to improve a lot how we can add functions...

crates/prql-compiler/src/tests/test.rs
max-sixty: And now it passes — I ran `cargo insta test --accept` and we can see the result now...

crates/prql-compiler/src/semantic/resolver/transforms.rs
```diff
@@ -1071,4 +1078,18 @@ mod tests {
             - Wildcard
         "###);
     }
+
+    #[test]
+    fn test_prql_version() {
+        println!(
```
max-sixty: Can we just run `.unwrap` to ensure it passes, rather than `println!`?

(or possibly we don't need this test given we have https://github.com/PRQL/prql/pull/3533/files#diff-35eb52ba10a796006603d51870d6532a65ad5d351ed21580bab2e0548c360a34R3406 — your call on adjusting vs removing)

max-sixty:
OK great — I think it's working.

Unless folks disagree, I would suggest that @Hulxv fixes the compiler error and then we merge, with `std.prql_version`. We can move to `prql.version` when the compiler is better set up to allow modules. Is that reasonable?

hulxv:
> OK great — I think it's working.
> 
> Unless folks disagree, I would suggest that @Hulxv fixes the compiler error and then we merge, with `std.prql_version`. We can move to `prql.version` when the compiler is better set up to allow modules. Is that reasonable?

Are there still any errors that need to be fixed? I think all good now

max-sixty:
Great, thank you @Hulxv !

Appreciate your patience. We'll use this what we learned here to make it easier to add functions...

If you'd like to add an entry to the Changelog in another PR, that would be welcome. It would also be great to add to the book (though I'm not sure exactly where atm...)

hulxv:
> If you'd like to add an entry to the Changelog in another PR, that would be welcome. It would also be great to add to the book (though I'm not sure exactly where atm...)

Should I add it to the features section or language?

max-sixty:
> Should I add it to the features section or language?

Features is great

PR #3539, opened 2023-09-18
internal: Create .gitignore for Zig

vanillajonathan:


max-sixty:
Nice!

One test we could add to CI is "are there any untracked files after running the build" — that would ensure we don't get these issues. There is probably an existing workflow that someone has created that we could add to the end of our jobs.

PR #3540, opened 2023-09-18
feat(Lezer grammar): Add scientific notation

vanillajonathan:
This adds scientific notation to the Lezer grammar with a test for `123e10`.


max-sixty:
Nice, thanks!

max-sixty:
One way to find gaps in the Lezer grammar would be to iterate through the files like https://github.com/prql/prql/blob/04b5174cba01bcc79837e69f0697486210d3d6dd/crates/prql-compiler/tests/integration/queries/invoice_totals.prql, and test whether they parse in Lezer (either snapshotting the results, or just checking they don't fail)

vanillajonathan:
Good advice, I managed to find some gaps!

PR #3542, opened 2023-09-19
feat(Lezer grammar): Improve grammar

vanillajonathan:
This patch relaxes the grammar so it allows multiple line breaks within arrays and tuples. It adds tests for it.

It also allows the use of the wildcard character `*` in identifiers which allows for the `/tests/integration/queries/distinct.prql` query.

It also fixes a incorrect reference in the grammar `Op_bin_only` when it was actually `op_bin_only`.


max-sixty:
Awesome! Very nice work @vanillajonathan !

PR #3543, opened 2023-09-19
feat(Lezer grammar): Add TypeScript definitions

vanillajonathan:
Add TypeScript definitions just as is done in upstream grammars.


max-sixty:
Are you familiar with this error:

```
  Unable to locate the project file. A project file (tsconfig.json or tsconfig.eslint.json) is required in order to use ts-standard.
```

from https://github.com/PRQL/prql/actions/runs/6240462983/job/16940618178?pr=3543#step:4:1934

vanillajonathan:
Well yeah those are needed for a TypeScript project, but this isn't a TypeScript project, it is a JavaScript project that offers TypeScript definitions, so I think that error should be ignored.

https://github.com/lezer-parser/javascript and https://github.com/lezer-parser/python/ doesn't provide any such files either.

max-sixty:
OK, that makes sense. We can add that linter to an ignore in https://github.com/prql/prql/blob/84979edc6d5c02d925a30020b24dd99e1bf3e87a/.mega-linter.yaml

Would be good for @eitsupi to check they're OK with this, if they're around — they know this well...

PR #3544, opened 2023-09-19
feat(Lezer grammar): Date and time support

vanillajonathan:
This fixes the broken date & time support and adds tests for it.

### DIFF
--- a/grammars/prql-lezer/src/prql.grammar
+++ b/grammars/prql-lezer/src/prql.grammar
@@ -28,7 +28,7 @@ expr_call { Expr | Func_call }
 Func_call { Ident ( (Named_arg | Assign | Expr))+ }
 
 // !term is required here, otherwise `[x+emp_no, -3]` is parsed as `x` & `+emp_no`
-term { !term Ident | Number | Tuple | Date | Nested_pipeline | op_unary term | String | R_string | S_string | F_string | Range | Array | Case }
+term { !term Ident | Number | Tuple | DateTime | Nested_pipeline | op_unary term | String | R_string | S_string | F_string | Range | Array | Case }
 
 Expr { term (( Op_bin ) term )* }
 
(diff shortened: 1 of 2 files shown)


grammars/prql-lezer/src/prql.grammar
```diff
@@ -53,11 +53,13 @@ Ident { ident_part ( "." (ident_part | "*"))* }
 Op_bin { op_bin_only | !term op_unary }
 
 @tokens {
+  Date { @digit+ "-" @digit+ "-" @digit+ }
+  Time { @digit+ ":" @digit+ ":" @digit+ ( "." @digit+ )? }
```
max-sixty, line 57: FYI this is a bit too strict — check out https://prql-lang.org/book/reference/syntax/literals.html#times — times can be shorter.

They can also include a timezone; e.g. `@2020-01-01T13:19:55-08:00`.

(That might be something good about pulling the examples from the broader repo — it would offer a wide scan of everything, and be easy to update. Wouldn't be trivial but wouldn't be too hard — we could pull from either the `.prql` files in https://github.com/prql/prql/blob/04b5174cba01bcc79837e69f0697486210d3d6dd/crates/prql-compiler/tests/integration/queries/constants_only.prql or — ambitious perhaps — parsing the PRQL from files such as https://github.com/prql/prql/blob/983bb3e2ca61685395334232ada3c01dae185162/web/book/tests/documentation/snapshots/documentation__book__reference__syntax__operators__wrapping-lines__0.snap#L3)

That said, this is still an improvement so am merging!

vanillajonathan: While looking at the dates and the time zone I saw that PRQL seem to not adhere to the [ISO 8601](https://en.wikipedia.org/wiki/ISO_8601) standard when it comes to the `Z` as PRQL requires a `+` before the `Z` but ISO does not.

max-sixty: Good spot! Fixed in https://github.com/PRQL/prql/pull/3545

PR #3546, opened 2023-09-19
feat(Lezer grammar): Add timezone and UTC support

vanillajonathan:
Also fixes time format in `hh:mm` format without seconds which are optional.


max-sixty:
Super!

PR #3547, opened 2023-09-19
test(Lezer grammar): Add test for arithmetics

vanillajonathan:


max-sixty:
Great!

max-sixty:
If you want to grow this, we could use some of the tools in https://discuss.codemirror.net/t/whats-the-best-to-test-and-debug-grammars/2542/5 (e.g. https://gist.github.com/msteen/e4828fbf25d6efef73576fc43ac479d2), to pretty-print large grammars...

How are you writing the outputs? If it's manual, then that does make changes more difficult, we could look into something like https://jestjs.io/docs/snapshot-testing to allow snapshotting.

(Probably doing one of those makes the other less important — if we have a few large tests, they're easier to update manually, if we have lots of small tests, then snapshotting is more important. Though we can also wait until we're using this in Jupyter before building out more infra, and you've had an excellent run at improving it already!)

vanillajonathan:
I am writing what to test (such as `1 + 1`) manual, then I run `npm run test` which says the test is failing and gives me the syntax tree that it expects, then I copy and paste that into the file and run the test again and it passes.

We now have a collection of syntax expressions (input to the parser) that we want the parser to be able to pass. We have syntax trees (output from the parser) that aren't necessarily always what we would like the trees to ideally look like, but we can tweak the grammar, generate a new parser, run the tests, and update the failing tests with the expected syntax trees once we look at them and deem that they reflect what we want them to look like.

I don't have any experience with Jest or snapshot testing.

The grammar still have issues remaining such as:
- `take 10..20` results in a parse error
- A comment on the line after a comment results in a weird syntax tree
- No support for triple quoted strings
- No support for escape sequences in strings

max-sixty:
> We now have a collection of syntax expressions (input to the parser) that we want the parser to be able to pass. We have syntax trees (output from the parser) that aren't necessarily always what we would like the trees to ideally look like, but we can tweak the grammar, generate a new parser, run the tests, and update the failing tests with the expected syntax trees once we look at them and deem that they reflect what we want them to look like.

Yes OK, this is very viable at the current scale. We can think about making it more robust much later.

> The grammar still have issues remaining such as:
> 
>     * `take 10..20` results in a parse error
> 
>     * A comment on the line after a comment results in a weird syntax tree
> 
>     * No support for triple quoted strings
> 
>     * No support for escape sequences in strings

Great observations! Those would be the top priorities, then. Probably in exactly that order...

Thank you!

PR #3549, opened 2023-09-20
feat(Lezer grammar): Fix support for ranges

vanillajonathan:
This adds support for ranges and lets you have `take 10..20` in your queries.

### DIFF
--- a/grammars/prql-lezer/src/prql.grammar
+++ b/grammars/prql-lezer/src/prql.grammar
@@ -99,7 +99,7 @@ Op_bin { op_bin_only | !term op_unary }
   // R, S & F strings have precedence over idents beginning with r / s / f (we could
   // use specialize but I think means we need to redefine `String` for each)
   @precedence { R_string, S_string, F_string, op_bin_only, ident_part }
-  
+
   @precedence { Range, Number }
 }
 
--- a/grammars/prql-lezer/src/prql.grammar
+++ b/grammars/prql-lezer/src/prql.grammar
@@ -38,7 +38,7 @@ Assign_call { ident_part "=" expr_call }
 Case_branch { Expr "=>" Expr }
 Array { "[" newline* Expr (("," newline* ) Expr)* ","? newline* "]" }
 // Possibly we could only accept case branches inside the Tuple?
-Case {  @specialize<ident_part, "case"> Tuple }
+Case { @specialize<ident_part, "case"> Tuple }
 
 Nested_pipeline { "(" newline* Pipeline ~ambig_newline newline? ")" }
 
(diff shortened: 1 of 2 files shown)


grammars/prql-lezer/src/prql.grammar
```diff
@@ -74,7 +74,7 @@ Op_bin { op_bin_only | !term op_unary }
   line_wrap { "\\" }
   wrapped_line { newline+ (Comment newline+)* line_wrap }
   newline { "\n" }
-  Range { ".." }
+  Range { @digit+ ".." @digit+ }
```
max-sixty, line 79: Because this can also be used to compile to `BETWEEN`, this can be any literal, and arguably any expression. It also shouldn't allow for spaces — is that possible with `![ ]`?

Could we test with `5..(-10)` and `@2022..@2023`?

(If some of these are very difficult, we can merge the partial version though)

grammars/prql-lezer/src/prql.grammar
```diff
@@ -53,13 +53,13 @@ Ident { ident_part ( "." (ident_part | "*"))* }
...
-  Date { @digit+ "-" @digit+ "-" @digit+ }
-  Time { @digit+ ":" @digit+ ( ":" @digit+ ( "." @digit+ )? )? }
+  date { @digit+ "-" @digit+ "-" @digit+ }
+  time { @digit+ ":" @digit+ ( ":" @digit+ ( "." @digit+ )? )? }
   // We can't seem to set the number of digits, so this will allow any
   // combination of digits & hyphens.
   // TODO: does excluding spaces here work? It seems to make a difference, but
   // then I think the `Number` rule doesn't allow spaces...
-  DateTime { "@" ![ ] ( Date | Time | Date "T" Time ( "Z" | ( "-" | "+" ) @digit+ ":" @digit+ )? ) }
+  DateTime { "@" ![ ] ( date | time | date "T" time ( "Z" | ( "-" | "+" ) @digit+ ":" @digit+ )? ) }
```
max-sixty, line 62: 👍

grammars/prql-lezer/src/prql.grammar
vanillajonathan: I don't know how to make it accept expressions instead of digits.

I also don't know about `![ ]` or even why our grammar allows implicit spaces, because the CodeMirror example grammar at https://github.com/codemirror/lang-example/blob/main/src/syntax.grammar doesn't.

Yeah, I would like this to be merged. I also found a way to add skipped tests which I could add for `5..(-10)` and `@2022..@2023` until we nail them.

max-sixty: I had a look but you're right, it isn't trivial...

max-sixty, line 79: Can we add a TODO?

```suggestion
  // TODO: Because this can also be used to compile to BETWEEN, ranges should
  // allow any literal, and arguably any expression. And if possible it shouldn't allow for spaces.
  Range { @digit+ ".." @digit+ }
```

max-sixty: Though we could use `Number` rather than `@digit+`?

vanillajonathan: Yeah, it seems we can use `Number` instead of `@digit+` but I don't know if that is desirable because then `123_456e10..123.00` would be valid which I am not sure it should be.

max-sixty: It's much better to be too permissive than too strict here, because false positives will be quickly corrected by the user (because it's invalid PRQL), but false negatives will continue to exist (because it's valid PRQL)

max-sixty: (that said, I'll merge and if you want we can make the change in another PR)

PR #3550, opened 2023-09-20
feat(highlight.js): New stuff

vanillajonathan:
This fixes the broken regex which caused some operators to be added as two separate tokens.

Adds a comment header to align with upstream highlight.js files.

Highlight built-in functions and data types.

Adds support for `Z` suffix in dates.

Adds the years, months and microseconds time units.

Adds support for the `~=` regex operator.

Marks `prql` (for version and target) as meta.

Adds support for for `r"strings"`.

Adds support for escape sequences such as `\n`.

Adds `#!` as a new docblock token with Markdown as sublanguage.

### DIFF
--- a/web/book/highlight-prql.js
+++ b/web/book/highlight-prql.js
@@ -40,15 +40,15 @@ formatting = function (hljs) {
   ];
 
   const DATATYPES = [
-    'bool',
-    'float', 
-    'int',
-    'int8',
-    'int16',
-    'int32',
-    'int64',
-    'text',
-    'timestamp'
+    "bool",
+    "float",
+    "int",
+    "int8",
+    "int16",
+    "int32",
+    "int64",
+    "text",
+    "timestamp",
   ];
 
   const TRANSFORMS = [
@@ -84,7 +84,7 @@ formatting = function (hljs) {
         begin: "#!",
         end: "$",
         subLanguage: "markdown",
-        relevance: 0
+        relevance: 0,
       },
       hljs.COMMENT("#", "$"),
       {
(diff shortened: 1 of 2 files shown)


web/book/highlight-prql.js
```diff
@@ -35,16 +66,31 @@ formatting = function (hljs) {
...
+      built_in: BUILTIN_FUNCTIONS,
       keyword: [...TRANSFORMS, ...BUILTIN_FUNCTIONS, ...KEYWORDS],
-      literal: "false true null ",
+      literal: "false true null",
+      type: DATATYPES,
     },
     contains: [
+      {
+        // docblock
+        begin: "#!",
```
max-sixty, line 89: Nice

max-sixty (approved):
Excellent! This is really good

web/book/highlight-prql.js
max-sixty: One nice addition to this would be changing a comment in the book to use this format (or adding another example), so that in the web book it'll be highlighted as markdown

vanillajonathan: Yeah, I sneaked that one in there, and it should work for at least for headers, bold, italic, and such, but since the token ends when the line ends, it probably doesn't work with multi-line Markdown stuff such as fenced code blocks, lists, and tables.

Also the the PRQL parser doesn't emit any tokens for `#!` so it doesn't get result in anything on the syntax tree which could be used for documentation generation.

I suppose we could extend the Lezer grammar with the [`docComment`](https://lezer.codemirror.net/docs/ref/#highlight.tags.docComment) token, and Ace with the [`block.documentation`](https://github.com/ajaxorg/ace/wiki/Creating-or-Extending-an-Edit-Mode#common-tokens) token.

max-sixty: > but since the token ends when the line ends, it probably doesn't work with multi-line Markdown stuff such as fenced code blocks, lists, and tables.

Ah good point...

I guess we could make it though! Basically allow repeated `\s*\n!#`, if that works...

PR #3562, opened 2023-09-22
feat(Lezer grammar): Add highlight.js

vanillajonathan:
Adds a `highlight.js` file as mentioned in the manual: https://lezer.codemirror.net/docs/ref/#highlight


max-sixty:
Great!

~To confirm, this isn't quite `highlight.js`, since that's a [specific project](https://highlightjs.org/), right? (Or tell me if I'm missing something, would obv be great to switch out our `highlight.js` grammars for this, but I don't think that's possible...?)~

Edit: I see, we've named the file `highlight.js`. Given that overlaps with the project, could we rename it to something like `prql_highlight.js`?

vanillajonathan:
I suppose we could, but upstreams all the parsers name it `highlight.js`.
https://github.com/lezer-parser/cpp
https://github.com/lezer-parser/css
https://github.com/lezer-parser/html
https://github.com/lezer-parser/javascript
https://github.com/lezer-parser/python
https://github.com/lezer-parser/xml

max-sixty:
> I suppose we could, but upstreams all the parsers name it `highlight.js`.

OK good point!

PR #3564, opened 2023-09-23
feat: Add Pygments to grammar README

vanillajonathan:


max-sixty:
@vanillajonathan that's great! Thank you very much!

I tried it on the demo page and it looks good...

vanillajonathan:
It's alright, but in the query `from pets` both `from` and `pets` gets the same token (`Name.Variable`). Although, that is the same behavior as the Elm parser. Also the same behavior as the Bash parser (although it outputs the `Text` token for both words instead of the `Name.Variable` token). It is a unlike the PowerShell parser and the SQL parser which emits `Keyword` and `Name`.

Notably absent is support for time units, e.g. `years`, `months`, `days`, etc.

max-sixty:
Yeah. I think it's nice to have the first transform be a different color, because it highlights the pipeline very clearly.

image

At the moment we do that by hardcoding the transform names. That's a presumption, since someone could define their own `from2` transform, which then wouldn't be highlighted. 

But I think it's a reasonable compromise, and for the simpler grammars which don't contain semantics, the correct tradeoff.

vanillajonathan:
Yeah, I think so too.

Pygments have a list of available tokens at https://pygments.org/docs/tokens/
It not always obvious which token to use to match the semantics, but I guess picking some different token even though the intended semantic is wrong would be a reasonable compromise just to have different tokens for the transforms and the parameters.

PR #3566, opened 2023-09-23
chore(highlightjs): Add metadata

vanillajonathan:
This adds the `requires` field to the metadata since we depend on Markdown for our docblocks, i.e. the `#!` comments.

https://highlightjs.readthedocs.io/en/latest/language-contribution.html?highlight=requires#add-language-metadata


max-sixty:
Looks good, thanks for the link

PR #3567, opened 2023-09-24
feat(Lezer grammar): Tokens for ints and floats

vanillajonathan:
This separates the `Number` token into separate tokens for integers and floats.


max-sixty:
Looks good! 

(Relative to your other PRs, I'm not sure this adds _that_ much — distinguishing between these is less impactful — but fine to include)

vanillajonathan:
The reason is that we can less ambiguously and more richly expose the grammar to the highlighter.

If you look at the tokens at:
https://lezer.codemirror.net/docs/ref/#highlight.tags.integer

You see that it has an dedicated token for `integer` and `float` so with this refactor we can make use of those. This would allow themes to style integers and floats differently.

max-sixty:
Yes that makes sense, but:

> This would allow themes to style integers and floats differently.

...I just think this is quite unlikely!

PR #3569, opened 2023-09-24
feat: Long Unicode escape sequence

vanillajonathan:
I like this turtle! 🐢

Previously we had the `\uXXXX` escape sequence, this patch introduces the `\U00XXXXXX` escape sequence so we can insert turtles and stuff.

### DIFF
--- a/crates/prql-parser/src/lexer.rs
+++ b/crates/prql-parser/src/lexer.rs
@@ -365,12 +365,10 @@ fn escaped_character() -> impl Parser<char, char, Error = Cheap<char>> {
                 .exactly(6)
                 .collect::<String>()
                 .validate(|digits, span, emit| {
-                    char::from_u32(u32::from_str_radix(&digits, 16).unwrap()).unwrap_or_else(
-                        || {
-                            emit(Cheap::expected_input_found(span, None, None));
-                            '\u{FFFD}' // unicode replacement character
-                        },
-                    )
+                    char::from_u32(u32::from_str_radix(&digits, 16).unwrap()).unwrap_or_else(|| {
+                        emit(Cheap::expected_input_found(span, None, None));
+                        '\u{FFFD}'
+                    })
                 }),
         )),
         (just('x').ignore_then(
--- a/crates/prql-parser/src/lexer.rs
+++ b/crates/prql-parser/src/lexer.rs
@@ -359,6 +359,20 @@ fn escaped_character() -> impl Parser<char, char, Error = Cheap<char>> {
                     })
                 }),
         )),
+        (just("U00").ignore_then(
+            filter(|c: &char| c.is_ascii_hexdigit())
+                .repeated()
+                .exactly(6)
+                .collect::<String>()
+                .validate(|digits, span, emit| {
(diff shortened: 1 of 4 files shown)


web/book/highlight-prql.js
```diff
@@ -72,7 +72,7 @@ formatting = function (hljs) {
 
   const CHAR_ESCAPE = {
     scope: "char.escape",
-    match: /\\\\|\\([bfnrt]|u\d{4})|x[0-9A-Fa-f]{2}/,
+    match: /\\\\|\\([bfnrt]|u\d{4})|U00[0-9A-Fa-f]{6}|x[0-9A-Fa-f]{2}/,
```
max-sixty, line 75: Unrelated — but `u\d{4}` looks like it should be using hex rather than digits...

crates/prql-parser/src/lexer.rs
```diff
@@ -359,6 +359,18 @@ fn escaped_character() -> impl Parser<char, char, Error = Cheap<char>> {
                     })
                 }),
         )),
+        (just("U00").ignore_then(
+            filter(|c: &char| c.is_ascii_hexdigit())
+                .repeated()
+                .exactly(6)
```
max-sixty, line 365: FWIW it might be easier to express these as just 8 characters, and then let errors fall through as with any other invalid code point, rather than restricting to a prefix of `00`...

max-sixty (approved):
Cool!

web/book/highlight-prql.js
vanillajonathan: Yeah, I think so too.

crates/prql-parser/src/lexer.rs
vanillajonathan: Yeah, maybe, I don't know.

web/book/highlight-prql.js
vanillajonathan: I just filed PR #3570 to fix this.

richb-hanover:
@vanillajonathan A request: Update the CHANGELOG to include examples of both Unicode and : `\U00XXXXXX` and `0xABCD` or whatever's correct. Many thanks!

max-sixty:
> @vanillajonathan A request: Update the CHANGELOG to include examples of both Unicode and : `\U00XXXXXX` and `0xABCD` or whatever's correct. Many thanks!

This would indeed be nice, but isn't strictly required — to the extent there's a tradeoff with the inclination to contribute — a higher volume of contributions compounds better vs. a higher quality changelog.

But agree it would be an improvement — either Changelog or Docs...

richb-hanover:
As an occasional documentation writer, it's enormously helpful to have a short example in the CHANGELOG that I can try to see if my draft description matches how it works.

In the absence of definitive info from the developer who added the feature, my options are to guess, to rummage through the commit and attempt to decode what it says, or to punt and let someone else write the final documentation. Thanks.

max-sixty:
Yeah, I do hear that. So it's a tradeoff.

I still end up on developer velocity being more important than documentation for most code. Particularly taking into account that for folks who don't necessarily have English as a first language, writing code is often easier than writing the docs...

For things that are important for understanding the language, I do go back and add docs if someone didn't add them.

PR #3570, opened 2023-09-24
feat(highlightjs): Add hex and long Unicode escape

vanillajonathan:
This add hex escapes using the `\xFF` escape sequence and long Unicode escape sequences using the `\U00FFFFFF` escape sequence.

It also adds a bunch of missing functions.


max-sixty:
Super!

PR #3571, opened 2023-09-25
docs: Example of hex escape sequences and long Unicode escape sequences

vanillajonathan:
This adds an escape sequence example to the change log as requested in #3569.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,8 +11,10 @@
 **Features**:
 
 - Add `std.prql_version` function to return PRQL version (@hulxv, #3533)
-- Add support for hex escape sequences in strings. (@vanillajonathan, #3568)
-- Add support for long Unicode escape sequences. (@vanillajonathan, #3569)
+- Add support for hex escape sequences in strings.
+  Example `"Hello \x51"`. (@vanillajonathan, #3568)
+- Add support for long Unicode escape sequences.
+  Example `"Hello \U0001F422"`. (@vanillajonathan, #3569)
 
 **Fixes**:
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,10 +11,10 @@
 **Features**:
 
 - Add `std.prql_version` function to return PRQL version (@hulxv, #3533)
-- Add support for hex escape sequences in strings.
-  Example `"Hello \x51"`. (@vanillajonathan, #3568)
-- Add support for long Unicode escape sequences.
-  Example `"Hello \U0001F422"`. (@vanillajonathan, #3569)
+- Add support for hex escape sequences in strings. Example `"Hello \x51"`.
+  (@vanillajonathan, #3568)
+- Add support for long Unicode escape sequences. Example `"Hello \U0001F422"`.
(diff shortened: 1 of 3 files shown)


max-sixty:
Thank you for doing this! And to @richb-hanover for suggesting.

---

Because we snapshot the results from the examples in the docs, this will require running `cargo insta test --accept` to accept the new snapshots, in order for the tests to pass.

It looks like your local setup might be in a good state recently @vanillajonathan , but if you're not quite there, I can help you get it set up, or I can run that command as a one-off so we can merge this.

richb-hanover:
Excellent! That's just what I was hoping for. Thanks.

vanillajonathan:
@max-sixty hmm... the snapshot fails

max-sixty:
I was confused for a bit but now I realize what's going on — this is actually a legit bug with our code gen.

For the moment, we can surround this code block with `no-fmt`

Running `echo 'from x | derive turtle = "\U0001F422"' | cargo run -p prqlc -- fmt` demonstrates this — our code gen creates `\u{...}` strings, because that's how rust represents them, and we don't override how we write strings (default impl [here](https://github.com/prql/prql/blob/04b5174cba01bcc79837e69f0697486210d3d6dd/crates/prql-compiler/src/codegen/mod.rs#L34-L38)).

The test that is failing ensures when we format PRQL, we can compile the result. But we can't in this case. So in another PR, we should either parse `\u{...}` strings (we could even move our unicode parsing to this if we think this format is better), or output the compatible format.

max-sixty:
In thinking about the best representation — [here's a survey](https://chat.openai.com/share/ee4d563b-ea23-47e4-945b-534b8a196f8f)[^1] of what other languages do. The more modern languages (Rust, Swift) seem to generally use brackets. It means there can be a single `u` prefix and then a variable number of characters inside the brackets.

So we could make that change to use brackets — I don't think it would be that much work to make the switch relative to what @vanillajonathan has already done. I don't have a strong view on the best representation. 

(Though we should make a decision before changing `codegen.rs`, since that's new work which would be wasted if we later changed to align with rust.) 

CC @aljazerzen @eitsupi @snth 

[^1]: I'm generally not a fan of "Here's what ChatGPT says", but in this case it is a useful survey of the langs, I'm not sending blindly

max-sixty:
@vanillajonathan I'll merge your code now that I've adjusted the test. Thank you!

aljazerzen:
Looking at the survey, I'm for the brackets (`\u{00AD}`) representation. Mainly because the plain one (`\u00AD`) has problems with "where does the notation stop".

max-sixty:
Yes, I'm tempted to agree.

@vanillajonathan let us know if that's something you'd be interested in doing. I actually think relative to the work you've done already, it's not that difficult a change. But if you feel it's a lot, then no pressure — there's no huge rush to make the change. And the current state is pretty good.

snth:
No strong feelings on this but the bracket notation seems good an unambiguous and I can't think of cases where you would want many in a row so the downside doesn't seem that big. If someone presents examples where this is a problem for languages like Japanese or Arabic then I might change my mind.

PR #3578, opened 2023-09-26
build: Bump MSRV to 1.70

max-sixty:


eitsupi (commented):
Is there any reason why you would want to switch to 1.70 immediately?

Windows on CRAN seems to still be running Rust 1.69, so if you do the 0.10.0 release soon, I'll have to do some work to fix MSRV...

crates/prqlc/Cargo.toml
```diff
@@ -5,9 +5,12 @@ name = "prqlc"
 edition.workspace = true
 license.workspace = true
 repository.workspace = true
-rust-version = "1.70.0"
+rust-version.workspace = true
```
eitsupi: I'm not sure if it's worth inheriting rust-version here.

CHANGELOG.md
```diff
@@ -30,7 +30,7 @@
 
 **Integrations**:
 
-- Bump `prqlc`'s MSRV to 1.70.0 (@eitsupi, #3521)
+- Bump `prql-compiler` & `prqlc`'s MSRV to 1.70.0 (@eitsupi, #3521)
```
eitsupi: Not linked to this PR.

eitsupi (changes requested):

max-sixty:
> Windows on CRAN seems to still be running Rust 1.69, so if you do the 0.10.0 release soon, I'll have to do some work to fix MSRV...

Ah, we can just wait for the slowest dependency then. No great rush, no reason to delay once the dependencies are on 1.70.0 though...

crates/prqlc/Cargo.toml
max-sixty: No strong view; it seemed more consistent...

max-sixty:
~@eitsupi do you happen to know whether winget requires a matching rust compiler? Debian does, but when I look at the winget manifest, it seems to take a binary: https://github.com/microsoft/winget-pkgs/pull/120030/files#diff-8a3a9ea01807574e4e83d54fadb778aafafb9002ab85cc894986ddc2b9ace7ceR15, suggesting that we don't need to wait for it to bump this?~

Sorry, not sure why I had a memory that this was about winget! The binding constraint is actually CRAN.

eitsupi:
In addition to Rust on Windows and macOS not being updated, Fedora on CRAN also hasn't been updated recently for some reason and appears to be stuck at Rust 1.69.

I hope it will be updated in the next R release which is scheduled for 2023-10-31...

max-sixty:
> I hope it will be updated in the next R release which is scheduled for 2023-10-31...

Cool, no rush from PRQL's side if not though!

eitsupi:
Sorry it took me so long to get here, I was wondering why Fedora's Rust on CRAN was out of date, but the cause was that it was still Fedora 36, which was out of support in the first place.
This is obviously a problem, so I sent an email to the mailing list.
<https://stat.ethz.ch/pipermail/r-package-devel/2023q4/010146.html>

Now that Ubuntu LTS rustc has also reached 1.70, it makes perfect sense to set MSRV to 1.70.

This PR is out of date, so open another one.

eitsupi:
Superseded by #3578.

PR #3588, opened 2023-10-01
feat(Lezer grammar): Rework grammar with semantic parse tree

vanillajonathan:
Previously the tests passed because they were crafted to pass, not because they reflected what the parse tree ought to look like. This patch reworks the grammar to output a more idiomatic parse tree.

Previously the parse tree looked like `Expr(Integer,Op_bin,Integer)` and now looks like `BinaryExpression(Integer,ArithOp,Integer)` or `BinaryExpression(Integer,CompareOp,Integer)`.

- A `Docblock` token is introduced.
- Time units now have their own `TimeUnit` token.
- The `highlight.js` file now maps the `Docblock`  and `TimeUnit` token to `t.docString` and `t.unit` respectively.
- The `@precedence` block has been moved up to the top of the file as per upstream convention.
- Tokens named with underscores have been renamed to camel case or pascal case as per upstream convention.
  - `Query_def` is now `QueryDefinition`. I am unsure about the name and consider `MetaDeclaration` or `PragmaDeclaration`. Named args are no longer optional.
  - `Def` is now `VariableDeclaration` as per upstream convention.
  - `Named_arg` is now `NamedArg`.
  - `Pipeline_stmt` is now `PipelineStatement`. The "Statement" suffix is by upstream convention.
  - `Tuple` and `Array` have been renamed to `TupleExpression` and `ArrayExpression` respectively as are the names in the Python grammar.
  - `Range` have been renamed to `RangeExpression`.
  - `Func_call` has been renamed to `CallExpression` as per upstream convention.
- `commaSep` is introduced as a internal generic token, it is currently only used by the `ArrayExpression` token. This token is a conventional token found in upstream grammars.
- The `BinaryExpression` token have been introduced along with the `LogicOp` and `CompareOp` tokens. The `Op_bin` token is gone. This is per upstream convention.
- The expression tokens are grouped together under the new internal `expression` token as per upstream convention.
- The `Interpolated_string` have been renamed to `interpolatedString` and is now a internal token.
- Tokens that resided below the `@token` block have been moved up as per upstream convention.
- Triple-quoted strings are now (kind of) supported.
- New tests
  - Array of integers
  - Array of floats
  - Comment
  - Docblock, i.e. `#!`
  - Variable declaration
  - Function declaration
  - Function declaration with two arguments
  - Simple pipeline
  - Coalesce operator
  - Triple-quoted single-quoted string, i.e. `'''`
  - Triple-quoted double-quoted string, i.e. `"""`


grammars/prql-lezer/src/prql.grammar
```diff
@@ -3,114 +3,153 @@
...
+NestedPipeline { "(" newline* Pipeline ~ambigNewline newline? ")" }
 
-Named_arg { ident_part ":" Expr }
-Assign { ident_part "=" Expr }
-Assign_call { ident_part "=" expr_call }
-Case_branch { Expr "=>" Expr }
-Array { "[" newline* Expr (("," newline* ) Expr)* ","? newline* "]" }
-// Possibly we could only accept case branches inside the Tuple?
-Case { @specialize<ident_part, "case"> Tuple }
+commaSep<expr> { newline* expr ("," newline* expr newline*)* ","? }
```
max-sixty, line 48: This is only used once, is it worth having it generic?

grammars/prql-lezer/src/prql.grammar
```diff
@@ -3,114 +3,153 @@
...
+LogicOp<expr> { expr }
+
+test {
+ testInner
+}
```
max-sixty:
```suggestion
test { testInner }
```

grammars/prql-lezer/src/prql.grammar
```diff
@@ -3,114 +3,153 @@
...
-Def { @specialize<ident_part, "let"> ident_part "=" (Nested_pipeline (newline+ | end) | Lambda) }
-
-Lambda { Lambda_param* "->" expr_call ( newline+ | end ) }
-Type_def { "<" Type_term ( "|" Type_term)* ">" }
-Type_term { ident_part Type_def? }
-Lambda_param { ident_part Type_def? (":" Expr)? }
-
 @external propSource prqlHighlight from "./highlight"
+
+@detectDelim
```
max-sixty, line 155: What does this do?

grammars/prql-lezer/src/prql.grammar
```diff
@@ -3,114 +3,153 @@
...
+  wrappedLine { newline+ (Comment newline+)* lineWrap }
   newline { "\n" }
   // TODO: Because this can also be used to compile to BETWEEN, ranges should
   // allow any literal, and arguably any expression.
-  Range { @digit+ ".." @digit+ }
+  RangeExpression { @digit+ ".." @digit+ }
   // Couldn't manage to do these & the interpolated as a template; couldn't
   // find how to negate a variable template
-  String { $["] !["]* $["] | $['] ![']* $['] }
+  String { '"""' ![Z]* '"""' | "'''" ![Z]* "'''" | $["] !["]* $["] | $['] ![']* $['] }
```
max-sixty: What are the `Z`s for?

grammars/prql-lezer/src/prql.grammar
```diff
@@ -3,114 +3,153 @@
...
-Case { @specialize<ident_part, "case"> Tuple }
+commaSep<expr> { newline* expr ("," newline* expr newline*)* ","? }
+LogicOp<expr> { expr }
+
+test {
+ testInner
+}
 
-Nested_pipeline { "(" newline* Pipeline ~ambig_newline newline? ")" }
+testInner { binaryTest | unaryTest | expression }
```
max-sixty, line 54: How come these are all called `test`?

max-sixty (approved):
This is excellent @vanillajonathan ! Nice work! It is much better, and I see you've standardized things.

I didn't quite understand all the `@` — that's OK though, I see the tests and the results look good.

I asked a couple of questions, and then we can merge...

grammars/prql-lezer/src/prql.grammar
vanillajonathan: That is debatable, but arguably. The `commaSep` token is commonly used upstreams.

It is used in the JavaScript, Rust, Python, C++, CSS, Java, PHP and Sass grammars, so most of the grammars except for HTML, XML, Markdown, JSON and Julia.

In other languages it is re-used more than once though. In Python for arguments and dictionaries, in JavaScript for arguments, objects and variable declarations, etc.

I do think it does a good job of clearly communicating intent while abstracting away implementation details. Maybe we could re-use it for tuples, or the `select` transform. I don't know.

grammars/prql-lezer/src/prql.grammar
vanillajonathan: Those are called coincidentally called `test` just because that is what they're called in the [Python grammar](https://github.com/lezer-parser/python/blob/main/src/python.grammar). They're not called that in the JavaScript or Rust parser though, so we could rename it to something else. Lezer is written in JavaScript, and the JavaScript  grammar have the most "stars" on GitHub, so I consider it the most authoritative.

grammars/prql-lezer/src/prql.grammar
vanillajonathan: It means any character except uppercase `Z`. It is a dirty workaround that I used because I don't know how to express any character except three consecutive double quotes.

While on the topic of strings, the [JavaScript grammar](https://github.com/lezer-parser/javascript/blob/main/src/javascript.grammar) have a `Escape` token. We would like that too. It can be mapped to [`escape`](https://lezer.codemirror.net/docs/ref/#highlight.tags.escape).

grammars/prql-lezer/src/prql.grammar
max-sixty: OK, cool!

grammars/prql-lezer/src/prql.grammar
vanillajonathan: It is used in most of the grammars, it is used in the CSS, Java, JavaScript, JSON, PHP, Python, Rust, Sass and XML grammars.

The documentation states:
> If you put a @detectDelim directive at the top level of your grammar file, the parser generator will automatically detect rule delimiter tokens, and create [openedBy](https://lezer.codemirror.net/docs/ref/#common.NodeProp%5EopenedBy) and [closedBy](https://lezer.codemirror.net/docs/ref/#common.NodeProp%5EclosedBy) props for rules where it finds them.

and...
> And finally, the parser generator can be asked to automatically infer matching delimiters with a @detectDelim directive. This will cause it to add [metadata](https://lezer.codemirror.net/docs/ref/#tree.NodeProp%5EclosedBy) to those node types, which the editor can use for things like bracket matching and automatic indentation.

grammars/prql-lezer/src/prql.grammar
max-sixty: OK, totally fine to have workarounds — but we have to have comments for these sorts of things, or the next person is going to have no idea why the `Z` is there!

I also don't completely understand how it's helping — why is it different from having "any character" — why does it treat `"` differently from "any character"? (quite possibly I'm misunderstanding...) 

Isn't it also going to break on any strict that contains `Z`?

Hope these questions are reasonable...

grammars/prql-lezer/src/prql.grammar
max-sixty: OK, maybe add a short comment saying that's why they're called `Test`, so it's not interpreted as something fundamentally different from the other tokens. Thanks...

grammars/prql-lezer/src/prql.grammar
max-sixty: Ah thanks! 

(and fine to just tell me to look at the docs in future!)

vanillajonathan:
> This is excellent @vanillajonathan ! Nice work! It is much better, and I see you've standardized things.

Thanks. Yes, I looked at upstream grammars and tried to adopt to their convention about naming, syntax and structure.

> I didn't quite understand all the @ — that's OK though, I see the tests and the results look good.

The `@digit` is a built-in character set that matches 0 to 9. ([reference](https://lezer.codemirror.net/docs/guide/#tokens))

The `@precedence` is a built-in block to define which token takes precedence during parsing when conflicts arise. Example the `Docblock` token must take precedence over `Comment` token since both of them start with the `#` character. ([reference](https://lezer.codemirror.net/docs/guide/#precedence))

`@top` is a built-in that defines the entry point to the grammar. ([reference](https://lezer.codemirror.net/docs/guide/#writing-a-grammar))

`@skip` is a built-in used to add tokens that can appear anywhere, such as comments and spaces. ([reference](https://lezer.codemirror.net/docs/guide/#skip-expressions))

`@specialize` is a built-in. ([reference](https://lezer.codemirror.net/docs/guide/#token-specialization))

`@external` is a built-in to load stuff from a JavaScript file, it can be used for advanced rules such as semicolon in JavaScript, or indention in Python, we just use it to load the highlighter as is commonly done in other upstream grammars. ([reference](https://lezer.codemirror.net/docs/guide/#external-tokens))

`@name` is a built-in used to set the node name to something different than what the token is named. ([reference](https://lezer.codemirror.net/docs/guide/#node-props))

`@isGroup` is a built-in used to create node groups. We use it to group all our expressions (such as `BinaryExpression`, `ArrayExpression`, `TupleExpression`, etc) under one internal `expression` token as is commonly done in upstream grammars. ([reference](https://lezer.codemirror.net/docs/guide/#external-tokens))

max-sixty:
> `@isGroup` is a built-in used to create node groups. We use it to group all our expressions (such as `BinaryExpression`, `ArrayExpression`, `TupleExpression`, etc) under one internal `expression` token as is commonly done in upstream grammars. ([reference](https://lezer.codemirror.net/docs/guide/#external-tokens))

OK great — this was the one I wasn't familiar with. Thanks for the explanation...

grammars/prql-lezer/src/prql.grammar
vanillajonathan: Okay, I added a comment about this.

grammars/prql-lezer/src/prql.grammar
vanillajonathan: You're right, I should have added a comment to make it more obvious. I have now added a comment, and also changed the `Z` to a 🍔 which is more obvious and less likely to be used.

I am not aware of any "any character" token in Lezer.

Yes, it is going to break on any string that contains `Z` (well now 🍔) but it's the price for the workaround, at least we got triple-quoted strings now.

max-sixty: Thanks!

You're right about there not being anything like `@any`. I hadn't realized that didn't exist.

FYI this might be helpful: https://discuss.codemirror.net/t/matching-all-text-until-a-specific-token-is-encountered-without-externaltokenizer/5573/3, which links to https://lezer.codemirror.net/docs/guide/#local-token-groups

max-sixty:
Thanks again @vanillajonathan !

vanillajonathan:
We have the `F_string`, `R_string`, and `S_string` tokens named with an underscore. Upstream does not use underscore in their token names. Upstream Rust grammar have a `RawString` token and the upstream Python grammar have a `FormatString` token.

We should have a `Boolean` token for the keywords `true` and `false`. Upstream expresses this as `boolean { @specialize[@name=Boolean]<identifier, "true" | "false"> }`.

We should handle the `null` keyword. Upstream expresses this as `kw<"null">`.

Right now `true`, `false` and `null` are outputted as `Ident` in the parse tree.

The statement `from invoice` is outputted as `Ident` and `Ident` in the parse tree. In PRQL you can do `select invoice.date` or `select invoice.*`, that asterisk makes it something of an "expression", maybe an `IdentifierExpression`? The upstream Python and JavaScript grammars have a `MemberExpression` token to represent things like `VariableName (("." | "?.") PropertyName`.

We need triple-quoted f-strings, r-strings and s-strings.

We need an `Escape` token for escape sequences (such as `\n`, etc).

max-sixty:
> We have the `F_string`, `R_string`, and `S_string` tokens named with an underscore. Upstream does not use underscore in their token names.

Fine to rename `FString` etc!

> We should have a `Boolean` token for the keywords `true` and `false`. Upstream expresses this as `boolean { @specialize[@name=Boolean]<identifier, "true" | "false"> }`.
> 
> We should handle the `null` keyword. Upstream expresses this as `kw<"null">`.
> 
> Right now `true`, `false` and `null` are outputted as `Ident` in the parse tree.

Yes, having these as keywords would be good. We don't necessarily need to distinguish them in more detail than "keyword", but it's OK to do so if you like.

> `select invoice.*`, that asterisk makes it something of an "expression", maybe an `IdentifierExpression`?

No strong view — I think it could be highlighted the same as an `Ident`

> We need triple-quoted f-strings, r-strings and s-strings.

Great! If these can be generic, even better... I had mixed success when I originally tried to do that.

> We need an `Escape` token for escape sequences (such as `\n`, etc).

Nice, yes.

---

Also feel free to take a swing at the change suggested at the bottom of https://github.com/PRQL/prql/pull/3571 — it would touch a lot of places, but I don't think it's that hard. It would be a nice next-PR to the rust parser.

PR #3594, opened 2023-10-02
revert: #3591

max-sixty:
Reverts PRQL/prql#3591

Actually possibly duckdb wasn't to blame — maybe a confluence of insta _and_ duckdb, given both passed pre-merge? Seems unlikely. Let's confirm whether reverting causes tests to pass...


max-sixty:
I think this may have been a non-deterministic test re duckdb, closing this for the moment.

PR #3597, opened 2023-10-03
feat: Add chroma grammar

vanillajonathan:
[chroma](https://github.com/alecthomas/chroma) is a syntax highlighting library written in Go and used by the static website generator [Hugo](https://gohugo.io/) and other software.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,8 +34,8 @@
 - [Pygments](https://pygments.org/), the generic syntax highlighting library now
   has syntax highlighting for PRQL. (@vanillajonathan, #3564)
 - [chroma](https://github.com/alecthomas/chroma), a syntax highlighting library
-  written in Go and used by the static website generator [Hugo](https://gohugo.io/).
-  (@vanillajonathan, #3597)
+  written in Go and used by the static website generator
+  [Hugo](https://gohugo.io/). (@vanillajonathan, #3597)
 
 **Internal changes**:
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,8 +31,11 @@
 **Integrations**:
 
 - Bump `prqlc`'s MSRV to 1.70.0 (@eitsupi, #3521)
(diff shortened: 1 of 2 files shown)


max-sixty:
Cool!

Do you know whether we can / should use this in place of `highlight.js` for the website? That is a Huge site...

vanillajonathan:
Yeah, I think it would be simple.

Hugo have a page on [syntax highlighting](https://gohugo.io/content-management/syntax-highlighting/) in the documentation.

We would need to wrap our code blocks like this:
```
{{< highlight prql }}
from customers
take 10
{{< / highlight >}}
```

However if we do some changes to the syntax of PRQL we would have to wait for it to get picked by upstream chroma and then Hugo before we could use the new stuff on our website. With our in-tree `highlight.js` we can patch and deploy it immediately.

max-sixty:
> However if we do some changes to the syntax of PRQL we would have to wait for it to get picked by upstream chroma and then Hugo before we could use the new stuff on our website. With our in-tree `highlight.js` we can patch and deploy it immediately.

Yes, that's a very good point...

I guess it wouldn't be possible to build from our own chroma fork?

I take your direction then — your work on these newer highlighters has been great, and it would be good to use them, rather than the older `highlight.js` styles. But i agree that it would be a disadvantage if it takes months to ship a change, and we can't show new syntax on the website as a result.

(I also don't know how much better `chroma` is than `highlight.js`. The Lezer one is really good now, but I haven't looked in as much detail on `chroma`...)

vanillajonathan:
I guess it would be possible to build our own Hugo from our own chroma fork, however it would require some effort, and I don't think it is worth it. I think `highlight.js` serves us well and that we should continue to use it.

I don't think `chroma` is better (or worse) than `highlight.js`, they're just different. We could use either but `highlight.js` allows us to move faster with changes.

max-sixty:
OK, that makes sense. Thanks @vanillajonathan !

PR #3599, opened 2023-10-04
feat!: `std.sql.read_csv` is compiled to `read_csv` in SQL by default

eitsupi:
Currently the PRQL's `read_csv` function compiles to `read_csv_auto`, which exists only in DuckDB, but Polars SQL has `read_csv`, and `read_csv` may be also added in DataFusion in the future (apache/arrow-datafusion#4850).

I feel it would be more appropriate for PRQL's `read_csv` to compile to `read_csv` by default and only in the DuckDB dialect to compile to `read_csv_auto`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,8 @@
 
 **Features**:
 
+- _Breaking_: The `std.sql.read_csv` function is now compiled to `read_csv`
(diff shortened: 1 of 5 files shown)


max-sixty:
Very nice, thanks!

PR #3601, opened 2023-10-04
chore: Add SCC to CHANGELOG.md

vanillajonathan:
[scc](https://github.com/boyter/scc), a source lines of code (SLOC) counter now has support for counting lines of code in `.prql` files since PR https://github.com/boyter/scc/pull/411.

Maybe we should have some place in the documentation to cover this?


max-sixty:
Cheers!

PR #3603, opened 2023-10-04
docs(book): add a section about division operators

eitsupi:
Previous discussion: #2426


web/book/src/reference/syntax/operators.md
```diff
@@ -34,6 +34,25 @@ operations and for function calls (see the discussion below.)
 |             or | <code>\|\|</code>           |     9      | left-to-right |
 |  function call |                             |     10     |               |
 
+## Division and integer division
+
+The `/` operator performs division that always returns a floot value, while the
```
max-sixty:
```suggestion
The `/` operator performs division that always returns a float value, while the
```

max-sixty (approved):
Super, thanks

PR #3608, opened 2023-10-05
web: Add CodeMirror demo

vanillajonathan:
Unfortunately we don't have any npm packages yet so instead of just using npm to install the packages you have build the `prql-lezer` package and copy the files.

CodeMirror doesn't use Lezer packages directly, instead they have language extensions that use the Lezer packages.

So upstream have a `lang-javascript`package ([npm](https://www.npmjs.com/package/@codemirror/lang-javascript)) that depends on the `lezer-javascript` package ([npm](https://www.npmjs.com/package/@lezer/javascript)).

Inside the `src` directory there is a directory called `lang-prql` which could be a npm package which could depend on the `prql-lezer` which in turn could also be a package.


web/prql-codemirror-demo/README.md
```diff
@@ -0,0 +1,18 @@
...
+
+## Instructions
+
+Build `prql-lezer` then copy the files from `/grammars/prql-lezer/dist/` to
+`src/lang-prql/prql-lezer/`.
+
+```
+mkdir src/lang-prql/prql-lezer
+cd grammars/prql-lezer/
+npm run build
```
not-my-profile, line 14: So this `npm run build` doesn't work yet because there doesn't yet exist a `package.json` in `grammars/prql-lezer/`?

I think it would be best to merge this demo when the instructions work ... would you mind turning `prql-lezer` into a package? (ideally in a separate PR) Note that it shouldn't be necessary to publish it on NPM just for the demo since I think the dependency could be specified with a [local path](https://docs.npmjs.com/cli/v9/configuring-npm/package-json#local-paths).

vanillajonathan: That `npm run buld` does work, because there does exist a `package.json` in `grammars/prql-lezer`.
https://github.com/PRQL/prql/blob/main/grammars/prql-lezer/package.json

I suppose I could fix the path in the instructions.

max-sixty: Let's add the instructions to a job in `bulid-web.yaml` (either copy-paste or put in `Taskfile.yml`), and then they'll run in CI.

Alternatively we can publish them outside the repo — I don't want to slow down anyone from offering examples — fine to put them in a gist.

If we put them in the repo, though, then there's a higher bar for ensuring that the description is correct, and the only way I know to do that is to automate the tests...

web/prql-codemirror-demo/src/codemirror.ts
```diff
@@ -0,0 +1,113 @@
+import {
```
max-sixty, line 2: Is this file copy-pasted from somewhere? We should link back to it if so...

web/prql-codemirror-demo/.gitignore
max-sixty, line 1: FWIW I think these should basically all be in the root gitignore; we shouldn't need this.. (but no big stress)

max-sixty (commented):
Looks cool @vanillajonathan !

web/prql-codemirror-demo/src/codemirror.ts
vanillajonathan: Yes, it is. This file is from the `basic-setup` repository.
https://github.com/codemirror/basic-setup/blob/main/src/codemirror.ts

We could also use the [`codemirror`](https://www.npmjs.com/package/codemirror) npm package instead, but this lets users pick and chose which extensions to load by modifying their own copy of the `codemirror.ts` file.

web/prql-codemirror-demo/README.md
vanillajonathan: Do you want it in `build-web.yaml` or in the `Taskfile.yml`.

Maybe we should publish `prql-lezer` and `lang-prql` as two packages on npm then have the demo depend on `lang-prql` which in turn would depend on `prql-lezer`?

web/prql-codemirror-demo/src/codemirror.ts
max-sixty: Let's link back to it, then

max-sixty: Great, thank you!

web/prql-codemirror-demo/README.md
max-sixty: It should run in `build-web.yaml`, and it can be nice to encapsulate it in `Taskfile.yml` and then call it from `build-web.yaml`, but it's a bit more work, so not strictly necessary.

> Maybe we should publish `prql-lezer` and `lang-prql` as two packages on npm then have the demo depend on `lang-prql` which in turn would depend on `prql-lezer`?

I agree with publishing the grammar as a package! What's the advantage of having the two packages?

vanillajonathan:
This is the Mega-Linter TypeScript thing again...

max-sixty:
> This is the Mega-Linter TypeScript thing again...

"Fixed"

max-sixty:
> "Fixed"

Sorry, I pushed this to the wrong branch! Let's see if this works now.

max-sixty:
This looks good now; it just needs some very light test to resolve the discussion at https://github.com/PRQL/prql/pull/3608#pullrequestreview-1661431645

web/prql-codemirror-demo/README.md
vanillajonathan: > I agree with publishing the grammar as a package! What's the advantage of having the two packages?

I am not sure what the rationale is behind it, but it is how it is done upstream (see [lang-javascript](https://github.com/codemirror/lang-javascript/blob/main/package.json#L35)). Perhaps the intention is for Lezer to be a general-purpose grammar definition which can be used by other software than CodeMirror and to make it possible for such software to depend on the package containing just the Lezer grammar without the CodeMirror-specific code, but I don't know.

vanillajonathan:
Okay, I added a job the `build-web` file that builds the demo.

web/prql-codemirror-demo/README.md
max-sixty: OK, I would probably side with doing a single package at first — it's less work, and it's easy to split it up later than it is to combine it, if we change our mind.

.github/workflows/build-web.yaml
```diff
@@ -70,3 +70,34 @@ jobs:
...
+        working-directory: web/prql-codemirror-demo/
+        run: |
+          mkdir src/lang-prql/prql-lezer
+          cp ../../grammars/prql-lezer/dist/* src/lang-prql/prql-lezer/
+      - name: Build CodeMirror demo
+        working-directory: web/prql-codemirror-demo/
+        run: bun run build
+      - uses: actions/upload-pages-artifact@v2.0.0
+        with:
+          path: web/prql-codemirror-demo/dist/
```
max-sixty, line 103: Nice work!

max-sixty:
Thanks @vanillajonathan !

Other things we could do if you want to continue your great run of contributions:
- Finish off the last bit of the `vite` change — would be great, was very close — maybe just fixing the taskfile
- Make a package for the `prql-lexer` grammar, so we could use it in a Jupyter extension
- I was impressed you'd done some rust PRs! You'd be welcome to do some more if you're interested to learn more rust...

not-my-profile:
Out of curiosity: is this new demo online somewhere at https://prql-lang.org/ so it can be tried out?

max-sixty:
> Out of curiosity: is this new demo online somewhere at https://prql-lang.org/ so it can be tried out?

No, only locally

vanillajonathan:
The pipeline uses `upload-pages-artifact` though.

max-sixty:
> The pipeline uses `upload-pages-artifact` though.

That's just for the website though, right? At https://github.com/PRQL/prql/blob/f4178e3c5e18c5b1e86a501eeb0e64c851f586e7/.github/workflows/build-web.yaml#L70-L72

So the codemirror demo doesn't get uploaded, unless I'm missing something.

(To the extent that moving the playground to codemirror is an option, that would be the ideal demo!)

vanillajonathan:
If you look below, at the end of the other job at the end of the file.

https://github.com/PRQL/prql/blob/f4178e3c5e18c5b1e86a501eeb0e64c851f586e7/.github/workflows/build-web.yaml#L101-L103

max-sixty:
Sorry, I somehow missed that.

How does GH know which artifact to use? I'm a bit worried that when we deploy this, the codemirror demo is going to overwrite the artifact from above (our website)...

vanillajonathan:
I don't know. The worry sounds reasonable. Maybe I should remove that?

max-sixty:
Yeah I think let's remove for the moment, if you're ok with that. Unless we can understand how it works and confirm it's working as intended — otherwise there's some small chance whole website might go down next time we release...

Thank you @vanillajonathan

PR #3613, opened 2023-10-06
web: Update mdbook-admonish assets to v3

max-sixty:
Would be great to have this managed along with other cargo dependencies, so we don't have breaking changes like this when CI pull a more recent versions. But not sure whether that's possible; mdbook runs outside cargo. Maybe some configuration of build dependencies?!

### DIFF
--- a/web/book/book.toml
+++ b/web/book/book.toml
@@ -5,7 +5,7 @@ multilingual = false
 title = "PRQL language book"
 
 [output.html]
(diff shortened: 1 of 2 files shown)


max-sixty:
This looks like one approach: https://github.com/o1-labs/proof-systems/pull/1006 — could possibly be modified to not use `make`

PR #3617, opened 2023-10-07
ci: Automerge toolchain updates

max-sixty:


eitsupi:
I don't think #3621 is auto-merged because GitHub Actions doesn't trigger another Action.

max-sixty:
Thanks! Adjusted in #3624...

PR #3625, opened 2023-10-08
build: git ignore `target/`

eitsupi:
Revert the change made by #1600

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,7 +1,7 @@
 .DS_Store
 
 dist
-target*/
+target/
 *.out
 *.log
 


max-sixty:
The existing state was useful for me because I use `target_ra` for rust analyzer's build, so it doesn't conflict with the other version. Is there a path that starts with `target` that we want to include?

max-sixty (commented):
The existing state was useful for me because I use target_ra for rust analyzer's build, so it doesn't conflict with the other version. Is there a path that starts with target that we want to include?

eitsupi:
> The existing state was useful for me because I use target_ra for rust analyzer's build, so it doesn't conflict with the other version. Is there a path that starts with target that we want to include?

I noticed that the target directory was not grayed out on VS Code's explorer.
There are no directories I want to include. Since #1600 said "This will be fixed in an upcoming version of rustup though.", I assumed that this is no longer needed, but if you need it, please close this PR.

max-sixty:
> Since #1600 said "This will be fixed in an upcoming version of rustup though.",

Ah OK, let's try this, then!

Issue #3619, opened 2023-10-07
window function could use local sort


jangorecki:
### What's up?

Having the following data `d`
```
  id value
1  1     2
2  2     3
3  3     4
4  6     2
5  5     3
6  4     4
```
How can I generate following SQL
```sql
SELECT
  *,
  AVG(value) OVER (
    ORDER BY
      id ROWS BETWEEN 2 PRECEDING AND CURRENT ROW
  ) AS sma3
FROM
  d
```

According to documentation it seems that I have to sort whole dataset rather than an input to windowing function, therefore query that is being generated by prql compiler includes an extra unwanted `ORDER BY id` at the end.
```r
library(prqlr)
d = data.frame(id=c(1,2,3,6,5,4), value=c(2,3,4,2,3,4))
sql = prql_compile("from d | sort id | window rolling:3 (derive {sma3 = average value})")
cat(sql)
#SELECT
#  *,
#  AVG(value) OVER (
#    ORDER BY
#      id ROWS BETWEEN 2 PRECEDING AND CURRENT ROW
#  ) AS sma3
#FROM
#  d
#ORDER BY
#  id
```

How can I define sorting locally for each window function call locally? so I don't have to clutter my SQL query with unwanted clauses.

snth:
Hi @jangorecki,

Great to see that you are trying out PRQL and posting questions.

You can put the `sort` inside the `window` pipeline.

So for your example:
```r
library(prqlr)
d = data.frame(id=c(1,2,3,6,5,4), value=c(2,3,4,2,3,4))
sql = prql_compile("from d | window rolling:3 (sort id | derive {sma3 = average value})")
cat(sql)
```

or something that you can execude in the playground:
```elm
[{id=1, value=2},
{id=2, value=3},
{id=3, value=4},
{id=6, value=2},
{id=5, value=3},
{id=4, value=4},]
window rolling:3 (
  sort id
  derive {sma3 = average value}
  )
```
which produces
```sql
WITH table_0 AS (
  ...
)
SELECT
  id,
  value,
  AVG(value) OVER (
    ORDER BY
      id ROWS BETWEEN 2 PRECEDING AND CURRENT ROW
  ) AS sma3
FROM
  table_0
ORDER BY
  id

-- Generated by PRQL compiler version:0.9.5 (https://prql-lang.org)
```

jangorecki:
Thanks for reply, that make sense, could be mentioned in documentation.
[...]

PR #3634, opened 2023-10-08
docs(book): documentation for window rows vs range

jangorecki:
- resolves TODO: example of rows vs range
- fix typo: row vs rows
- improves documentation as requested in #3619: provide example of `sort` inside `window`

Possibly it is overly verbose, but it is better like that, so there is no space for misinterpretation.

### DIFF
--- a/web/book/src/reference/stdlib/transforms/window.md
+++ b/web/book/src/reference/stdlib/transforms/window.md
@@ -94,17 +94,21 @@ window range:-2..0 (
 | 7       | 20    | 15       | 20        |
 | 8       | 22    | 17       | 21        |
 
-We can see that rows having `time_id` of 5 and 6 are missing in example data; we can say there are gaps in our data time series data.
+We can see that rows having `time_id` of 5 and 6 are missing in example data; we
+can say there are gaps in our data time series data.
 
 When computing SMA3 for the fifth row (`time_id==7`) then:
+
 - "rows" will compute average on 3 rows (`time_id` in `3, 4, 7`)
-- "range" will compute average on single row only (`time_id==7`)
-Then when computing SMA3 for the sixth row (`time_id==8`) then:
+- "range" will compute average on single row only (`time_id==7`) Then when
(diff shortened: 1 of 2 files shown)


eitsupi:
Thanks for the contribution!
In my opinion, this new code block is evaluable, so I removed the `no-eval` tag.
And I ran tests and added the snapshot.

jangorecki:
Yes code block is evaluable. I decided to not eval it because generated SQL doesn't really gives any more clarity over rows-vs-range, and due the to amount of lines it could possibly add more confusion, but I am fine either way.

eitsupi:
> Yes code block is evaluable. I decided to not eval it because generated SQL doesn't really gives any more clarity over rows-vs-range, and due the to amount of lines it could possibly add more confusion, but I am fine either way.

Yes, I understand the intent.
Perhaps the ideal here would be to be able to generate tables from prql code blocks using sqlite (or duckdb, I think the easiest way is to paste the Markdown output of duckdb CLI), but I think that would be a lot of work to try to implement in Rust.
(Of course, with knitr and prqlr this can be accomplished quickly, but that's the domain of <https://github.com/eitsupi/querying-with-prql>)

jangorecki:
Ideally would be to have a spoiler button that shows SQL if desired. Anyway, I just pasted output from playground and formatted the same way as other tables are formatted in the book.

max-sixty (approved):
I think this is excellent! Thank you @jangorecki 

FWIW I agree with @eitsupi that having the SQL there is useful. While it's a bit noisy in this case, since most of the documentation is about the behavior rather the translation, it limits how confused someone can get :)

max-sixty:
Re having the results of queries in the book — https://github.com/PRQL/prql/issues/1895

snth:
This is great! Thank you @jangorecki . `ROWS` vs `RANGE` is one of the most confusing things about window functions and SQL in general. When I try to recall my experience learning of learning this in SQL and the RDBMs docs I read at the time, they didn't make it that much clearer whereas this is an excellent example that explains it clearly! 🙏

PR #3635, opened 2023-10-08
docs(readme): remove dbt-prql and split prqlc from prql-compiler

eitsupi:
- Exclude dbt-prql since it is no longer updated.
- prqlc was separated from prql-compiler a few months ago.


max-sixty (approved):
Great!

PR #3636, opened 2023-10-08
feat(Lezer grammar): Add new tokens

vanillajonathan:
This adds a new token called `VariableName` that we expose in variable declarations using the `let` keyword. We use this token for highlighting the variable name in CodeMirror.

- `F_string` have been renamed to ~`FormatString`~ `FString` to match the upstream Python grammar.
- `R_string` have been renamed to ~`RawString`~ `RString` to match the upstream Rust grammar.
- `S_string` have been renamed to ~`ServerString`~ `SString` to match upstream convention.
- `Ident` have been renamed to `Identifier` to match SQL grammar.
- The `ArithOp`, `CompareOp` and `LogicOp` tokens are now highlighted.
- More tests.


max-sixty:
> This adds a new token called `VariableName` that we expose in variable declarations using the `let` keyword. We use this token for highlighting the variable name in CodeMirror.

This sounds good!

---

> * `F_string` have been renamed to `FormatString` to match the upstream Python grammar.
> 
> * `R_string` have been renamed to `RawString` to match the upstream Rust grammar.
> 
> * `S_string` have been renamed to `ServerString` to match upstream convention.

But I think "format string" is actually something slightly different, namely in python `"{}.{}".format(name, suffix)`, rather than with a leading `f`. Is that not your understanding? https://docs.python.org/3/library/string.html#formatstrings

---

For s-string, if we really had to put a word there, it would be "SQL String". But adding another name to something means we need to learn two names. And "s-string" is quite clear in form. 

OTOH, "SQL String" does describe its function quite clearly, so we could change its name in PRQL. CC @PRQL/prql-team 

---

Overall, I think we should match names of things in the grammar with PRQL itself, and only change the name to match other projects if we change it everywhere.

max-sixty (commented):
> This adds a new token called `VariableName` that we expose in variable declarations using the `let` keyword. We use this token for highlighting the variable name in CodeMirror.

This sounds good!

---

> * `F_string` have been renamed to `FormatString` to match the upstream Python grammar.
> 
> * `R_string` have been renamed to `RawString` to match the upstream Rust grammar.
> 
> * `S_string` have been renamed to `ServerString` to match upstream convention.

But I think "format string" is actually something slightly different, namely in python `"{}.{}".format(name, suffix)`, rather than with a leading `f`. Is that not your understanding? https://docs.python.org/3/library/string.html#formatstrings

---

For s-string, if we really had to put a word there, it would be "SQL String". But adding another name to something means we need to learn two names. And "s-string" is quite clear in form. 

OTOH, "SQL String" does describe its function quite clearly, so we could change its name in PRQL. CC @PRQL/prql-team 

---

Overall, I think we should match names of things in the grammar with PRQL itself, and only change the name to match other projects if we change it everywhere.

vanillajonathan:
> But I think "format string" is actually something slightly different, namely in python `"{}.{}".format(name, suffix)`, rather than with a leading `f`. Is that not your understanding? https://docs.python.org/3/library/string.html#formatstrings

I wanted to get away from using underscores in token names in order to match conventions of upstream grammars which use camel case and pascal case. `FormatString` is name of the token used by the upstream Python grammar for `f"test"` (try on https://lezer-playground.vercel.app/)

> For s-string, if we really had to put a word there, it would be "SQL String". But adding another name to something means we need to learn two names. And "s-string" is quite clear in form.
> 
> OTOH, "SQL String" does describe its function quite clearly, so we could change its name in PRQL. CC @PRQL/prql-team
> 
> Overall, I think we should match names of things in the grammar with PRQL itself, and only change the name to match other projects if we change it everywhere.

I just assumed that the S in "s-string" was for server, but I guess I my assumption was wrong. Still, I feel that "ServerString" is a better name than "SQL String" because I see SQL as just one of potentially many query languages that PRQL could transpile to.

max-sixty:
> `FormatString` is name of the token used by the upstream Python grammar for `f"test"` (try on https://lezer-playground.vercel.app/)

I'm guessing that an f-string is formatted the same as a format string in the python grammar. But our f-string is not actually a format string, and so we shouldn't name it the same as that, particularly when it's incompatible with the PRQL language.

> I wanted to get away from using underscores in token names in order to match conventions of upstream grammars which use camel case and pascal case

Unless you feel quite strongly, how about doing `FString` or `fString` for all of these, rather than changing the name?

vanillajonathan:
> I'm guessing that an f-string is formatted the same as a format string in the python grammar. But our f-string is not actually a format string, and so we shouldn't name it the same as that, particularly when it's incompatible with the PRQL language.

No `'{1}'.format(a, b)` is a `CallExpression(MemberExpression(ArgList)))`.

> Unless you feel quite strongly, how about doing `FString` or `fString` for all of these, rather than changing the name?

We cannot change it to `fString` because with a lower case it is not exported, so it must start with an uppercase letter. We can name it `FString` or `Fstring`though. Either would be fine with me.

max-sixty:
> We cannot change it to `fString` because with a lower case it is not exported, so it must start with an uppercase letter. We can name it `FString` or `Fstring`though. Either would be fine with me.

I think `FString` would be great, that would be standard!



> No `'{1}'.format(a, b)` is a `CallExpression(MemberExpression(ArgList)))`.

OK interesting. Though the python docs (link above) do call a "format string" something different from "f string"...

vanillajonathan:
Okay, I have now renamed:
- `FormatString` to `FString`
- `RawString` to `RString`
- `ServerString` to `SString`

PR #3638, opened 2023-10-08
refactor: Unicode with curly braces

vanillajonathan:
This takes off where #3571 and #3569 left off, and changes the syntax for the Unicode escape sequence to use curly braces because maybe you wanted that instead?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -41,6 +41,10 @@
   [Hugo](https://gohugo.io/). (@vanillajonathan, #3597)
 - [scc](https://github.com/boyter/scc), a source lines of code counter now has
   support for `.prql` files.
+- [gcloc](https://github.com/JoaoDanielRufino/gcloc) a source lines of code
+  counter now has support for `.prql` files. (@vanillajonathan)
+- [cloc](https://github.com/AlDanial/cloc) a source lines of code counter now
+  has support for `.prql` files. (@AlDanial)
 
 **Internal changes**:
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,7 +16,7 @@
 - Add `std.prql_version` function to return PRQL version (@hulxv, #3533)
 - Add support for hex escape sequences in strings. Example `"Hello \x51"`.
   (@vanillajonathan, #3568)
-- Add support for long Unicode escape sequences. Example `"Hello \U0001F422"`.
+- Add support for long Unicode escape sequences. Example `"Hello \u{01F422}"`.
   (@vanillajonathan, #3569)
 
 **Fixes**:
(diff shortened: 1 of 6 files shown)


crates/prql-parser/src/lexer.rs
```diff
@@ -347,29 +347,19 @@ fn escaped_character() -> impl Parser<char, char, Error = Cheap<char>> {
         just('n').to('\n'),
         just('r').to('\r'),
         just('t').to('\t'),
-        (just('u').ignore_then(
+        (just("u{").ignore_then(
```
max-sixty, line 350: The canonical way would be to use `.delimited_by`, but this is OK given it seems to work and these can't be nested

CHANGELOG.md
```diff
@@ -40,11 +40,7 @@
   written in Go and used by the static website generator
   [Hugo](https://gohugo.io/). (@vanillajonathan, #3597)
 - [scc](https://github.com/boyter/scc), a source lines of code counter now has
-  support for `.prql` files. (@vanillajonathan)
-- [gcloc](https://github.com/JoaoDanielRufino/gcloc) a source lines of code
-  counter now has support for `.prql` files. (@vanillajonathan)
-- [cloc](https://github.com/AlDanial/cloc) a source lines of code counter now
-  has support for `.prql` files. (@AlDanial)
```
max-sixty: Did we intend to remove these?

max-sixty (approved):
Awesome! Thanks @vanillajonathan !

max-sixty:
For the tests to pass, we need to remove the `no-fmt` in the example at `reference/syntax/strings/quoting-and-escape-characters`, since the formatting now works...

CHANGELOG.md
vanillajonathan: No, we do not intend to remove those. Good spotted, my mistake.

vanillajonathan:
> For the tests to pass, we need to remove the `no-fmt` in the example at `reference/syntax/strings/quoting-and-escape-characters`, since the formatting now works...

Thanks!

crates/prql-parser/src/lexer.rs
vanillajonathan: Oh, I didn't know about that one. We could use that one instead if you want, but then we can't use `at_least` and `at_most`. I noticed that the Rust compiler is friendly enough to actually tell you that you can at most use 6 characters if you try to use more than that.

max-sixty: I think my message was unclear. We can still do the `at_least` / `at_most`. The `.delimited_by` would be for the brackets. Here's an example of it: https://github.com/PRQL/prql/blob/33b71715136380255482e902c8afe92bd5a8fe10/crates/prql-parser/src/expr.rs#L114

CHANGELOG.md
```diff
@@ -40,7 +40,7 @@
   written in Go and used by the static website generator
   [Hugo](https://gohugo.io/). (@vanillajonathan, #3597)
 - [scc](https://github.com/boyter/scc), a source lines of code counter now has
-  support for `.prql` files. (@vanillajonathan)
+  support for `.prql` files.
```
max-sixty, line 43: Did we intend to remove this?

max-sixty:
This is great, so I'll merge, but fine to do any small refinements as a follow up...

max-sixty:
One small follow-up that would be nice is to be explicit that we've changed the syntax from `uXXXX` to `u{XXXXX}` (better expressed, hopefully!)

CHANGELOG.md
vanillajonathan: No. Thanks for spotting.

PR #3652, opened 2023-10-10
feat(Monarch grammar): Add token postfix and data types

vanillajonathan:
From the Monaco website:
> Optional postfix attached to all returned tokens. By default this attaches the language name so in the CSS you can refer to your specific language. For example, for the Java language, we could use `.identifier.java` to target all Java identifiers specifically in CSS.

https://microsoft.github.io/monaco-editor/monarch.html


max-sixty:
Cheers @vanillajonathan !

PR #3654, opened 2023-10-11
feat: Add hexadecimal numerical notation

vanillajonathan:
This lets you use hexadecimal notation to express numbers so you can do:
```elm
from products
filter status_flag == 0xff
```

### DIFF
--- a/crates/prql-parser/src/lexer.rs
+++ b/crates/prql-parser/src/lexer.rs
@@ -157,14 +157,18 @@ pub fn ident_part() -> impl Parser<char, String, Error = Cheap<char>> + Clone {
 }
 
 fn literal() -> impl Parser<char, Literal, Error = Cheap<char>> {
-    let hex_notation = just("0x").ignore_then(
-        filter(|c: &char| c.is_ascii_hexdigit())
-            .repeated()
-            .at_least(1)
-            .at_most(12)
-            .collect::<String>()
-            .try_map(|digits, _| Ok(Literal::Integer(i64::from_str_radix(&digits, 16).unwrap()))),
-    ).labelled("number");
+    let hex_notation = just("0x")
+        .ignore_then(
+            filter(|c: &char| c.is_ascii_hexdigit())
+                .repeated()
+                .at_least(1)
+                .at_most(12)
+                .collect::<String>()
+                .try_map(|digits, _| {
+                    Ok(Literal::Integer(i64::from_str_radix(&digits, 16).unwrap()))
+                }),
+        )
+        .labelled("number");
 
     let exp = one_of("eE").chain(one_of("+-").or_not().chain::<char, _, _>(text::digits(10)));
 
--- a/crates/prql-parser/src/lexer.rs
+++ b/crates/prql-parser/src/lexer.rs
@@ -157,6 +157,15 @@ pub fn ident_part() -> impl Parser<char, String, Error = Cheap<char>> + Clone {
 }
 
 fn literal() -> impl Parser<char, Literal, Error = Cheap<char>> {
+    let hex_notation = just("0x").ignore_then(
+        filter(|c: &char| c.is_ascii_hexdigit())
+            .repeated()
+            .at_least(1)
+            .at_most(12)
+            .collect::<String>()
+            .try_map(|digits, _| Ok(Literal::Integer(i64::from_str_radix(&digits, 16).unwrap()))),
+    ).labelled("number");
+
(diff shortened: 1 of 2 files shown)


eitsupi (changes requested):
Thanks, could you add tests for this?

vanillajonathan:
> Thanks, could you add tests for this?

Done!

eitsupi (approved):
Thanks for the update!

max-sixty (approved):
Looks good! Thanks @vanillajonathan !

Would be great to have this as a line in an example on the numbers page in the book, fine to do as a follow-up

crates/prql-parser/src/lexer.rs
```diff
@@ -157,6 +157,19 @@ pub fn ident_part() -> impl Parser<char, String, Error = Cheap<char>> + Clone {
...
+            filter(|c: &char| c.is_ascii_hexdigit())
+                .repeated()
+                .at_least(1)
+                .at_most(12)
+                .collect::<String>()
+                .try_map(|digits, _| {
+                    Ok(Literal::Integer(i64::from_str_radix(&digits, 16).unwrap()))
+                }),
+        )
+        .labelled("number");
```
max-sixty, line 171: Good rust! Am impressed!

aljazerzen, line 171: Ideally, this should be the part of the `number` parser below.

As of now, it won't parse these:

```
+0x001
-0x001
0x0000_0001
```

vanillajonathan: Yeah, I would like it to be able to parse those too. I don't know how to put it in the `number` parser,. but I've been thinking maybe it should emit a `Literal::HexadecimalInteger(i64)` instead, that way AST would have richer semantic knowledge which would be useful if one were to transpile to a target that support such numerical notations.

https://dev.mysql.com/doc/refman/8.0/en/hexadecimal-literals.html
https://mariadb.com/kb/en/hexadecimal-literals/

aljazerzen: > Yeah, I would like it to be able to parse those too

The chumsky parser declarations can be a bit tricky. I think it should be possible to add a `chain<char, _, _>(just("0x").or_not().flatten())` before integer parsing.

... but that would also allow things like this: `0x00.01`, which we probably don't want to allow.

> I've been thinking maybe it should emit a `Literal::HexadecimalInteger(i64)`

The meaning of `HexadecimalInteger` and `Integer` would be equal, right? It's just a different representation. Let's try to keep the AST minimal and representation-independent.

vanillajonathan: Currently `50` becomes `Literal::Integer(50)` and `0xFF` becomes `Literal::Integer(255)` so in the SQL output they show up as `50` and `255` respectively so currently there is no way for the target to know the original representation hence the target cannot adjust the representation.

If target had information about the original representation, such as a  `Literal::HexadecimalInteger` token, it could adjust the SQL output to `0xFF` instead of `255`.

An alternative to creating a new `Literal::HexadecimalInteger` token would be to extend the existing `Literal::Integer` token with a enum, i.e. `Literal::Integer(i64, NotationKind)` so it would be constructed as `Literal::Integer(255, NotationKind::Hex)`.

aljazerzen: What I'm saying, is that `50` (parsed into `Literal::Integer(50)`) and `0x32` (also parsed into `Literal::Integer(50)`) are semantically equivalent. There is no difference in their meaning if compiled to either `50` or `0x32`.

I see that I might make sense to retain information about the original representation, so the compiler does not change `0x32` into `50` and the compilation results are easier to follow. In most cases, I would agree that we should make the compilation results more readable, but in this case, I think that less complexity in the compiler out weights the readability of the output.

Complexity in the compiler would come from all the places where we unpack integers (for example when doing `take 4`).

(Long term, compilation should be transparent, meaning that people don't even look at generated SQL, but just execute it)

max-sixty: FWIW I tend to agree with @aljazerzen that the current implementation is correct to parse into a number quite early — I don't think it's that helpful to retain much information about whether it was `0x32` or `50`. The goal of parsing is to compress representations into their semantic form, and in that case the semantics are identical. We can retain a link to the source with a `span` when we want a link to the source.

> Ideally, this should be the part of the `number` parser below.

Yes, agree, this would be a good follow-up @vanillajonathan . I'm not so sure about allowing underscores, but we should be able to have these two:

```
+0x001
-0x001
```

(these would make good tests)

PR #3659, opened 2023-10-11
docs: Add table of escape sequences

vanillajonathan:
I hope this is the right place, else feel free to move it somewhere else.

### DIFF
--- a/web/book/src/reference/syntax/strings.md
+++ b/web/book/src/reference/syntax/strings.md
@@ -63,7 +63,7 @@ implementation.
 
 ## Escape sequences
 
-Unless an `'r'` prefix is present, escape sequences in string literals are
+Unless an `r` prefix is present, escape sequences in string literals are
 interpreted according to rules similar to those used by Standard C. The
 recognized escape sequences are:
 
--- a/web/book/src/reference/syntax/strings.md
+++ b/web/book/src/reference/syntax/strings.md
@@ -60,3 +60,21 @@ and PRQL doesn't currently account for
 these differences. Please open issues with any difficulties in the current
 implementation.
 ```
+
+## Escape sequences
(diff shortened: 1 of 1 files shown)


web/book/src/reference/syntax/strings.md
```diff
@@ -60,3 +60,23 @@ and PRQL doesn't currently account for
 these differences. Please open issues with any difficulties in the current
 implementation.
 ```
+
+## Escape sequences
+
+Unless an `'r'` prefix is present, escape sequences in string literals are
```
max-sixty:
```suggestion
Unless an `r` prefix is present, escape sequences in string literals are
```

I think clearer without the quotes, given the `r` doesn't have quotes...

web/book/src/reference/syntax/strings.md
```diff
@@ -60,3 +60,23 @@ and PRQL doesn't currently account for
...
+interpreted according to rules similar to those used by Standard C. The
+recognized escape sequences are:
+
+| Escape Sequence    | Meaning                       |
+| ------------------ | ----------------------------- |
+| `\`&lt;newline&gt; | Backslash and newline ignored |
```
max-sixty: I don't completely get this one — should it be `\\n`?

max-sixty (approved):
Nice! Thanks. I added one question and suggestion.

---

One way of showing these very clearly is having them in an example, like:

```elm
from foo
derive {
  single_quote = '\'',
  double_quote = '\"',
  ... etc
```

...because then we also get the test for free; there's no way of making a mistake.

(but this is already great, and potentially clearer for beginners)

web/book/src/reference/syntax/strings.md
vanillajonathan: Oh, it was meant for string continuation on the next line, but PRQL doesn't have this since strings can span multiple lines anyway without a continuation token, so I removed it.

PR #3661, opened 2023-10-11
feat: Add binary numerical notation

vanillajonathan:
This lets you use binary notation to express numbers so you can do:
```elm
from foo
filter bar == 0b1111000011110000
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,8 @@
   (@vanillajonathan, #3568)
 - Add support for long Unicode escape sequences. Example `"Hello \u{01F422}"`.
   (@vanillajonathan, #3569)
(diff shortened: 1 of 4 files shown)


max-sixty:
Nice!

max-sixty:
(would be great to have these as a line in the docs too!)

vanillajonathan:
> (would be great to have these as a line in the docs too!)

Thanks for the reminder. I added it to the docs now.

max-sixty:
Thanks! Will need a `cargo insta test --accept` to update the docs snapshot (also `task test-rust` should work)

vanillajonathan:
Yeah, I just noticed. Fixed.

max-sixty:
Looking really good, thanks @vanillajonathan

PR #3665, opened 2023-10-12
docs: add docs about `std.prql_version` and more test

eitsupi:
I think the most effective use of the `prql_vesion` function is in an environment where PRQL results are returned as tables, such as ClickHouse.
In such a case, the fastest way to call this function is with a query like `[{version = prql_version}]`, so I want to add this to the book and tests.

### DIFF
--- a/crates/prql-compiler/src/tests/test.rs
+++ b/crates/prql-compiler/src/tests/test.rs
@@ -3417,6 +3417,21 @@ fn prql_version() {
 }
 
 #[test]
+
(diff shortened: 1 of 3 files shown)


max-sixty (approved):
Great docs!

One small downside of these is that the tests will fail when we change the version, and require re-snapshotting. Though that's less of a downside now we bump the version after releasing, since it happens off the critical path.

(We could add a phrase in the release instructions to run `task test-rust` after bumping the version)

PR #3609, opened 2023-10-06
web: Add Vite to playground

vanillajonathan:
@richb-hanover in #3459 wanted Vite on the playground. He can have it!

PR #3666, opened 2023-10-12
revert: #3609

max-sixty:
Sorry to merge and then revert, but I don't think this is quite ready yet. My fault — I should have checked more thoroughly before merging.

@vanillajonathan immediate things that don't seem to work:
- `npm run preview` from the playground path opens a browser that just states `Cannot GET /`
- `task run-playground` calls into `npm start`, which doesn't exist now

---

From a git perspective, the way to fix this is to run `git revert {this commit}`, where `{this commit}` is the hash of the commit this PR creates. Then you'll have the code as though #3609 is still open. Then if we can fix those things, open a new PR with that code, then we'll be good. Let me know anything I can do to help.


vanillajonathan:
It doesn't work:

    $ git revert 170b80f9b98b715ba6e108e6b6584e71662da7b7
    fatal: bad object 170b80f9b98b715ba6e108e6b6584e71662da7b7
    $ git revert 170b80f
    fatal: bad revision '170b80f'

The `npm start` needs to be replaced with `npm run dev`. 

When I run `npm run preview` on GitHub Codespaces it asks me and I click on the Open in browser button and it runs. I don't get "Cannot GET /".

max-sixty:
It would be `git revert 37db5f2` — that's the hash of the commit on `main`

max-sixty:
> When I run `npm run preview` on GitHub Codespaces it asks me and I click on the Open in browser button and it runs. I don't get "Cannot GET /".

Ah OK cool — possibly there's something up with my build. 

Could you confirm this is still the behavior, and then I'll help debug if we still see a difference?

max-sixty:
After re-reverting this _and_ merging #3667, `npm run dev` works great!!

`npm run preview` still doesn't work for me. What's the difference between `dev` and `preview`? Do we need the `preview` command?

vanillajonathan:
`npm run dev` runs Vite in development mode where it compiles any changes on fly and immediately reloads them. Very convenient for development, while `npm run preview` serves it up for preview from the compiled `dist` directory. It seems you need to run `npm run build` before you can run `npm run preview`.

max-sixty:
> `npm run dev` runs Vite in development mode where it compiles any changes on fly and immediately reloads them. Very convenient for development, while `npm run preview` serves it up for preview from the compiled `dist` directory. It seems you need to run `npm run build` before you can run `npm run preview`.

Thanks for the explanation, great. 

So let's just wire up `task run-playground` to `npm run dev`, update the Readme, and then I think we should be good. 

(and then the monarch break we can handle separately)

vanillajonathan:
Could you fix this or help me out with this?
I am having some problems, when I run `git revert 37db5f2` I get only the `.mega-linter.yaml` file but not the rest of the files.

max-sixty:
No stress — git can be a pain. In this case, there a conflict with that file.

Try this:

```
git revert 37db5f2 -Xours
```

That will resolve the merge conflict in favor of "our" branch automatically

vanillajonathan:
Yeah, that resolved the merge conflict but then I have no files to commit:

    nothing to commit, working tree clean

max-sixty:
Once that command has run, the branch has the previous `vite` changes. So from there we need to fix the Taskfile, push, and open a PR. Does that make sense?

(This is advanced git, which I know is confusing...)

vanillajonathan:
Yeah, that makes sense, but when I try to commit, the only thing it wants to commit is the Taskfile. Is that how it should be?

max-sixty:
The existing code is already committed by the `revert` command; so yes, that's good!

PR #3667, opened 2023-10-12
revert: #3652

max-sixty:
Part of the reason the Vite playground wasn't working is actually that the Monarch grammar has a small error: `Uncaught Error: prql: the @ match target 'constants' is not defined, in rule: tokenizer.root: [a-z_$][\w$]*`

So reverting, and we can re-revert and merge when fixed

Hope that's OK @vanillajonathan

### DIFF
--- a/web/playground/src/workbench/prql-syntax.js
+++ b/web/playground/src/workbench/prql-syntax.js
@@ -20,22 +20,8 @@ const LITERALS = ["null", "true", "false"];
 const def = {
   // Set defaultToken to invalid to see what you do not tokenize yet
   // defaultToken: 'invalid',
-  tokenPostfix: ".prql",
 
   keywords: [...TRANSFORMS, ...BUILTIN_FUNCTIONS, ...KEYWORDS, ...LITERALS],
-
-  typeKeywords: [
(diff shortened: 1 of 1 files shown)


max-sixty:
Would be great to have a test that would have caught this — not sure of the best way of doing that — is there a command that will test that the playground will start correctly?

vanillajonathan:
Well you could run `npm run build` for Vite to build the playground, and you could run `npm run dev` for Vite to build and run it for development where it will refresh on changes, or you can run `npm run preview` for Vite to build and serve a preview of the build from the `dist/` directory.

max-sixty:
> Well you could run `npm run build` for Vite to build the playground, and you could run `npm run dev` for Vite to build and run it for development where it will refresh on changes, or you can run `npm run preview` for Vite to build and serve a preview of the build from the `dist/` directory.

(I think this is probably in reference to https://github.com/PRQL/prql/pull/3666)

OK great — so `npm run build && npm run preview` works for me!

So maybe we just need to update the `Taskfile` to re-revert the `vite` PR.

vanillajonathan:
Yeah!

PR #3672, opened 2023-10-14
feat: Add octal numerical notation

vanillajonathan:
This lets you use octal notation to express numbers so you can do:
```elm
from files
filter permissions == 0o777
```

### DIFF
--- a/crates/prql-parser/src/lexer.rs
+++ b/crates/prql-parser/src/lexer.rs
@@ -189,12 +189,13 @@ fn literal() -> impl Parser<char, Literal, Error = Cheap<char>> {
         .then_ignore(just("_").or_not())
         .ignore_then(
             filter(|&c| ('0'..='7').contains(&c))
-            })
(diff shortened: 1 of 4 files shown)


crates/prql-parser/src/lexer.rs
```diff
@@ -185,6 +185,27 @@ fn literal() -> impl Parser<char, Literal, Error = Cheap<char>> {
...
+        .ignore_then(
+            filter(|c: &char| {
+                *c == '0'
+                    || *c == '1'
+                    || *c == '2'
+                    || *c == '3'
+                    || *c == '4'
+                    || *c == '5'
+                    || *c == '6'
+                    || *c == '7'
```
max-sixty:
```suggestion
            filter(|&c| ('0'..='7').contains(&c))
```
?

max-sixty (approved):
Looks good, thanks! I suggested one small optimization...

PR #3673, opened 2023-10-14
internal: Don't parse leading unaries in numbers

max-sixty:
Should we be parsing these as part of the numbers? My guess is that we should just do it in one place, and so `-5` and `-foo` are parsed similarly.

### DIFF
--- a/crates/prql-parser/src/lexer.rs
+++ b/crates/prql-parser/src/lexer.rs
@@ -195,9 +195,7 @@ fn literal() -> impl Parser<char, Literal, Error = Cheap<char>> {
         .chain::<char, _, _>(filter(|c: &char| c.is_ascii_digit()))
         .chain::<char, _, _>(filter(|c: &char| c.is_ascii_digit() || *c == '_').repeated());
 
-    let number = one_of("+-")
-        .or_not()
-        .chain::<char, _, _>(integer)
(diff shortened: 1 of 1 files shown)


max-sixty:
Note this also _reduces_ the priority of integrating the recent number representations into `number`, CC @vanillajonathan

aljazerzen (approved):
Yup, we statically evaluate 
```
Unary { op: Neg, expr: Literal(Integer(x)) }
```
... to:
```
Literal(Integer(-x))
```
... anyway, so this simplification has no effect on the compiler behavior.

---

We'd still like to parse `0x1000_0000` too.

vanillajonathan:
In Python and Rust you can also write it as `0b_0000_0000`.

Python also provides friendly diagnostic messages:
```pycon
>>> 0b3
  File "<stdin>", line 1
    0b3
      ^
SyntaxError: invalid digit '3' in binary literal
>>> 0o8
  File "<stdin>", line 1
    0o8
      ^
SyntaxError: invalid digit '8' in octal literal
>>> 0xg
  File "<stdin>", line 1
    0xg
     ^
SyntaxError: invalid hexadecimal literal
```

max-sixty:
Nice! Would be great if we could do something similar!

PR #3674, opened 2023-10-15
devops: Push devcontainer when requested

max-sixty:
I realize this was _never_ pushing the devcontainer...

Let's see if this works


max-sixty:
I also cleared up the workflows a bit

PR #3675, opened 2023-10-15
docs(CHANGELOG): update of Quarto VS Code extension

eitsupi:
See quarto-dev/quarto#300

Currently PRQL code blocks do not have syntax highlighting and there is no option completion for code blocks, but the new version should apply proper syntax highlighting (if PRQL VS Code is installed) and auto-completion of options.

We will be able to add a page on this to the documentation after the new version of the Quarto VS Code extension is released.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -54,6 +54,10 @@
   has support for `.prql` files. (@AlDanial)
 - [gocloc](https://github.com/hhatto/gocloc) a source lines of code counter now
   has support for `.prql` files. (@vanillajonathan)
+- [The Quarto VS Code extension](https://marketplace.visualstudio.com/items?itemName=quarto.quarto)
+  supports editing PRQL code blocks
+  ([`prqlr`](https://prql-lang.org/book/project/bindings/r.html) is required to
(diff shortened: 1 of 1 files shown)


max-sixty:
Awesome!

To the extent we can somehow bundle `prqlr` with the extension, that would be cool. But may not be ready too detect the correct R environment etc...

eitsupi:
The need for `prqlr` is not directly related to the VS Code extension, as it is only needed when rendering via the `knitr` engine by the Quarto CLI.
(In short, the Quarto VS Code extension only helps with editing)

max-sixty:
> as it is only needed when rendering via the `knitr` engine by the Quarto CLI.

Yes totally (but having it render correctly is cool, so if it's costless to install...)

eitsupi:
In fact, installation costs are quite high...

eitsupi:
It works!

image

Perhaps we can consider developing this to display the prql-vscode preview screen on the side like this:
https://quarto.org/docs/authoring/diagrams.html#authoring

max-sixty:
Really cool!! 



> Perhaps we can consider developing this to display the prql-vscode preview screen on the side like this:

Yes definitely, that sounds good

Issue #3630, opened 2023-10-08
`read_csv` and `read_parquet` should be compiled to the `file()` function in ClickHouse?


eitsupi:
https://clickhouse.com/docs/en/sql-reference/table-functions/file

It might be worth converting something like this:

```elm
let read_parquet = source -> s"file({source:0}, 'Parquet')"
let read_csv = source -> s"file({source:0}, 'CSV')"
```

PR #3676, opened 2023-10-15
test: add integration tests for the `std.sql.read_csv` function

eitsupi:
Related to #3630

Add a test for the `read_csv` function.
The current targets are DuckDB, ClickHouse (unimplemented) and GlareDB (unimplemented and not yet able to run the test).


crates/prql-compiler/tests/integration/main.rs
```diff
@@ -35,7 +35,13 @@ trait IntegrationTest {
     fn get_connection(&self) -> Option<DbConnection>;
     // We sometimes want to modify the SQL `INSERT` query (we don't modify the
     // SQL `SELECT` query)
-    fn import_csv(&mut self, protocol: &mut dyn DbProtocol, csv_name: &str, runtime: &Runtime);
+    fn import_csv(
+        &mut self,
+        protocol: &mut dyn DbProtocol,
+        csv_dir: &str,
+        csv_name: &str,
```
max-sixty: These two could be a single path and then we extract the dir & name from it (but not a big deal)

crates/prql-compiler/tests/integration/main.rs
```diff
@@ -155,36 +171,32 @@ impl IntegrationTest for Dialect {
...
+                        &format!(
+                            "COPY {csv_name} FROM '{csv_dir}/{csv_name}.csv' (AUTO_DETECT TRUE);"
+                        ),
                         runtime,
                     )
                     .unwrap();
             }
             Dialect::SQLite => {
-                let path = get_path_for_table(csv_name);
+                let path = format!("{csv_dir}/{csv_name}.csv");
```
max-sixty: Using slashes rather than the path API presumably won't work on windows, but I'm not sure we really mind about that...

max-sixty (approved):
Very nice, thanks!

crates/prql-compiler/tests/integration/main.rs
eitsupi: Generally, slashes work on Windows. It was probably ancient versions that did not work.
https://stackoverflow.com/questions/1589930/so-what-is-the-right-direction-of-the-paths-slash-or-under-windows

For example, it is already used in the following places, but it works, right?
https://github.com/PRQL/prql/blob/f3a750e498aadd86b4ebe168bbfc0ec0b55f7c87/web/book/tests/documentation/website.rs#L16

max-sixty: Ah cool! TIL...

Good point re it already working there!

crates/prql-compiler/tests/integration/main.rs
```diff
@@ -348,7 +345,10 @@ fn test_rdbms() {
             }
             let dialect = con.dialect;
             let options = Options::default().with_target(Sql(Some(dialect)));
-            let mut rows = prql_compiler::compile(&prql, &options)
+            let replaced_prql = regex::Regex::new("data_file_root")
+                .unwrap()
+                .replace_all(&prql, &con.data_file_root);
```
eitsupi: It would be quite wasteful to do this for every query, but perhaps not a problem? (I was also thinking of doing this only if the filename contains `read_` as a workaround)

crates/prql-compiler/tests/integration/queries/read_csv.prql
```diff
@@ -0,0 +1,5 @@
+# sqlite:skip
+# postgres:skip
+# mysql:skip
+# clickhouse:skip (Not implemented yet)
+from (read_csv "data_file_root/media_types.csv")
```
max-sixty, line 5: I think it's really good that we're testing this — one of the most helpful things our integration tests can do!

crates/prql-compiler/tests/integration/main.rs
max-sixty: I think it's OK, and less code than checking for the filename (which not be much cheaper...)

One small change would be to move the Regex creation outside the loop — plausibly that's an order of magnitude more expensive than the replace itself.

crates/prql-compiler/tests/integration/main.rs
```diff
@@ -328,14 +328,22 @@ fn test_rdbms() {
...
             .file_stem()
             .and_then(|s| s.to_str())
             .unwrap_or_default();
 
-        let prql = fs::read_to_string(path).unwrap();
+        let mut prql = fs::read_to_string(path).unwrap();
+
+        if re.is_match(&prql) {
+            is_contains_re = true;
+        }
```
max-sixty:
```suggestion
        let is_contains_data_root = re.is_match(&prql);
```

No strong view on the names etc — the main adjustment is a small change to set `X=Y` rather than `if X then Y = true`

crates/prql-compiler/tests/integration/main.rs
```diff
@@ -328,14 +328,22 @@ fn test_rdbms() {
         con.setup_connection(runtime);
     }
 
+    // Each connection has a different data_file_root, so we need to replace.
+    let re = regex::Regex::new("data_file_root").unwrap();
+    let mut is_contains_re = false;
```
max-sixty:
```suggestion
```

PR #3678, opened 2023-10-15
revert: Vite-again

vanillajonathan:
This reverts commit 37db5f23b85ce4e29b08f73b1f7ab8d73c44dc54.

And it fixes the Taskfile so that it runs Vite.


max-sixty:
I corrected the readme too

max-sixty:
Thanks!

Sorry the revert was more git work than anticipated. I should have checked the original more thoroughly. (though I can't promise it won't happen again — I do like to merge sooner and then revert if something isn't quite right, rather than leave PRs hanging for days...)

max-sixty:
This deserves an entry in the Changelog — it's a big change! Thanks @vanillajonathan

PR #3679, opened 2023-10-15
refactor: cleanup

aljazerzen:
- refactor: cleanup literal Display trait
- refactor: move Error into prql_ast
- refactor: move generic
- refactor!: stop leaking chumsky prql_parser
- refactor: move parser tests into own file


crates/prql-ast/src/expr/literal.rs
```diff
@@ -20,3 +20,62 @@ pub struct ValueAndUnit {
...
+        }
+        Ok(())
+    }
+}
+
+fn quote_string(s: &str, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+    let s = escape_all_except_quotes(s);
+
+    if !s.contains('"') {
+        return write!(f, "\"{s}\"");
```
max-sixty:
```suggestion
        return write!(f, r#""{s}""#);
```

(personal preference, feel free to ignore)

crates/prql-ast/src/expr/literal.rs
```diff
@@ -20,3 +20,62 @@ pub struct ValueAndUnit {
...
+    }
+
+    if !s.contains('\'') {
+        return write!(f, "'{s}'");
+    }
+
+    // when string contains both single and double quotes
+    // find minimum number of double quotes
+    let mut quotes = "\"\"\"".to_string();
```
max-sixty:
```suggestion
    let mut quotes = r#""""""#".to_string();
```

crates/prql-ast/src/expr/literal.rs
```diff
@@ -20,3 +20,62 @@ pub struct ValueAndUnit {
...
+    while s.contains(&quotes) {
+        quotes += "\"";
+    }
+    write!(f, "{quotes}{s}{quotes}")
```
max-sixty, line 68: Because we're only adding one quote for every three in `s`, does this work correctly?

crates/prql-ast/src/error.rs
```diff
@@ -0,0 +1,107 @@
+//! Error message produced by the compiler.
+// Should be in some prql_common crate, but until we need that, it can reside here.
```
max-sixty, line 2: FYI I thought this blog post was interesting, which I read yesterday — I haven't really thought about it like this before, given I've never worked on a _huge_ rust project. 

https://blog.danhhz.com/compile-times-and-code-graphs

max-sixty (approved):
I haven't reviewed it all, because I'm not sure what's new vs a move. But looks great generally.

(though I'm going to try and be better about really understanding the code, rather than "LGTM" and then it becoming too far for me to follow)

crates/prql-ast/src/expr/literal.rs
aljazerzen: This is basically encasing s into number of quotes that is the largest number of quotes in s + 1.

So if s is `hello"""""""world`, there are 7 quotes, and this would be encased into 8 quotes.

aljazerzen:
It all just refactoring, apart for improvements to the quoting code and a few new tests.

crates/prql-ast/src/expr/literal.rs
max-sixty: Sorry, that's obvious now. I was misreading `s` vs. `quotes`...

crates/prql-ast/src/expr/literal.rs
```diff
@@ -61,7 +61,7 @@ fn quote_string(s: &str, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
 
     // when string contains both single and double quotes
     // find minimum number of double quotes
-    let mut quotes = "\"\"\"".to_string();
+    let mut quotes = r#""""""#".to_string();
```
max-sixty: Sorry, I realize I had a stray `"` there...

aljazerzen:
We need to do something about our string literals.

There is quite a lot of stuff supported, even though not strictly necessary. We have:
- "-quoted strings,
- '-quoted strings,
- escapes, so we are able to express the following: `"hello'`,
- triple and n-ple quotes strings with either ' or ".

We can enforce stricter rules and provide nicer error messages, the need to be backward compatible is not that strong yet.

Also, our lexer does not conform to these ideas: https://matklad.github.io/2023/08/01/on-modularity-of-lexical-analysis.html

max-sixty:
I think the thing to cut would be one of the `'` or `"` representations. Is there anything else we could / should cut?

max-sixty:
> Also, our lexer does not conform to these ideas: https://matklad.github.io/2023/08/01/on-modularity-of-lexical-analysis.html

Yes I sent this around when working on the line wrapping character! Ours doesn't do this. But folks thought it was a worthwhile tradeoff...

aljazerzen:
Could we forbid multiline strings and triple quotes strings in favor of this syntax:

```
let a = "
    "this is the first line
    "  second line, indented with two spaces
  "  third line, also indented with two spaces
 
```

This is what zig does: https://ziglang.org/documentation/master/#String-Literals-and-Unicode-Code-Point-Literals.

The syntax that I propose differs in
- usage of `"` instead of `//` and
- the leading `"` on the first line, which is needed because our newlines have semantic meaning.

max-sixty:
Interesting!

How would the string finish? Would we allow subsequent `"` on a line?

aljazerzen:
Multiline string extends a long as the first non-whitespace is `"`. I think this is how Zig does it?

max-sixty:
Yes interesting. I'm very open to it.

Possibly we could add a character in front of the `"`, such as `\`, to make it a) definitely a multiline string and not an unclosed string, b) look less like a syntax error to folks who would otherwise see this as a syntax error...

```
let a = \"
    \"this is the first line
    \"  second line, indented with two spaces
  \"  third line, also indented with two spaces
 ```

(not sure this specific suggestion is great, even though I think the goals above are reasonable...

**Comment by aljazerzen** [comment]
Hmm, these backslashes look strange... It feels like we are escaping the PRQL syntax :D

But your concern is a valid: it does look very much like an unclosed string.

We could get around that with this approach too:

```
let a = "an unclosed string
        -------------------
                        \__ error: unclosed string
Hint: if you want a multi-line string, use this:
      ```
      let a = "
        "an unclosed string
      ```
```

max-sixty:
Yes totally, that'd be fine too.

Can we think about it for a couple of days? Thoughts from @PRQL/prql-team ?

snth:
> Could we forbid multiline strings and triple quotes strings in favor of this syntax:

That example looks odd to me and I'm not understanding what advantage it brings?

I like this [d-string](https://discuss.python.org/t/d-string-vs-str-dedent/35907) proposal that @max-sixty posted to the #language-design channel on Discord recently.

---
Should we move the string quoting discussion to a separate issue?

PR #3681, opened 2023-10-15
build: justfile

aljazerzen:


.prettierignore
```diff
@@ -1,8 +1,17 @@
+**/*.rs
 **/*.min.*
+
 /web/book/theme/highlight.js
+/web/playground/build
+/web/website/public
+/web/book/book
```
max-sixty, line 7: Modern versions of prettier already respect `.gitignore`s IIRC. I guess no problem with having more here, but it shouldn't be required

justfile
```diff
@@ -0,0 +1,52 @@
+pull-request: fmt prqlc-test
```
max-sixty, line 1:
```suggestion
@# For context — we generally use `task` for common tasks, but some developers prefer `just`, so we're trying this out for a few tasks

pull-request: fmt prqlc-test
```

justfile
```diff
@@ -0,0 +1,52 @@
...
+
+    @ echo '--- cargo fmt ---'
+    @ cargo fmt --all
+
+    @ echo '--- prettier ---'
+    @ prettier --write . \
+            --config=.prettierrc.yaml \
+            --ignore-path=.prettierignore \
+            --ignore-unknown \
+            --log-level=warn
```
max-sixty, line 19: No objection — your file — but surely pre-commit is better for these sorts of things, rather than rewriting them in `just`?

justfile
```diff
@@ -0,0 +1,52 @@
...
+
+
+prqlc-lint:
+    @echo '--- clippy ---'
+    @cargo clippy --all --fix --allow-staged
+
+
+target := 'x86_64-unknown-linux-gnu'
```
max-sixty, line 27: Yeah, as an example — this is coded up to be really flexible in the Taskfile, whereas this will only work on linux

justfile
aljazerzen: I don't like precommit because:
- it interferes with my committing workflow with git hooks,
- it is slow,
- I had trouble with installation (it was downloading stuff each time and writing to my home dir)

My goal when writing this was to have something that is equivalent to running our pipelines. So it could (maybe) (someday) replace the action.

The idea is that nix provides the environment (downloaded, compiled and cached) and just executes a few commands to determine the CI status. This has a few benefits:
- executable locally,
- no downloading, booting up clusters, calling external services in the pipeline,
- no automatic commits to my branch that I then accidentally override, interfere with commit messages, prevent me for fast-forward push

Also, depending on a proprietary build system is not wise, even if is provided by the most open-source friendly company (GitHub).

justfile
aljazerzen: I was translating this from GH actions and provided the default for target here.

justfile
max-sixty: I think you're underrating `pre-commit`!

If we only run `pre-commit install-hooks` (and not `pre-commit install`, which I also find annoying, and IMO is a bad suggestion), then it only runs on `pre-commit run -a`, and doesn't interrupt any commit workflow. I've never had issues with installation (not denying that you did, but might be worth giving it another go)

---

Totally agree it would be awesome to be able to run CI locally, very much so. But I think that's a broader topic than pre-commit — my main claim here is that writing our own `prettier` script, or expression to find `dbg` or lines with whitespace is worse than pre-commit.

aljazerzen: But what's the value proposition of precommit then? It invokes prettier? It has a few hooks predefined?

I did use it for a few days and realized it's mostly just calling other tools...

max-sixty: > I did use it for a few days and realized it's mostly just calling other tools...

Totally! 😁 

It includes everything to install a tool on any OS — e.g. `actionlint` is a few lines rather than us figuring out to how to get it to install on each system (inc Windows, which me & you might not care about, but some do). Tools can define how they're run themselves, so we don't need to write our own scripts — the `prettier` example is a good one. It keeps everyone on the same version (underrated!). It's a common approach across projects.

Honestly think it's great, and quite different from your points about running CI locally — which I strongly agree with...

Issue #3671, opened 2023-10-14
DevContainer:  Incompatible assets installed: required mdbook-admonish assets version '^2.0.0', but found '3.0.0'.


richb-hanover:
### What happened?

I opened `b48a6f130` (from today) in VSCode to use the DevContainer. I let it rebuild the project. Everything went fine. I ran `task run-playground` and it worked as expected. 

But when I ran `task run-book` I got the following errors:

1. I got an mdbook-admonish version warning
2. After installing mdbook-admonish, I got a compiler warning.

But it `task run-book` worked fine afterwards

Here's a narrated output:
```
vscode ➜ /workspaces/prql (main) $ task run-book
task: [run-book] mdbook serve --port=3000
2023-10-14 22:30:47 [INFO] (mdbook::book): Book building has started
[2023-10-14T22:30:47Z ERROR mdbook_admonish] Fatal error: ERROR:
      Incompatible assets installed: required mdbook-admonish assets version '^2.0.0', but found '3.0.0'.
      Please run `mdbook-admonish install` to update installed assets.
      For more information, see: https://github.com/tommilligan/mdbook-admonish#semantic-versioning
[2023-10-14T22:30:47Z ERROR mdbook_admonish]   - ERROR:
      Incompatible assets installed: required mdbook-admonish assets version '^2.0.0', but found '3.0.0'.
      Please run `mdbook-admonish install` to update installed assets.
      For more information, see: https://github.com/tommilligan/mdbook-admonish#semantic-versioning
2023-10-14 22:30:47 [ERROR] (mdbook::utils): Error: The "admonish" preprocessor exited unsuccessfully with exit status: 1 status
task: Failed to run task "run-book": exit status 101
```
So I typed `mdbook-admonish install` and got another error
```
vscode ➜ /workspaces/prql (main) $ mdbook-admonish install
[2023-10-14T22:32:11Z INFO  mdbook_admonish::install] Reading configuration file './book.toml'
[2023-10-14T22:32:11Z ERROR mdbook_admonish] Fatal error: can't read configuration file './book.toml'
[2023-10-14T22:32:11Z ERROR mdbook_admonish]   - can't read configuration file './book.toml'
[2023-10-14T22:32:11Z ERROR mdbook_admonish]   - No such file or directory (os error 2)
```
Then I cd'd to _web/book_ and that ran fine
```
vscode ➜ /workspaces/prql (main) $ cd web/book
[...]

PR #3682, opened 2023-10-15
ci: Ensure devcontainer is pushing correctly

max-sixty:
I'm seeing an error around `true` at https://github.com/PRQL/prql/actions/runs/6525681773


eitsupi:
I fear this may be over complicating things. Could you please explain your intent?

.github/workflows/tests.yaml
```diff
@@ -153,6 +153,13 @@ jobs:
           github.ref == 'refs/heads/main' && github.event_name == 'push' }}" >>
           "$GITHUB_OUTPUT"
 
+      - id: devcontainer-build
+        run:
+          echo "run=${{ steps.devcontainer-push.outputs.run == 'true' ||
+          steps.changes.outputs.devcontainer-build == 'true' ||
+          steps.changes.outputs.nightly-schedule == 'true' }}" >>
+          "$GITHUB_OUTPUT"
```
eitsupi, line 161: Perhaps this is related to the fact that GitHub Actions have a bool type but Shell does not.
Are you sure that choosing `'true'` instead of `true` here is correct?

max-sixty: Here, inside `${{ }}` IIUC _everything_ is stringified. So I think we always want `== 'true'`...

.github/workflows/release.yaml
```diff
@@ -263,4 +263,6 @@ jobs:
     if: github.event_name == 'release'
     uses: ./.github/workflows/build-devcontainer.yaml
     with:
+      # TOOD: I think having this as a bool is OK, but check a recent release
+      # from 0.10.0 and confirm that it is indeed pushing the image.
       push: true
```
max-sixty, line 268: Here, I'm less sure — the last release ran with this being false: https://github.com/PRQL/prql/actions/runs/6211228138/job/16860225191#step:8:8

max-sixty:
> I fear this may be over complicating things. Could you please explain your intent?

Yes for sure.

The main issue is that we haven't had a devcontainer be pushed since August. Hence https://github.com/PRQL/prql/issues/3671.

GHA has some very awkward and (to me) complicated rules around when values are typed as bools vs strings, so this tries to:
- Clean those up
- Encapsulate the logic a bit more, similar to some of the other jobs in `tests.yaml`

I kinda agree it's over complicated, though...

max-sixty:
There are lots of commits here, because the only way of testing is by pushing! So I did at least confirm that it passes `true` through to the docker action when it should ...

Unless you disagree, I'd be up for merging and ensuring this works (i.e. does it actually push the image next time that it should), even if we later simplify it.

eitsupi (approved):
Ok, thanks for working on this. Try merging...

max-sixty:
This does seem to work: https://github.com/PRQL/prql/actions/runs/6527982444/job/17723556755#step:8:8

PR #3683, opened 2023-10-15
refactor: reorganize dirs

aljazerzen:
Combine compiler "things" into its own dir.

This basically says "bindings are not PRQL bindings, but prqlc bindings". Similar for packages/.

The caveat is that I've shortened the `prqlc/crates/prql-parser/` to `prqlc/crates/parser`, but kept the crate name `prql-parser`. We can update this, but I don't want to change the published names "prql-compiler" and "prqlc".


Cargo.toml
```diff
@@ -1,13 +1,18 @@
...
-  "crates/*",
-  "bindings/prql-elixir/native/prql",
-  "bindings/prql-java",
-  "bindings/prql-js",
-  "bindings/prql-lib",
-  "bindings/prql-python",
+  "prqlc/crates/ast",
+  "prqlc/crates/compiler",
+  "prqlc/crates/compiler-macros",
+  "prqlc/crates/parser",
```
max-sixty: I would note from `@matklad` in https://matklad.github.io/2021/08/22/large-rust-workspaces.html

> Don’t succumb to the temptation to strip common prefix from folder names. If each crate is named exactly as the folder it lives in, navigation and renames become easier. Cargo.tomls of reverse dependencies mention both the folder and the crate name, it’s useful when they are exactly the same.

prqlc/crates/compiler/README.md
```diff
@@ -1,17 +1,18 @@
...
-`prql-compiler` contains the implementation of PRQL's compiler, written in Rust.
+`prqlc` is the reference implementation of a compiler from PRQL to SQL, written
+in Rust.
 
 For more on PRQL, check out the [PRQL website](https://prql-lang.org) or the
 [PRQL repo](https://github.com/PRQL/prql).
 
 For more usage examples and the library documentation, check out the
-[`prql-compiler` documentation](https://docs.rs/prql-compiler/latest/prql_compiler/).
+[`prqlc_main` documentation](https://docs.rs/prqlc_main/latest/prql_compiler/).
```
max-sixty: What's `prqlc_main`?

max-sixty (approved):
I'm OK with this — it's purely internal.

I'm pro-the approach described at https://matklad.github.io/2021/08/22/large-rust-workspaces.html, but given your contributions I would weigh your vote very highly here.

prqlc/crates/compiler/README.md
aljazerzen: I had prql-compiler renamed to `prqlc/crates/main`.

.config/vscode-recommended/settings.json
```diff
@@ -12,6 +12,8 @@
     "**/Thumbs.db": true,
     "web/book/book": true,
     "web/build": true,
-    ".direnv": true
+    ".direnv": true,
+    ".pytest_cache": true,
+    ".mypy_cache": true
```
max-sixty, line 17: FWIW I have these in my global settings rather than per-workspace...

max-sixty:
I see the upgrades have caused merge conflicts. They should be mostly done (they come in each Monday), but feel free to pause them if it's causing you trouble.

max-sixty:
@aljazerzen lmk if you want a hand merging this! I worry that as more changes come from elsewhere, git might not be smart enough and the merge will be really difficult.

I don't think there's much we can "split off" — you already merged things like the justfile separately...

aljazerzen:
Yeah, that'd be great, if you have the time.

max-sixty:
> Yeah, that'd be great, if you have the time.

You merged 2 hours later! I'm too online, but not _that_ online. :D

aljazerzen:
I though more things need fixing :D

eitsupi:
Great work!

aljazerzen:
@eitsupi I was a bit worried you'd have objections to the reorganization, since you have done a lot of work on the bindings, so hearing praises from you means a lot!

PR #3690, opened 2023-10-17
ci: Make space on runner image

max-sixty:
Unfortunately we've been running out of space in the past few days when running GHA. I had been clearing the cache and hoping that it was due to stale cache files. But even that doesn't work in https://github.com/PRQL/prql/actions/runs/6540530244/job/17760594651.

This was suggested (link inline). Otherwise we could try splitting the "build extra targets" from the integration tests, which require pulling docker images.


eitsupi:
You may wan to see this.
https://github.com/apache/arrow/blob/4011058f4a56bdcf160f46373355ffa0e22bcd2c/ci/scripts/util_free_space.sh

max-sixty:
> You may wan to see this. https://github.com/apache/arrow/blob/4011058f4a56bdcf160f46373355ffa0e22bcd2c/ci/scripts/util_free_space.sh

Thanks a lot!

eitsupi:
Hmm, this might be overkill here since the script took nearly 3 minutes to run.

eitsupi:
The new version of DuckDB is taking a long time to compile, so I'll merge it for now.
We can come back to it later if you feel the script is overkill.

max-sixty:
> Hmm, this might be overkill here since the script took nearly 3 minutes to run.

Ah, that's unfortunate. We could try only if there's no cache hit; possibly when there's no cache hit then it needs more disk space to build...

I'm surprised disk space is in such short supply...

PR #3694, opened 2023-10-17
feat(Lezer grammar): Highlight the equals operator

vanillajonathan:
This exposes a token named `Equals` for the equals operator `=`, we then map the `Equals` token to the Lezer highlighting tag [`definitionOperator`](https://lezer.codemirror.net/docs/ref/#highlight.tags.definitionOperator).


max-sixty (approved):
Cool!

PR #3696, opened 2023-10-17
feat(Lezer grammar): Add Boolean token

vanillajonathan:
This exposes a token named `Boolean` for `true` and `false`, we then map the `Boolean` token to the Lezer highlighting tag [`bool`](https://lezer.codemirror.net/docs/ref/#highlight.tags.bool).


max-sixty:
Looks good!

PR #3698, opened 2023-10-18
feat(Lezer grammar): Highlight null, this and that

vanillajonathan:
This exposes `null`, `this` and `that` as keywords which we map to the Lezer highlighting tag [`null`](https://lezer.codemirror.net/docs/ref/#highlight.tags.null) and [`self`](https://lezer.codemirror.net/docs/ref/#highlight.tags.self) respectively.

This `@precedence` block has been split. Parse tree tests have been added for `null`, `this` and `that`.


aljazerzen (approved):

max-sixty:
Thanks @vanillajonathan !

PR #3700, opened 2023-10-18
revert: check github links

aljazerzen:
Followup for #3683

Because I've moved a bunch of files, links to github.com repo main branch were broken.
So I've disabled the check, merged and now they should work again.


max-sixty:
Cool, thanks for fixing!

PR #3703, opened 2023-10-18
feat!: any type

aljazerzen:
Also renames:
- std.every to std.all, ref #2940,
- std.any to std.some.

### DIFF
--- a/prqlc/prql-compiler/src/sql/std.sql.prql
+++ b/prqlc/prql-compiler/src/sql/std.sql.prql
@@ -34,7 +34,7 @@ let stddev = column -> s"STDDEV({column:0})"
 let all = column -> s"BOOL_AND({column:0})"
 
 @{window_frame=true, coalesce="FALSE"}
-let some = column -> s"BOOL_OR({column:0})"
+let any = column -> s"BOOL_OR({column:0})"
 
 @{window_frame=true, coalesce="''"}
 let concat_array = column -> s"STRING_AGG({column:0}, '')"
@@ -218,7 +218,7 @@ module sqlite {
   let all = column -> s"MIN({column:0}) > 0"
 
   @{window_frame=true, coalesce="FALSE", binding_strength=6}
-  let some = column -> s"MAX({column:0}) > 0"
+  let any = column -> s"MAX({column:0}) > 0"
 
   @{window_frame=true, coalesce="''"}
   let concat_array = column -> s"GROUP_CONCAT({column:0}, '')"
--- a/prqlc/prql-compiler/src/sql/std.sql.prql
+++ b/prqlc/prql-compiler/src/sql/std.sql.prql
@@ -31,10 +31,10 @@ let average = column -> s"AVG({column:0})"
 let stddev = column -> s"STDDEV({column:0})"
 
 @{window_frame=true, coalesce="TRUE"}
-let every = column -> s"BOOL_AND({column:0})"
+let all = column -> s"BOOL_AND({column:0})"
 
 @{window_frame=true, coalesce="FALSE"}
-let any = column -> s"BOOL_OR({column:0})"
+let some = column -> s"BOOL_OR({column:0})"
 
 @{window_frame=true, coalesce="''"}
 let concat_array = column -> s"STRING_AGG({column:0}, '')"
@@ -215,10 +215,10 @@ module glaredb {
 
 module sqlite {
   @{window_frame=true, coalesce="TRUE", binding_strength=6}
-  let every = column -> s"MIN({column:0}) > 0"
+  let all = column -> s"MIN({column:0}) > 0"
 
   @{window_frame=true, coalesce="FALSE", binding_strength=6}
-  let any = column -> s"MAX({column:0}) > 0"
+  let some = column -> s"MAX({column:0}) > 0"
 
   @{window_frame=true, coalesce="''"}
   let concat_array = column -> s"GROUP_CONCAT({column:0}, '')"
--- a/prqlc/prql-compiler/src/codegen/pl.rs
+++ b/prqlc/prql-compiler/src/codegen/pl.rs
@@ -73,6 +73,7 @@ impl WriteSource for TyKind {
                 r += &(*func.return_ty).as_ref().write(opt)?;
                 Some(r)
             }
+            Any => Some("any".to_string()),
         }
     }
 }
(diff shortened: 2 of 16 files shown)


.envrc
```diff
@@ -0,0 +1 @@
+use flake .#full
```
max-sixty: No objection, but confirming this is intentional

aljazerzen: It was not

aljazerzen: switching between my old branch which has it committed deletes my uncommitted file on master branch

prqlc/prql-compiler/src/sql/std.sql.prql
```diff
@@ -31,10 +31,10 @@ let average = column -> s"AVG({column:0})"
 let stddev = column -> s"STDDEV({column:0})"
 
 @{window_frame=true, coalesce="TRUE"}
-let every = column -> s"BOOL_AND({column:0})"
+let all = column -> s"BOOL_AND({column:0})"
 
 @{window_frame=true, coalesce="FALSE"}
-let any = column -> s"BOOL_OR({column:0})"
+let some = column -> s"BOOL_OR({column:0})"
```
max-sixty: I don't have an immediate solution, given it sounds like we want `any` for types, but I do think that `any` is a much better function name than `some`:
- convention
- `some` generally means "more than one"

prqlc/prql-compiler/src/sql/std.sql.prql
```diff
@@ -31,10 +31,10 @@ let average = column -> s"AVG({column:0})"
 let stddev = column -> s"STDDEV({column:0})"
 
 @{window_frame=true, coalesce="TRUE"}
-let every = column -> s"BOOL_AND({column:0})"
+let all = column -> s"BOOL_AND({column:0})"
```
max-sixty, line 34: Nice, I do think this is better!

max-sixty:
Should the commit be renamed to `feat!` given I think it's breaking?

If you have the patience, a changelog entry would be great! But if you're on a roll, leave it and I'll do it on release

prqlc/prql-compiler/src/sql/std.sql.prql
aljazerzen: > some generally means "more than one"

I'd say it means "one or more".

In any case, I do agree that "any" is a better name than "some".

aljazerzen: ... but we need a name for the "all encapsulating type". In typescript this is called `any` and in Python it is `Any`. So there is no debate here.

aljazerzen:
I'm cherry-picking from my branch, will add changelog later if no one beats me to it.

prqlc/prql-compiler/src/sql/std.sql.prql
aljazerzen: For reference:

Rust has `Iterator.any`: https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.any

JavaScript has `Array.some`: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/some

Python has `any`: https://docs.python.org/3/library/functions.html#any

Pandas has `any`: https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.any.html

Dyplr has `cumany`: https://dplyr.tidyverse.org/reference/cumall.html

Postgres has neither (or both if you count range functions): (see first note below the table) https://www.postgresql.org/docs/9.5/functions-aggregate.html

max-sixty: Yeah — the root issue is that most languages can have `any` as both a function and a type, and so `any` is used for both.

How viable do you think that is for us?

aljazerzen: Not viable at all. I hope that the need for having `any` type in the language at all will subdue when we add generics.

Right now, I need any type for expressing `array of anything` or `tuple of anything`, but this will go away when we have generics, because you'd write `array of T` or `tuple of T`.

So maybe, for now, any type can be named `anytype` and `some` can be renamed back to `any`. And we do a rename again later, if it turns out there really is a need for any type.

max-sixty: Yes, that's a great idea

snth: From earlier in the thread I was thinking to myself, why not use `anything` for the type name so that we can keep `any` for the function. Then I got to [comment](https://github.com/PRQL/prql/pull/3703#discussion_r1364379675) with

>  I need any type for expressing `array of anything` or `tuple of anything`, 

So, @aljazerzen you also defaulted to that as your type name in your description so I think that would be a good natural choice.

I also think that the function reference to `any` will probably be used more by most people than the `anything` type reference so it makes sense to use the more common, shorter term for that.

snth: > So maybe, for now, any type can be named `anytype` and some can be renamed back to `any`. And we do a rename again later, if it turns out there really is a need for any type.

+1 on this for either `anytype` or `anything`.

eitsupi:
Maybe this will help solve #3127?

prqlc/prql-compiler/src/sql/std.sql.prql
eitsupi: Timely topic, I learned today that polars changed horizontal `any` to `any_horizontal`.
https://github.com/eitsupi/querying-with-prql/commit/4dcdfd6b58d7860ae6d29f9299ef1175f309061e
https://pola-rs.github.io/polars/py-polars/html/reference/expressions/api/polars.any_horizontal.html

dplyr has `if_any` for this.
https://dplyr.tidyverse.org/articles/colwise.html#filter

---

By the way, I think the equivalent here in dplyr is base R's `base::any`.
`dplyr::cumany` is a Window function.

``` r
any(FALSE, TRUE, FALSE)
#> [1] TRUE
dplyr::cumany(c(FALSE, TRUE, FALSE))
#> [1] FALSE  TRUE  TRUE
```

<sup>Created on 2023-10-19 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

---

I also think that the `any` function is written more often then any type, so I think a type name like `anytype` makes sense.

aljazerzen: Great, this means we've made the right choice.

PR #3707, opened 2023-10-18
feat(Lezer grammar): Highlight `let`

vanillajonathan:
This exposes `let` as a keyword which we map to the Lezer highlighting tag [`definitionKeyword`](https://lezer.codemirror.net/docs/ref/#highlight.tags.definitionKeyword).


max-sixty:
Cool!

PR #3708, opened 2023-10-18
ci: Remove `actions/upload-pages-artifact`

vanillajonathan:
Remove `actions/upload-pages-artifact` as per discussion in #3608.


max-sixty:
Thanks @vanillajonathan

max-sixty:
I think the (now reverted) change never hit the `web` branch, so it's likely OK, and we'll release soon

PR #3710, opened 2023-10-19
web: Use offcanvas for menu

vanillajonathan:
This uses Bootstrap's [offcanvas](https://getbootstrap.com/docs/5.3/components/navbar/#offcanvas) menu, just like the Bootstrap website. I think this is much better than our current implementation which I think works poorly and annoys me.


web/website/themes/prql-theme/layouts/partials/header.html
```diff
@@ -12,15 +12,28 @@
...
-          <li>
-            {{ print "[" .Name "]" "("  .URL ")" | markdownify }}
-          </li>
-        {{ end }}
-      </ul>
-      <i class="bx bx-list-ul mobile-nav-toggle"></i>
+    <nav id="navbar" class="navbar navbar-expand-lg">
+      <button class="navbar-toggler border-0 bg-white" type="button" data-bs-toggle="offcanvas" data-bs-target="#offcanvasNavbar" aria-controls="offcanvasNavbar" aria-label="Toggle navigation">
+        <svg viewBox="0 0 16 16" style="width: 1.5rem; height: 1.5rem">
+          <path d="M3 9.5a1.5 1.5 0 1 1 0-3 1.5 1.5 0 0 1 0 3zm5 0a1.5 1.5 0 1 1 0-3 1.5 1.5 0 0 1 0 3zm5 0a1.5 1.5 0 1 1 0-3 1.5 1.5 0 0 1 0 3z"></path>
```
max-sixty, line 18: Is there any info on what all these are?

max-sixty (approved):
I tried this out, looks very reasonable. Text maybe gets very slightly smaller, but very similar. And mobile seems to get a bit better

web/website/themes/prql-theme/layouts/partials/header.html
vanillajonathan: `navbar-expand-lg` is a class which sets the breakpoint for when it should change from links to a button that brings up a menu. In this case, it is `-lg` which is short for large which means large screen.

The button have some default styling, that we override with `border-0` and `bg-white`, then follows a couple of data attributes which is used by the Bootstrap JavaScript for listening  to events and configure the JavaScript which is used to handle `onclick` events to bring up the menu.

Then follows a `svg` element which is used for the triple-dot icon, which is defined by the `path` element.

max-sixty: Ah, so the triple dot icon is defined in text as an svg... OK, that sort of thing is helpful as a comment (though maybe this is more obvious to those who are more used to HTML?)

vanillajonathan: An inline SVG is quite obvious to those who know HTML, however I found a better way to do this without inline SVG, so I'll follow up with a PR.

Issue #3709, opened 2023-10-19
Field Report from using Vite build...


richb-hanover:
### What's up?

Here's a report of using commit b48a6f13 (18 Oct 2023) on Mac M2 (amd). It's working well, but there are a few rough edges that might be sanded off...

TL;DR:
- `postCreateCommand` gives error that doesn't seem to harm things
- `task run-playground` works, but opens on different port (3173) than expected. Works fine
- Vite.js is SO MUCH FASTER
- `task run-book` seems to compile things twice; opens on port 3000
- `task run-website` fails the first time, works  the second. Opens on port 1313 
- Did Playground, Book, and Website all run on different ports in earlier versions? I thought they all used 3000...

These nits aside, this is a terrific enhancement.

## Starting the Dev Container
Start up Dev Container in VSCode and see this error message. Dev Container seems to operate OK, though...

```
Running the postCreateCommand from devcontainer.json...

[65453 ms] Start: Run in container: /bin/sh -c task install-maturin && task install-pre-commit && pre-commit install-hooks
[65453 ms] Start: Run in container: /bin/sh -c task install-npm-dependencies
Running install-npm-dependencies from devcontainer.json...
task: [install-npm-dependencies] npm install --global prettier prettier-plugin-go-template
[###############...] | reify:prettier-plugin-go-template: http fetch GET 200 https://registry.npmjs.org/prettier-plugin-go-template/-/[###############...] | reify:prettier-plugin-go-template: http fetch GET 200 https://registry.npmjs.org/prettier-plugin-go-template/-/[###############...] | reify:prettier-plugin-go-template: http fetch GET 200 https://registry.npmjs.org/prettier-plugin-go-template/-/[###############...] | reify:prettier-plugin-go-template: http fetch GET 200 https://registry.npmjs.org/prettier-plugin-go-template/-/[###############...] | reify:prettier-plugin-go-template: http fetch GET 200 https://registry.npmjs.org/prettier-plugin-go-template/-/[###############...] | reify:prettier-plugin-go-template: http fetch GET 200 https://registry.npmjs.org/prettier-plugin-go-template/-/[###############...] | reify:prettier-plugin-go-template: http fetch GET 200 https://registry.npmjs.org/prettier-plugin-go-template/-/[###############...] | reify:prettier-plugin-go-template: http fetch GET 200 https://registry.npmjs.org/prettier-plugin-go-template/-/

added 3 packages in 2s

1 package is looking for funding
  run `npm fund` for details
[...]

PR #3712, opened 2023-10-20
devops: Don't change user for devcontainer

max-sixty:
From #3709

### DIFF
--- a/.devcontainer/base-image/Dockerfile
+++ b/.devcontainer/base-image/Dockerfile
@@ -2,8 +2,7 @@
 
 FROM mcr.microsoft.com/devcontainers/rust:1-1-bullseye
 
-# ========= Install cargo-tools for non-root user (vscode) =========
-USER vscode
+# ========= Install cargo-tools =========
 
 ARG cargo_crates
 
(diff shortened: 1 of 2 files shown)


eitsupi:
What does this change mean?
There is no point in this in my understanding.

max-sixty:
If we build as `root`, and then connect as `root` — would that plausibly prevent the error in #3709? I'm not confident here at all...

...but we also have multiple users in our image, and they're not working together well on Mac, and it seems unlikely to be a bug with VS Code. So hence the attempt to simplify...

eitsupi:
Please read the official documentation.
Which users are used in the container is recorded in the metadata (Docker image label) and is set upstream.
It is not related to this repository.
Overwriting with the root user completely destroys the operation on the Linux file system including WSL2 (unless you want it)

I don't understand at all why you would change the file here and ask me for a review without following my advice and opening the issue upstream

max-sixty:
> I don't understand at all why you would change the file here and ask me for a review without following my advice and opening the issue upstream

Can you help me understand why this is an upstream issue? I'm all for logging issues where they belong. But are

PR #3715, opened 2023-10-20
test: move sql tests into a separate binary

aljazerzen:
This should speed-up the compilation when you only update a single test,
as the compiler does not need to be updated.

On my computer it went from 3.5s to 2.1s. Not really significant.


max-sixty:
Yes, this is a tradeoff of more binaries — which are quite slow given link times — vs bigger compilation units (discussed in https://matklad.github.io/2021/02/27/delete-cargo-integration-tests.html)

On balance, I think this probably makes sense. While often we'll be updating `prql-compiler` together with the tests — and that will take slightly longer — it is much faster to only update the tests now.

(I also think it's arguably slightly better conceptually organized...)

aljazerzen:
You've linked to this post a few times and I don't think this PR goes against its advice - it is removing unit tests and converting them to a single large integration test.

max-sixty:
> You've linked to this post a few times and I don't think this PR goes against its advice - it is removing unit tests and converting them to a single large integration test.

Agree!

PR #3718, opened 2023-10-20
web: Button without inline svg

vanillajonathan:
This is a follow up to #3710 that uses an icon from the Box Icons font instead of an inline SVG element.


max-sixty:
Nice, thanks!

PR #3719, opened 2023-10-20
refactor: resolver

aljazerzen:
- move decl into ir
- split resolver into separate files
- cleanup imports
- context -> root_mod


max-sixty (approved):
Looks good! 

I took a look through. One question — not important — is: would having `span_map` on `Module`, and then having `root_module` just be a `Module` make that more general? I'm not sure whether finding `main_relation` is then easier — the `root_module` can have its `main`, as can sub-modules. (I would not expect this would be worth changing, I'm mostly asking to expand my understanding of this part of the code)

aljazerzen:
It only makes sense for `span_map` to be global. It is used to lookup a span of node quickly.

I don't see why finding main would be easier? If it were, why could it not work on current setup? The complications of resolving `main_path` stem from the fact that user might be referring to `blah.inner_module.main` variable or just `blah.inner_module`.

max-sixty:
Yes OK, that makes sense. 

I think "finding main would be easier" was actually a bad way of expressing what I meant — I more meant that it could be more general — an inner module could have its own `main` (which we then expose as itself), and that could percolate all the way up.

But what you say makes sense, thanks

aljazerzen:
I don't understand what you mean. This inner module (what is now `Module`, right?) would have a field `main`?

max-sixty:
Sorry — I don't think my questions are that informed. I'm trying to get more acquainted with the parts of the code I don't know well — but asking mediocre questions isn't a great way of doing that. Let me reflect a bit on how I'm approaching this.

PR #3720, opened 2023-10-20
feat(Lezer grammar): Add bin, oct and hex notation

vanillajonathan:
This adds support for binary, octal and hexadecimal notation along with tests.

This also allows plus and minus signs after the `e` in scientific notation.


max-sixty:
Very nice!

PR #3722, opened 2023-10-21
test: combine a few more tests

aljazerzen:
Move most of the parser and ast tests into the prql-compiler test binary.

Add `lib.test = false` into a few `Cargo.toml`, which will skip the unit tests by default, but will still run them when calling with `--all-targets` flag.


max-sixty:
I realize some of the diff here was in https://github.com/PRQL/prql/pull/3723. I'd added comments there and won't move them over, even though they're about the tests.

aljazerzen:
>   >    which will skip the unit tests by default, but will still run them when calling with --all-targets flag.
> How do we think about skipping tests by default?
> 
> I had thought that having all "standard-ish" tests run with cargo test was an important convention. I can imagine being a bit confused (and maybe a bit frustrated if I were a newer contributor and didn't know about it, or I forgot) that something is failing but cargo test is all passing. If we want to run fewer tests for speed, could we have a (just / task / xtask / etc) command that runs those?
> 
> ...though I recognize we do skip --test-external-dbs by default, without --all-features, but they are much much more expensive.

That's true, having `cargo test` run all the "standard" tests would be a nice default.

But also, using features to disable tests, makes it compile all dependencies, compile the test, produce and run the binary and then execute 0 tests.

So I would suggest we rather have the defaults run the "important" tests and use `cargo test --all-targets` to run the full test suite.

max-sixty:
> But also, using features to disable tests, makes it compile all dependencies, compile the test, produce and run the binary and then execute 0 tests.

The only tests we disabled with features are the ones that are very expensive to run — `duckdb` and external DBs. And we don't compile any dependencies for these.

I still don't get the logic for excluding some tests from `cargo test`?

aljazerzen:
The idea is that running `cargo test` runs only the important tests, with one or two binaries instead of running 6 different binaries where 4 of them contain <15 tests.

aljazerzen:
Oh, you are right, we don't compile dependencies for integration tests if `test-dbs` feature is not set. Ok, it's much better then.

aljazerzen:
But the `integration` binary is still being produced.

aljazerzen:
In any case, I don't have a strong opinion. These test reorganization commits were something I thought was a good idea and since I had the freedom of working on a fork, I could do it without the lengthily discussion about its benefits and shortcomings. And now I'm paying the "discussion debt" :'(

max-sixty:
> In any case, I don't have a strong opinion. These test reorganization commits were something I thought was a good idea and since I had the freedom of working on a fork, I could do it without the lengthily discussion about its benefits and shortcomings. And now I'm paying the "discussion debt" :'(

I agreed with the test reorg (I did say that!).

I disagree with ignoring a bunch of tests by default. 

I'm really trying not to slow you down (e.g. I said my discussion in https://github.com/PRQL/prql/pull/3719 wasn't productive). But we do somehow have to agree on things if multiple people are going to work on the repo. I really would hope that discussions can lead to better decisions.

What would you want me to do here, for example? I don't know of any other repo that excludes tests like this. If you're averse to discussion (and you generally seem don't seem that way! but maybe I'm misinterpreting), then why choose the radical default vs. the standard one? It's possible to write a short command to exclude the tests when you want that (I've already set a similar one up in taskfile).

aljazerzen:
I would not say that running only the fast tests by default is the "radical" default. It may not be standard (well, I'm not even sure about this), but it also makes quite a lot of sense.

The driving factor to make the change was that I wasn't able to figure out how to compile and run only some tests, but not the others, in a single cargo command, so they are compiled in parallel. The problem is that if I use `--lib`, this will compile unittests for all libraries, while I'd want only `prql-compiler`'s, and not `prqlc-parser`'s.

---

I'm not averse to discussion, I do know that there will be push-back on changing the defaults like this and only way to get it in main, would be a discussion with a resolution that we are all at least somewhat satisfied with.

What I'm saying is that these are commits that were not created under such scrutiny.

As I see, the change of which tests are run by default will not find a consensus, so it is easier for me to revert it.

max-sixty:
I see, thanks. I think I intially misinterpreted your comment re requiring discussion as a complaint rather than an observation.

I can try tomorrow to find a command that excludes those. Strongly agree re the importance of fast dev loops...

PR #3723, opened 2023-10-21
refactor: Ty::new

aljazerzen:


prqlc/prqlc-parser/Cargo.toml
```diff
@@ -10,6 +10,7 @@ version.workspace = true
 
 [lib]
 doctest = false
+test = false
```
max-sixty, line 13: We lost some tests here — https://github.com/PRQL/prql/blob/a9d81a429b6a2c940e94fcf177c402842c05f25b/prqlc/prqlc-parser/src/lexer.rs#L507

Should we move them or enable them here?

When working on a sub-project, it can be nice to have the tests loop be extremely fast, because we only need to compile that crate (and plausibly have the tests next to the code?). OTOH it adds to the number of binaries we need to create to run all the tests (I won't link to the blog post this time :) )

Cargo.toml
```diff
@@ -6,7 +6,6 @@ members = [
   "prqlc/prql-compiler-macros",
   "prqlc/prqlc",
   "prqlc/prql-compiler/examples/compile-files", # An example
-  "prqlc/tests_misc",
```
max-sixty, line 9: I wasn't confident enough that I was correct to do this myself, but I do agree collapsing this seems good...

prqlc/prqlc/Cargo.toml
```diff
@@ -8,6 +8,10 @@ repository.workspace = true
 rust-version = "1.70.0"
 version.workspace = true
 
+[[bin]]
+name = "prqlc"
+test = false
```
max-sixty, line 13: Do we lose some tests here too? https://github.com/PRQL/prql/blob/8a230fc0f1d2a9c2f88682ab447d6ed28fecaea0/prqlc/prqlc/src/cli.rs#L505

max-sixty:
> which will skip the unit tests by default, but will still run them when calling with --all-targets flag.

How do we think about skipping tests by default? 

I had thought that having all "standard-ish" tests run with `cargo test` was an important convention. I can imagine being a bit confused (and maybe a bit frustrated if I were a newer contributor and didn't know about it, or I forgot) that something is failing but `cargo test` is all passing. If we want to run fewer tests for speed, could we have a (`just` / `task` / `xtask` / etc) command that runs those?

...though I recognize we do skip `--test-external-dbs` by default, without `--all-features`, but they are much much more expensive.

prqlc/prqlc-parser/Cargo.toml
aljazerzen: These tests are still run with `cargo test --lib` or `cargo test --all-targets`.

I've left them here, because they cannot be tested via public API of prqlc-parser crate (well, it could be done probably, but they would have to be rewritten a bit).

prqlc/prqlc/Cargo.toml
aljazerzen: We do yes, similar to prqlc-parser

prqlc/prqlc-parser/Cargo.toml
max-sixty: > I've left them here, because they cannot be tested via public API of prqlc-parser crate (well, it could be done probably, but they would have to be rewritten a bit).

What's the principle here? Are we excluding internal-only tests by default?

PR #3724, opened 2023-10-21
feat(Lezer grammar): Add escape sequence

vanillajonathan:
This adds support for escape sequences though a new token called `Escape` which we map to the Lezer highlighting tag [`escape`](https://lezer.codemirror.net/docs/ref/#highlight.tags.escape). A test is included.

This also does away with the any-char-but-hamburger workaround.

### DIFF
--- a/grammars/prql-lezer/src/prql.grammar
+++ b/grammars/prql-lezer/src/prql.grammar
@@ -124,7 +124,7 @@ LambdaParam { identPart TypeDefinition? (":" expression)? }
   identPart { identifierChar (identifierChar | "_" | @digit )* }
 
   hex { @digit | $[a-fA-F] }
-  
+
   Integer {
     @digit ( @digit | "_" )* ("e" ("+" | "-")? Integer)? |
     "0x" (hex | "_")+ |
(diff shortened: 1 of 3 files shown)


max-sixty:
This looks good! I'll merge. I left a question for the future.

grammars/prql-lezer/src/prql.grammar
```diff
@@ -127,6 +137,14 @@ LambdaParam { identPart TypeDefinition? (":" expression)? }
   // underscores and allows no digit before the decimal point.
   space { " "+ }
 
+  Escape {
+    "\\" ("x" hex hex | "u" "{" hex+ "}" | ![xu])
```
max-sixty, line 141: How would we handle `\n` at the moment?

vanillajonathan: `\n` is given the `Escape` token and highlighted because `![xu]` matches any character that is not `x` or `u`. Unfortunately this includes `z` and others. This is the way the upstream JavaScript grammar does this though. Although on a second thought, I could replace that with `$[bfnrt]`.

max-sixty: > `\n` is given the `Escape` token and highlighted because `![xu]` matches any character that is not `x` or `u`.

Ah yes. Great!

PR #3730, opened 2023-10-22
feat(Lezer grammar): Add nested pipeline

vanillajonathan:
This PR adds the token `NestedPipeline` to the list of expressions in order to get the included test to pass.

It also removes a redundant set of parenthesis in the `CallExpression` token, and makes the escape sequence matching more strict.


max-sixty:
Nice — good change, great test!

PR #3731, opened 2023-10-22
tweak(ci): Remove additional npm step

max-sixty:
I'm not sure we need this now


max-sixty:
It does seem to be required. I don't quite understand, and a 40% chance that nightly tests fail tonight — I will address if so...

PR #3735, opened 2023-10-23
test: run unit tests by default for all crates

aljazerzen:
Ref #3722


max-sixty:
Thanks!! I added #3736 as an equivalent!

Issue #3474, opened 2023-09-05
Module System Redux


snth:
### What's up?

Hi,

As mentioned on the last dev-call, I have some ideas for the module system that I would like to discuss. The discussion in #3316 also illustrates that there may be a broader need for this.

## tl;dr

PRQL, and hence the prqlc compiler, is different from languages like Rust, Go, Zig, C, C++, ... in that while compiled, it compiles to SQL rather than some bytecode.

  - PRQL often need to operate within environments without a file system, e.g. in DuckDB, ClickHouse, Python, R, DBeaver, ...
  - prqlc usually doesn't have to compile large systems or applications (like systems languages like Rust, Zig, C, or C++ might have to), but rather just single analytical queries. This implies that execution speed and parallelizability are unlikely to be as big of a concern as for those systems languages.
  - Therefore different tradeoffs for the module system are likely to be appropriate.

---
Disclaimer: This is a first draft and I may edit this later.

## Background

Triggered by a recent [HackerNews post](https://news.ycombinator.com/item?id=37265097) on [Why htmx Does Not Have a Build Step](https://htmx.org/essays/no-build-step/) I reflected on this again. (Disclaimer: I'm a big fan of htmx and have found their thinking and approach great and very clear headed in what I've seen.) The top comment from the HN thread also resonated with me:

> I liked Julia Evans' take on this: [https://jvns.ca/blog/2023/02/16/writing-javascript-without-a...](https://jvns.ca/blog/2023/02/16/writing-javascript-without-a-build-system/)
>""" My goal is that if I have a site that I made 3 or 5 years ago, I’d like to be able to, in 20 minutes:
>
> - get the source from github on a new computer
>
> - make some changes
>
> - put it on the internet """
> 
> Front-end build scripts make sense for projects that are constantly developed by a team that is actively ready to incrementally  fix any problems that come up.
> 
> For sites that only get maintained every year or so (and I have plenty of those myself) my experience is that they often cause more problems than they solve.
> 
> I also found that I started really enjoying JavaScript development again once I gave myself permission to ignore the npm/webpack/etc script world entirely and just write regular code that runs in browsers!

[...]

PR #3739, opened 2023-10-24
feat: cli commands `collect` and `expand-pl`

### LINKED ISSUE #3316: Nesting variables
### What's up?

The way we generate our PRQL it would help a lot if we could use `let` inside another `let`.

So far it has been a minor inconvenience but today I had to resort to using `from` with a 70-line s-string to get around it.

Would've been much easier with PRQL :)

aljazerzen:
Add `collect` commands to the CLI that does the amalgamation described in #3474.

Removes the check for module declarations, which closes #3316

Improves the debug output of resolving and adds an intermediate step `expand-pl`, which shows de-sugared PL which is passed to the resolver.


max-sixty:
`collect` is great, thanks a lot!

PR #3742, opened 2023-10-24
feat(Lezer grammar): Allow tabs

vanillajonathan:
This allows the use tabs instead of spaces to separate stuff.

It also highlights the data type when using type annotations and adds a test for type annotations when using lambdas. This is done by mapping our `TypeTerm` token to the Lezer highlighting tag [`typeName`](https://lezer.codemirror.net/docs/ref/#highlight.tags.typeName).


max-sixty:
Yes, great on both!

PR #3746, opened 2023-10-26
feat(Lezer grammar): Highlight functions

vanillajonathan:
This PR introduces a token called `ArgList`. Such a token is common in upstream grammars (such as C++, JavaScript, Python, PHP and Rust). Those languages use it together with `(`, `,` and `)` which we don't have, but its nice have token to represent the semantics of arguments anyway. Upstream grammars don't have any separate token for the function names so I followed and did the same. The highlight for functions is applied by "CallExpression/Identifier" which highlights all the `Identifier` tokens within a `CallExpression`, but since we now have a `ArgList` token to group the other `Identifier` tokens just the immediate `Identifier` token gets highlighted.

This PR also changes `pipe+` to `pipe` so that one and only one pipe character (`|`) can be used. This prevents `from foo ||||| select bar`.

This PR also changes the formatting to be more similar to that of the upstream grammars.

### DIFF
--- a/grammars/prql-lezer/test/misc.txt
+++ b/grammars/prql-lezer/test/misc.txt
@@ -101,7 +101,7 @@ from foo | select bar
 
 ==>
 
-Query(Statements(PipelineStatement(Pipeline(CallExpression(Identifier,Identifier),CallExpression(Identifier,Identifier)))))
+Query(Statements(PipelineStatement(Pipeline(CallExpression(Identifier,ArgList(Identifier)),CallExpression(Identifier,ArgList(Identifier))))))
 
 # Derive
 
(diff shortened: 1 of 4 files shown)


grammars/prql-lezer/test/misc.txt
```diff
@@ -124,7 +124,7 @@ group customer_id (
 
 ==>
 
-Query(Statements(PipelineStatement(Pipeline(CallExpression(Identifier,Identifier,NestedPipeline(Pipeline(CallExpression(Identifier,TupleExpression(CallExpression(Identifier,Identifier))))))))))
+Query(Statements(PipelineStatement(Pipeline(CallExpression(Identifier,ArgList(Identifier,NestedPipeline(Pipeline(CallExpression(Identifier,ArgList(TupleExpression(CallExpression(Identifier,ArgList(Identifier)))))))))))))
```
max-sixty, line 127: Potentially one thing we could do is to collapse some of these; e.g. maybe `PipelineStatement` & `Pipeline` are always together and one can be silent. (But no real cost, not high priority)

max-sixty (approved):
Looks good!

PR #3751, opened 2023-10-27
web: Fix vite base

max-sixty:
This doesn't fix all the problems when running locally, but it's possible that's the Hugo server. This seems correct regardless, so will attempt this

### DIFF
--- a/web/playground/package-lock.json
+++ b/web/playground/package-lock.json
@@ -34,6 +34,7 @@
       "license": "Apache-2.0",
       "devDependencies": {
         "chai": "^4.3.6",
+        "cross-env": "^7.0.3",
         "mocha": "^10.0.0"
       }
     },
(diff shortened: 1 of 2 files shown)


max-sixty:
This works on GitHub pages. It actually doesn't work by running:
- `task build-web`
- `cd web/website; hugo serve`

...I think possibly because we need to configure MIME types in Hugo?!

vanillajonathan:
https://vitejs.dev/guide/build.html#public-base-path
https://vitejs.dev/config/shared-options.html#base

I saw you set it to `/playground/playground`. Shouldn't this be `/playground/`?

max-sixty:
> I saw you set it to `/playground/playground`. Shouldn't this be `/playground/`?

We actually host the playground app at https://prql-lang.org/playground/playground/, and then the title bar shows at https://prql-lang.org/playground (open to other ways of doing it!)

PR #3755, opened 2023-10-27
refactor(Lezer grammar): Trim the tree

vanillajonathan:
This PR trims the tree to do away with the redundant tokens `Statement` and `PipelineStatement`. We do this by just changing `Statement` to `statement` and `PipelineStatement` to `pipelineStatement` which makes them internal to the grammar and does not expose them in the tree. This is a quick and safe way to fix it for now, but if we want we could rework the grammar to do away with the internal `pipelineStatement` token altogether.

The upstream JavaScript grammar also have a internal token named `statement`.

### DIFF
--- a/grammars/prql-lezer/src/prql.grammar
+++ b/grammars/prql-lezer/src/prql.grammar
@@ -11,15 +11,15 @@
   or @left
 }
 
-@top Query { Statements }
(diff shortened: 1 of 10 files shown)


max-sixty:
Great, thanks @vanillajonathan !

PR #3756, opened 2023-10-29
fix(Lezer grammar): Highlight CompareOp

vanillajonathan:
This PR changes the tests to use `CallExpression` because we don't want expressions at the root. The grammar still allows it tho, but we should fix that.

Changing the tests resulted in that the operator tests breaking which led me to change `expresssion` inside the `CallExpression` token to `test` which made the test pass, and which also made the comparison operators properly highlight in CodeMirror when used in the context of a call expression.

This PR also adds a test for two subsequent comments which is something that produced a weird tree before we got rid of the `Statement` and `PipelineStatement` tokens.

### DIFF
--- a/grammars/prql-lezer/test/arithmetics.txt
+++ b/grammars/prql-lezer/test/arithmetics.txt
@@ -1,39 +1,39 @@
 # Plus
 
-10 + 10.5
+test 10 + 10.5
 
 ==>
 
-Query(Pipeline(BinaryExpression(Integer,ArithOp,Float)))
+Query(Pipeline(CallExpression(Identifier,ArgList(BinaryExpression(Integer,ArithOp,Float)))))
 
 # Minus
 
-10 - 10
+test 10 - 10
 
 ==>
 
-Query(Pipeline(BinaryExpression(Integer,ArithOp,Integer)))
+Query(Pipeline(CallExpression(Identifier,ArgList(BinaryExpression(Integer,ArithOp,Integer)))))
 
 # Multiply
 
-10 * 10
+test 10 * 10
 
 ==>
 
-Query(Pipeline(BinaryExpression(Integer,ArithOp,Integer)))
+Query(Pipeline(CallExpression(Identifier,ArgList(BinaryExpression(Integer,ArithOp,Integer)))))
 
 # Divide
 
-10 / 10
+test 10 / 10
 
 ==>
 
-Query(Pipeline(BinaryExpression(Integer,ArithOp,Integer)))
+Query(Pipeline(CallExpression(Identifier,ArgList(BinaryExpression(Integer,ArithOp,Integer)))))
 
 # Multiple ops
 
-10 + 10 + 10
+test 10 + 10 + 10
 
 ==>
 
-Query(Pipeline(BinaryExpression(BinaryExpression(Integer,ArithOp,Integer),ArithOp,Integer)))
+Query(Pipeline(CallExpression(Identifier,ArgList(BinaryExpression(BinaryExpression(Integer,ArithOp,Integer),ArithOp,Integer)))))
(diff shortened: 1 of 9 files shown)


grammars/prql-lezer/test/arithmetics.txt
```diff
@@ -1,39 +1,39 @@
 # Plus
 
-10 + 10.5
+test 10 + 10.5
```
max-sixty: Are all these `test`s supposed to be here? Sorry if I'm missing something...

vanillajonathan: Yes, they are suppose to be there. They move the expression from the root into a expression within a call expression where "test" is the name of a hypothetical transform function and where `10 + 10.5` is the argument to that function. `test` could be anything, such as `foo` or `filter`.

max-sixty: I see. Because we no longer allow a full query with just `10 + 10.5`, and these tests parse a full query?

Can we do a search / replace for `test` -> `filter`? `test` might be a bit confusing; it seems like it might be specifically related to the word `test` given it's a test.

It's up to you, but you might want to think about either parsing lots of small expressions (i.e. the existing code), or fewer bigger full queries. Because it's more difficult to manage if we have lots of small queries — we have all the overhead of `Query(Pipeline(CallExpression(Identifier,ArgList(` just to see what `2+2` gives us. We also have queries that aren't actually valid — there's no `from foo` — but are bigger than simple expressions.

But it's up to you, no pressure from me; I hope that's reasonable feedback.

vanillajonathan: Well the PRQL compiler doesn't allow a full query with just `10 + 10.5` but the grammar does, but I would like change the grammar so that it does not.

Yes, we can change `test` to `filter`. The grammar doesn't care.
I used `test` since it "tests the expression". I could use `filter` but technically `filter 10 + 10.5` isn't a valid PRQL query, not that the grammar cares.

The query tree in the tests do get bigger, but that is the way it have to be if we want the Lezer grammar to be more strict and adhere to what the PRQL compiler accepts. We could let the grammar allow expressions at root even though the PRQL compiler does not if we wanted though.

Small queries have been great for iteratively improving the Lezer grammar while it was rather primitive but the grammar have now reached a more mature state where it makes sense to add bigger queries too which I will do.

vanillajonathan: @max-sixty I am awaiting the merge

vanillajonathan: @max-sixty ping!

vanillajonathan: @max-sixty I am looking to get this merged

max-sixty: Sorry! I appreciate the patience.

max-sixty:
I'll merge because I trust your direction @vanillajonathan 

But more broadly — I would caution against moving away from tests which can test small individual examples, such as expressions, since it means that even small tests need lots of context, and the output of something very small is quite large ("`Query(Pipeline(CallExpression(Identifier,ArgList(BinaryExpression(Integer,ArithOp,Float)))))`")

If you prefer to have tests which test a full query, an alternative to many small tests is a small number of larger tests, which each test as much as possible ("Write tests, not too many, mostly integration"). For example, a few big queries that have lots of examples in them. In that approach, when we make a change to our code, we can add a expression to a large test. That's often less maintenance when the code we're testing doesn't have as many layers that we want to individually verify.

vanillajonathan:
I like the small tests, they're a great way to progressively hack at the Lezer grammar until it falls into places, and they're good for catching regressions.

I intend to keep the small tests, but here they got a bit bigger because I wanted to the Lezer grammar (which is context sensitive) to reflect real PRQL.

Big tests have previously not been feasible because the grammar wasn't yet mature and exhaustive enough. It's getting there though.

Soon a larger test will be added but small tests are good for pinpointing regressions which can be tricky to pinpoint in larger tests.

max-sixty:
Great, thanks @vanillajonathan

PR #3765, opened 2023-10-31
docs: Update README.md for November 2023

richb-hanover:
Good stuff appeared in 0.10.0 - we need to talk about it.


max-sixty:
Awesome — thanks so much @richb-hanover !

PR #3774, opened 2023-11-08
fix: make SQL rel vars unique within a query

### LINKED ISSUE #3713: Generated SQL contains relation var name clashes
### What happened?

No error was emitted, but incorrect sql code was generated

### PRQL input

```elm
from t
join t (==x)
```

### SQL output

```SQL
SELECT
  t.*,
  t.*
FROM
  t
  JOIN t ON t.x = t.x

```

### Expected SQL output

```SQL
-- Maybe something like:
SELECT
  t.*,
  _t_alias.*
FROM
  t
  JOIN t AS _t_alias ON t.x = _t_alias.x
-- but that wouldn't be very useful since one could not use _t_alias in the rest of the query
```

otherwise, prql could simple throw an error

### MVCE confirmation

- [X] Minimal example
- [X] New issue

### Anything else?

_No response_

aljazerzen:
Mostly fixes #3713

### DIFF
--- a/prqlc/prql-compiler/src/sql/gen_query.rs
+++ b/prqlc/prql-compiler/src/sql/gen_query.rs
@@ -276,7 +276,7 @@ fn translate_relation_expr(relation_expr: RelationExpr, ctx: &mut Context) -> Re
 
     Ok(match relation_expr.kind {
         RelationExprKind::Ref(tid) => {
-            let decl = ctx.anchor.table_decls.get_mut(&tid).unwrap();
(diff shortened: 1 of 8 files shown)


max-sixty (approved):
Awesome, thanks a lot!

(test failures look like it's just updating the doc snapshots...)

aljazerzen:
No, the generated SQL for loop was actually incorrect. Loop was done a bit sloppily, so when I added code that was mutating names after they had been assigned, it stopped working.

It's nice that we have quite a lot of tests that catch stuff like this.

PR #3775, opened 2023-11-08
fix: lineage input name clashes

### LINKED ISSUE #3713: Generated SQL contains relation var name clashes

aljazerzen:
Closes #3713

### DIFF
--- a/prqlc/prql-compiler/src/ir/pl/lineage.rs
+++ b/prqlc/prql-compiler/src/ir/pl/lineage.rs
@@ -46,7 +46,7 @@ pub enum LineageColumn {
 
     /// All columns (including unknown ones) from an input (i.e. `foo_table.*`)
     All {
-        input_name: String,
+        input_id: usize,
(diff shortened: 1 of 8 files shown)


max-sixty (approved):
Nice, thanks!

PR #3776, opened 2023-11-08
fix: handle panic of relation literals containing exprs

### LINKED ISSUE #3664: functions in array literals cause panic
This works (Awesome!)

```elm
from [{version=prql_version}]
```

This causes panic.

```elm
from [{a=(1+1)}]
```

Error message from the edge version prqlc:

```log
Message:  called `Result::unwrap()` on an `Err` value: RqOperator { name: "std.add", args: [Expr { kind: Literal(Integer(1)), span: Some(0:10-11), alias: None, id: Some(274), target_id: None, target_ids: [], ty: Some(Ty { kind: Primitive(Int), name: None }), lineage: None, needs_window: false, flatten: false }, Expr { kind: Literal(Integer(1)), span: Some(0:12-13), alias: None, id: Some(275), target_id: None, target_ids: [], ty: Some(Ty { kind: Primitive(Int), name: None }), lineage: None, needs_window: false, flatten: false }] }
Location: crates\prql-compiler\src\semantic\lowering.rs:342

Backtrace omitted. Run with RUST_BACKTRACE=1 environment variable to display it.
Run with RUST_BACKTRACE=full to include source snippets.
```

It would be great if we got a kind error message.

aljazerzen:
Closes #3664

### DIFF
--- a/prqlc/prql-compiler/src/semantic/lowering.rs
+++ b/prqlc/prql-compiler/src/semantic/lowering.rs
@@ -338,10 +338,16 @@ impl Lowerer {
                                 .into_tuple()
                                 .unwrap()
                                 .into_iter()
-                                .map(|element| element.kind.into_literal().unwrap())
-                                .collect()
+                                .map(|element| {
+                                    element.try_cast(
+                                        |x| x.into_literal(),
+                                        Some("relation literal"),
+                                        "literals",
+                                    )
+                                })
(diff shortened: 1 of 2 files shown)


max-sixty (approved):
Super, thanks a lot!

(and for the recent burst, I'll hopefully be back on it next week too...)

PR #3780, opened 2023-11-09
fix: coalesce and window functions

### LINKED ISSUE #3587: Window function on aliased column produces invalid SQL
### What happened?

In the following, the `COALESCE(SUM(a), 0) OVER (...)` is incorrect and should probably be `COALESCE(SUM(a) OVER (...), 0)`.

It's also curious why the `COALESCE` is only added to the aliased version `cumsum_b` and not for `cumsum_a`?

### PRQL input

```elm
[{a=1}, {a=2}]
select {a, b=a}
window rows:..0 (
  derive {
    cumsum_a=(sum a),
    cumsum_b=(sum b)
  }
)
```

### SQL output

```SQL
WITH table_0 AS (
  SELECT
    1 AS a
  UNION
  ALL
  SELECT
    2 AS a
)
SELECT
  a,
  a AS b,
  SUM(a) OVER (ROWS BETWEEN UNBOUNDED PRECEDING AND CURRENT ROW) AS cumsum_a,
  COALESCE(SUM(a), 0) OVER (ROWS BETWEEN UNBOUNDED PRECEDING AND CURRENT ROW) AS cumsum_b
FROM
  table_0

```

### Expected SQL output

```SQL
WITH table_0 AS (
  SELECT
    1 AS a
  UNION
  ALL
  SELECT
    2 AS a
)
SELECT
  a,
  a AS b,
  SUM(a) OVER (ROWS BETWEEN UNBOUNDED PRECEDING AND CURRENT ROW) AS cumsum_a,
  COALESCE(SUM(a) OVER (ROWS BETWEEN UNBOUNDED PRECEDING AND CURRENT ROW), 0) AS cumsum_b
FROM
  table_0

```

### MVCE confirmation

- [X] Minimal example
- [X] New issue

### Anything else?

We previously had a discussion about whether these "identity" elements should be added as defaults for the aggregation functions. I don't know if I felt strongly about it at the time but I think I would probably prefer to not have it and rathe be able to add these myself if I wanted. With the `??` operator it's easy enough. For example here it could easily be done for `cumsum_a` with `cumsum_a=(sum a) ?? 0` which produces valid SQL.

aljazerzen:
Closes #3587

### DIFF
--- a/prqlc/prql-compiler/src/sql/gen_expr.rs
+++ b/prqlc/prql-compiler/src/sql/gen_expr.rs
@@ -395,9 +395,10 @@ pub(super) fn translate_cid(cid: CId, ctx: &mut Context) -> Result<ExprOrSource>
                 let window = compute.window.clone();
                 let span = compute.expr.span;
 
+                let prev_wf = ctx.query.window_function;
                 ctx.query.window_function = window.is_some();
                 let expr = translate_expr(compute.expr.clone(), ctx)?;
-                ctx.query.window_function = false;
+                ctx.query.window_function = prev_wf;
 
                 if let Some(window) = window {
                     translate_windowed(expr, window, ctx, span)?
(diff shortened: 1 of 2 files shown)


max-sixty:
Thanks a lot for doing all these! I'm sure they're more work than the LOC suggest...

max-sixty (approved):

PR #3786, opened 2023-11-10
feat: parse types separately

aljazerzen:
Before this PR, we type parsing was very minimalistic and relied
on expr parsing and then coersing the value expr into a type expression.
While this is quite elegant, it had its own challanges, for example:
- union was expressed with || operator, which was desugared into std.or
  function call. This meant that to resolve a union type, we had to do a
  lookup for std.or. Which had to be resolved *before* any other types.
- it is impossible to separate lexical analysis from the semantic resolver,
  since type inference relies on semantic data,
- a few resolution code chunks were needlessly complex because of this
  (i.e. custom code for disabling type checking for type expressions),
- pl::ExprKind had to support expressing a set (of integers for example),
  which could be later coerced into a type.

So now, types are parsed separately, have separate AST nodes and live in
their own namespace.


eitsupi:
It seems that this change causes stack overflow on Windows (#2857) again?
https://github.com/PRQL/prql/actions/runs/6840454202/job/18599730371
```log
thread 'long_query' panicked at 'snapshot assertion for 'long_query' failed in line 134', C:\Users\runneradmin\.cargo\registry\src\index.crates.io-6f17d22bba15001f\insta-1.34.0\src\runtime.rs:563:9
```

Issue #3111, opened 2023-07-26
using `select` after `group col (take n)` causes a crash/panic for certain targets (have `DISTINCT ON`)


cornservant:
### What happened?

The compiler panics when compiling the PRQL input given below.

I am using `prqlc 0.9.2` on `Manjaro Linux x86_64`.

### PRQL input

```elm
prql target:sql.postgres
# prql target:sql.clickhouse
# prql target:sql.duckdb

from a
join b (b.a_id == a.id)
group {a.id} (
  sort b.x
  take 1
)
select {a.id, b.y}
```

### SQL output

```SQL
--
```

### MVCE confirmation

- [X] Minimal example
- [X] New issue

### Anything else?

The compiler does not crash when

- adding the column `b.x` to the final select statement, OR
- changing the target language to anything other than `sql.postgres`, `sql.clickhouse`, `sql.duckdb`

Playground-friendly version of the offending input:

```prql
prql target:sql.postgres
# prql target:sql.clickhouse
# prql target:sql.duckdb

from customers
join invoices (invoices.customer_id == customers.customer_id)
group {customers.customer_id} (
  sort {- invoices.total}
  take 1
)
sort { customers.customer_id }
select { customers.customer_id, invoices.billing_country }
```

stderr output with `RUST_LOG=trace`: https://gist.github.com/exit91/410a49f7d7b8d7ccf14b3dbf781e7bec

max-sixty:
Thanks for the excellent bug report @exit91

eitsupi:
Adding `select` after the #2711 example does not seem to compile correctly.

Like:

```elm
prql target:sql.duckdb

from employees
group city (
  take 1
)
select last_name
```

Expected:

```elm
WITH table_0 AS (
  SELECT
    DISTINCT ON (city) *
  FROM
    employees
  ORDER BY
    city
)
SELECT
  last_name
FROM
  table_0
```

max-sixty:
FYI slightly smaller example for the panic:

```elm
    prql target:sql.postgres

    from a
    group {a.id} (
      sort a.x
      take 1
    )
    select {a.id}
```

max-sixty:
This cropped up again in #3460 

@aljazerzen would you have any direction on this?

mklopets:
FWIW, this is blocking our upgrade from 0.8.1, meaning we've got an almost 4.5 month-old version running in prod

eitsupi:
[...]

Issue #2182, opened 2023-03-14
Implement `DISTINCT ON` for DuckDB


max-sixty:
### What's up?

We should translate to `DISTINCT ON` for DuckDB & Postgres where we want to `take 1` within a group, but can't use `DISTINCT`. It's much higher perf in DuckDB, and slightly higher in Postgres:

https://github.com/duckdb/duckdb/issues/6709

max-sixty:
I have a partial branch, but I think it's waiting on https://github.com/sqlparser-rs/sqlparser-rs/issues/806. 

It's possible we could fool the parser (which, tbc, we use as a serializer, it's great!) into thinking `ON` were a function, but to the extent that's forthcoming, using a native version would be ideal...

eitsupi:
sqlparser is now updated...

max-sixty:
Would be closed by https://github.com/PRQL/prql/pull/2244

eitsupi:
Closed by #2244

eitsupi:
Reopened by #3788
Blocker: #3111

PR #3788, opened 2023-11-11
fix: disable compiling to `distinct on`

eitsupi:
A workaround for #3111, reopen #2182

This feature is really great but unfortunately needs to be disabled at this time as it causes a fatal bug. (Ideally, the user should be able to enable it, but unfortunately I can't think of an easy way to do that other than Rust's feature flag, so I think disabling it is easier for now.)


max-sixty (approved):
@eitsupi thanks a lot for this. 

Sorry I was dragging — I've been out for a bit and hadn't realized quite how easy it was...

PR #3792, opened 2023-11-13
fix: distinct on

### LINKED ISSUE #2182: Implement `DISTINCT ON` for DuckDB

### LINKED ISSUE #3111: using `select` after `group col (take n)` causes a crash/panic for certain targets (have `DISTINCT ON`)

aljazerzen:
Closes #3111
Closes #2182

This is the full log of what I've done:

take problematic example and paste it into _a.prql

```
prql target:sql.duckdb

from tab1
group col1 (
take 1
)
derive foo = 1
select foo
```

cargo run -p prqlc -- compile _a.prql

It works, probably because DISTINCT ON is disabled?

Search for #3111 in codebase, find it in dialects.rs, remove the workaround. 

cargo run -p prqlc -- compile _a.prql

panic, unreachable! expects all `ON` contents to be SelectItem::UnnamedExpr, what are they?

change into:

```
                .map(|item| match item {
                    SelectItem::UnnamedExpr(expr) => expr,
                    _ => unreachable!("{item:?}"),
                })
```

run again,

```
Message:  internal error: entered unreachable code: ExprWithAlias { expr: CompoundIdentifier([Ident { value: "col1", quote_style: None }]), alias: Ident { value: "_expr_0", quote_style: None } }
```

Can't we just unwrap the aliased expr?

```
                .map(|item| match item {
                    SelectItem::UnnamedExpr(expr) => expr,
                    SelectItem::ExprWithAlias { expr, .. } => expr,
                    _ => unreachable!("{item:?}"),
                })
```

Compiles to:

```
WITH table_0 AS (
  SELECT
    DISTINCT ON (col1) NULL
  FROM
    tab1
)
SELECT
  1 AS foo
FROM
  table_0
```

What's the NULL doing there?
Looking at the source of sqlparser, it is not comming from Distinct::On.

Adding a `dbg!(projection);` reveals that the projections are empty, so we inject NULL for making sure that there is at least one select items in a query.
https://github.com/prql/PRQL/blob/19c4c0bfd38d4216788d4d892b615ca6ba93cd99/prqlc/prql-compiler/src/sql/gen_projection.rs#L162-L166
So this is actaully fine, because we don't pull any columns from the distinct table.

This query now works as it should. Let's add it into the test suite.

Reading this comment:

```
// FIXME: this "works" but is very hacky — it's using the
// `translate_select_item` to translate each id into an Expr — is that
// correct?
//
// Do we know we won't have more than one DistinctOn? We have that for
// `Distinct`. But this will panic if we have more than one.
// And if there the `SelectItem` is not an `UnnamedExpr`, it'll panic.
```

1. We don't know how many DistinctOn there will be. That's because code here:
   https://github.com/prql/PRQL/blob/19c4c0bfd38d4216788d4d892b615ca6ba93cd99/prqlc/prql-compiler/src/sql/srq/anchor.rs#L347-L358
   Checks if the pipeline should split when a Distinct or DistinctOn is encountered
   when stepping the pipeline from the back.
   It checks if the following pipeline contains some transforms, but because it is not
   checking if it contains Distinct or DistinctOn, it will not split for the second such
   transform.
   For Distinct, this is ok, because applying distinct multiple times is the same a applying it once,
   and the code later on just checks if there is any number of Distinct in the pipeline.
   For DistinctOn, this is not ok, because they columns to be distinct on and if columns of the second
   pipeline are not a subset of the columns of the first, second transform is doing work that has not been
   done from the first. We also cannot have two DistinctOn in one SELECT, so we should split the pipeline
   if this happends.

   ```
        SqlTransform::DistinctOn(_) => contains_any(
            following,
            [
                "From",
                "Join",
                "Compute",
                "Filter",
                "Aggregate",
                "Sort",
                "Take",
                "DistinctOn",
            ],
        ),
        SqlTransform::Distinct => contains_any(
            following,
            [
                "From",
                "Join",
                "Compute",
                "Filter",
                "Aggregate",
                "Sort",
                "Take",
            ],
        ),
   ```
   Now, we can be sure that each atomic pipeline gets at most one DistinctOn.
   But there might be many Distinct, followed by one DistinctOn.
   
   
2. We don't need the whole `translate_select_item`.
   We could use just `translate_cid` because we don't need any aliases
   that are created by `translate_select_item`.

   Also a few things: 
   - because we are sure of the assumption that there will be only one DistinctOn,
     we change:
     ```diff 
     - distinct_ons.into_iter().exactly_one().unwrap()
     + distinct_ons.into_iter().exactly_one()?
     ```
     If this break, it is not a user error, it's a bug.
   - if `translate_select_item` fails, it is not a bug, it might be an user error:
     ```
     - .map(|id| translate_select_item(id, ctx).unwrap())
     - .collect::<Vec<_>>(),
     + .map(|id| translate_select_item(id, ctx))
     + .collect::<Result<Vec<_>>>()?,
     ```
     (I don't like handling error in iterators, this would be easier with a for loop)
   - if we use `translate_cid`, we don't have to `match` on SelectItem.

```
    Some(sql_ast::Distinct::On(
        distinct_ons
            .into_iter()
            .exactly_one()
            .unwrap()
            .into_iter()
            .map(|id| translate_cid(id, ctx).map(|x| x.into_ast()))
            .collect::<Result<Vec<_>>>()?,
    ))
```

cargo test --test=sql 

cargo insta review

just pull-request

Let's also test the original example:

cargo run -p prqlc -- compile _a.prql

```
prql target:sql.duckdb

from a
join b (b.a_id == a.id)
group {a.id} (
  sort b.x
  take 1
)
select {a.id, b.y}
```

Panic, https://github.com/prql/PRQL/blob/5a9cec37b11ce2bfd95a19e34df6cef5e9557de2/prqlc/prql-compiler/src/sql/gen_expr.rs#L439

It is not finding column name that should be generated prior to this step.
Which column is this?

`dbg!(cid);` and re-run

[prqlc/prql-compiler/src/sql/gen_expr.rs:439] cid = column-3

Ok, cid=3, but which columnis this?

cargo run -p prqlc -- resolve _a.prql

```
def:
  version: null
  other:
    target: sql.duckdb
tables:
- id: 0
  name: null
  relation:
    kind: !ExternRef
    - b
    columns:
    - !Single a_id
    - !Single x
    - !Single y
    - Wildcard
- id: 1
  name: null
  relation:
    kind: !ExternRef
    - a
    columns:
    - !Single id
    - Wildcard
relation:
  kind: !Pipeline
  - !From
    source: 1
    columns:
    - - !Single id
      - 0
    - - Wildcard
      - 1
    name: a
  - !Join
    side: Inner
    with:
      source: 0
      columns:
      - - !Single a_id
        - 2
      - - !Single x
        - 3
      - - !Single y
        - 4
      - - Wildcard
        - 5
      name: b
    filter:
      kind: !Operator
        name: std.eq
        args:
        - kind: !ColumnRef 2
          span: 0:585-591
        - kind: !ColumnRef 0
          span: 0:595-599
      span: 0:585-599
  - !Take
    range:
      start: null
      end:
        kind: !Literal
          Integer: 1
        span: null
    partition:
    - 0
    sort:
    - direction: Asc
      column: 3
  - !Select
    - 0
    - 4
  - !Select
    - 0
    - 4
  columns:
  - !Single id
  - !Single y
```

It's `x`, introduced in relational instance `riid=0` in `Join`.

If I remember correctly, we have functions that generate names for
all columns that need it. Maybe they are not looking at Take.sort.direction?

https://github.com/prql/PRQL/blob/5a9cec37b11ce2bfd95a19e34df6cef5e9557de2/prqlc/prql-compiler/src/sql/srq/gen_query.rs#L269-L279

Jup, what's probably happening is that because the Take.sort columns are not 
part of output SELECT items, we don't call `ensure_column_name` on them.
So they don't have a name set.

```
fn ensure_names(transforms: &[SqlTransform], ctx: &mut AnchorContext) {
    let empty = HashSet::new();
    for t in transforms {
        match t {
            SqlTransform::Super(Transform::Sort(_)) => {
                for r in anchor::get_requirements(t, &empty) {
                    ctx.ensure_column_name(r.col);
                }
            }
            SqlTransform::DistinctOn(columns) => {
                for cid in columns {
                    ctx.ensure_column_name(*cid);
                }
            }
            _ => (),
        }
    }
}
```

cargo run -p prqlc -- compile _a.prql

still panic, same location

is `ensure_names()` even called?

```
dbg!(transforms);
```

cargo run -p prqlc -- compile _a.prql

```
[prqlc/prql-compiler/src/sql/srq/gen_query.rs:271] transforms = [
    Super(
        Select(
            [
                column-0,
                column-4,
            ],
        ),
    ),
    From(
        RIId(
            0,
        ),
    ),
    Join {
        side: Inner,
        with: RIId(
            1,
        ),
        filter: Expr {
            kind: Operator {
                name: "std.eq",
                args: [
                    Expr {
                        kind: ColumnRef(
                            column-2,
                        ),
                        span: Some(
                            0:585-591,
                        ),
                    },
                    Expr {
                        kind: ColumnRef(
                            column-0,
                        ),
                        span: Some(
                            0:595-599,
                        ),
                    },
                ],
            },
            span: Some(
                0:585-599,
            ),
        },
    },
    Sort(
        [
            ColumnSort {
                direction: Asc,
                column: column-0,
            },
            ColumnSort {
                direction: Asc,
                column: column-3,
            },
        ],
    ),
    DistinctOn(
        [
            column-0,
        ],
    ),
]
```

Ah! The column-3, is not in `DistinctOn`, but in preceeding `Sort`.
And it is not found because we are matching on `SqlTransform::Super(Sort)`, while this
function is recieving already-compiled `SqlTransform::Sort`.

So we must match on `Sort` too. And why is this even calling `get_requirements`? It would be easier
(and eqivalent) to just loop over sort columns.

```
fn ensure_names(transforms: &[SqlTransform], ctx: &mut AnchorContext) {
    for t in transforms {
        if let SqlTransform::Super(Transform::Sort(columns)) | SqlTransform::Sort(columns) = t {
            for r in columns {
                ctx.ensure_column_name(r.column);
            }
        }
    }
}
```

cargo run -p prqlc -- compile _a.prql

just pull-request

### DIFF
--- a/prqlc/prql-compiler/src/sql/dialect.rs
+++ b/prqlc/prql-compiler/src/sql/dialect.rs
@@ -195,10 +195,7 @@ impl DialectHandler for PostgresDialect {
     }
 
     fn supports_distinct_on(&self) -> bool {
-        // https://www.postgresql.org/docs/current/sql-select.html
-        // TODO: switch to true when the bug is fixed
-        // https://github.com/PRQL/prql/issues/3111
-        false
+        true
     }
 }
 
@@ -258,10 +255,7 @@ impl DialectHandler for ClickHouseDialect {
     }
 
     fn supports_distinct_on(&self) -> bool {
-        // https://clickhouse.com/docs/en/sql-reference/statements/select/distinct
-        // TODO: switch to true when the bug is fixed
-        // https://github.com/PRQL/prql/issues/3111
-        false
+        true
     }
 }
 
@@ -304,10 +298,7 @@ impl DialectHandler for DuckDbDialect {
     }
 
     fn supports_distinct_on(&self) -> bool {
-        // https://duckdb.org/docs/sql/query_syntax/select.html#distinct-on-clause
-        // TODO: switch to true when the bug is fixed
-        // https://github.com/PRQL/prql/issues/3111
-        false
+        true
     }
 }
 
--- a/prqlc/prql-compiler/src/sql/gen_query.rs
+++ b/prqlc/prql-compiler/src/sql/gen_query.rs
@@ -119,24 +119,14 @@ fn translate_select_pipeline(
     let distinct = if is_distinct {
         Some(sql_ast::Distinct::Distinct)
     } else if !distinct_ons.is_empty() {
-        // FIXME: this "works" but is very hacky — it's using the
-        // `translate_select_item` to translate each id into an Expr — is that
-        // correct?
-        //
-        // Do we know we won't have more than one DistinctOn? We have that for
-        // `Distinct`. But this will panic if we have more than one.
-        // And if there the `SelectItem` is not an `UnnamedExpr`, it'll panic.
         Some(sql_ast::Distinct::On(
             distinct_ons
                 .into_iter()
-                .exactly_one()?
+                .exactly_one()
+                .unwrap()
                 .into_iter()
-                .map(|id| translate_select_item(id, ctx).unwrap())
-                .map(|item| match item {
-                    SelectItem::UnnamedExpr(expr) => expr,
-                    _ => unreachable!(),
-                })
-                .collect::<Vec<sql_ast::Expr>>(),
+                .map(|id| translate_cid(id, ctx).map(|x| x.into_ast()))
+                .collect::<Result<Vec<_>>>()?,
         ))
     } else {
         None
--- a/prqlc/prql-compiler/src/sql/srq/anchor.rs
+++ b/prqlc/prql-compiler/src/sql/srq/anchor.rs
@@ -344,7 +344,20 @@ fn is_split_required(transform: &SqlTransform, following: &mut HashSet<String>) 
             following,
             ["From", "Join", "Compute", "Filter", "Aggregate", "Sort"],
         ),
-        SqlTransform::Distinct | SqlTransform::DistinctOn(_) => contains_any(
+        SqlTransform::DistinctOn(_) => contains_any(
+            following,
+            [
+                "From",
+                "Join",
+                "Compute",
+                "Filter",
+                "Aggregate",
+                "Sort",
+                "Take",
+                "DistinctOn",
+            ],
+        ),
+        SqlTransform::Distinct => contains_any(
             following,
             [
                 "From",
--- a/prqlc/prql-compiler/src/sql/srq/gen_query.rs
+++ b/prqlc/prql-compiler/src/sql/srq/gen_query.rs
@@ -1,6 +1,5 @@
 //! This module is responsible for translating RQ to SRQ.
 
-use std::collections::HashSet;
 use std::str::FromStr;
 
 use anyhow::Result;
@@ -267,11 +266,10 @@ fn compile_loop(pipeline: Vec<SqlTransform>, ctx: &mut Context) -> Result<Vec<Sq
 }
 
 fn ensure_names(transforms: &[SqlTransform], ctx: &mut AnchorContext) {
-    let empty = HashSet::new();
     for t in transforms {
-        if let SqlTransform::Super(Transform::Sort(_)) = t {
-            for r in anchor::get_requirements(t, &empty) {
-                ctx.ensure_column_name(r.col);
+        if let SqlTransform::Super(Transform::Sort(columns)) | SqlTransform::Sort(columns) = t {
+            for r in columns {
+                ctx.ensure_column_name(r.column);
             }
         }
     }
--- a/prqlc/prql-compiler/tests/integration/queries/window.prql
+++ b/prqlc/prql-compiler/tests/integration/queries/window.prql
@@ -1,4 +1,7 @@
 # mssql:skip Conversion("cannot interpret I64(Some(1)) as an i32 value")', connection.rs:200:34
+# duckdb:skip problems with DISTINCT ON (duckdb internal error: [with INPUT_TYPE = int; RESULT_TYPE = unsigned char]: Assertion `min_val <= input' failed.)
+# clickhouse:skip problems with DISTINCT ON
+# postgres:skip problems with DISTINCT ON
 from tracks
 group genre_id (
   sort milliseconds
--- a/prqlc/prql-compiler/tests/sql/sql.rs
+++ b/prqlc/prql-compiler/tests/sql/sql.rs
@@ -1663,7 +1663,7 @@ fn test_distinct() {
 }
 
 #[test]
-fn test_distinct_on() {
+fn test_distinct_on_01() {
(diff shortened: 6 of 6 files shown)


eitsupi:
Great work, thanks!

It would be great if a new release was made soon after this was merged.

max-sixty:
I love this! 

I've read through several times already. It's actually super helpful and revealing to have the whole journey. I'm sure it took lots of time, I appreciate the investment...

I think I could have probably done 60% of it without spinning. There are a couple of "then I remembered" moments, or equivalent, which I wouldn't have remembered. But there are also bits like "cid=3, let's find out what that is" which show that many of the steps don't require much context.

aljazerzen:
Great to hear that my log is useful :D I tried to be explicit in "what was the question" and "where should I go look now" and it seems it works.

Yes, there are cases of "I remembered" that people can learn in a codebase only by working on it or asking.

snth:
I agree with Max that this thought-log is super useful!

I just skimmed over it now but I've bookmarked it to come back to when I get a chance to work on the compiler myself some time.

Thank you for writing this out @aljazerzen !

PR #3809, opened 2023-11-15
build: Rename `clib` binary name

### LINKED ISSUE #3800: Build output filename collision
### What's up?

I coincidentally saw this warning, which only seems to happen on windows (??) at https://github.com/PRQL/prql/actions/runs/6855982279/job/18642276364

```
warning: output filename collision.
The lib target `prqlc` in package `prqlc-clib v0.10.1 (D:\a\prql\prql\prqlc\bindings\clib)` has the same output filename as the bin target `prqlc` in package `prqlc v0.10.1 (D:\a\prql\prql\prqlc\prqlc)`.
Colliding filename is: D:\a\prql\prql\target\x86_64-pc-windows-msvc\debug\prqlc.pdb
The targets should have unique names.
Consider changing their names to be unique or compiling them separately.
This may become a hard error in the future; see <https://github.com/rust-lang/cargo/issues/6313>.
```

Should we name the `clib` output to `prqlc-lib`?

max-sixty:
Closes #3800


prqlc/bindings/clib/Cargo.toml
```diff
@@ -9,9 +9,9 @@ rust-version.workspace = true
 version.workspace = true
 
 [lib]
-name = "prqlc"
 crate_type = ["staticlib", "cdylib"]
 doctest = false
+name = "prqlc_lib"
```
eitsupi, line 14: Just a question: why `_lib`, not `_clib`?

max-sixty: No view! 

@aljazerzen what would you prefer?

aljazerzen: I guess `_lib`? I don't know...

max-sixty:
It'll auto-merge, but feel free to pause if we want to decide on the name. It's also very easy to change now we know where all the locations are!

PR #3810, opened 2023-11-15
ci: Release .deb and .rpm packages

vanillajonathan:
This adds .deb and .rpm packages for prqlc as release assets on GitHub.


.github/workflows/release.yaml
eitsupi, line 1: Could it be changed to create packages using pre-built prqlc to avoid duplicate builds? (like the `winget-release` job)

eitsupi (changes requested):

max-sixty (approved):
This looks good, but I don't know the steps that well, maybe @eitsupi would have more context here?

.github/workflows/release.yaml
```diff
@@ -79,18 +79,15 @@ jobs:
...
     steps:
-      - name: 📂 Checkout code
-        uses: actions/checkout@v4
-      - name: 👷 Build prql-compiler
-        run: cargo build --release -p prqlc
       - name: Copy files into .deb package
         run: |
           mkdir -p .debpkg/usr/bin
-          cp target/release/prqlc .debpkg/usr/bin/prqlc
+          cp target/x86_64-unknown-linux-musl/release/prqlc .debpkg/usr/bin/prqlc
```
eitsupi: Could you use matrix to make package for arm64 as well?

.github/workflows/release.yaml
```diff
@@ -105,19 +102,22 @@ jobs:
 
             PRQL is a modern language for transforming data — a simple,
             powerful, pipelined SQL replacement.
+      - name: Release
+        if: github.event_name == 'release'
+        uses: softprops/action-gh-release@v1
+        with:
+          files: prqlc_*.deb
```
eitsupi, line 138: Could you please upload these using the upload artifact action so we can confirm that it is building successfully with nightly builds and PRs?

Like:
https://github.com/PRQL/prql/blob/8ed38fbef546289c846a2820e79356f8ac91a318/.github/actions/build-prqlc/action.yaml#L89-L93

.github/workflows/release.yaml
```diff
@@ -105,19 +102,22 @@ jobs:
...
     steps:
-      - name: 📂 Checkout code
-        uses: actions/checkout@v4
-      - name: 👷 Build prqlc
-        run: cargo build --bin prqlc --release
       - name: Copy files into .rpm package
         run: |
           mkdir -p .rpmpkg/usr/bin
-          cp target/release/prqlc .rpmpkg/usr/bin/prqlc
+          cp target/x86_64-unknown-linux-musl/release/prqlc .rpmpkg/usr/bin/prqlc
```
eitsupi: Ditto.

.github/workflows/release.yaml
```diff
@@ -135,6 +135,12 @@ jobs:
             PRQL is a modern language for transforming data — a simple,
             powerful, pipelined SQL replacement.
           license: Apache-2.0
+      - run: rm prqlc-debuginfo-*.rpm
+      - name: Release
+        if: github.event_name == 'release'
+        uses: softprops/action-gh-release@v1
+        with:
+          files: prqlc-*.rpm
```
eitsupi, line 199: Ditto.

eitsupi (changes requested):

.github/workflows/release.yaml
vanillajonathan: Okay, I think I've done this now. I now upload the `rpm` and `deb` artifact.

.github/workflows/release.yaml
vanillajonathan: Okay, I think I've done this now. I now upload the `rpm` and `deb` artifact.

.github/workflows/release.yaml
```diff
@@ -79,18 +79,23 @@ jobs:
...
     steps:
-      - name: 📂 Checkout code
-        uses: actions/checkout@v4
-      - name: 👷 Build prql-compiler
-        run: cargo build --release -p prqlc
       - name: Copy files into .deb package
         run: |
           mkdir -p .debpkg/usr/bin
-          cp target/release/prqlc .debpkg/usr/bin/prqlc
+          cp target/*/release/prqlc .debpkg/usr/bin/prqlc
```
eitsupi: I think we need to use `actions/download-artifact` to download the artifact.
https://github.com/PRQL/prql/blob/8ed38fbef546289c846a2820e79356f8ac91a318/.github/actions/build-prqlc/action.yaml#L89-L97

vanillajonathan: So I tried that now, but I am struggling with this. Any more ideas? Wanna give it a try? The PR is open for edits by maintainers.

eitsupi: Look for other repositories doing similar things. For example I know bendsql does something similar with this job.
https://github.com/datafuselabs/bendsql/blob/b9033d327ab21e8fb18ead2613d848cabb3b53a3/.github/workflows/release.yml#L97-L124

.github/workflows/release.yaml
vanillajonathan: Good idea!

.github/workflows/release.yaml
```diff
@@ -79,18 +81,28 @@ jobs:
...
-    runs-on: ubuntu-latest
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - os: ubuntu-latest
+            target: x86_64-unknown-linux-musl
+          - os: ubuntu-latest
+            target: aarch64-unknown-linux-musl
```
eitsupi: Obviously only Linux is the OS here, so I think we can specify only `target` without using include.

Something like:

```suggestion
      matrix:
         target:
           - x86_64-unknown-linux-musl
           - aarch64-unknown-linux-musl
```

.github/workflows/release.yaml
```diff
@@ -105,19 +117,40 @@ jobs:
...
```
eitsupi: Ditto

```suggestion
      matrix:
         target:
           - x86_64-unknown-linux-musl
           - aarch64-unknown-linux-musl
```

.github/workflows/release.yaml
```diff
@@ -79,18 +81,28 @@ jobs:
...
+      - name: Download a Build Artifact
+        uses: actions/download-artifact@v2
+        with:
+          name: prqlc-${{ matrix.target }}
       - name: Copy files into .deb package
         run: |
+          tar -xf prqlc-*.tar.gz
           mkdir -p .debpkg/usr/bin
-          cp target/release/prqlc .debpkg/usr/bin/prqlc
+          cp prqlc .debpkg/usr/bin/prqlc
```
eitsupi: Why not use `mv` instead of `cp`?

.github/workflows/release.yaml
```diff
@@ -105,19 +117,40 @@ jobs:
...
+      - name: Download a Build Artifact
+        uses: actions/download-artifact@v2
+        with:
+          name: prqlc-${{ matrix.target }}
       - name: Copy files into .rpm package
         run: |
+          tar -xf prqlc-*.tar.gz
           mkdir -p .rpmpkg/usr/bin
-          cp target/release/prqlc .rpmpkg/usr/bin/prqlc
+          cp prqlc .rpmpkg/usr/bin/prqlc
```
eitsupi: Ditto.

eitsupi (changes requested):

eitsupi:
Looking at the artifacts, it seems that only amd64 deb packages exist...

.github/workflows/release.yaml
vanillajonathan: Okay, I'll do that.

.github/workflows/release.yaml
```diff
@@ -79,18 +81,27 @@ jobs:
...
 
   build-deb-package:
-    # TODO: currently this doesn't publish it as an artifact; would be a welcome contribution
-    # to add that (and eventually as a release asset)
-    runs-on: ubuntu-latest
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [ubuntu-latest]
```
eitsupi:
```suggestion
    runs-on: ubuntu-latest]
    strategy:
      fail-fast: false
      matrix:
```

.github/workflows/release.yaml
```diff
@@ -105,19 +116,39 @@ jobs:
...
+          files: prqlc_*.deb
 
   build-rpm-package:
-    # TODO: This doesn't publish the rpm yet, that would be a welcome follow-up (even as a CI artifact)
-    runs-on: ubuntu-latest
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [ubuntu-latest]
```
eitsupi:
```suggestion
    runs-on: ubuntu-latest
    strategy:
      fail-fast: false
      matrix:
```

eitsupi (dismissed):

max-sixty:
FYI this is looking nice! Clever use of the existing build @vanillajonathan !

vanillajonathan:
> FYI this is looking nice! Clever use of the existing build @vanillajonathan !

It was @eitsupi who suggested using the existing build.
This build fails with some codecov check, but I don't know why.
This looks good now? Can we merge?

@eitsupi This is good?

max-sixty:
> This build fails with some codecov check, but I don't know why.

This is a mistake — you can ignore it.

~I think that probably the number of lines increased which means the % of lines checked went down. Ideally we could change the codecov settings so it only considered rust code...~ It does think there are [actual coverage changes](https://app.codecov.io/gh/prql/prql/pull/3810/indirect-changes), which is surely a mistake. I'll try merging `main` to see if that helps...

eitsupi:
Please check the uploaded artifacts.
As I wrote in https://github.com/PRQL/prql/pull/3810#issuecomment-1816621205, it looks like it was not uploaded correctly. (I don't know if I can really build deb/rpm packages this way in the first place)

vanillajonathan:
@eitsupi Both `artifact-deb` and `artifact-rpm` are uploaded. I downloaded `artifact-deb`, extracted it and was able to install it on my system.

The release step was skipped because it is only ran when `if: github.event_name == 'release'`.

I think it looks good.

eitsupi:
It looks like the arm64 versions are not included in the artifact.
Probably overwritten because the name has not been changed due to architectural differences.

Also, I think that the one for arm64 is also named as amd64 because the architecture is not specified in the first place.

https://github.com/PRQL/prql/actions/runs/6906907289/job/18793515022#step:5:36
https://github.com/PRQL/prql/actions/runs/6906907289/job/18793514701#step:5:146

.github/workflows/release.yaml
```diff
@@ -79,46 +81,92 @@ jobs:
...
+          mv prqlc .debpkg/usr/bin/prqlc
           chmod +x .debpkg/usr/bin/prqlc
+      - name: Set variables
+        run: |
+          if [ "${{ matrix.target }}" == "aarch64-unknown-linux-musl" ]; then
+            echo "package_arch=arm64" >> $GITHUB_ENV
+          elif [ "${{ matrix.target }}" == "x86_64-unknown-linux-musl" ]; then
+            echo "package_arch=amd64" >> $GITHUB_ENV
+          else
+            echo "::error::Unknown architecture: ${{ matrix.target }}"
```
max-sixty, line 112:
```suggestion
            echo "::error::Unknown architecture: ${{ matrix.target }}"
            exit 1
```
?

.github/workflows/release.yaml
```diff
@@ -79,46 +81,92 @@ jobs:
...
+        with:
+          name: prqlc-${{ matrix.target }}
       - name: Copy files into .deb package
         run: |
+          tar -xf prqlc-*.tar.gz
           mkdir -p .debpkg/usr/bin
-          cp target/release/prqlc .debpkg/usr/bin/prqlc
+          mv prqlc .debpkg/usr/bin/prqlc
           chmod +x .debpkg/usr/bin/prqlc
+      - name: Set variables
```
max-sixty: Let's either use a descriptive name or no name

```suggestion
      - name: Set arch variable
```

max-sixty:
Looking v good, thanks for refining it @vanillajonathan

vanillajonathan:
@eitsupi Well spotted, I missed that. I was able to fix that by specifying the `arch` parameter for the `build-deb-action`.

We can't build anything else than `x86_64` with `build-rpm-action` though. https://github.com/jiro4989/build-rpm-action/issues/6

eitsupi:
How about using <https://github.com/goreleaser/nfpm> like bendsql?
https://github.com/datafuselabs/bendsql/blob/b9033d327ab21e8fb18ead2613d848cabb3b53a3/.github/workflows/release.yml#L97-L124

vanillajonathan:
We could do that, but then I would have to rework the pipeline and it's a bit more code. I think I finally got the pipeline working, so I would like just to merge it and consider this good enough until there is a demand for `.rpm` packages compiled for ARM.

eitsupi:
Since `jiro4989/build-rpm-action` doesn't seem to be widely used and doesn't have enough functionality, I'm not sure if it's worth using here.
I hope it's fixed now because I don't believe that someone will do it in the future if it's not fixed now, but I won't stop merging if @max-sixty want to merge right away.

I'm also in favor of just not adding rpm build artifacts to a release if we want to keep code small in the first place (did someone want this?)

max-sixty:
I would indeed encourage merging incremental improvements...

ARM would be nice to have, and nfpm looks good. But I would vote to merge given it sounds like @vanillajonathan is happy with it, and that it does make incremental improvements (and is clever in how it shares the build...)

max-sixty:
@eitsupi no rush if you're aware but FYI it's waiting on your "request changes" :)

eitsupi:
Sorry for the delay.

max-sixty:
Thanks @vanillajonathan & @eitsupi !

PR #3814, opened 2023-11-17
refactor: type validation

aljazerzen:


prqlc/prql-compiler/tests/integration/main.rs
```diff
@@ -26,6 +26,7 @@ fn test_sql_examples_generic() {
     // We're currently not testing for each dialect, as it's a lot of snapshots.
     // We can consider doing that if helpful.
     glob!("queries/**/*.prql", |path| {
+        log::debug!("testing {path:?}");
```
max-sixty, line 29: FYI I filed an issue upstream about this last week, since I think we had lost the ability to see the test that's currently failing. (I know we went back & forth on the design of this, I definitely remember that having `description` _used_ to work...)

max-sixty:
Looks great!

PR #3826, opened 2023-11-20
feat: exclude group by columns from partition

### LINKED ISSUE #3490: What columns does `group` pass through?
### What's up?

Part of the reason I was confused in #2947 is that it appears that the `group` transform doesn't retain its [key_columns](https://prql-lang.org/book/reference/stdlib/transforms/group.html) (or at least, I don't know how to retain them).

The example below joins a bunch of tables, with the aim of summarizing the names of the tracks for each city (and country).

Here's the output of that query (as-is) from the Playground:

name | ct1
-- | --
For Those About To Rock (We Salute You) | 1
Balls to the Wall | 2
Fast As a Shark | 1
Restless and Wild | 1

Note that `billing_city` and `billing_country` do not appear to be in the table passed to the second `group` transform

And un-commenting either `sort` or `take 1` leaves only the `ct1` column.

Is this intended behavior? Thanks.

```
from inv=invoices
join t=tracks (item.track_id==t.track_id)

group {inv.billing_city, inv.billing_country} (

  group {t.name} (
    aggregate {
    ct1 = count t.name,
    }
  #sort {-ct1}
  #take 1
  )
)
```

aljazerzen:
Closes #3490

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@
 
 **Language**:
 
+- _Breaking_: Exclude `group`'s `by` columns from the partition. See #3490
+
 **Features**:
 
 **Fixes**:
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2898,6 +2898,7 @@ dependencies = [
  "csv",
  "duckdb",
  "enum-as-inner",
+ "env_logger",
  "glob",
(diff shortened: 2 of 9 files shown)


max-sixty:
Thanks!

Issue #3822, opened 2023-11-19
Rename "output.sql", "output.arrow", etc. tabs in Playground


richb-hanover:
### What's up?

Two thoughts about the Playground:

1. I recommend we rename the four tabs of the Playground to something like this. Any better choices? If not, I'll create a PR.

   * Compiled SQL
   * Query Results
   * PL yaml
   * RQ yaml

2. The first two tabs are useful for newcomers. The last two are inscrutable (to me, at least). I assume they were enormously valuable in early development days for checking the generated code. Are they still useful? Do they justify taking up screen space in the Playground?

Thanks as always

aljazerzen:
I agree, this would make more sense. Current design tries to mimic vscode editor tabs, but this is probably not obvious to all.

The last two tabs are still useful - I occasionally use them to quickly debug something on the main branch.

max-sixty:
Agree re renaming — care to take a swing at a PR @richb-hanover ?

(I find the other tabs useful at times too, I would prefer to retain...)

richb-hanover:
OK. I will raise a PR to update the names in the tabs in the next week or so.

You've confirmed all four tabs have value: I will leave them all. Thanks

richb-hanover:
I believe PR #3838 fixes this

Issue #3807, opened 2023-11-15
"output.arrow" tab hides error message in Playground


richb-hanover:
### What happened?

Using a modified version of the query from https://github.com/PRQL/prql/issues/3490#issuecomment-1712962873, SOMETIMES I cannot see the error message in the Playground. (The modification is to paste in the full query, then comment out the second `aggregate` transform, to get the query shown below.)

When it fails, I only see a small strip of red at the bottom of the left pane, containing the word "Error".

But now I cannot make it happen - it shows the full error message as expected. Darn.

### PRQL input

```elm
from inv=invoices
join item=invoice_items (==invoice_id)
join t=tracks (item.track_id==t.track_id)

group {inv.billing_country, t.name} (
  aggregate (c=count t.name)
)

group {billing_country} (
  # aggregate (
  #   sort (-c)
  #   take 1
  # )
)
```

### SQL output

```SQL
N/A
```

### Expected SQL output

```SQL
N/A
```

### MVCE confirmation

- [X] Minimal example
- [X] New issue

richb-hanover:
I have tried changing window sizes, and other silly stuff, but can't make it happen. I'll snag a screen shot if I see it again.

max-sixty:
Haha, OK — yeah I couldn't reproduce either. I believe you it happened! 😄 

Let's reopen when it does!

richb-hanover:
@max-sixty  I was able to trigger this error in the Playground. Here's a screenshot.

<img width="483" alt="image" src="https://github.com/PRQL/prql/assets/1094930/11591292-eed8-4f11-8965-b3e39177b81e">

To reproduce: (Original message had steps swapped...)

1. Paste the query below (ignore the non-sensical SQL)
2. Remove one of the "="  Full error message appears
3. Click "output.arrow" tab. Presto! Only "Error:" appears
4. It seems related to which tab is showing - the other two hide part of the message
```
from invoices

select ! total == 0.99
```

eitsupi:
I can't repro.

![image](https://github.com/PRQL/prql/assets/50911393/074a26d1-f6d3-45b6-bf1b-68d6f8a824b4)
![image](https://github.com/PRQL/prql/assets/50911393/8146986d-9c66-45c4-9c1e-4a55278b5481)

richb-hanover:
My apologies - original instructions swapped steps 2 & 3. You should paste, remove a "=", then observe the error message in the Playground, then click the "output.arrow".

[...]

PR #3838, opened 2023-11-22
feat: Fix Playground tabs and error message

richb-hanover:
Two easy fixes:

1. Rename Playground tabs as suggested in #3822 

   Tab names use `&nbsp;` to prevent the label from getting broken onto separate lines if the tab width gets narrow. Also change background color of inactive tabs to `dimgrey` to set them off from active tab.

2. Fix height of the error message depending on #3807 

   The `error-pane` \<div> was adjusting its height based on the number of lines in the right-hand pane. Long panes caused the `error-pane` to be short. I was lucky to notice the `flex-shrink` CSS property in the Inspector. It appears to default to 1, but when set to 0, seems to allow the \<div> to grow to contain the message, while growing to no more than 50% of the main pane's height.

### DIFF
--- a/web/playground/src/output/Output.jsx
+++ b/web/playground/src/output/Output.jsx
@@ -29,10 +29,10 @@ class Output extends React.Component {
     return (
       <div className="tab">
         <div className="tab-top">
-          <Tab text="output.sql" name="sql" parent={this.props} />
-          <Tab text="output.arrow" name="arrow" parent={this.props} />
-          <Tab text="output.pl.yaml" name="pl" parent={this.props} />
-          <Tab text="output.rq.yaml" name="rq" parent={this.props} />
+          <Tab text="Compiled SQL" name="sql" parent={this.props} />
+          <Tab text="Query Results" name="arrow" parent={this.props} />
+          <Tab text="PL.yaml" name="pl" parent={this.props} />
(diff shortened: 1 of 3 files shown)


web/playground/src/output/Output.jsx
```diff
@@ -29,10 +29,10 @@ class Output extends React.Component {
...
       <div className="tab">
         <div className="tab-top">
-          <Tab text="output.sql" name="sql" parent={this.props} />
-          <Tab text="output.arrow" name="arrow" parent={this.props} />
-          <Tab text="output.pl.yaml" name="pl" parent={this.props} />
-          <Tab text="output.rq.yaml" name="rq" parent={this.props} />
+          <Tab text="Compiled&nbsp;SQL" name="sql" parent={this.props} />
+          <Tab text="Query&nbsp;Results" name="arrow" parent={this.props} />
+          <Tab text="PL.yaml" name="pl" parent={this.props} />
+          <Tab text="RQ.yaml" name="rq" parent={this.props} />
```
max-sixty, line 35: Much better! Thanks!

max-sixty (approved):
Looks good! The error panel works on my end now.

The gray for "Rename" / "Save" / "Copy to c" maybe looks a bit out of place, but overall it does look better, let's deploy and then we can iterate if we want to refine...

image

PR #3845, opened 2023-11-24
feat: infer types of transforms

aljazerzen:
Improve type inference, mainly for transforms. This is a step toward joining Lineage and Ty, but not quite there yet.

I don't think this has any behavioral effects, so it is not really a "feature", but it is not just a refactor either. Only changes that might be observable by users are new types errors where we didn't detect them previously.

### DIFF
--- a/prqlc/prql-compiler/src/ir/pl/fold.rs
+++ b/prqlc/prql-compiler/src/ir/pl/fold.rs
@@ -74,8 +74,8 @@ pub fn fold_expr_kind<T: ?Sized + PlFold>(fold: &mut T, expr_kind: ExprKind) -> 
     Ok(match expr_kind {
         Ident(ident) => Ident(ident),
         All { within, except } => All {
-            within,
-            except: fold.fold_exprs(except)?,
+            within: Box::new(fold.fold_expr(*within)?),
+            except: Box::new(fold.fold_expr(*except)?),
         },
         Tuple(items) => Tuple(fold.fold_exprs(items)?),
         Array(items) => Array(fold.fold_exprs(items)?),
--- a/prqlc/prql-compiler/src/semantic/resolver/transforms.rs
+++ b/prqlc/prql-compiler/src/semantic/resolver/transforms.rs
@@ -593,10 +593,6 @@ impl Resolver<'_> {
         // now, we need wrap the result into a closure and replace
         // the dummy node with closure's parameter.
 
-        // extract reference to the dummy node
-        // let mut tbl_node = extract_ref_to_first(&mut pipeline);
-        // *tbl_node = Expr::new(ExprKind::Ident("x".to_string()));
-
         // validate that the return type is a relation
         // this can be removed after we have proper type checking for all std functions
         let expected = Some(Ty::relation(vec![TupleField::Wildcard(None)]));
--- a/prqlc/prql-compiler/src/semantic/resolver/transforms.rs
+++ b/prqlc/prql-compiler/src/semantic/resolver/transforms.rs
@@ -631,10 +631,10 @@ fn fold_by_simulating_eval(
 }
 
 impl TransformCall {
-    pub fn infer_type(&self, root_mod: &RootModule) -> Result<Lineage> {
+    pub fn infer_lineage(&self, root_mod: &RootModule) -> Result<Lineage> {
         use TransformKind::*;
 
-        fn ty_frame_or_default(expr: &Expr) -> Result<Lineage> {
+        fn lineage_or_default(expr: &Expr) -> Result<Lineage> {
             expr.lineage
                 .clone()
                 .ok_or_else(|| anyhow!("expected {expr:?} to have table type"))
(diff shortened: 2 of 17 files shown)


prqlc/prql-compiler/src/ir/pl/fold.rs
```diff
@@ -74,8 +74,8 @@ pub fn fold_expr_kind<T: ?Sized + PlFold>(fold: &mut T, expr_kind: ExprKind) ->
     Ok(match expr_kind {
         Ident(ident) => Ident(ident),
         All { within, except } => All {
-            within,
-            except: fold.fold_exprs(except)?,
+            within: Box::new(fold.fold_expr(*within)?),
+            except: Box::new(fold.fold_expr(*except)?),
```
max-sixty, line 78: Quick Q — are these boxing to try and fix the stack issue? Or just helpful per se?

prqlc/prql-compiler/src/semantic/resolver/transforms.rs
```diff
@@ -454,6 +452,79 @@ impl Resolver<'_> {
...
+
+    /// Figure out the type of a function call, if this function is a *special function*.
+    /// (declared in std module & requires special handling).
+    pub fn infer_type_of_special_func(
+        &mut self,
+        transform_call: &TransformCall,
+    ) -> Result<Option<Ty>> {
+        // Long term plan is to make this function obsolete with generic function parameters.
+        // In other words, I hope to make our type system powerful enough to express return
+        // type of all std module functions.
```
max-sixty, line 464: 🙌

prqlc/prql-compiler/src/semantic/resolver/transforms.rs
```diff
@@ -482,131 +553,140 @@ fn into_literal_range(range: (Expr, Expr)) -> Result<(Option<i64>, Option<i64>)>
...
+        let mut pipeline = self.fold_expr(pipeline)?;
+
+        self.root_mod.module.stack_pop(NS_PARAM).unwrap();
+
+        // now, we need wrap the result into a closure and replace
+        // the dummy node with closure's parameter.
+
+        // extract reference to the dummy node
+        // let mut tbl_node = extract_ref_to_first(&mut pipeline);
+        // *tbl_node = Expr::new(ExprKind::Ident("x".to_string()));
```
max-sixty:
```suggestion
```
?

prqlc/prql-compiler/tests/sql/bad_error_messages.rs
```diff
@@ -158,8 +158,9 @@ fn invalid_lineage_in_transform() {
     sort -val
   )
   "###).unwrap_err(), @r###"
-  Error: "Invalid lineage in group, verify the contents of the group call"
-  "###);
+    Error: expected a pipeline that resolves to a table, but found `internal std.sub`
+    ↳ Hint: are you missing `from` statement?
```
max-sixty, line 162: I guess not exactly correct but I reckon we should be quite liberal in hints — fine if we have false positives

prqlc/prql-compiler/tests/sql/sql.rs
```diff
@@ -3329,7 +3363,10 @@ fn test_exclude_columns() {
       tracks
     "###
     );
+}
 
+#[test]
+fn test_exclude_columns_07() {
```
max-sixty, line 3369: Interesting — do you prefer them like this?

One advantage of doing it this way is seeing exactly what fails, particularly in non-insta tests which don't `FORCE_PASS`...

max-sixty (approved):
Looks good! (though for transparency I didn't review in detail, I can do that if at all helpful)

prqlc/prql-compiler/src/ir/pl/fold.rs
aljazerzen: Not exactly - I've changed `within: Ident` into `within: Expr` and `except: Vec<Expr>` into `except: Expr`. I have a general rule that every reference to Expr in AST should be either a `Vec` or a `Box`, because `Expr` is really big and without `Box`, `ExprKind` would have to grow substantially.

prqlc/prql-compiler/tests/sql/bad_error_messages.rs
aljazerzen: Yeah, they will surly improve with better and better typing.

prqlc/prql-compiler/tests/sql/sql.rs
aljazerzen: That was because one of the tests was panicking and it wasn't saying which one.

prqlc/prql-compiler/src/semantic/resolver/transforms.rs
aljazerzen: I don't even remember why this was commented out or what it was supposed to do... It should go.

prqlc/prql-compiler/src/ir/pl/fold.rs
max-sixty: Perfect, that makes sense

prqlc/prql-compiler/tests/sql/sql.rs
max-sixty: Yeah, I aesthetically prefer them as one big test, but then sometimes I find it doesn't give quite enough info...

PR #3846, opened 2023-11-24
fix: Fix precedence

### LINKED ISSUE #3835: a-(b+c) becomes a-b+c
### What happened?

wrong operator precedence

### PRQL input

```elm
from foo
select a - (b + c)
```

### SQL output

```SQL
SELECT
  a - b + c
FROM
  foo
```

### Expected SQL output

```SQL
SELECT
  a - (b + c)
FROM
  foo
```

### MVCE confirmation

- [x] Minimal example
- [x] New issue

### Anything else?

_No response_

max-sixty:
Closes #3835, sorry for the mistake. The problem (though not the solution) seems obvious in retrospect.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,8 @@
 
 **Fixes**:
 
+- Fix an issue with arithmetic precedence (@max-sixty, #3846)
+
 **Documentation**:
 
 **Web**:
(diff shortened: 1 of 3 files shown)


max-sixty:
Should I do a small release for this?

PR #3847, opened 2023-11-24
feat(Lezer grammar): Unary operator

vanillajonathan:
This adds support for unary operators (`+` and `-`) so we can parse `sort { -name }`.

It also adds a test for a more complete query, however it still lacks a couple of lines from the query because of gaps in the parser so there is work remaining to be done.


max-sixty:
Nice, thanks!

PR #3849, opened 2023-11-25
feat(Lezer grammar): Highlight properties

vanillajonathan:
This renames the `AssignCall` token to `Property` which seems more common in upstream tokens, albeit there it is inconsistently named and sometimes named `Property` (JSON) and `PropertyDeclaration` (JavaScript), etc.

This also adds a token with the name `PropertyName` which we map to the Lezer highlighting tag [`propertyName`](https://lezer.codemirror.net/docs/ref/#highlight.tags.propertyName).

### DIFF
--- a/grammars/prql-lezer/src/highlight.js
+++ b/grammars/prql-lezer/src/highlight.js
@@ -12,6 +12,7 @@ export const prqlHighlight = styleTags({
   Boolean: t.bool,
   Integer: t.integer,
   Float: t.float,
+  PropertyName: t.propertyName,
   TypeTerm: t.typeName,
   Escape: t.escape,
   String: t.string,
--- a/grammars/prql-lezer/src/highlight.js
+++ b/grammars/prql-lezer/src/highlight.js
@@ -12,7 +12,7 @@ export const prqlHighlight = styleTags({
   Boolean: t.bool,
   Integer: t.integer,
   Float: t.float,
-  PropertyName: t.propertyName,
+  DeclarationItem: t.propertyName,
   TypeTerm: t.typeName,
   Escape: t.escape,
   String: t.string,
--- a/grammars/prql-lezer/src/prql.grammar
+++ b/grammars/prql-lezer/src/prql.grammar
@@ -28,17 +28,17 @@ pipe { "|" | ~ambigNewline newline }
 
 TupleExpression { "{" newline* tupleItem (("," newline*) tupleItem)* ","? newline* "}" }
 
-tupleItem { AssignTupleItem | expression | CallExpression | CaseBranch }
+tupleItem { DeclarationTuple | expression | CallExpression | CaseBranch }
 
 // Ideally we would force a space after `Identifier` to prevent an invalid s-string
 // being parsed as a CallExpression, e.g. `s"{a"` -> `s` & `"{a"'. But we
 // can't seem to force a space because it's in our skip, and I can't see
 // a way of changing the skip expression to only specialize on a single item
-CallExpression { Identifier ArgList { (NamedArg | Parameter | test)+ } }
+CallExpression { Identifier ArgList { (NamedArg | Declaration | test)+ } }
 
 NamedArg { ArgumentName { identPart } ":" expression }
-Parameter { ParameterName { identPart } "=" expression }
-AssignTupleItem { PropertyName { identPart } "=" expression }
+Declaration { DeclarationItem { identPart } "=" expression }
+DeclarationTuple { DeclarationItem { identPart } "=" expression }
 CaseBranch { expression "=>" expression }
 // Possibly we could only accept case branches inside the TupleExpression?
 CaseExpression { @specialize<identPart, "case"> TupleExpression }
--- a/grammars/prql-lezer/src/prql.grammar
+++ b/grammars/prql-lezer/src/prql.grammar
@@ -28,7 +28,7 @@ pipe { "|" | ~ambigNewline newline }
 
 TupleExpression { "{" newline* tupleItem (("," newline*) tupleItem)* ","? newline* "}" }
 
-tupleItem { Property | expression | CallExpression | CaseBranch }
+tupleItem { AssignTupleItem | expression | CallExpression | CaseBranch }
 
 // Ideally we would force a space after `Identifier` to prevent an invalid s-string
 // being parsed as a CallExpression, e.g. `s"{a"` -> `s` & `"{a"'. But we
@@ -38,7 +38,7 @@ CallExpression { Identifier ArgList { (NamedArg | Assign | test)+ } }
 
 NamedArg { identPart ":" expression }
 Assign { identPart "=" expression }
-Property { PropertyName { identPart } "=" expression }
+AssignTupleItem { PropertyName { identPart } "=" expression }
 CaseBranch { expression "=>" expression }
 // Possibly we could only accept case branches inside the TupleExpression?
 CaseExpression { @specialize<identPart, "case"> TupleExpression }
--- a/grammars/prql-lezer/src/prql.grammar
+++ b/grammars/prql-lezer/src/prql.grammar
@@ -28,7 +28,7 @@ pipe { "|" | ~ambigNewline newline }
 
 TupleExpression { "{" newline* tupleItem (("," newline*) tupleItem)* ","? newline* "}" }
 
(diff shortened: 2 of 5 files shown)


max-sixty (commented):
I agree `AssignCall` is not great, but I don't think this is a `Property` either.

Do we need to distinguish between `Assign` & `AssignCall`?

---

Overall, I think it's more important to stay aligned with the PRQL parser than the names that languages like JS use. If the names in the PRQL parser can be improved, that's very welcome. But in this case, we don't really have objects, and so we don't have properties... Or am I missing something?

vanillajonathan:
How about the name `Column` or `Field`? Please provide suggestions if you have any better name.

Good question about `Assign` and `AssignCall`. I think it depends on whether they're semantically the same thing. If they are then they do not need to be distinguished, but if they are semantically different then it would be good to have separate Lezer tokens for them as that allows us to highlight them differently by mapping them to different Lezer highlighting tags in the `highlight.js` file and would also allow us to [lint](https://codemirror.net/examples/lint/) them differently.

max-sixty:
How about `variable` or `declaration`? That's what we call them in the book: https://prql-lang.org/book/reference/declarations/variables.html. ~They're not always columns (for example, a function), so `column` isn't ideal either.~

---

> Good question about `Assign` and `AssignCall`. I think it depends on whether they're semantically the same thing.

I think we used to differentiate between

```elm
from x
derive foo = (round 2 g)
```

and

```elm
from x
derive {foo = round 2 g}
```

...since only the latter can be outside a tuple, because of our rules about parentheses (https://prql-lang.org/book/reference/syntax/operators.html?parentheses#parentheses). So possibly that's why we have two types here. We can merge them if the parser can still handle that difference...

vanillajonathan:
We cannot use `variable` or `declaration` (lowercase) because then they would be internal tokens, so they would have to be `Variable` or `Declaration`.

We already have a `VariableDeclaration` token which is composed of the `let` keyword, the variable name (a token named `VariableName`), the equals sign, and a nested pipeline. Adding a `Variable` token could be confusing when we already have a `VariableDeclaration` and a `VariableName` token. The `Declaration` is rather ambiguous.

Perhaps `TupleAtom` or `TupleVariable`?

max-sixty:
> Perhaps `TupleAtom` or `TupleVariable`?

These aren't always Tuples though! e.g. `derive foo = bar`

I have to admit that `Assign` seems reasonable! I can't think of a better term than those at the moment. `Alias` could be OK...

vanillajonathan:
But we have an `Assign` token, the one I renamed to `Property` was the `AssignCall` token. Should I rename it back to `AssignCall`?

max-sixty:
Ah, I see what you mean. 

`AssignTupleItem` / `AssignTuple` could be reasonable. I agree the `Call` part of it isn't that clear. It was probably there because `derive foo = (round 2 g)` contains a function call...

(and looks like at the moment it's the same as `Assign` — though actually `derive foo = round 2 g` should be invalid, and I'm guessing it's not at the moment, since the current `Assign` allows a function call, such as `round 2 g`. But this isn't the most important thing...)

vanillajonathan:
`AssignTupleItem` sounds alright for now.

grammars/prql-lezer/src/highlight.js
```diff
@@ -12,6 +12,7 @@ export const prqlHighlight = styleTags({
   Boolean: t.bool,
   Integer: t.integer,
   Float: t.float,
+  PropertyName: t.propertyName,
```
max-sixty: Should we change this too?

vanillajonathan: That is the name of the identifier left of the equal signs.

    AssignTupleItem { PropertyName { identPart } "=" expression }
    
So the name of the token in the highlight.js file matches the name of the token in the grammar file. So this works but if you don't like the name of the token, maybe you want it changed to `TupleItemName` or `TupleItemKey`?

Maybe I could rename `AssignTupleItem` to just `TupleItem`?

grammars/prql-lezer/src/prql.grammar
```diff
@@ -38,7 +38,7 @@ CallExpression { Identifier ArgList { (NamedArg | Assign | test)+ } }
 
 NamedArg { identPart ":" expression }
 Assign { identPart "=" expression }
-AssignCall { identPart "=" expression }
+AssignTupleItem { PropertyName { identPart } "=" expression }
```
max-sixty: I definitely think that `AssignTupleItem` & `Assign` should be formatted the same. So we could name the LHS something like `Alias` or `Assignee` or `LHS` (but not property, we don't have properties) — if we do we should update them both...

vanillajonathan: I agree that `Assign` should have the `identPart` wrapped in a public token so it is exposed as a child token of `Assign` because it increases the semantic richness of the grammar.

`Assign` is just used in the context of `CallExpression` so it could be renamed `Parameter` and have the child token `ParameterName`.

The grammar rules for the `Assign` token and the `AssignTupleItem` token would be equal. I don't know if it makes sense to have them as separate tokens or merge them into token that we re-use. `Parameter` and `ParameterName` would be unsuitable names for tokens inside a `TokenExpression` though, so probably keep them as separate tokens.

max-sixty: OK! Happy to have a look at the next version...

grammars/prql-lezer/src/prql.grammar
```diff
@@ -34,10 +34,10 @@ tupleItem { AssignTupleItem | expression | CallExpression | CaseBranch }
...
 // can't seem to force a space because it's in our skip, and I can't see
 // a way of changing the skip expression to only specialize on a single item
-CallExpression { Identifier ArgList { (NamedArg | Assign | test)+ } }
+CallExpression { Identifier ArgList { (NamedArg | Parameter | test)+ } }
 
-NamedArg { identPart ":" expression }
-Assign { identPart "=" expression }
+NamedArg { ArgumentName { identPart } ":" expression }
+Parameter { ParameterName { identPart } "=" expression }
 AssignTupleItem { PropertyName { identPart } "=" expression }
```
max-sixty: TBC, these two are very very similar and should be formatted exactly the same, so very likely `ParameterName` and `PropertyName` should be the same...

```
Parameter { ParameterName { identPart } "=" expression }
AssignTupleItem { PropertyName { identPart } "=" expression }
```

max-sixty: And `Parameter` & `AssignTupleItem` should be very similar — I think both named `Assign` in some way

vanillajonathan: You earlier said we don't have properties, so do you want both both `Parameter` and `AssignTupleItem` to have a child token named `ParameterName`. I am not sure `ParameterName` is a good name for the tuple item since parameters are for f unctions, not for tuples.

Also, I don't like the name "assign" because it is a verb.

So maybe `AssignTupleItem` should just be called `TupleItem`?

max-sixty: But it's not any `TupleItem` — it only applies to assignments.

---

I don't want to dissuade you from making small incremental changes — they often add up to great changes! And overall you've made lots of great contributions to this repo.

But I think this one may not be worthwhile — it must be taking up a decent amount of your time, and it has been the most commented issue from me in the past week. Unless we have a way that is obviously better, I might suggest we leave this for a while, see how other changes go, and we could come back to it later. WDYT?

vanillajonathan: Then maybe it is a `TupleItemDeclaration`?

Well, this one enables highlighting of that token. Maybe you could just say some names, and we'll use that, we can always change it later.

max-sixty: > Then maybe it is a `TupleItemDeclaration`?

Yes this works! So we have something like

```diff
- Assign { identPart "=" expression }
+ Declaration { DeclarationItem { identPart } "=" expression }`
- AssignCall { identPart "=" expression }
+ DeclarationTuple { DeclarationItem { identPart } "=" expression }
```

`DeclarationItem` could also be `Variable` or `Alias`. I think it should be the same in both so they're formatted the same

WDYT?

max-sixty:
OK great! Thanks for pushing through

PR #3850, opened 2023-11-26
refactor: make Debug of pl::Expr much more compact

aljazerzen:


max-sixty (commented):
Great! Fan of terser representations, it's surprising how helpful a quick snapshot is when debugging, as opposed to scrolling around...

PR #3862, opened 2023-11-29
docs: Add repology to README

vanillajonathan:
Some projects include a repology thing like this. We could too if we wanted.

[Packaging status](https://repology.org/project/prqlc/versions)


eitsupi:
I think it is already included in prqlc's README.

PR #3863, opened 2023-11-29
refactor: wildcard resolution and pl::Expr::target_ids

aljazerzen:
This PR has partially been cherry-picked from an old branch of mine and partially rewritten anew. I've worked on this for a few weeks now, so I forgot the initial motivation of why I wanted to change this :D

It does simplify:
- the wildcard resolution (this opinion might be biased since I spent so much time debugging it and I know how it works in depth now)
- `Expr::target_id` and `Expr::target_ids`, which are now a single number.

I *think* I wanted this because it made lineage inference independent of the declarations in root module: https://github.com/PRQL/prql/pull/3863/files#diff-d2aed3357385f80cd4517d8ca5492ad503c33c13b8afbd9e219666f57f947459L264

As an additional upside, I changed functions implemented specifically for lineage into functions that manipulate the AST, which means that I don't have to duplicate this behavior in type inference.

### DIFF
--- a/justfile
+++ b/justfile
@@ -30,12 +30,12 @@ prqlc-lint:
 # Test prqlc
 packages := '--package=prqlc-ast --package=prqlc-parser --package=prql-compiler --package=prqlc'
 prqlc-test-fast:
-    cargo clippy --all-targets {{packages}} -- -D warnings
-
     # cargo insta test, but allowing multiple --package arguments
     INSTA_FORCE_PASS=1 cargo test --locked {{packages}}
 
     cargo insta review
+
+    cargo clippy --all-targets {{packages}} -- -D warnings
 
 prqlc-test:
     cargo clippy --all-targets {{packages}} -- -D warnings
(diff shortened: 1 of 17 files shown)


justfile
```diff
@@ -30,13 +30,13 @@ prqlc-lint:
 # Test prqlc
 packages := '--package=prqlc-ast --package=prqlc-parser --package=prql-compiler --package=prqlc'
 prqlc-test-fast:
-    cargo clippy --all-targets {{packages}} -- -D warnings
-
     # cargo insta test, but allowing multiple --package arguments
     INSTA_FORCE_PASS=1 cargo test --locked {{packages}}
```
max-sixty, line 34: FYI this is now fixed upstream in `cargo-insta` so you can use it directly with multiple packages!

aljazerzen: Just in time for me to discover cargo-nextest :D

max-sixty: Also supported by `cargo-insta`! :D

max-sixty:
Some of the changes in test outputs look good!

aljazerzen:
I realized now that I didn't even mention why I'm changing this! I've edited the original message.

PR #3864, opened 2023-11-29
fix(Playground): Handle compiler crashing

### LINKED ISSUE #3848: Panic in Playground shows bad error message
### What happened?

When editing a query in the Playground (using 1759a867 from 25 Nov 2023) the query looks OK when pasted in. But editing `inv.name` in the nested `group`, the displayed error message is incorrect.
- changing to `in.name` correctly shows an error
- adding the "v" back doesn't change the error
- adding another "v" displays the correct error
- removing the extraneous "v" doesn't change the displayed error

I made a video of the behavior... https://youtube.com/shorts/uB66piyGihI?feature=share

NB: It helps to reload the Playground before starting the experiment

### PRQL input

```elm
from inv=invoices

group { inv.billing_city } (

  group { item.name } (
    aggregate {
      ct1 = count inv.name,
    }
  )
)
```

### SQL output

```SQL
N/A
```

### Expected SQL output

```SQL
N/A
```

### MVCE confirmation

- [X] Minimal example
- [ ] New issue

### Anything else?

_No response_

vanillajonathan:
The playground just caught any exception then called `JSON.parse` to deserialize the `inner` property. That was a problem since sometimes we get a `Error` thrown (when the compiler returns an error), and sometimes a `RuntimeError` thrown (when the compiler crashes).

So now we check if the exception is an `RuntimeError` and then we know the compiler crashed and there is nothing to JSON deserialize.

Fixes #3848

### DIFF
--- a/web/playground/src/workbench/Workbench.jsx
+++ b/web/playground/src/workbench/Workbench.jsx
@@ -74,7 +74,8 @@ class Workbench extends React.Component {
     } catch (e) {
       if (e instanceof WebAssembly.RuntimeError) {
         this.setState({
-          prqlError: "A compiler bug encountered. Please report this to https://github.com/PRQL/prql/issues/new/choose",
+          prqlError:
+            "A compiler bug encountered. Please report this to https://github.com/PRQL/prql/issues/new/choose",
         });
         return;
       }
--- a/web/playground/src/workbench/Workbench.jsx
+++ b/web/playground/src/workbench/Workbench.jsx
@@ -74,7 +74,7 @@ class Workbench extends React.Component {
     } catch (e) {
       if (e instanceof WebAssembly.RuntimeError) {
         this.setState({
-          prqlError: "The compiler crashed! Report this as a bug!",
+          prqlError: "A compiler bug encountered. Please report this to https://github.com/PRQL/prql/issues/new/choose",
         });
         return;
       }
--- a/web/playground/src/workbench/Workbench.jsx
+++ b/web/playground/src/workbench/Workbench.jsx
@@ -72,6 +72,11 @@ class Workbench extends React.Component {
       this.setState({ prqlError: null });
       this.monaco.editor.setModelMarkers(this.editor.getModel(), "prql", []);
     } catch (e) {
+      if (e instanceof WebAssembly.RuntimeError) {
+        this.setState({ prqlError: "The compiler crashed! Report this as a bug!" });
+        return;
+      }
+
       const errors = JSON.parse(e.message).inner;
       this.setState({ prqlError: errors[0].display });
 
--- a/web/playground/src/workbench/Workbench.jsx
+++ b/web/playground/src/workbench/Workbench.jsx
@@ -73,7 +73,9 @@ class Workbench extends React.Component {
       this.monaco.editor.setModelMarkers(this.editor.getModel(), "prql", []);
     } catch (e) {
       if (e instanceof WebAssembly.RuntimeError) {
-        this.setState({ prqlError: "The compiler crashed! Report this as a bug!" });
+        this.setState({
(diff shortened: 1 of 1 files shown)


web/playground/src/workbench/Workbench.jsx
```diff
@@ -72,6 +72,13 @@ class Workbench extends React.Component {
       this.setState({ prqlError: null });
       this.monaco.editor.setModelMarkers(this.editor.getModel(), "prql", []);
     } catch (e) {
+      if (e instanceof WebAssembly.RuntimeError) {
+        this.setState({
+          prqlError: "The compiler crashed! Report this as a bug!",
```
aljazerzen:
```suggestion
          prqlError: "A compiler bug encountered. Please report this to https://github.com/PRQL/prql/issues/new/choose",
```

aljazerzen (approved):

max-sixty:
Awesome @vanillajonathan !

Is it really only this code — did you test it locally?

vanillajonathan:
I tested it on GitHub Codespaces and it seems to work fine for me.

max-sixty:
I'm still not seeing an error message — 
image

Are others seenig something different?

vanillajonathan:
Hmm, this is weird. I don't get any error message either on the deployed playground, but I do get the error message when I open the `main` branch in GitHub Codespaces and run `cd web/playground`, `npm i`, `npm run dev`.

richb-hanover:
Using 9cc5bc8c in a Dev Container on my Intel Macbook (12.6.7), I do see the compiler error message:

image

I wonder about modifying the error describe a good process for reporting an error. Perhaps something like this:

```
this.setState({
prqlError:
            "A compiler bug was encountered. Please copy/paste the PRQL into a new bug report at:\n" +
            "   https://github.com/PRQL/prql/issues/new/choose",
        });
```
If the wording is OK, I'm happy to submit this as a PR...

@vanillajonathan - Thanks for such a speedy fix!

max-sixty:
> Hmm, this is weird. I don't get any error message either on the deployed playground, but I do get the error message when I open the `main` branch in GitHub Codespaces and run `cd web/playground`, `npm i`, `npm run dev`.

You're completely correct!

This is surprising — I'll see if I can figure it out...

max-sixty:
If I run ~`npm preview`~ `vite preview` or `npm run preview`, then it seems to not work — do you see that also? 

Could this be that it operates differently in `dev`?

richb-hanover:
Both these seem to display the Playground when executed in my Dev Container:

1. `task run-playground`
2. `cd web/playground; npm i; npm run dev`

`npm preview` doesn't seem to be a legal command. `cd web/playground` and `npm run preview` seems to start the backend http server, but the browser is sent to http://127.0.0.1:4173/ (different port; no path) and the browser displays `Cannot GET /`

max-sixty:
> `npm preview` doesn't seem to be a legal command.

Sorry, edited above

vanillajonathan:
@max-sixty You need to run `npm run build` before you can run `npm run preview`. I do get the error message when I run `npm run preview`.

@richb-hanover You need to append `/playground/playground` to the URL in the web browser.

max-sixty:
> @max-sixty You need to run `npm run build` before you can run `npm run preview`. I do get the error message when I run `npm run preview`.

Sorry! You're completely correct!

It's confusing why this isn't working in the deployed version, then

max-sixty:
~So we have exactly the same versions + we can rule out any deployment issues, I'm going to try publishing this commit, rather than the usual backport to web.~

~This is from a fork so doesn't work. I'll start a release, we're somewhat due for one anyway~

OK actually this resolved! Thanks again @vanillajonathan !

PR #3868, opened 2023-11-29
test: move a few tests from unit to integration

aljazerzen:
This blocks #3762, because we are snapshoting a struct that has different fields on Windows.

So I've made these tests into regular SQL tests.


max-sixty:
LGTM

PR #3869, opened 2023-11-30
tweak: Remove unused items

max-sixty:
I think this is showing warnings in the nightly compiler, so shows up on PRs when `check-unused-dependences` is run


aljazerzen (approved):
Hmm, strange that clippy does not report this...

PR #3865, opened 2023-11-29
test: combine prql-compiler test binaries, test_each_file

aljazerzen:
Followup for #3715

This PR does two things:
- combines `prqlc/prql-compiler/tests/integration` and `prqlc/prql-compiler/tests/sql` into a single test binary,
- replaces `insta::glob!` with `test_each_file::test_each_path!`, which is a macro that generates separate tests:

```
prql-compiler::integration:
    bad_error_messages::array_instead_of_tuple
    bad_error_messages::empty_interpolations
...
    error_messages::test_ambiguous
    error_messages::test_ambiguous_inference
...
    parser::check_invalid_version
    parser::check_valid_version
...
    queries::compile::aggregation   <-- all of these used to be a single test
    queries::compile::arithmetic   
    queries::compile::cast
    queries::compile::constants_only
    queries::compile::distinct
    queries::compile::distinct_on
    queries::compile::genre_counts
    queries::compile::group_all
    queries::compile::group_sort
    queries::compile::invoice_totals
    queries::compile::loop_01
    queries::compile::pipelines
    queries::compile::read_csv
    queries::compile::set_ops_remove
    queries::compile::sort
    queries::compile::switch
    queries::compile::window
    queries::fmt::aggregation
    queries::fmt::arithmetic
    queries::fmt::cast
    queries::fmt::constants_only
    queries::fmt::distinct
    queries::fmt::distinct_on
    queries::fmt::genre_counts
    queries::fmt::group_all
    queries::fmt::group_sort
    queries::fmt::invoice_to
...
```

This means that when a test panics:
- it is easy to see which test file panicked,
- other tests will run regardless.

I've also started using `cargo-nextest`, which has nicer printout and cuts run time of all tests from 7.5sec to 4.8sec. Not much, but it's something.

PR #3871, opened 2023-11-30
test: test-db organization and for_each_path!

aljazerzen:
Followup for #3865


prqlc/prql-compiler/tests/integration/queries.rs
```diff
@@ -8,123 +8,145 @@ use prql_compiler::sql::Dialect;
...
+                "{} {} {}",
+                test_name,
+                left_dialect,
+                right_dialect
+            );
         }
-    })
+
+        with_settings!({ input_file => prql_path }, {
+            assert_snapshot!(test_name, left_text, &prql)
```
max-sixty, line 194: `description` isn't that helpful given it doesn't seem to show on failures within `glob!`, though is worth _removing_? (from https://github.com/PRQL/prql/pull/3871/files#diff-194853534e9688ae9ff624c9936b82108b3fa8aea5d5bb02dc503047395b2ef4L123)

(and possibly now we're not using `glob!`, it does show?)

prqlc/prql-compiler/tests/integration/queries.rs
```diff
@@ -8,123 +8,145 @@ use prql_compiler::sql::Dialect;
...
-    use crate::dbs::IntegrationTest;
+    fn init_connections() -> Mutex<Vec<DbConnection>> {
+        let con_dir = "./tests/integration/dbs/connections";
 
-    let runtime = &*crate::dbs::RUNTIME;
+        let mut connections = Vec::new();
+        for con_file in fs::read_dir(con_dir).unwrap() {
+            let con_file = con_file.unwrap();
+            let con_toml = fs::read_to_string(con_file.path()).unwrap();
+            let con_cfg: ConnectionCfg = toml::from_str(&con_toml)
```
max-sixty: I'm fine with this but FWIW I'm generally less keen on breaking things that don't change often / don't require modularity into config files — it makes them further away from the code, means LSPs can't resolve them, adds a bunch more files...

And here, we also have the configs listed in two places — structs and a config file. It would be more reasonable if we could make the rust code totally generic and then the DBs were defined in the config file. If someone comes to the code new, I think it would be harder to add a config than it was.

(Not suggesting we do the work to change this, so I know this isn't immediately impactful, tell me if it feels like peanut-gallery critique, especially given our recent relative contributions...)

max-sixty (approved):

prqlc/prql-compiler/tests/integration/queries.rs
aljazerzen: Yeah, I might have over-engineered this. It was very strange to be implementing a trait directly on `Dialect` and I couldn't find where we defined which databases to connect to.

Because I was looking for some kind of declaration, I made one. Let me inline this to be define in the code.

prqlc/prql-compiler/tests/integration/queries.rs
aljazerzen: That's a removal by mistake! I don't know which fields show-up where, so I'm mostly copy-pasting stuff around here :D

prqlc/prql-compiler/tests/integration/queries.rs
```diff
@@ -58,23 +58,68 @@ mod results {
...
+                    url: "host=localhost user=glaredb dbname=glaredb port=6543".to_string(),
+                },
+            },
+            ConnectionCfg {
+                dialect: Dialect::MsSql,
+                data_file_root: "/tmp/chinook".to_string(),
+
+                protocol: DbProtocol::MsSql,
+            },
+        ];
```
max-sixty, line 117: Awesome, thanks!

aljazerzen: Initial idea was to find a test generator macro that I would point at two directories and it would then generate a test for each combination of the two files.

PR #3877, opened 2023-12-01
build: migrate just to task

aljazerzen:
I've moved my scripts from just to task.

I would wish to be able to run most of the tests that we have in GH actions locally and this is a step into that direction.

I hope that multiple taskfiles are not problematic.


prqlc/Taskfile.yaml
```diff
@@ -0,0 +1,68 @@
...
+          rg '\s+$' --files-with-matches --glob '!*.rs' . \
+          | xargs -I _ sh -c "echo Removing trailing whitespace from _ && sd '[\t ]+$' '' _"
+
+      - cmd: |
+          # rustfmt
+          cargo fmt {{.packages_core}} {{.packages_addon}} {{.packages_bindings}}
+
+      - cmd: |
+          # no dbg
+          rg 'dbg!' --glob '*.rs' . --no-heading
```
max-sixty, line 29: 100% fine to have these. We do have some of the fast ones like the greps & trailing whitespace in `pre-commit`, just as an FYI (having both is reasonable)

prqlc/Taskfile.yaml
```diff
@@ -0,0 +1,68 @@
...
+          --config=../.prettierrc.yaml \
+          --ignore-path=../.prettierignore \
+          --ignore-unknown \
+          --log-level=warn
+
+  test-fast:
+    desc: A fast test used for feedback during compiler development
+    cmds:
+      - cmd: |
+          INSTA_FORCE_PASS=1 cargo nextest run {{.packages_core}} --no-fail-fast
```
max-sixty, line 43: Nice, I haven't tried `nextest`, will try this!

prqlc/Taskfile.yaml
```diff
@@ -0,0 +1,68 @@
...
+
+  test:
+    desc: A full test of prqlc
+    cmds:
+      - cmd:
+          cargo nextest run {{.packages_core}} {{.packages_addon}}
+          {{.packages_bindings}}
+
+      - cmd:
+          cargo clippy --all-targets {{.packages_core}} {{.packages_addon}}
```
max-sixty, line 57: Great, I don't think we used to have this (and too slow for pre-commit), so great we have it here

prqlc/Taskfile.yaml
```diff
@@ -0,0 +1,68 @@
...
+          {{.packages_bindings}} -- -D warnings
+
+      - cmd:
+          cargo test --package prql-compiler --features=test-dbs
+          --test=integration -- queries::results
+
+  pull-request:
+    desc: All checks that run within GH actions for a pull request
```
max-sixty:
```suggestion
    desc: Most checks that run within GH actions for a pull request
```
(We don't have external test dbs etc)

max-sixty (approved):
Awesome, thanks!

We can add these to the `test-taskfile.yaml` IYL — adds some overhead, but they did have a tendency to break when folks changed something otherwise.

Totally reasonable to have multiple Taskfiles. It's also OK to combine them, as you prefer.

Thanks!

PR #3881, opened 2023-12-03
feat(Lezer grammar): Add self equals operator

vanillajonathan:
I wanted to add support for the self equals operator to support statements such as:

    join c=customers (==customer_id)

We already have a `UnaryExpression` token for expressions like `+1`and `-1`. Is the self equals operator a unary expression? Lexically it looks like a unary expression, but it is a short hand for what is a binary expression as it translates to something like `a.customer_id == b.customer_id`.

Either way, I added it as a `UnaryExpression`, not sure if that's right, but I hope so. So `==customer_id` gets parsed into the following tree:

    UnaryExpression {
      CompareOp
      Identifier
    }

With this PR the Lezer grammar can parse `filter ==customer_id` which is **not a valid** PRQL query and it cannot parse `join c=customers (==customer_id)` which is a valid PRQL query because the second argument to the `join` function is a `NestedPipeline` which currently only can contain a `Pipeline` which currently only can contain a `CallExpression`.

Any ideas?


max-sixty:
> Is the self equals operator a unary expression?

It's a Unary Operator, yes. So I think we can treat `(==customer_id)` exactly the same as `(-customer_id)`.

Does that make sense?

vanillajonathan:
> > Is the self equals operator a unary expression?
> 
> It's a Unary Operator, yes. So I think we can treat `(==customer_id)` exactly the same as `(-customer_id)`.
> 
> Does that make sense?

Yes, that makes sense, and this PR handles `==customer_id` the same as `+customer_id`, both are `UnaryExpression`'s. But `(customer_id)` and `(1)` and `(-1)` are still not handled.

What exactly is `(+1)` in the sense of a grammar/parse tree?

Currently a `NestedPipeline` token can only contain a `Pipeline` token which can only contain a  `CallExpression`. 

`+1`, `1`, `customer_id`, `-customer_id` and `==customer_id` are all expressions. Expressions are currently not allowed within a `Pipeline` nor within a `NestedPipeline`.

max-sixty:
> What exactly is `(+1)` in the sense of a grammar/parse tree?

One way of looking at how the compiler handles this is the `PL.yaml` in the playground:

```elm
from artists
derive x = +5
```

```yaml
- VarDef:
    kind: Main
    name: main
    value:
      Pipeline:
        exprs:
          - FuncCall:
              name:
                Ident:
                  - from
              args:
                - Ident:
                    - artists
          - FuncCall:
              name:
                Ident:
                  - derive
              args:
                - Unary:
                    op: Add
                    expr:
                      Literal:
                        Integer: 5
                  alias: x
  span: 1:0-26
```



> `+1`, `1`, `customer_id`, `-customer_id` and `==customer_id` are all expressions. Expressions are currently not allowed within a `Pipeline` nor within a `NestedPipeline`.

OK interesting — will the Lezer grammar not let us compile something like:

```elm
from artists
derive x = 5 + (1)
```

?

vanillajonathan:
Yeah, I looked at the PL.

No, it will not let us compile that because in the pipeline `derive x = 5 + (1)` the `(1)` is an `Integer` token within a `Pipeline` token and the `Pipeline` token currently only allows a `CallExpression`.

max-sixty:
> No, it will not let us compile that because in the pipeline `derive x = 5 + (1)` the `(1)` is an `Integer` token within a `Pipeline` token and the `Pipeline` token currently only allows a `CallExpression`.

I see, good spot.

If we can allow that to work, by allowing parentheses around & within expressions, that would be very good. Lezer is stricter about ambiguities than our parser in rust, so this might not be trivial, but it does seem quite important.

Do you know if there are other big gaps in the parser? I hadn't known that there was anything as big as this.

One way of discovering gaps is to check whether it can parse our queries at https://github.com/PRQL/prql/blob/a9d81a429b6a2c940e94fcf177c402842c05f25b/prqlc/prql-compiler/tests/integration/queries/arithmetic.prql. I looked at whether any of these queries had parentheses in expressions — I think possibly none of the queries have this, surprisingly.

vanillajonathan:
> If we can allow that to work, by allowing parentheses around & within expressions, that would be very good. Lezer is stricter about ambiguities than our parser in rust, so this might not be trivial, but it does seem quite important.

Then we might get a collision with nested pipelines since they use parenthesis too. An alternative would be to allow expressions within nested pipelines.

How would we know what is an expression surrounded by parenthesis and how would we know what is a pipeline?

> Do you know if there are other big gaps in the parser? I hadn't known that there was anything as big as this.

A couple of small stuff:
- No support for annotations, i.e. `@{binding_strength=100}`.
- No interpolation within f-strings.
- No support for parameters, i.e. `$1`.
- No support for modules, i.e. `module foo { ... }`.
- The `DateTime` token is not mapped to any Lezer highlighting tag.

But the big issue is the newline handling which does not allow empty lines or comments within two pipelines.

Example:
```elm
from foo
# comment
select bar
```

and
```elm
from foo

select bar
```

max-sixty:
> A couple of small stuff:
> 
>     * No support for annotations, i.e. `@{binding_strength=100}`.
> 
>     * No interpolation within f-strings.
> 
>     * No support for parameters, i.e. `$1`.
> 
>     * No support for modules, i.e. `module foo { ... }`.
> 
>     * The `DateTime` token is not mapped to any Lezer highlighting tag.

Nice list, thanks for putting this together. I agree those are mostly small. Would be nice to solve them in time.

---

> But the big issue is the newline handling which does not allow empty lines or comments within two pipelines.

Ah, good spot. That + expressions seem like the top priority, we should fix those before rolling this out to playground / jupyter extension (which I still haven't done, but hoping to make progress over the holidays here)

max-sixty:
> Then we might get a collision with nested pipelines since they use parenthesis too. An alternative would be to allow expressions within nested pipelines.
> 
> How would we know what is an expression surrounded by parenthesis and how would we know what is a pipeline?

I think Lezer has an approach for this — it accepts ambiguity if precedence is defined. So it would try and parse as an expression, and then if that fails, it would backtrack and parse as a nested pipeline. Is this familiar to you? I think possibly you've already done some work with precedence on this?

grammars/prql-lezer/test/operators.txt
```diff
@@ -62,10 +62,18 @@ filter foo ?? bar
...
 
 sort { -name }
 
 ==>
 
 Query(Pipeline(CallExpression(Identifier,ArgList(TupleExpression(UnaryExpression(ArithOp,Identifier))))))
+
+# Unary operator ==
+
+filter ==customer_id
```
max-sixty: I don't think this is correct PRQL — it should require parentheses

vanillajonathan: Yes, I know, but thanks for pointing it out which helped me get back on track. The reason I left the parentheses out was because the parser didn't support those at the time, but now with PR #3906 it does, and which is why this PR was a draft, because things weren't ironed out yet.

grammars/prql-lezer/test/full_queries.txt
```diff
@@ -7,7 +7,20 @@ derive {
...
+  }
+)
+sort {-sum_income}
+take 10
+join c=customers (==customer_id)
+derive name = f"{c.last_name}, {c.first_name}"
+select {
+  c.customer_id, name, sum_income
+}
+derive db_version = s"version()"
```
max-sixty, line 22: Nice!

vanillajonathan: Yes, but unfortunately it misses two lines in the group aggregate.

PR #3883, opened 2023-12-03
feat: add filter based on a list of values

### LINKED ISSUE #993: filter based on a list of values
Hi team,
Is the SQL `IN` method currently supported in PRQL? I didn't manage to find a single example in the book.

```sql
WHERE column_name IN (value1, value2, ...)
```

PrettyWood:
closes #993

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,8 @@
 
 **Features**:
 
+- The filter `std.in` now supports a list of values (@PrettyWood, #3883)
+
 **Fixes**:
 
 - Fix an issue with arithmetic precedence (@max-sixty, #3846)
@@ -23,6 +25,8 @@
 - Bump `prql-compiler`'s MSRV to 1.70.0 (@eitsupi, #3876)
 
 **New Contributors**:
+
+- @PrettyWood, with #3883
 
 ## 0.10.1 — 2023-11-14
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@
 
 **Features**:
 
-- The filter `std.in` now supports a list of values (@PrettyWood, #3883)
+- The `std.in` function now supports a list of values (@PrettyWood, #3883)
 
 **Fixes**:
 
--- a/prqlc/prql-compiler/src/semantic/resolver/transforms.rs
+++ b/prqlc/prql-compiler/src/semantic/resolver/transforms.rs
@@ -238,6 +238,14 @@ impl Resolver<'_> {
 
                 let [pattern, value] = unpack::<2>(func.args);
 
+                if let ExprKind::Array(mut in_values) = pattern.kind {
+                    in_values.insert(0, value);
+                    return Ok(Expr::new(ExprKind::RqOperator {
+                        name: "std.in".to_string(),
+                        args: in_values.clone(),
+                    }));
+                }
+
                 let pattern = match try_restrict_range(pattern) {
                     Ok((start, end)) => {
                         let start = restrict_null_literal(start);
--- a/prqlc/prql-compiler/src/semantic/resolver/transforms.rs
+++ b/prqlc/prql-compiler/src/semantic/resolver/transforms.rs
@@ -238,11 +238,13 @@ impl Resolver<'_> {
 
                 let [pattern, value] = unpack::<2>(func.args);
 
-                if let ExprKind::Array(mut in_values) = pattern.kind {
-                    in_values.insert(0, value);
+                if let ExprKind::Array(in_values) = pattern.kind {
+                    let mut col_and_values: Vec<Expr> = Vec::with_capacity(in_values.len() + 1);
+                    col_and_values.push(value);
+                    col_and_values.extend(in_values);
                     return Ok(Expr::new(ExprKind::RqOperator {
                         name: "std.in".to_string(),
-                        args: in_values.clone(),
+                        args: col_and_values,
                     }));
                 }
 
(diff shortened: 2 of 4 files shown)


max-sixty:
This looks great! Thanks @PrettyWood and welcome to PRQL...

I'll ask @aljazerzen to have a look before merging. (We could also add a changelog + update docs, but that can be a separate PR.)

PrettyWood:
> welcome to PRQL

@max-sixty Thank you! This project is really amazing! Congrats for all the hard work put on this library
We are considering PRQL to rewrite part of our query generation tool in our company! I am planning to spend quite some time on the internals of PRQL to deeply understand how it works and hopefully be able to help a bit on some other features.

> This looks great! 

Does it? I really don't like my `in_values.insert(0, value);` where I inject the column name expression in the args to use them back afterwards. I feel like it's super hacky and there is probably a (way!) better way of achieving this.
This is why I opened it as draft. I'm still reading the code and try to understand the whole transformation workflow
(btw thank you for the very clear ARCHITECTURE.md)

max-sixty:
> Does it? I really don't like my `in_values.insert(0, value);` where I inject the column name expression in the args to use them back afterwards. I feel like it's super hacky and there is probably a (way!) better way of achieving this

:)

Yeah, I somewhat agree — but the hackiness is limited in scope, and I actually didn't know we could do this with such little code. I would generally favor adding small amounts of code like this when it moves the functionality forward, though pausing before we implement the whole compiler this way. Possibly @aljazerzen has different views (or has an elegant way of doing it!)

aljazerzen:
> I really don't like my `in_values.insert(0, value);`

This is fine :D If you want more idiomatic Rust, some of these might look better:

```rust
let mut new_vals = Vec::new();
new_vals.push(value);
new_vals.extend(in_values);

let new_vals = [vec![value], in_values].concat();

let new_vals: Vec<_> = Some(value)
    .into_iter()
    .chain(in_values.into_iter())
    .collect();
```

prqlc/prql-compiler/src/semantic/resolver/transforms.rs
```diff
@@ -238,6 +238,14 @@ impl Resolver<'_> {
 
                 let [pattern, value] = unpack::<2>(func.args);
 
+                if let ExprKind::Array(mut in_values) = pattern.kind {
+                    in_values.insert(0, value);
+                    return Ok(Expr::new(ExprKind::RqOperator {
+                        name: "std.in".to_string(),
+                        args: in_values.clone(),
+                    }));
```
aljazerzen: This clashes with what I envisioned `RqOperator` to be (but I don't think this is documented anywhere, so bear with me):

- they should not be polymorphic (their parameters should take only a single primitive type),
- they should have a fixed number of arguments.

Note that `std.in` in `std.prql` is a PRQL function, while `name: "std.in"` you list here refers to the "RQ operator" (RQ means Relational Query Intermediate Representation).

The idea is to have rich, polymorphic PRQL functions (`std.in` can take a range or an array) that are resolved to simple RqOperators that are easy to implement.

Your work is in line with this, but I would ask for two changes:

1. rename the RqOperator to `std.array_in` or `std.array_contains` or something similar.
  This way, we will not be tempted to squeeze any other types into its parameters.

2. make the number of arguments fixed. This is hard to do, since we want arrays to have a variable number of arguments and because we don't have array in the RQ. Here I would suggest:
   - to add `rq::ExprKind::Array`, https://github.com/prql/PRQL/blob/main/prqlc/prql-compiler/src/ir/rq/expr.rs#L20,
   - lower `pl::ExprKind::Array` into `rq::ExprKind::Array` here: https://github.com/prql/PRQL/blob/main/prqlc/prql-compiler/src/semantic/lowering.rs#L886
   - unpack the `rq::ExprKind::Array` in the `process_in_values` function,
   - in any other place in SQL backend, reject the `rq::ExprKind::Array` saying "unsupported feature"

   This is a bit of work, so I'm fine with merging this PR with a comment saying "TODO: instead of unpacking the array, compile it to rq::ExprKind::Array".

aljazerzen (approved):
Looks good, nice work!

prqlc/prql-compiler/src/semantic/resolver/transforms.rs
PrettyWood: > This is fine :D If you want more idiomatic Rust, some of these might look better

Yes totally! I first wanted your input on the prepended column expression in args. Glad it's ok as a first step

> This is a bit of work, so I'm fine with merging this PR with a comment saying "TODO: instead of unpacking the array, compile it to rq::ExprKind::Array".

Thank you for the very detailed explanation. Much clearer to me.
I changed the `std.in` into `std.array_in` already to be more explicit and added the TODO comment.
I'm more than happy to try to work on the second point this week in another PR

eitsupi (commented):
Awesome work!

If you don't mind, could you write about this change in the Changelog? (Or someone can do it later)

PrettyWood:
Thank you @eitsupi! Please tell me if the changelog looks ok to you

CHANGELOG.md
```diff
@@ -8,6 +8,8 @@
 
 **Features**:
 
+- The filter `std.in` now supports a list of values (@PrettyWood, #3883)
```
eitsupi: Maybe ``The `std.in` function`` is more better?
(I think "filter" is ambiguous in this context.)

PrettyWood: fixed :)

eitsupi: Thanks!

prqlc/prql-compiler/src/semantic/resolver/transforms.rs
PrettyWood: Just to be sure. You would want something like
```rs
Ok(Expr::new(ExprKind::RqOperator {
    name: "std.array_in".to_string(),
    args: vec![value, Expr::new(ExprKind::Array(in_values))],
}));
```

instead of 

```rs
Ok(Expr::new(ExprKind::RqOperator {
    name: "std.array_in".to_string(),
    args: col_and_values,
}));
```

right?

PR #3884, opened 2023-12-04
refactor: have fixed number of args for `RqOperator`

PrettyWood:
following @aljazerzen advice in [#3883](https://github.com/PRQL/prql/pull/3883#discussion_r1413535136) here is the follow up PR

> make the number of arguments fixed. This is hard to do, since we want arrays to have a variable number of arguments and because we don't have array in the RQ. Here I would suggest:
>        
>        * to add `rq::ExprKind::Array`, https://github.com/prql/PRQL/blob/main/prqlc/prql-compiler/src/ir/rq/expr.rs#L20,
>        * lower `pl::ExprKind::Array` into `rq::ExprKind::Array` here: https://github.com/prql/PRQL/blob/main/prqlc/prql-compiler/src/semantic/lowering.rs#L886
>        * unpack the `rq::ExprKind::Array` in the `process_in_values` function,
>        * in any other place in SQL backend, reject the `rq::ExprKind::Array` saying "unsupported feature"
>        
>        This is a bit of work, so I'm fine with merging this PR with a comment saying "TODO: instead of unpacking the array, compile it to rq::ExprKind::Array".

2 questions:
1) is it the right approach?
2) if yes what should I do with the failing test? Should I change `infer_complexity_expr` to return a `Result` in order to propagate a clean `Err(Error::new(Reason::Unexpected ...))` upstream?

This test now fails
```rs
#[test]
fn array_instead_of_tuple() {
    // Particularly given this used to be our syntax, this could be clearer
    // (though we do say so in the message, which is friendly!)
    assert_display_snapshot!(compile(r###"
    from e=employees
    select [e.first_name, e.last_name]
    "###).unwrap_err(), @r###"
    Error:
       ╭─[:3:12]
       │
     3 │     select [e.first_name, e.last_name]
       │            ─────────────┬─────────────
       │                         ╰─────────────── unexpected `[this.e.first_name, this.e.last_name]`
       │
       │ Help: this is probably a 'bad type' error (we are working on that)
    ───╯
    "###);
}
```
since we now `panic!`

```console
---- bad_error_messages::array_instead_of_tuple stdout ----
thread 'bad_error_messages::array_instead_of_tuple' panicked at 'unsupported expr kind: array', prqlc/prql-compiler/src/sql/srq/anchor.rs:570:35
```

### DIFF
--- a/prqlc/prql-compiler/src/sql/gen_expr.rs
+++ b/prqlc/prql-compiler/src/sql/gen_expr.rs
@@ -117,6 +117,7 @@ pub(super) fn translate_expr(expr: Expr, ctx: &mut Context) -> Result<ExprOrSour
             }
             super::operators::translate_operator_expr(expr, ctx)?
         }
+        ExprKind::Array(_) => panic!("unsupported expr kind: array"),
     })
 }
 
@@ -149,18 +150,23 @@ fn process_null(name: &str, args: &[Expr], ctx: &mut Context) -> Result<sql_ast:
 
 /// Translates into IN (v1, v2, ...) if possible
 fn process_array_in(args: &[Expr], ctx: &mut Context) -> Result<sql_ast::Expr> {
-    let col_expr = args.first().expect("The column expr is always prepended");
-    let expr = Box::new(translate_expr(col_expr.clone(), ctx)?.into_ast());
-    let list: Vec<sql_ast::Expr> = args
-        .iter()
-        .skip(1) // skip the column expr
-        .map(|a| Ok(translate_expr(a.clone(), ctx)?.into_ast()))
-        .collect::<Result<Vec<_>>>()?;
-    Ok(sql_ast::Expr::InList {
-        expr,
-        list,
-        negated: false,
-    })
+    match args {
+        [col_expr @ Expr {
+            kind: ExprKind::ColumnRef(_),
+            ..
+        }, Expr {
+            kind: ExprKind::Array(in_values),
+            ..
+        }] => Ok(sql_ast::Expr::InList {
+            expr: Box::new(translate_expr(col_expr.clone(), ctx)?.into_ast()),
+            list: in_values
+                .iter()
+                .map(|a| Ok(translate_expr(a.clone(), ctx)?.into_ast()))
+                .collect::<Result<Vec<sql_ast::Expr>>>()?,
+            negated: false,
+        }),
+        _ => panic!("args to `std.array_in` must be a column ref and an array"),
+    }
 }
 
 fn process_concat(expr: &Expr, ctx: &mut Context) -> Result<sql_ast::Expr> {
--- a/prqlc/prql-compiler/src/sql/gen_expr.rs
+++ b/prqlc/prql-compiler/src/sql/gen_expr.rs
@@ -14,7 +14,7 @@ use crate::ir::pl::{self, Ident, Literal};
 use crate::ir::rq::*;
 use crate::sql::srq::context::ColumnDecl;
 use crate::utils::{OrMap, VALID_IDENT};
-use crate::{Error, Span, WithErrorInfo};
+use crate::{Error, Reason, Span, WithErrorInfo};
 use prqlc_ast::expr::generic::{InterpolateItem, Range};
 
 use super::gen_projection::try_into_exprs;
@@ -117,7 +117,13 @@ pub(super) fn translate_expr(expr: Expr, ctx: &mut Context) -> Result<ExprOrSour
             }
             super::operators::translate_operator_expr(expr, ctx)?
         }
-        ExprKind::Array(_) => panic!("unsupported expr kind: array"),
+        ExprKind::Array(_) => {
+            return Err(Error::new(Reason::Unexpected {
+                found: "array of values (not supported here)".to_string(),
+            })
+            .with_span(expr.span)
+            .into());
+        }
     })
 }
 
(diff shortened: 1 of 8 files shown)


aljazerzen:
1. Yes it is
2. `infer_complexity` is trying to figure out to to materialize an array (i.e. can it appear in a GROUP BY? Can it appear in WHERE?).
   
   In this PR, we have no plans to compile to actual SQL arrays (since they vary between dialects **a lot**), but just use them as a container when using `std.array_contains`. 

   This means that `infer_complexity` is inconsequential: we can just return the highest complexity, since this is the safest option.

prqlc/prql-compiler/src/sql/gen_expr.rs
```diff
@@ -117,6 +117,7 @@ pub(super) fn translate_expr(expr: Expr, ctx: &mut Context) -> Result<ExprOrSour
             }
             super::operators::translate_operator_expr(expr, ctx)?
         }
+        ExprKind::Array(_) => panic!("unsupported expr kind: array"),
```
aljazerzen: Let's produce a proper error here.

PrettyWood: Done in [752a127](https://github.com/PRQL/prql/pull/3884/commits/752a1279ee01d8198f661516608d1709387672a4). Feedback welcome :)
Thank you for your review

max-sixty: Looks great, thanks!

prqlc/prql-compiler/tests/integration/bad_error_messages.rs
```diff
@@ -90,9 +90,7 @@ fn array_instead_of_tuple() {
        │
      3 │     select [e.first_name, e.last_name]
        │            ─────────────┬─────────────
-       │                         ╰─────────────── unexpected `[this.e.first_name, this.e.last_name]`
-       │
-       │ Help: this is probably a 'bad type' error (we are working on that)
+       │                         ╰─────────────── unexpected array of values (not supported here)
```
max-sixty: FWIW we could now promote this from this file to `error_messages.rs`, as it's now pretty good (but in another PR is fine)

PrettyWood: it costs nothing [4f60434](https://github.com/PRQL/prql/pull/3884/commits/4f604341df408d84ecb6796fe4a0746c130af836) :)

prqlc/prql-compiler/tests/integration/error_messages.rs
```diff
@@ -101,6 +101,24 @@ fn test_errors() {
     "###);
 }
 
+#[test]
+fn array_instead_of_tuple() {
+    // Particularly given this used to be our syntax, this could be clearer
+    // (though we do say so in the message, which is friendly!)
```
max-sixty:
```suggestion
```

PrettyWood: of course [786d90e](https://github.com/PRQL/prql/pull/3884/commits/786d90e262ab5795d5516267144f4d54d7840810)

prqlc/prql-compiler/src/sql/srq/anchor.rs
```diff
@@ -566,7 +566,8 @@ pub fn infer_complexity_expr(expr: &Expr) -> Complexity {
         rq::ExprKind::ColumnRef(_)
         | rq::ExprKind::Literal(_)
         | rq::ExprKind::SString(_)
-        | rq::ExprKind::Param(_) => Complexity::Plain,
+        | rq::ExprKind::Param(_)
+        | rq::ExprKind::Array(_) => Complexity::Plain,
```
aljazerzen: With "highest complexity" I meant `Aggregation`, not `Plain`. Although I don't think it matters at this stage.

It would matter if we supported things like this:

```elm
    from employees
    derive allowed_titles = case [
        is_guest => ["Sales Manager"],
        true => ["Sales Manager", "Sales Support Agent"],
    ]
    group (title | in allowed_titles) (aggregate {count this})
```

Now `allowed_titles`, is a proper column in the relation and we want to prevent this:

```
SELECT COUNT(*)
FROM employees
GROUP title IN (CASE
   WHEN is_guest THEN ARRAY["Sales Manager"]
   ELSE ARRAY["Sales Manager", "Sales Support Agent"]
END)
```
... because I don't think having `CASE` in `GROUP` is supported.

Setting complexity to be highest possible (`Aggregation`) would force the pipeline to be split, so `allowed_titles` is defined in a CTE, assigned an alias which is used in the next CTE in GROUP BY.

aljazerzen (approved):

prqlc/prql-compiler/src/sql/srq/anchor.rs
PrettyWood: [ed9c135](https://github.com/PRQL/prql/pull/3884/commits/ed9c135686f985b05af0de66d384130ecde185a3)
Right! Now I understand sorry 😬
Thanks for the clear explanation

prqlc/prql-compiler/tests/integration/error_messages.rs
max-sixty: Thanks! (FYI you can also hit "Commit suggestion" in GH and it'll add a commit with the change for you :) )

PrettyWood:
Last question on the whole "filter on list of values"
Should I add an example in `web/book/src/reference/stdlib/transforms/filter.md` ?
I feel like it's as important as the filtering in range

aljazerzen:
Docs would be great, yes. I don't see `std.in` as only to be used with filtering, but to compute a boolean in general. So I would make a new page under `stdlib/` and maybe link to it from `filter.md`.

PrettyWood:
Ok. Should I do that in this PR?

aljazerzen:
As you wish. Our only policy on commits is "is it an improvement?", regardless of how big or small.

aljazerzen:
I've given you write access to this repo, merge when you think it's done.
(you might have to accept it via email)

PrettyWood:
Wow thanks! I will merge this PR and open a new one dedicated to documentation

PR #3892, opened 2023-12-05
docs: Fix syntax in example

k-nut:
Since release 0.9.0, tuples are enclosed in braces, not brackets. This commit adapts the example to be valid PRQL with the latest version again.
Additionally, since 0.9.0 `count` requires and argument which is also added to the example.


aljazerzen:
Thank you.

I guess we've missed this when we changed the syntax: https://github.com/PRQL/prql/pull/2612/files#diff-2e40721d10c62757d1bd6be7a6a5f71ee4fa1bdea2fb51c8dfc068a29bb7ad81

Quickly looking at the posts, some of the others also contain outdated syntax.

PR #3896, opened 2023-12-05
feat(Lezer grammar): Highlight DateTime

vanillajonathan:
This maps the `DateTime` token to the Lezer highlighting tag [`color`](https://lezer.codemirror.net/docs/ref/#highlight.tags.color). Lezer doesn't have any highlighting tag for dates or times, but we definitely do want to somehow highlight the `DateTime` token so we'll just pick an relatively arbitrary one. The `color` tag belongs to the set of literal tags. In the real world the end result is an improvement.

### DIFF
--- a/grammars/prql-lezer/src/highlight.js
+++ b/grammars/prql-lezer/src/highlight.js
@@ -12,6 +12,7 @@ export const prqlHighlight = styleTags({
   Boolean: t.bool,
   Integer: t.integer,
   Float: t.float,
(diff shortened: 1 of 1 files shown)


max-sixty:
OK! I guess we can refine when we start using this more. Possibly this should appear similar to a string (I doubt there should be a specific color for dates, but let's see)

Would be good to add a comment in the code for this sort of thing, so it's clear that it's arbitrary.

PR #3900, opened 2023-12-06
refactor(Lezer grammar): Rename TypeTerm

vanillajonathan:
This PR renames the `TypeTerm` token to `TypeName` to align with upstream grammars.

Upstream JavaScript and Java grammars names it `TypeName` while the Rust and C++ grammars names it `TypeIdentifier`.

`TypeName` is nice because it maps nicely to the Lezer highlighting tag named [`typeName`](https://lezer.codemirror.net/docs/ref/#highlight.tags.typeName).


max-sixty:
Yes agree!

PR #3901, opened 2023-12-07
feat(Lezer grammar): Add parameters

vanillajonathan:
This PR adds support for parameters, i.e. `$1` through a new token named `Parameter`. I am unsure whether it should be mapped as the Lezer highlighting tag [`processingInstruction`](https://lezer.codemirror.net/docs/ref/#highlight.tags.processingInstruction) or `t.special(t.variable)` or even something else, I guess it doesn't matter that much. Either way it improves the grammar as it fills a gap.


max-sixty:
Great! 

For the tags — yeah, none of the tags seem ideal, `processingInstruction` included. Maybe we leave them as blander options, even if they overlap, and then refine later.

I'll merge now though, since the grammar improvement is 👍

PR #3902, opened 2023-12-07
docs: add info on std.in

PrettyWood:
Following https://github.com/PRQL/prql/pull/3884#issuecomment-1840146416
Feedback welcome

Screenshot 2023-12-07 at 14 56 40

Screenshot 2023-12-07 at 14 56 57


web/book/src/reference/stdlib/transforms/filter.md
```diff
@@ -18,7 +18,16 @@ from employees
...
+filter (department | in ["IT", "HR"])
+```
+
 ```prql
 from employees
 filter (age | in 25..40)
 ```
+
+```admonish note
+You can also use `in` operator to [declare variables or functions](../compute)
```
max-sixty: I think maybe this is a bit superfluous, or possibly we could phrase it differently. Because any function can be used as part of a variable or function, no? There's nothing specific to `in`?

web/book/src/reference/stdlib/transforms/filter.md
```diff
@@ -18,7 +18,16 @@ from employees
 filter (age > 25 || department != "IT")
 ```
 
+```prql
+from employees
+filter (department | in ["IT", "HR"])
```
max-sixty, line 23: 👍

web/book/src/reference/stdlib/compute/README.md
```diff
@@ -0,0 +1,19 @@
...
+
+from employees
+select {
+    first_name,
+    last_name,
+    hobby,
+    adult = is_adult age,
+}
+filter (is_cool hobby)
+```
```
max-sixty: Very good example! (Though maybe the irony of `is_cool` might be lost on some! Maybe we make that blander like "writes code" or something...)

I think we could put this at the top of https://prql-lang.org/book/reference/declarations/functions.html, rather than make a new page?

web/book/src/reference/stdlib/transforms/filter.md
PrettyWood: I tend to agree. I tried to follow @aljazerzen comment but maybe I didn't understand properly what was expected

web/book/src/reference/stdlib/compute/README.md
PrettyWood: Haha yeah there is always a risk with jokes. Ok I'll do that 👍

PrettyWood: Update done

max-sixty:
Very nice, thank you @PrettyWood !

Issue #1775, opened 2023-02-11
[book] Standard Library page: expand explanation & give examples


richb-hanover:
The [Standard Library page](https://prql-lang.org/book/language-features/standard-library.html) of the Book offers a list of the functions that are available.

But I find the notation used on that page fairly impenetrable. For example: `func min <scalar|column> column -> null` - I can guess that `min` takes a scalar or column, but what's the second "column" doing there? And does it matter if it returns `null`? And what about `intersect` - what's going on there?

Perhaps that page could include a tutorial on the notation for the function definitions, and maybe a few more examples. Thanks.

Issue #1761, opened 2023-02-10
[book] Document the standard library


vanillajonathan:
https://prql-lang.org/book/language-features/standard-library.html

The standard library page in the documentation just includes the source code for `std.prql`.

I would like fuller documentation; a paragraph for each function that describes what it does.

max-sixty:
Agree. This would make a nice contribution. Especially since we added types to the stdlib, it's not really in a user-presentable form.

Even if we just had a table of functions and their SQL equivalent, that would be much nicer...

vanillajonathan:
Yes, I am unable to read and understand the `std.prql` file.

I see some stuff like `from<table>` and I was thinking those are generics? But I have a feeling they're not. I am confused.

max-sixty:
The important parts of the stdlib to explain are the functions — `min` / `max` etc. So if someone wanted to make a table of those, that would be welcome. The `foo<table>` is a type annotation; that's not needed to make a contribution here.

vanillajonathan:
This type annotation syntax is very foreign to me, and I can imagine to others as well. In my head that syntax is for generics.

aljazerzen:
The long-term solution would be to introduce "structured comments" to the language. Something like `///` in Rust or `/**` in Java. These comments could then be extracted and converted to HTML which would be inserted into the book (or be a standalone site), similar to what rustdoc are doing javadoc.

I don't think we have the manpower to do this now, but as a start, we can document stdlib *in stdlib.prql*:

```elm
## Finds minimum value of a column
##
## Example:
## ```
## from tracks
## select (min milliseconds)
## ```
func min <scalar || column> column -> null
```

Issue #217, opened 2022-03-20
Fill out stdlib


max-sixty:
We have a [stdlib](https://github.com/max-sixty/prql/blob/main/src/stdlib.prql), written by @qorrect , which provides some standard functions by using s-strings.

We could fill this out with some more standard functions — e.g. `EXP`, `MOD`, string functions, etc. 

Possibly there is even a definition of these somewhere?

aljazerzen:
The problem here is that different dialects have different functions (see datetime and such). I tried to find SQL standard (ISO/IEC 9075) in the parts of the standard I manged to find there was no mentions of functions like CONCAT. This leads me to think there is no explicit standard...

@qorrect @max-sixty 
Does anyone know if SQL has a list of standard functions defined somewhere?

For dialect specific stdlib, we can extract list of all function from docs (or even source code?), as said in #301.

max-sixty:
The best reference I can find is http://users.atw.hu/sqlnut/sqlnut2-chp-4-sect-4.html, but it doesn't strictly give ANSI functions.

There's also http://www.contrib.andrew.cmu.edu/~shadow/sql/sql1992.txt, but that only seems to give 5 functions!

```

         Function

         Specify a value derived by the application of a function to an
         argument.

         Format

         <set function specification> ::=
                COUNT <left paren> <asterisk> <right paren>
              | <general set function>

         <general set function> ::=
                <set function type>
                    <left paren> [ <set quantifier> ] <value expression> <right paren>


         <set function type> ::=
              AVG | MAX | MIN | SUM | COUNT

         <set quantifier> ::= DISTINCT | ALL
```

So possibly there isn't a way of doing this systematically and we should just build out the functions that people want, maybe with something like Postgres' manual as a guide...

max-sixty:
Ah, it's possible to buy (?!): https://www.iso.org/standard/63555.html

There are also lists of "Reserved words": https://en.wikipedia.org/wiki/SQL_reserved_words, but that's a large superset of functions

aljazerzen:
The two links you've sent are quite old and they don't seem exhaustive, but may serve as a checklist of things to add.

9075 has 11 parts (9075-1, 9075-2, 9075-3, ...). I don't know which (if any) contains a list of functions, since I've only managed to get my hands on 9075-1.

Possibly, we may be able to find it in genesis library, but I cannot comment on that :D

It seems like scraping docs of postres would be the most efficient way of doing this...

max-sixty:
[...]

PR #3909, opened 2023-12-09
feat: add main mathematical functions

PrettyWood:
With all the analytics DB, working on math and dates operations is key.
I will try to work on date / time / durations... types but still need some time for onboarding.
In the meantime if you're ok with this, I would like to first add math functions that are available for almost every dialects. In another PR I'll have a deeper look at string functions and I reckon it should cover pretty much https://github.com/PRQL/prql/issues/217 was opened for.
I'll work in a next PR on adding documentation for https://github.com/PRQL/prql/issues/1761 (similar to https://github.com/PRQL/prql/issues/1775)


prqlc/prql-compiler/src/sql/std.sql.prql
```diff
@@ -60,6 +60,34 @@ let rank_dense = -> s"DENSE_RANK()"
...
+let sqrt = column -> s"SQRT({column:0})"
+let degrees = column -> s"DEGREES({column:0})"
+let radians = column -> s"RADIANS({column:0})"
+let cos = column -> s"COS({column:0})"
+let acos = column -> s"ACOS({column:0})"
+let sin = column -> s"SIN({column:0})"
+let asin = column -> s"ASIN({column:0})"
+let tan = column -> s"TAN({column:0})"
+let atan = column -> s"ATAN({column:0})"
+let pow = column exponent -> s"POW({column:0}, {exponent:0})"  # TODO: create an operator for that (** or ^ ?)
```
PrettyWood: If you're ok with this PR, I reckon we should have some kind of operator for `pow`
Which one would you prefer?

max-sixty (approved):
Super — this is v helpful, we should have done this a while ago.

They all look fine to me. I guess folks aren't using all of these (`atan`!) _that_ much, but no harm in having them...

prqlc/prql-compiler/src/sql/std.sql.prql
max-sixty: Yes agree. No particularly strong view on `^` vs `**`, WDYT? (our parser can cope with the initial ambiguity of `**` fine)

PrettyWood: I personally would go with `**` for 3 reasons
- it's in python which is super used in data world (also in JS)
- `^` makes me think of bitwise XOR in Rust and Python
- ~It feels more comfortable to type on a QWERTY ^^ I prefer to type twice a key than use shift~ nvm I'm stupid

max-sixty: (Maybe `**` uses up fewer unique characters, in case we want to use `^` for something else)

Edit: I just saw your reply, OK — let's go with `**`...

max-sixty:
In another PR, we could add some of these to our integration tests. I think they're simple enough that I'm fairly confident they work, but nice to have coverage of them across dialects. A couple of times we've found inconsistencies between dialects this way.

If you also want to add them to the unit tests and/or docs, that would be good. Again, am quite confident they work, so not a top priority.

PrettyWood:
@max-sixty I think it's ready for review :)

PrettyWood:
@max-sixty I should be good to review :)

prqlc/prql-compiler/tests/integration/sql.rs
```diff
@@ -38,6 +38,58 @@ fn test_stdlib() {
...
+      ACOS(salary) AS salary_acos,
+      SIN(salary) AS salary_sin,
+      ASIN(salary) AS salary_asin,
+      TAN(salary) AS salary_tan,
+      ATAN(salary) AS salary_atan,
+      POW(salary, 2) AS salary_pow1,
+      POW(salary, 2) AS salary_pow2
+    FROM
+      employees
+    "#
```
max-sixty, line 87: Nice!

max-sixty:
Great re adding `**` here!

We need to set the associativity here: https://github.com/PRQL/prql/blob/de975cf6807c5f5026f39f224e573b6e604ff704/prqlc/prql-compiler/src/sql/gen_expr.rs#L890-L896

It should be right associative (the only one). And we should add a test that the precedence works; `5 * 4 ** 3 ** 2` should be equivalent to `5 * (4 ** 9)`. (it's OK if it passes it through to SQL; it's important it doesn't insert parentheses that shouldn't be there)

If you prefer to push that to another PR, that's totally fine; we should just comment it out from the parser so it doesn't work yet.

CHANGELOG.md
```diff
@@ -9,6 +9,10 @@
 **Features**:
 
 - The `std.in` function now supports a list of values (@PrettyWood, #3883)
+- Most standard mathematical functions are now supported: `abs`, `floor`,
+  `ceil`, `pi`, `exp`, `ln`, `log10`, `log`, `sqrt`, `degrees`, `radians`,
+  `cos`, `acos`, `sin`, `asin`, `tan`, `atan` and `pow`.<br/>Note: `pow` can be
```
max-sixty: If you have the patience to make a change — we can just put this as a new item, no need to add the HTML tag here. But no big stress.

eitsupi: I think `\` can be used instead of `<br/>` here for soft line breaking.

PrettyWood: I could use double spaces at the end for a soft line breaking but some linters remove them.
I just added a new item as suggested :)

PrettyWood:
> We need to set the associativity

Very good point I totally forgot to test that! I'll update

PrettyWood:
> It should be right associative

You're completely right. I'll work on it tonight though I'll have to do some extra work because it's not a `BinaryOperator` (not available in SQL parser)

CHANGELOG.md
eitsupi: > I could use double spaces at the end for a soft line breaking but some linters remove them.

Pls use `\`

PrettyWood: Yup sorry done in 1659506

PrettyWood:
@max-sixty So after some dive in I'm a bit perplex on how I should make the right associativity work.
What we want is for
```
c ** a ** b
```
to be translated into
```
POW(c, POW(a, b))
```
instead of
```
POW(POW(c, a), b)
```
Since there is no "power" operator in Postgres, we can't delegate that to the DB and there is nothing on sqlparser that handles that directly. I tried to use `BinaryOperator::Custom(String::from("**"))` for example and set right associativity on it, it doesn't work.
IMO what we want is for `c ** a ** b` to be translated into `(c ** a) ** b` before that but I would need some guidance 😬

max-sixty:
OK, good point... this is actually a bit harder than I thought — it's not only adjusting the SQL writer, it's also adjusting the parser here: https://github.com/PRQL/prql/blob/054d29e06b5ff7c830e45bfd383c141db5a3436f/prqlc/prqlc-parser/src/expr.rs#L250, I think.

(Previously everything had been left-associative, so all the work I'd done here had been on the SQL writer)

I can come back to this and look more. In the meantime, can we comment out the change in the parser so `**` doesn't work yet? We can leave the rest of the code there so it'll flow through when we do fix it.

prqlc/prqlc-parser/src/lexer.rs
```diff
@@ -35,6 +35,7 @@ pub enum Token {
     Or,          // ||
     Coalesce,    // ??
     DivInt,      // //
+    Pow,         // **
```
max-sixty, line 38: If we just comment this out, but leave literally everything else, I think that'll be fine. (I do think we should prioritize correctness, so I would vote against merging `**` when it doesn't get the associativity right)

max-sixty: If we just comment this out, but leave almost everything else, I think that'll be fine. (I do think we should prioritize correctness, so I would vote against merging ** when it doesn't get the associativity right)

It's fine to leave it in the syntax files etc


```suggestion
//    Pow,         // **
```

web/book/src/reference/syntax/operators.md
```diff
@@ -26,13 +26,14 @@ operations and for function calls (see the discussion below.)
...
 |          unary | `-` `+` `!` `==`            |     2      |               |
 |          range | `..`                        |     3      |               |
-|            mul | `*` `/` `//` `%`            |     4      | left-to-right |
-|            add | `+` `-`                     |     5      | left-to-right |
-|        compare | `==` `!=` `<=` `>=` `<` `>` |     6      | left-to-right |
-|       coalesce | `??`                        |     7      | left-to-right |
-|            and | `&&`                        |     8      | left-to-right |
-|             or | <code>\|\|</code>           |     9      | left-to-right |
-|  function call |                             |     10     |               |
+|            pow | `**`                        |     4      |               |
```
max-sixty:
```suggestion
<!-- |            pow | `**`                        |     4      |               | -->
```

grammars/prql-lezer/src/prql.grammar
```diff
@@ -165,7 +165,7 @@ LambdaParam { identPart TypeDefinition? (":" expression)? }
   Comment { "#" ![\n]* }
   @precedence { Docblock, Comment }
 
-  //op_bin_only { "*" | "/" | "//" | "%" | "!=" | ">=" | "<=" | "~=" | ">" | "<" | "??" | "&&" | "||" }
+  //op_bin_only { "*" | "/" | "//" | "%" | "**" | "!=" | ">=" | "<=" | "~=" | ">" | "<" | "??" | "&&" | "||" }
```
max-sixty: @vanillajonathan should we remove this line?

max-sixty (approved):

PrettyWood:
Ok I removed the `**` operator I think we're good!
One last remark: should I write `pow` as `pow = exponent column` instead of `pow = column exponent` ?
I'm a bit torn between:
- SQL that uses `POW(col, 2)`
- PRQL where I would prefer to write `derive { pow_col = col | pow 2 }`

max-sixty:
> PRQL where I would prefer to write `derive { pow_col = col | pow 2 }`

I think this is correct — the `col` generally goes last, because piping puts the argument last, and is the thing that's normally piped.

We do something similar for `round` etc...

prqlc/prql-compiler/src/sql/std.sql.prql
```diff
@@ -168,6 +201,12 @@ module mssql {
   @{binding_strength=11}
   let div_f = l r -> s"({l} * 1.0 / {r})"
 
+  # Mathematical functions
+  # https://learn.microsoft.com/en-us/sql/t-sql/functions/mathematical-functions-transact-sql
+  let ceil = column -> s"CEILING({column:0})"
+  let ln = column -> s"LOG({column:0})"
+  let pow = column exponent -> s"POWER({column:0}, {exponent:0})"
```
max-sixty, line 208: So I guess we swap here... (sorry, should have spotted this before)

Then feel free to merge!

PrettyWood: Perfect I do that

max-sixty: OK cool — so we should swap these in another PR, to `pow = exponent column`, so then `foo | pow 2` compiles to `POWER(foo, 2)`; is that consistent with what you were thinking? (I think maybe we only did the swap at https://github.com/PRQL/prql/pull/3909/files#diff-d948a1b9efe8b58ca00f24357849721f96c0957d4555d28937b148491220c29eR89 ?)

PrettyWood: FK sorry I did the modification but forgot the mssql overload ;( I'll fix that

PrettyWood: https://github.com/PRQL/prql/pull/3911

vanillajonathan:
It would be pretty convenient if all the math functions could be grouped together under a math namespace.

Like `Math` in JavaScript.
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Math

PrettyWood:
Yup @vanillajonathan I thought about it and agree with you I'd like that. We could have the same for string utils.
This would be a bit more explicit and avoid name collisions

max-sixty:
Yes, I agree...

PR #3913, opened 2023-12-10
feat: add string functions in a `string` module

### LINKED ISSUE #3238: PRQL equivalent of `ilike`
### What's up?

Hi, pqrl looks great. Thank you.

How would I go about implement the equivalent of SQL`ilike` or `like` operators?

In particular I want simple string contains, e.g. `table.foobar ilike '%foo%'`.

I've looked through the docs and searched the code, but can't find an obvious solution. Obviously I can use a regex, but that will be much less performant than a simple "string contains" query.

PrettyWood:
Continue work on #217
closes #3238

Screenshot 2023-12-12 at 19 01 41


prqlc/prql-compiler/src/semantic/std.prql
```diff
@@ -215,6 +215,15 @@ let from_text = input<text> `noresolve.format`:csv -> <relation> internal from_t
 ## String functions
 let lower = column -> <text> internal std.lower
 let upper = column -> <text> internal std.upper
+let ltrim = column -> <text> internal std.ltrim
+let rtrim = column -> <text> internal std.rtrim
+let trim = column -> <text> internal std.trim
+let length = column -> <int> internal std.length
```
max-sixty: Fine for now, but eventually for `length` probably we want it within a `string` namespace; given how general the term is

prqlc/prql-compiler/src/sql/std.sql.prql
```diff
@@ -93,8 +93,25 @@ let round = n_digits column -> s"ROUND({column:0}, {n_digits:0})"
...
 let upper = column -> s"UPPER({column:0})"
+let ltrim = column -> s"LTRIM({column:0})"
+let rtrim = column -> s"RTRIM({column:0})"
+let trim = column -> s"TRIM({column:0})"
+let length = column -> s"CHAR_LENGTH({column:0})"
+let substring = offset length column -> s"SUBSTRING({column:0}, {offset:0}, {length:0})"
+let replace = pattern replacement column -> s"REPLACE({column:0}, {pattern:0}, {replacement:0})"
+let starts_with = prefix column -> s"{column} LIKE CONCAT({prefix}, '%')"
+let contains = substr column -> s"{column} LIKE CONCAT('%', {substr}, '%')"
+let ends_with = suffix column -> s"{column} LIKE CONCAT('%', {suffix})"
```
max-sixty: These are nice functions to have.

But I think these ones are likely to only work with dialects that use single quotes and `CONCAT`.

Instead, could we use an [f-string](https://prql-lang.org/book/reference/syntax/f-strings.html) to construct the pattern, and then I think it'll work for everything? (but would need to check)

prqlc/prql-compiler/tests/integration/sql.rs
```diff
@@ -3255,7 +3291,7 @@ fn test_case() {
         r###"
     from tracks
     select category = case [
-        length > avg_length => 'long'
+        this.length > avg_length => 'long'
```
max-sixty: 👍

prqlc/prql-compiler/src/semantic/std.prql
PrettyWood: Yes I'm planning to first open a PR to add the `math` module
And for this one I plan to add a `str` module
Would that be ok?

So we would have `derive { x | str.lower | str.endswith "foo" }`

prqlc/prql-compiler/src/sql/std.sql.prql
PrettyWood: I wanted to do that but how do you mix f-strings and s-strings? 😅

prqlc/prql-compiler/src/semantic/std.prql
PrettyWood: A module `str` has been added. Do you prefer `string`?

max-sixty: Perfect re a `str` module!

No particularly strong view on `str` vs `string`. Possibly we should pick the one that is _not_ the name of the type.

I'd also previously thought that "string" was a bit too "engineer-y", and most folks think a string is something used for tying. So we currently have `from_text`, for example. But I'm not sure we're going to win that battle! So I'm fine with any of those — `str` / `string` / `text`.

Any thoughts from others?

prqlc/prql-compiler/src/sql/std.sql.prql
max-sixty: Ah good point, that would require https://github.com/PRQL/prql/pull/3280. Then we could have something like `s"{column LIKE f'%{suffix}'}"`.

Can we maybe only add to the dialects that support it? Or that's a lot of work? I would emphasize moving quickly, but also we should try not to have things that we know don't work...

We could also leave those ones commented out, and I could take a another go at #3280 (but not sure whether #3280 is easy or difficult)

PrettyWood: I would love to have #3280 implemented but I'm afraid it's a lot of work.
I took a look at all the the supported dialects, only SQLite doesn't support `CONCAT` so I'll go with [080c120](https://github.com/PRQL/prql/pull/3913/commits/080c120efc67191c1d8f6cdfcdca59dddd79d4c5) if that's ok for you

max-sixty: Looks great! Thanks for filling those out

prqlc/prql-compiler/tests/integration/sql.rs
```diff
@@ -166,13 +177,34 @@ fn test_stdlib_string() {
...
+      name LIKE CONCAT('%', 'pika') AS name_ends_with
+    FROM
+      employees
+    "#
+    );
+}
+
+#[rstest]
+#[case::generic(sql::Dialect::Generic, "LIKE CONCAT('%', 'pika', '%')")]
+#[case::sqlite(sql::Dialect::SQLite, "LIKE '%' || 'pika' || '%'")] // `CONCAT` is not supported in SQLite
```
max-sixty, line 192: Interesting. I had heard of `rstest` but haven't used it before. (I use pytest a lot, which I think is similar)

Happy to try this! (Maybe @aljazerzen has thoughts?)

(alternatively, we could have a normal loop and supply different names for the snapshot name)

PrettyWood: I'm a huge fan of pytest and the parametrize + fixtures mechanisms.
And I'm a huge fan of rstest for the exact same reasons. It is quite similar and I feel like it would be a great value to cover many dialects since prql will probably need more tests to ensure compatibility across all the supported DBs. In my company we support 5 SQL databases/DWs and everything is parametrized on that.
But I can definitely remove that if you prefer

max-sixty: Very happy to try it for a while at least!

aljazerzen: I'm +1 for rstest - on the macro-expansion level it does basically the same thing as `for_each_file!`, but offers much more flexibility.

aljazerzen (approved):
Nice work.

Again, it would be nice to also have test-dbs tests.

PrettyWood:
@aljazerzen @eitsupi @vanillajonathan @richb-hanover Following [this comment](https://github.com/PRQL/prql/pull/3913#discussion_r1423052597). Sorry to ping you but since you are main contributors on this repo, your opinion matters for the name of the module that would contain main functions on strings
I went with `str` (so usage would be `str.upper`, `str.ends_with`). But we could go with `string` (not a huge fan since it just adds more characters for nothing) or `text` (maybe more enduser friendly I don't know)
Thank you 🙏

eitsupi:
Thanks for mentioning me (and thanks for your great works!).
I don't feel uncomfortable with `str` because I think the abbreviation `str` is pretty common, but `string` might be better for sure.

prqlc/prql-compiler/src/semantic/std.prql
vanillajonathan: It seems like most languages prefer `string` or `String` over `str` and `String` or `text`.

- C# uses `string.Format`
- D uses `std.string.isNumeric("123")`
- Elixir uses `String.first("elixir")`
- F# uses `String.length "Hello"`
- Go uses `strings.ToLower("Hello")`
- Java uses `String.format`
- JavaScript uses `String.fromCharCode(65)`
- Kotlin uses `String.format("Hello")`
- Lua uses `string.lower("Hello")`
- PowerShell uses `[string]::Format("Hello")`
- Python uses `str.format`
- Ruby uses `String::try_convert('Hello')`
- Rust uses `String::from("Hello, world!");`
- Scala uses `String.format("Hello")`
- VB.NET uses `String.Format`

vanillajonathan:
The `substring` function is interesting, perhaps it could be used for indexing and slicing with brackets #1811

richb-hanover:
I'll weigh in to support the full name `string`. Abbreviations become awkward because you don't always remember what to leave out (for example, why is it `ls`?)

PrettyWood:
Ok I'll change for `string` you're already a majority 😄 This name can always be changed later
EDIT: done! @max-sixty do you agree with this last change? If yes I think we're good to merge

max-sixty:
> EDIT: done! @max-sixty do you agree with this last change? If yes I think we're good to merge

Great!

max-sixty:
Thanks @PrettyWood ! These are great. 

Indeed — I think when developing the compiler, it's easy for us to focus too much on "do we have a good algorithm for right-associative precedence", etc — but actually adding an equivalent for `LIKE` is much more important for most users. So these sorts of PRs have lots of impact.

aljazerzen:
As a followup, we might want to reconsider `type text` and `std.from_text`.

max-sixty:
> As a followup, we might want to reconsider `type text` and `std.from_text`.

I am OK with changing these if folks prefer...

vanillajonathan:
Yeah, I think we should change those, because `text` is pretty much only used in the context of SQLite, in the rest of the computing world, it is `string`.

See: https://github.com/PRQL/prql/pull/3913#discussion_r1424049816

snth:
Sorry for coming late to this (I was away for a while) but I would throw in my vote in favour of `text` or `str`. I agree with @PrettyWood that `string` just seems to add extra characters for no benefit.

I also think our earlier discussions around `text` vs `string` still hold and `text` would be my preferred option.

PrettyWood:
For me the big downside of text is bytes vs strings. How do you differentiate them? I agree that often people want strings but not always! and even here there are subtleties like encoding, chars vs graphemes, ...

snth:
I think `text` would pretty much be a synonym for unicode strings. If you want bytes then you would need another type like `bytea`. I lived through the upheaval around this in the Python 2 to 3 transition but it's not something I've ever encountered in database work. Is it much of an issue there?

The closest I came was when working with MS SQL 15 years ago, you would sometimes get unexpected ordering depending on what collation you chose for your database. Is that still much of an issue these days? It's not something I've encountered in a long time so I really can't say.

PrettyWood:
To be honest my interest in PRQL is to simplify the refactoring of a custom DSL we have that translates many steps into Mongo, Polars and many SQL dialects.
If PRQL is first meant for data scientists/analysts who are not developers, maybe choosing `text` is better.
[DuckDB went with that approach as well](https://duckdb.org/docs/sql/functions/char)
I'm not familiar with the primary target of PRQL so I might be biased

vanillajonathan:
Well, `string` is so widely prevalent in the computing world that I would think it probably is familiar even with non-developers such a data scientists and analysts.

aljazerzen:
I strongly agree with @snth: regardless of how it is named, `type string` or `type text` should contain UTF-8 encoded text. For arbitrary bytes there should be another type.

But I'm impartial on the name. A quick survey of exactly 1 person showed that `string` is not familiar to biologists, who tend to know the basics of R. On the other hand, `string` really is prevalent in the computing world.

max-sixty:
A quick survey of a couple of others reinforces that "string" is actually not that well-known. From a program manager at SpaceX (i.e. someone who is close to lots of technical things without writing code herself):

> Not a lot to me. Something about a string of logic, but that's about it for me

Issue #1797, opened 2023-02-11
Add hash functions


vanillajonathan:
### Discussed in https://github.com/PRQL/prql/discussions/1768

<div type='discussions-op-text'>

<sup>Originally posted by **vanillajonathan** February 10, 2023</sup>
Maybe there could be hash functions?

Like [SHA-3](https://en.wikipedia.org/wiki/SHA-3), BLAKE or ChaCha20? 

Example:
```rust
let foo = sha3_512("Hello world!")
```
</div>

max-sixty:
We'd need to:

- figure out the correct interface — do we want a `hash` and then an parameter for the type of `hash`? Or we just offer `sha2` at first, with only a parameter for the value?
- Understand how various dialects implement these. We don't need to cover every single one, but we need to cover most of them for the cases we support
- Write the code in the compiler to get the correct function for each dialect in https://github.com/PRQL/prql/blob/f56b258605fdde961927df3a6b0d6651c89dc647/prql-compiler/src/sql/dialect.rs, and link it to the function.

Brief note:

> let foo = sha3_512("Hello world!")

This would be `let foo = sha_bar "Hello world!"`, no parentheses

PR #3916, opened 2023-12-11
refactor: add a `math` module

PrettyWood:
Screenshot 2023-12-11 at 13 26 52


PrettyWood:
@vanillajonathan Your feedback is more than welcome!

richb-hanover:
@PrettyWood COOL! How do people "include" the `math...` or `str...` modules? (Or are they built-in?)

And probably a little farther afield, how can people define new modules? I know this was planned, but (I think) disabled. Where does it stand now? Thanks again.

vanillajonathan:
I think this looks very promising!

I've mentioned hash functions in #1768 and #1797. There could be a `hash` namespace, example `hash.sha1` and `hash.sha3_512`.

We have two string functions; `upper` and `lower`, those could be in a `string` namespace.

PrettyWood:
@richb-hanover `math` and `str` modules are available automatically since they are in fact `std.math` and `std.str`.
@vanillajonathan I added in #3913 the `str` module that contains `lower`, `upper` and other string related functions
If this whole approach is validated, we could definitely add other modules like `hash`
(personally I really want to work on a module for date + time management afterwards because for BI purposes, it's almost mandatory)

CHANGELOG.md
```diff
@@ -12,6 +12,7 @@
 - Most standard mathematical functions are now supported: `abs`, `floor`,
   `ceil`, `pi`, `exp`, `ln`, `log10`, `log`, `sqrt`, `degrees`, `radians`,
   `cos`, `acos`, `sin`, `asin`, `tan`, `atan` and `pow` (@PrettyWood, #3909)
+- Those mathematical functions are in the `math` module (@PrettyWood, #3916)
```
max-sixty: If you have the patience — let's integrate this with the item above, since they can be handled as one change. Fine to list more than one PR under one change.

max-sixty:
I'm kinda surprised this was so easy to implement! A testament to the module work that @aljazerzen has done!

I'm definitely +1. Let's wait for @aljazerzen to have a look and then merge!

CHANGELOG.md
PrettyWood: Done in [ccc424d](https://github.com/PRQL/prql/pull/3916/commits/ccc424d2c6872cb5761d5ea44a4f3b515684c9aa)

prqlc/prql-compiler/src/sql/operators.rs
```diff
@@ -123,20 +123,24 @@ fn find_operator_impl(
...
 ) -> Option<(&pl::Func, Option<i32>, bool, Option<String>)> {
     let operator_name = operator_name.strip_prefix("std.").unwrap();
-
-    let operator_name = pl::Ident::from_name(operator_name);
+    let operator_ident = pl::Ident::from_path(
+        operator_name
+            .split('.')
+            .map(String::from)
+            .collect::<Vec<_>>(),
+    );
```
aljazerzen, line 131: `std.sql.prql` was not meant to have submodules for things other than dialect handling, but it is clearer this way.

aljazerzen (approved):
Looks good. It would be nice to have an `test-dbs` integration test, so make sure the results are what we expect them to be on all dbms.

PrettyWood:
>It would be nice to have an test-dbs integration test, so make sure the results are what we expect them to be on all dbms.

@aljazerzen Totally agree ! This is why I introduced `rstest` in the `str` (or `text` or `string`) submodule.
My plan is to use that (and maybe some macro on top of it) to test the outputs for all the supported dialects

aljazerzen:
`test-dbs` does more than that: it executes the queries on a sample chinook database.

This does not run by default, because it takes a while to compile duckdb, but you can run it with:
```
cargo test --package=prql-compiler --test=integration --features=test-dbs
```

PrettyWood:
@aljazerzen Ah ok right! I'll add some in another PR. Thanks

max-sixty:
> `test-dbs` does more than that: it executes the queries on a sample chinook database.

Yes the advantage of these is that it'll test a) on a real database, b) in CI (or locally using Docker) it'll test against every supported database. So things like quotes / `CONCAT` we can let the tests check.

(some details under "integration tests" here: https://prql-lang.org/book/project/contributing/development.html

PR #3923, opened 2023-12-11
fix: make `add` and `sub` less strict

### LINKED ISSUE #3127: standard functions sometimes can't be used after the `as` function

PrettyWood:
Following https://github.com/PRQL/prql/issues/3127#issuecomment-1849032396

closes #3127


max-sixty (approved):
Great, thanks!

PR #3924, opened 2023-12-11
fix(Lezer grammar): Comment between pipelines

vanillajonathan:
This PR allows comments and blank lines between pipelines, example:
```elm
from foo
# filter x == y
select bar
```
and
```elm
from foo

select bar
```


max-sixty (approved):
Super!

PR #3934, opened 2023-12-13
chore: improve integration tests

PrettyWood:
## Feedbacks on integration tests

When I first started to work on integration tests, I didn't know there was the great
`prqlc/prql-compiler/tests/integration/dbs/README.md` file that explains clearly how to run those tests.
I first had an issue by running
```shell
cargo nextest run --all-targets -F "test-dbs,test-dbs-external
```
and got bitten by the "it doesn't handle multiple processes"
Fortunately I saw the message
```
# test-dbs runs database setup when the connection is established,
# and because nextest runs test in separate processes, this happens on every test.
# To prevent mutliple setups running at once, we set max-threads to 1.
```
in `nextest.toml` and managed to run them properly

I also had issues with just starting the `mssql` container but this is due to my M1 and I found this issue that was already
linked in the docker-compose (https://github.com/microsoft/mssql-docker/issues/668#issuecomment-1436802153)
Thank you for that!

Once MSSQL was up and running, I thought everything would work. I was wrong 😬

The first time I run the test I got a timeout on the MSSQL server

```shell
test queries::results::aggregation has been running for over 60 seconds
.test queries::results::aggregation ... FAILED

failures:

---- queries::results::aggregation stdout ----
thread 'queries::results::aggregation' panicked at 'called `Result::unwrap()` on an `Err` value: Tls("connection closed via error")', prqlc/prql-compiler/tests/integration/dbs/protocol/mssql.rs:32:18
```

I also got once the error
```shell
thread 'queries::results::aggregation' panicked at 'Failed to insert row: no such table: albums.my

Caused by:
    Error code 1: SQL error or missing database
```

I don't remember if I stopped a test during execution or whatever but I had to dig a bit in the code
that it's just a side effect of the MySQL test that copies the csv files into `xyz.my.csv`.
And since they are git ignored, you don't see anything in a `git status`.
Anyway easy fix:

```shell
find . -type f -name '*.my.csv' -delete
```

### So how do we fix the MSSQL timeout error?

It appears there is an issue with the TLS handshake on M1 (see https://github.com/prisma/tiberius/issues/65)
I propose the switch to `rustls` since it doesn't require headers or runtime libraries.
(it conflicts with `async_native_tls` and is in the default features henece the need to deactive
those: see https://github.com/prisma/tiberius/issues/317)

And when I went with this approach the handshake was done! Great so it works finally!!!
OR NOT

```
error[E0277]: the trait bound `BigDecimal: tiberius::FromSql<'_>` is not satisfied
   --> prqlc/prql-compiler/tests/integration/dbs/protocol/mssql.rs:64:48
    |
64  | ...                   .get::<BigDecimal, usize>(i)
    |                                          ^^^^^ the trait `tiberius::FromSql<'_>` is not implemented for `BigDecimal`
    |
    = help: the following other types implement trait `tiberius::FromSql<'a>`:
              bool
              i16
              i32
              i64
              u8
              f32
              f64
              Uuid
            and 5 others
note: required by a bound in `tiberius::Row::get`
   --> /Users/ericjolibois/.cargo/registry/src/index.crates.io-6f17d22bba15001f/tiberius-0.12.2/src/row.rs:375:12
    |
373 |     pub fn get<'a, R, I>(&'a self, idx: I) -> Option<R>
    |            --- required by a bound in this associated function
374 |     where
375 |         R: FromSql<'a>,
    |            ^^^^^^^^^^^ required by this bound in `Row::get`

For more information about this error, try `rustc --explain E0277`.
error: could not compile `prql-compiler` (test "integration") due to previous error
```

Ahhhh right! By setting `default-features = false`, I forgot to add `tds73` which is also enabled by default
Annnnnnd TADA!

### Improvements

When you run a test and see
```console
test queries::compile::math ... ok
```
you don't know which dialects have been tested
And with a little `dbg!()` macro you see that `MSSQL` is not tested by default!

```console
[prqlc/prql-compiler/tests/integration/queries.rs:164] test_name = "math"
[prqlc/prql-compiler/tests/integration/queries.rs:164] dialect = DuckDb
[prqlc/prql-compiler/tests/integration/queries.rs:164] test_name = "math"
[prqlc/prql-compiler/tests/integration/queries.rs:164] dialect = Postgres
[prqlc/prql-compiler/tests/integration/queries.rs:164] test_name = "math"
[prqlc/prql-compiler/tests/integration/queries.rs:164] dialect = MySql
[prqlc/prql-compiler/tests/integration/queries.rs:164] test_name = "math"
[prqlc/prql-compiler/tests/integration/queries.rs:164] dialect = ClickHouse
[prqlc/prql-compiler/tests/integration/queries.rs:164] test_name = "math"
[prqlc/prql-compiler/tests/integration/queries.rs:164] dialect = GlareDb
```
So you NEED to add `# mssql:test` on top of the file to run it. That's weird compared to the others...
But then I looked at the code and saw that level is "UNSUPPORTED". So makes sense but again you need to know to know ;)

And the second point is just to improve/update a bit the documentation

-----------------------------------------------------------------------------------

So I would go with those changes in this PR if you're ok with it

- [x] make integration tests pass on M1 (for that I would like to know if it doesn't break anything for tests on Linux)
- [x] improve documentation

### DIFF
--- a/.config/nextest.toml
+++ b/.config/nextest.toml
@@ -10,8 +10,8 @@ slow-timeout = { period = "2s", terminate-after = 4 }
 
 [test-groups.test-dbs]
 # test-dbs runs database setup when the connection is established,
-# and becuase nextest runs test in separate processes, this happens on every test.
-# To prevent mutliple setups running at once, we set max-threads to 1.
+# and because nextest runs test in separate processes, this happens on every test.
+# To prevent multiple setups running at once, we set max-threads to 1.
 # Ideally, we could run tests in parallel and they would use a locking mechanism to see if
 # the database has already been setup. For now, we can use cargo test instead.
 max-threads = 1
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -23,7 +23,7 @@ version = "0.7.6"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "fcb51a0695d8f838b1ee009b3fbf66bda078cd64590202a864a8f3e8c4315c47"
 dependencies = [
- "getrandom 0.2.9",
+ "getrandom",
  "once_cell",
  "version_check",
 ]
@@ -36,7 +36,7 @@ checksum = "2c99f64d1e06488f620f932677e24bc6e2897582980441ae90a671415bd7ec2f"
 dependencies = [
  "cfg-if",
  "const-random",
- "getrandom 0.2.9",
+ "getrandom",
  "once_cell",
  "version_check",
 ]
@@ -313,18 +313,6 @@ dependencies = [
  "num",
  "regex",
  "regex-syntax",
-]
-
-[[package]]
-name = "async-native-tls"
-version = "0.4.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "d57d4cec3c647232e1094dc013546c0b33ce785d8aeb251e1f20dfaf8a9a13fe"
-dependencies = [
- "futures-util",
- "native-tls",
- "thiserror",
- "url",
 ]
 
 [[package]]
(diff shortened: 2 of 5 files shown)


PrettyWood:
@max-sixty Here you go 😄

max-sixty (approved):
Thanks a lot for the report, and for the the PR!

Issue #366, opened 2022-04-16
Date to string function


max-sixty:
In SQL, formatting dates is not easy or elegant; e.g. getting the year from a date, and differs between DBs:

- Postgres uses `TO_CHAR('2018-12-12', 'YYYY-MM')`
- MySQL & BigQuery uses `DATE_FORMAT('2018-12-12', '%y-%m')`

I quite like the python f-string approach of just `f"{date:%y-%m}"`

This is (maybe the first) issue where implementing for multiple DBs won't be trivial — I guess we are going to have to choose an approach and then write a mapping to the different date formats — more difficult than just switching out a function.

aljazerzen:
I think that this formatting approach is very convenient and is the way to go.

But it's true that is no small feat to implement for different dialects. Because even the format strings are different, we will have to parse them and convert to required dialect.

But a question opens: would we want `f"{date:%y-%m}"` or `f"{date:YYYY-MM}"`? Or something else?

max-sixty:
> But a question opens: would we want `f"{date:%y-%m}"` or `f"{date:YYYY-MM}"`? Or something else?

I'm not a great fan of the `%`, but it does seem to be more standard. Go has a [different approach](https://programming.guide/go/format-parse-string-time-date-example.html) that is very readable, but less writable.

That link also has the Java approach, which is not bad at all, but I don't know whether it has broad adoption.

snth:
I really like this f-string formatting suggestion and agree that it's the way to go. I've been using the f-strings quite a bit already and was thinking how great it would be if they could also do formatting of dates and numbers.

vanillajonathan:
I really like the f-string approach too.

However, maybe it should be easy format the date according to a standard instead of having to specify the format yourself. There are date standards such as  ISO 8601, RFC 2822 and RFC 3339.

There could be methods on a date object.
Example:
```rust
let string1 = date.to_rfc2822_string()
let string2 = date.to_rfc3339_string()

snth:
I'm familiar with ISO 8601 and e.g. Python has `.isoformat()` methods.

I'm not familiar with RFC 2822 and RFC 3339 so I think we would need better names for that. What do these formats look like?

vanillajonathan:
Yeah, sure the names could be very different, those were just some examples for how it could be done with methods, but it could also be by done by having a method take in a `DateFormat` enum:
```rust
let string1 = date.to_string(DateFormat.RFC2822)
[...]

PR #3942, opened 2023-12-15
feat: add date to string function

PrettyWood:
tackles #366 (not writing "closes" since I first want to validate the approach)

Hello everyone
I would like to first validate the approach taken here for the "date to string" function
If this is validated I will continue working on it

- [x] start with a basic date to string function that handles multiple dialects
- [ ] add more options for "date format"
- [ ] support this format in f-strings
- [ ] add integration tests
- [ ] add documentation
- [ ] add changelog


PrettyWood:
@max-sixty @eitsupi @aljazerzen The idea would be to use in PRQL the chrono datetime format (see https://docs.rs/chrono/latest/chrono/format/strftime/index.html)
And then we just map the inner types of chrono to the expected format for a given dialect
(See https://github.com/PRQL/prql/blob/6afdf10e0c10bde47c15bd9e8ffabecd7d35fae2/prqlc/prql-compiler/src/sql/dialect.rs#L234-L257)
WDYT?

prqlc/prql-compiler/src/sql/gen_expr.rs
```diff
@@ -104,6 +104,32 @@ pub(super) fn translate_expr(expr: Expr, ctx: &mut Context) -> Result<ExprOrSour
                 }
                 "std.concat" => return Ok(process_concat(&expr, ctx)?.into()),
                 "std.array_in" => return Ok(process_array_in(args, ctx)?.into()),
+                "std.date.to_string" => {
+                    if let [date_format_exp @ Expr {
+                        kind: ExprKind::Literal(Literal::String(date_format)),
+                        ..
+                    }, col_expr] = &args[..]
```
aljazerzen, line 111: This unpacking is a crucial detail. It assumes that the format is a string literal.

It might have been defined in a separate variable, but it will not work if you do:
```
d | date.to_string ('%Y' + '%M')
```
(I'm using `+` as string concatenation here).

I think this restriction is ok for now (since we do have the same restriction on `from_text` and a few others), but I'd be nice if we'd have a plan for the future:

If this is some expression that we cannot statically evaluate in the compiler, we need to parse it in the database. With SQL. On different databases. This will be hard to do.

Probably the best approach to do this would be to write this parsing and mapping in PRQL.

prqlc/prql-compiler/src/sql/dialect.rs
```diff
@@ -322,6 +561,27 @@ mod tests {
...
+    fn test_translate_date(
+        #[case] prql_date_format: &str,
+        #[case] dialect: impl DialectHandler + 'static,
+        #[case] postgres_date_format: &str,
+    ) {
+        assert_eq!(
+            dialect.translate_prql_format(prql_date_format),
+            postgres_date_format.to_string()
+        );
+    }
```
aljazerzen: I'd prefer a much tests as possible in the integration test instead of unit test.

aljazerzen:
Interesting. The language-design part of this PR is "how do we express date formats in PRQL" and your answer is [strftime](https://docs.rs/chrono/latest/chrono/format/strftime/index.html).

I like it.

prqlc/prql-compiler/src/sql/gen_expr.rs
max-sixty: Yes good point.

Though how would we do this in PRQL directly?

@aljazerzen do you think it's feasible to have the function take the "materialized" value of an argument while being in rust? So `d | date.to_string (f"%{foo}-%M")`, where `foo="%Y"` would work?

prqlc/prql-compiler/src/sql/dialect.rs
max-sixty: (I agree on having integration tests, though I also find that initially having them as very small unit tests makes them easier to iterate on, though possibly that's a difference in programming styles — I'm more of the "write something wrong and change it until it works" approach!)

max-sixty (commented):
This would be a great feature to have.

Nice use of `chrono` for the front end of this.

It's surprising how different the dialects are in their interfaces here! That means it's more work now, but also very helpful. (I find LLMs useful for this sort of survey FWIW, here's [GPT-4's survey](https://chat.openai.com/share/d63a0874-e703-466b-a378-ecb82f256b15))

If we agree on the approach, I would be quite happy to merge something for a single database, to break the work up more. (Even if we didn't want it partially working, we could disable it in PRQL's interface, while still having the code in `main`, until we have more databases working).

prqlc/prql-compiler/src/sql/dialect.rs
PrettyWood: Ofc @aljazerzen it's the plan (see the description)
I already made an integration test but just went with some unit tests as well to be faster and make sure the approach would be validated

prqlc/prql-compiler/src/sql/gen_expr.rs
PrettyWood: I didn't understand how you plan to do that sorry

PrettyWood:
Great that you agree on the Chrono approach. I can add full support for duckdb to be able to test in the playground. And we can complete in future PRs for the other dialects + add integration tests, doc and f-string syntax

prqlc/prql-compiler/src/sql/gen_expr.rs
aljazerzen: I also don't understand what you mean @max-sixty 

Let me try to explain the problem on another example. Say that we have this query:
```
from dates_to_display
select {my_date, my_format}
select {std.date.to_string my_date my_format}
```
The problem is now `my_format`, which is a string column in the database that we have no insight into at compile time.

If we want to allow usages like this, we must implement the conversion from `Year: %Y` to `TO_CHAR('Year: YYYY', ...)` in SQL. So we must parse the fstrtime format and produce the database's format string using what's available in the target database's dialect.

Because this 'parsing and producing' code is complex, we don't want to write or maintain it for each database separately. Instead, we could implement it using PRQL. The problem here is that PRQL is not powerful enough to express such algorithms with ease.

---

In any case, looking at strftime format, it seems like it is the most widespread time-formatting language, so I don't object adding it into PRQL's std module, regardless of the cost of the full implementation.

prqlc/prql-compiler/src/sql/gen_expr.rs
```diff
@@ -104,6 +104,32 @@ pub(super) fn translate_expr(expr: Expr, ctx: &mut Context) -> Result<ExprOrSour
...
+                                        )),
+                                        span: date_format_exp.span,
+                                    },
+                                    col_expr.clone(),
+                                ],
+                            },
+                            ..expr
+                        };
+                        return super::operators::translate_operator_expr(expr, ctx);
+                    }
```
aljazerzen, line 131: Add an else branch that produces an error saying "function std.date.to_string, param format expected a string literal".

Also add the following test:
```
from dates_to_display
select {my_date, my_format}
select {std.date.to_string my_date my_format}
```

aljazerzen:
@PrettyWood consider this approach validated, it will be a great addition to the language.

PrettyWood:
@max-sixty @aljazerzen I'll keep this PR as a reference for all the dialects.
I managed to have integration tests working with every dialects except sqlite and glaredb.
SQLite doesn't support day or month names so we'll need to do some magic with something like
```
case cast (strftime('%w', '2022-01-01') as integer)
  when 0 then 'Sunday'
  when 1 then 'Monday'
  when 2 then 'Tuesday'
  when 3 then 'Wednesday'
  when 4 then 'Thursday'
  when 5 then 'Friday'
  else 'Saturday' end as weekday
  ```

Anyway I'll open another PR with just DuckDB for now to have a least the foundations

File: prqlc/prql-compiler/src/sql/gen_expr.rs (hunk 1 above)
Comment (max-sixty) [reply prqlc/prql-compiler/src/sql/gen_expr.rs:111]: Sorry if I'm being confusing.

I'm also a bit confused!

> The problem is now `my_format`, which is a string column in the database that we have no insight into at compile time.

The date format string needs to be in the query itself, in the SQL dialects I'm familiar with. We can't do:

```sql
SELECT 
  dt, 
  TO_CHAR(dt, fmt_column) as dt_str   # fmt_column is a column
FROM artists
```

---

The way that we _can_ get a dynamic date format string is by forming it in PRQL:

```elm
let dt_string = "%Y-%m" 

from artists
select {dt, dt_str = (dt | date.to_string dt_string)}
```

I (I think mistakenly) thought you were claiming that wouldn't work, and I was responding by asking if it were possible. After re-reading the messages, I think we both believe this is possible.

---

Maybe if there's still confusion, would you mind clarifying the example that wouldn't work?

max-sixty:
> I managed to have integration tests working with every dialects except sqlite and glaredb.

Amazing!

> SQLite doesn't support day or month names so we'll need to do some magic with something like

Yeah. If you're actively using SQLite and would like this, it would be nice to have. But if a dialect is so far behind such that we need to write this ourselves, my sense is that it's not worth slowing down other dialects to bring everything along. (But not confident and ofc open to thoughts, there is some advantage to consistency)

max-sixty (commented):
One brief comment — there do seem like there are lots of lines that are missed by tests. While we don't have some rule of 100% coverage, if we can spray some snapshot tests with a large number of date format strings, that's probably worth doing.

An LLM might be able to do that for you, lmk if you want a hand.

PrettyWood:
@max-sixty Target a high coverage and add tests for every new fix or feature is a nice hygiene that we already have in my company or other OSS repos where I contribute a lot. So don't worry I don't plan to open a PR with 40% of coverage ;)
This PR was just a way to have your feedback and see if we could have something working for most dialects. I'll now open smaller PRs with a clear scope and plan to have (almost) full coverage for sure on those ones :)

As of SQLite I agree I guess we can forget about it for now and already bring a lot of value to most prod DBs and DWs

PR #3951, opened 2023-12-16
feat: add basic date to string for duckdb

PrettyWood:
Now that #3942 is validated, let's start to split the work in different PRs.
This PR adds basic support for date to string for DuckDB

Do you think we should handle more specifiers? (see https://docs.rs/chrono/latest/chrono/format/strftime/index.html)
I reckon the most common ones are in this PR. We can always add some afterwards.


prqlc/prql-compiler/tests/integration/error_messages.rs
```diff
@@ -267,3 +266,56 @@ fn test_ambiguous_inference() {
...
+  "#).unwrap_err(), @r###"
+    Error:
+       ╭─[:4:11]
+       │
+     4 │   select {std.date.to_string my_date my_format}
+       │           ──────────────────┬─────────────────
+       │                             ╰─────────────────── `std.date.to_string` only supports a string literal as format
+    ───╯
+    "###);
+}
```
max-sixty, line 321: Great errors!

prqlc/prql-compiler/src/sql/dialect.rs
```diff
@@ -187,7 +202,11 @@ impl dyn DialectHandler {
     }
 }
 
-impl DialectHandler for GenericDialect {}
+impl DialectHandler for GenericDialect {
+    fn translate_chrono_item(&self, _item: Item) -> Result<String> {
+        bail!("Date formatting requires a dialect")
```
max-sixty, line 207: I would guess that the `%Y`-like strings are the most common. I would probably vote to default to that for the generic.

Though this is one of the first times where the ecosystem is fairly evenly split, so no strong view

(And doesn't need to be in this PR)

max-sixty (approved):
Looks great!

I would think we don't need to start with that many date specifiers — that the vast majority come from just a few...

PrettyWood:
@max-sixty Do you know why there is an error on x86_64? I can take a VM to try but if by any chance you work on Linux and/or have a clue it would be super nice 🙏

prqlc/prql-compiler/src/sql/dialect.rs
PrettyWood: I don't know to be honest
- BigQuery has `%M` for the minutes when MySQL has `%i`
- Clickhouse supports both percentage (MySQL-like) and Joda syntax (MSSQL-like)
- DuckDB is very close to `chrono` syntax so why not but some very common specifiers like the one for the full month name is `%B` when it's `%W` in MySQL...
- GlareDB doesn't support it yet I guess. I digged into datafusion doc and asked on discord without any response
- MySQL yes it works
- MSSQL is `yyyy`
- Postgres (and hence Snowflake) is `YYYY`

I think it's dangerous to have defaults that are for only 1/3 of the supported DBs (and even for them we'll have issues for others)
This is why I chose this approach

max-sixty: Yeah, that is even more balkanized than I thought before...

max-sixty:
> @max-sixty Do you know why there is an error on x86_64? I can take a VM to try but if by any chance you work on Linux and/or have a clue it would be super nice 🙏

This is because there's a snapshot which isn't used. You can repro locally with https://github.com/PRQL/prql/blob/4a3e43f67fb71a9682c929f585d8efa0e1582723/Taskfile.yml#L298.

It's only on the linux one because some tests are excluded on windows; though we could add it to the Mac tests too...

PrettyWood:
I have
```
info: no unreferenced snapshots found
```
And when I saw the error on the CI the test fails and so this message appears
But the real error is 
```
 ---- queries::results::date_to_string stdout ----
thread 'queries::results::date_to_string' panicked at 'swap_remove index (is 0) should be < len (is 0)', library/alloc/src/vec/mod.rs:1399:13
```

max-sixty:
> But the real error is

Ah sorry! Yes, I'm looking

max-sixty:
It seems like there are no results for a query. I can't repro locally (though am on Mac).

We could try skipping if there are no results. Though I worry that something else is causing this.

```diff
diff --git a/prqlc/prql-compiler/tests/integration/queries.rs b/prqlc/prql-compiler/tests/integration/queries.rs
index d4b4111b..87459f1b 100644
--- a/prqlc/prql-compiler/tests/integration/queries.rs
+++ b/prqlc/prql-compiler/tests/integration/queries.rs
@@ -181,21 +181,22 @@ fn run(prql_path: &Path) {
             results.push((dialect, result));
         }
 
-        // insta::allow_duplicates!, but with reporting of which two cases are not matching.
-        let (left_dialect, left_text) = results.swap_remove(0);
-        for (right_dialect, right_text) in results {
-            similar_asserts::assert_eq!(
-                left_text,
-                right_text,
-                "{} {} {}",
-                test_name,
-                left_dialect,
-                right_dialect
-            );
+        // insta::allow_duplicates!, but with reporting of which two cases are
+        // not matching.
+        if let Some((_, left_text)) = results.first() {
+            for (right_dialect, right_text) in results.iter().skip(1) {
+                similar_asserts::assert_eq!(
+                    left_text,
+                    right_text,
+                    "{} {} {}",
+                    test_name,
+                    right_dialect,
+                    right_dialect
+                );
+            }
+            with_settings!({ input_file => prql_path }, {
+                assert_snapshot!(test_name, left_text, &prql)
+            })
         }
-
-        with_settings!({ input_file => prql_path }, {
-            assert_snapshot!(test_name, left_text, &prql)
-        })
     }
 }
```

~I'll turn on the nightly tests, which include other platforms, to see whether it's a CI vs a Linux thing...~ edit: actually only the linux one runs the external tests — so this isn't helpful. But it is probably from the external tests...

PrettyWood:
Thank you very much for your help. The external tests work on my mac so it's hard to debug.
I just tried something random because I see nothing obvious at first sight.
I'll take a VM tomorrow if needed and check if I can reproduce

max-sixty:
Great, that worked! As ever, feel free to merge when you're ready

prqlc/prql-compiler/src/sql/dialect.rs
aljazerzen: > balkanized

Haha, first time I seen this term, but it describes the state here nicely :D

----

I'd say that for the generic dialect, we avoid confusing introducing a combination of the formats of other databases and just use strftime format (i.e. don't translate), since generic dialect is not meant to work at all, but it should be something that is displayed in the docs.

PR #3952, opened 2023-12-17
test: Run in-process DB tests with `test-dbs-external`

max-sixty:
I think this might be causing the issues at https://github.com/PRQL/prql/pull/3951 — we're not running the in-process tests at the moment, because we're only enabling the `test-dbs-external` feature.

This PR allows runs all db tests with the `test-dbs-external` feature.

### DIFF
--- a/prqlc/prql-compiler/tests/integration/dbs/mod.rs
+++ b/prqlc/prql-compiler/tests/integration/dbs/mod.rs
@@ -41,10 +41,8 @@ pub struct ConnectionCfg {
 impl DbConnection {
     pub fn new(cfg: ConnectionCfg) -> Option<DbConnection> {
         let protocol = match &cfg.protocol {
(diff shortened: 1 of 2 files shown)


PrettyWood:
Ah riiiight I was running `cargo t -F "test-dbs,test-dbs-external"` on my machine
Just saw that on the CI it's only `cargo t -F "test-dbs-external"` and indeed I have the error
```
thread 'queries::results::date_to_string' panicked at 'swap_remove index (is 0) should be < len (is 0)', library/alloc/src/vec/mod.rs:1399:13
```
Nice catch

max-sixty:
@aljazerzen this needs your review it seems (ref our discussion on Discord...)

max-sixty:
> @aljazerzen this needs your review it seems (ref our discussion on Discord...)

I added myself as an override, so we could get this merged FYI

PR #3954, opened 2023-12-17
fix: escape quotes and percentage in date to string for duckDB

PrettyWood:
Also
- fix the tests (the output was in the snapshot so the expected was useless and it was more confusing than anything)
- remove `%-M` which is not for all dialects and not super common
- force UTC as timezone for the date to string repr


prqlc/prql-compiler/src/sql/dialect.rs
```diff
@@ -322,31 +322,37 @@ impl DialectHandler for DuckDbDialect {
...
+                (Numeric::Month, Pad::Zero) => "%m",
+                (Numeric::Day, Pad::None) => "%-d",
+                (Numeric::Day, Pad::Zero) => "%d",
+                (Numeric::Hour, Pad::None) => "%-H",
+                (Numeric::Hour, Pad::Zero) => "%H",
+                (Numeric::Hour12, Pad::Zero) => "%I",
+                (Numeric::Minute, Pad::Zero) => "%M",
+                (Numeric::Second, Pad::Zero) => "%S",
+                (Numeric::Nanosecond, Pad::Zero) => "%f", // Microseconds
+                _ => unimplemented!("Unsupported chrono numeric item: {:?}", item),
```
max-sixty: I think this will panic — is that right? We should instead return an error so it's returned to the user. 

Every panic is a bug...

Does that make sense? Happy to provide more guidance!

max-sixty: (I realize that we previously had `unimplemented!`, I should have spotted that...)

max-sixty (approved):
Looks good, thanks, fine to merge. Would be great to fix the `unimplemented!` either here or in another PR.

prqlc/prql-compiler/src/sql/dialect.rs
PrettyWood: you're right! [7e00918](https://github.com/PRQL/prql/pull/3954/commits/7e009180030ac5d05781f2f843f1cb45eb7cc349)

PrettyWood: Thanks you're right! [cad094c](https://github.com/PRQL/prql/pull/3954/commits/cad094c9030d6e422f27cbb2b7c44a4deffec444)

PrettyWood:
While writing the doc I saw that I forgot to support `%%`. Now it's done :)

prqlc/prql-compiler/src/sql/dialect.rs
```diff
@@ -185,12 +212,17 @@ pub(super) trait DialectHandler: Any + Debug {
     /// (see https://docs.rs/chrono/latest/chrono/format/strftime/index.html)
     fn translate_prql_date_format(&self, prql_date_format: &str) -> Result<String> {
         Ok(StrftimeItems::new(prql_date_format)
-            .map(|item| self.translate_chrono_item(item))
+            .map(|item| {
+                if !is_supported_chrono_item(&item) {
+                    bail!("PRQL doesn't support this date format item")
+                }
+                self.translate_chrono_item(&item)
+            })
```
max-sixty: I think the more rust-y thing to do here is to change the `translate_chrono_item` `_` branch to a `bail!`, remove the `is_supported_chrono_item` check, and then it should all flow through.

Then we also don't need to have two similar lists. Does that make sense?

PrettyWood: For me it's not a question of rusty or not rusty but more of "where is it decided that it's supported"
And in my mind it was in the `is_supported_chrono_item` and we could write the documentation based on those specifiers.
Even if a dialect can do more, we don't support it natively in PRQL (people can always do a `s-string`)
This was a way to have exactly the same body for all the `translate_chrono_item` methods for all the dialects

I can change it if you want but this is why I went with this approach

PrettyWood: If we don't do that, we need to `bail!` with the same error message for all the dialects. And make sure they implement exactly the same specifiers.
But maybe I'm taking the wrong approach or there is a nicer way.
Maybe we just accept that the dialects won't have the same support and some will have almost everything available and others very few 🤷 In which case the `bail!` can be indeed at dialect level
Let me know what you think

max-sixty: The problem with the currently proposed implementation is that if we make a small mistake — i.e. the code in `is_supported_chrono_item` covers a slightly different set of cases than a dialect's `translate_chrono_item` — then we get a panic (and we should never panic, that's always a bug). So we lose exhaustiveness — a big advantage of pattern matching.

> If we don't do that, we need to `bail!` with the same error message for all the dialects.

The approach I'm thinking is that we can write any message we want in each dialect (e.g. at https://github.com/PRQL/prql/pull/3954/files#diff-65a3c21d281e36b6fff6cc3e5ffdb821f86f3a54388bf07f67fdde4c17f6e90dR378). And we don't have any `is_supported_chrono_item` function.

> And make sure they implement exactly the same specifiers.

I agree, this requires some coordination. But IIUC — we have to do that now; it's just that if we make a mistake, we get a panic. (Very possible I'm misunderstanding, let me know)

> And in my mind it was in the `is_supported_chrono_item` and we could write the documentation based on those specifiers.

I agree we lose the canonical set of identifies — that's a disadvantage (although it wasn't enforced previously, it was just somewhere to write them). I think probably the canonical source can be a query which is then tested across every dialect + serves as documentation (we don't currently run our docs through every dialect, but we could quite easily add a [tag](https://github.com/PRQL/prql/blob/b1d54576eca2ed2f90eb8d77fe31c20214215127/web/book/src/lib.rs#L43) which would then run the query over every dialect, or just copy & paste with our integration tests for the moment...)

WDYT?

PrettyWood: Thank you very much for your feedback. All the specifiers that I added in the doc are tested in the integration tests so ok let's do that. I'll go back to the approach with one bail per dialect :+1:

prqlc/prql-compiler/src/sql/dialect.rs
```diff
@@ -323,30 +323,28 @@ impl DialectHandler for DuckDbDialect {
...
+            Item::Numeric(Numeric::Nanosecond, Pad::Zero) => "%f".to_string(), // Microseconds
+            Item::Fixed(Fixed::ShortMonthName) => "%b".to_string(),
+            Item::Fixed(Fixed::LongMonthName) => "%B".to_string(),
+            Item::Fixed(Fixed::ShortWeekdayName) => "%a".to_string(),
+            Item::Fixed(Fixed::LongWeekdayName) => "%A".to_string(),
+            Item::Fixed(Fixed::UpperAmPm) => "%p".to_string(),
+            Item::Fixed(Fixed::RFC3339) => "%Y-%m-%dT%H:%M:%S.%fZ".to_string(),
+            Item::Literal(literal) => literal.replace('\'', "''").replace('%', "%%"),
+            Item::Space(spaces) => spaces.to_string(),
+            _ => bail!("PRQL doesn't support this format specifier"),
```
max-sixty, line 346: Super!

max-sixty (approved):

PR #3955, opened 2023-12-17
feat: add `date.to_text` for MySQL, Postgres, MSSQL and Clickhouse + doc

PrettyWood:
tackles most of https://github.com/PRQL/prql/issues/366


web/book/src/reference/syntax/date-time/format-specifiers.md
```diff
@@ -0,0 +1,45 @@
+# Date & time format specifiers
+
+PRQL specifiers for date and time formatting is a subset of specifiers used by
+[`chrono`](https://docs.rs/chrono/latest/chrono/format/strftime/index.html).
```
max-sixty, line 4: Nice

web/book/src/reference/syntax/date-time/index.md
```diff
@@ -0,0 +1,46 @@
+# Date and time
+
+You already know that PRQL uses the `@2020-01-01` syntax to declare dates. To
```
max-sixty:
```suggestion
We already know that PRQL uses the `@2020-01-01` syntax to declare dates. To
```

Tiny point — I've generally tried to avoid the second person, so we can try and get a consistent voice through the docs. (Though have not always been successful!)

web/book/src/reference/syntax/date-time/index.md
```diff
@@ -0,0 +1,46 @@
...
+This function allows to convert a date into a string. Since there are many
+possible date representations, `to_string` takes a `format` parameter that
+describes thanks to [specifiers](./format-specifiers.md) how the date or
+timestamp should be structured.
+
+```admonish info
+Since all RDBMS have different ways to format dates and times, PRQL **requires an explicit dialect** to be specified
+```
+
+```admonish warning
```
max-sixty:
```suggestion
```admonish info
```

max-sixty (approved):
Awesome, great docs!

snth:
On the language design aspect (should this comment go in #366 or here since that issue is closed?), when we created `from_text` in #286, we decided at the time that `string` was too computer sciency and we went with `text` instead, i.e. `from_text` rather than `from_string` (see https://github.com/PRQL/prql/issues/286#issuecomment-1332656023).

Therefore should this function be called `date.to_text` rather than `date.to_string`?

PrettyWood:
@snth We discussed this matter recently https://github.com/PRQL/prql/pull/3913#issuecomment-1852364749 and `string` has been chosen as the go to. It can always be reverted and I agree we should be consistent (either `string` everywhere or `text` everywhere)

max-sixty:
> Therefore should this function be called `date.to_text` rather than `date.to_string`?

I was the primary advocate for the `from_text`...

On reflection, I think that was:
- Marginal / a bit ambitious
- Much better suited to `from_X` than here, because `from_X` involved a big block of text creating a relation, whereas this is creating a single value of type `str`

So I wouldn't really advocate for it here, and happy to align on `string` if that's the consensus

---

(OK, @PrettyWood just commented, I'll still post this but echoing his sentiment)

CHANGELOG.md
```diff
@@ -21,6 +21,9 @@
   `length`, `substring`, `replace`. Utility functions `starts_with`, `contains`
   and `ends_with` are also available.\
   Those functions are in the `string` module (@PrettyWood, #3913)
+- Formatting a date to a string is now available for Clickhouse, DuckDB, MySQL,
+  MSSQL and Postgres. A new `date` module has been added with the `to_string`
+  function (@PrettyWood, #3951, #3954 & 3955)
```
```suggestion
  function (@PrettyWood, #3951, #3954 & #3955)
```

web/book/src/reference/syntax/date-time/format-specifiers.md
```diff
@@ -0,0 +1,45 @@
...
+| `%F`  | `2001-07-08`                  | Year-month-day format (ISO 8601). Same as `%Y-%m-%d`             |
+|       |                               |                                                                  |
+|       |                               | **TIME SPECIFIERS:**                                             |
+| `%H`  | `00`                          | Hour number (00-23)                                              |
+| `%k`  | ` 0`                          | Same as `%H` but space-padded. Same as `%_H`.                    |
+| `%I`  | `12`                          | Hour number in 12-hour clocks (01--12), zero-padded to 2 digits. |
+| `%p`  | `AM`                          | `AM` or `PM` in 12-hour clocks.                                  |
+| `%M`  | `34`                          | Minute number (00-59), zero-padded to 2 digits.                  |
+| `%S`  | `60`                          | Second number (00-59), zero-padded to 2 digits.                  |
+| `%f`  | `264900`                      | Number of microseconds since last whole second                   |
```
max-sixty:
```suggestion
| `%f`  | `264900`                      | Number of microseconds since last whole second{footnote: this is different from chrono, for which `%f` represents nanoseconds}                   |
```

snth:
I think our earlier discussion on `from_text` was sensible and we shouldn't be that quick to dismiss it.

The list in https://github.com/PRQL/prql/pull/3913#discussion_r1424049816 is impressive but are those languages aiming at the same kind of users as us? I would argue they are more aimed at developers while we are targeting Data Analysts who often come from a less computer science heavy background.

I keep coming back to Power Query M Language because I think it's a great example of a language that's aimed at the same kind of data munging that we target. Power Query M Language has a [**Text** module](https://learn.microsoft.com/en-us/powerquery-m/text-functions) and in the Date module there are [Date.FromText](https://learn.microsoft.com/en-us/powerquery-m/date-fromtext) and [Date.ToText](https://learn.microsoft.com/en-us/powerquery-m/date-totext) functions.
 
[Pandas](https://pandas.pydata.org/docs/user_guide/text.html#string-methods) and [Polars](https://pola-rs.github.io/polars/user-guide/expressions/strings/) both use a `.str` namespace for their text manipulation functions (see https://pandas.pydata.org/docs/user_guide/text.html#string-methods and https://pola-rs.github.io/polars/user-guide/expressions/strings/).

I'm -1 on `string`.

max-sixty (approved):
Looking really good!

PrettyWood:
@max-sixty I let you decide if you want to merge for the release or not

max-sixty:
> @max-sixty I let you decide if you want to merge for the release or not

For sure! I was waiting for you to hit the button but happy to ofc

PR #3960, opened 2023-12-18
fix: Micro / Nanoseconds

max-sixty:
Should this be `%n`? From the docs: `%n 	Nanosecond as a decimal number, zero-padded on the left.`

(Very possible I'm missing something)


PrettyWood:
Yeah this was my [original implementation](https://github.com/PRQL/prql/pull/3942/files#diff-65a3c21d281e36b6fff6cc3e5ffdb821f86f3a54388bf07f67fdde4c17f6e90dR792) but to be consistent with all other DBMS, I chose to display the microseconds (only DuckDB supports nanoseconds)

If not what we can do is choose the `%.6f` in the ISO format to enforce a consistent behaviour and not implement "nanoseconds". But then in the documentation it means we support `%.6f` but not `%f` which is the same for every DBMS except DuckDB 😆

max-sixty:
Ah yes, I see here: https://github.com/PRQL/prql/pull/3955/files#diff-65a3c21d281e36b6fff6cc3e5ffdb821f86f3a54388bf07f67fdde4c17f6e90dR234

And I see it's in the docs! I added a footnote [here](https://github.com/PRQL/prql/pull/3955/files#r1430626593), so closing!

PR #3961, opened 2023-12-19
devops: Update ics file for new dev call time

max-sixty:


max-sixty:
FYI for it to update for me, I had to delete the old one before opening this file

Issue #3963, opened 2023-12-19
Parentheses omitted from sql output in non-commutative operations


syko:
### What happened?

Attempting to do math fails at BEMDAS in even some pretty basic cases. The compiler seems to omit some parentheses from the output in scenarioes where they are needed.

This seems to be somewhat dialect-dependent. `sql.generic` seems to be perform the worst.

### PRQL input

```elm
from foo
derive {
  a = 10 - (10 - 2),
  b = 10 - (10 + 2),
  c = 10 / (10 / 2),
  d = 10 / (10 * 2),
}
```

### SQL output

```SQL
-- sql.generic
SELECT
  *,
  10 - 10 - 2 AS a,
  10 - 10 + 2 AS b,
  10 / 10 / 2 AS c,
  10 / 10 * 2 AS d
FROM
  foo


-- sql.postgres
SELECT
  *,
  10 - 10 - 2 AS a,
  10 - 10 + 2 AS b,
  (10 * 1.0 / (10 * 1.0 / 2)) AS c,
  (10 * 1.0 / 10 * 2) AS d
FROM
  foo
```

### Expected SQL output

```SQL
SELECT
  *,
  10 - (10 - 2) AS a,
  10 - (10 + 2) AS b,
  10 / (10 / 2) AS c,
  10 / (10 * 2) AS d
FROM
  foo
```

### MVCE confirmation

- [X] Minimal example
- [X] New issue

richb-hanover:
Good bug report. Thanks.

This is a bug in the deployed Playground, it's [fixed in main](https://github.com/PRQL/prql/pull/3846) and [could be rolled into a new release](https://github.com/PRQL/prql/issues/3929) fairly soon

NB I use the [Dev Container](https://prql-lang.org/book/project/contributing/development.html#option-3-use-a-dev-container) to stay up to date with the latest, and to test this...

PrettyWood:
@richb-hanover The bug is still here on `main` (see #3968)
(`a` and `b` were indeed good thanks to #3846 but not `c` and `d`)

richb-hanover:
Yikes! You're right. The addition/subtraction paren's are correct, multiply/divide are not... (See generated SQL below)

@syko Sorry for the misinformation...

```
SELECT
  *,
  10 - (10 - 2) AS a,
  10 - (10 + 2) AS b,
  10 / 10 / 2 AS c,
  10 / 10 * 2 AS d
FROM
  foo

[...]

PR #3968, opened 2023-12-19
fix: ensure parentheses are set for sub expr

PrettyWood:
Proposal for #3963
I don't know if this is the right fix, there is maybe a better way.
I understand that it may add some unneeded parentheses but honestly I feel like it's way more important to have the right result than the cleanest query possible.
+ sometimes some extra parentheses make stuff more explicit (for example when you choose a column that is already an operation that you divide)


max-sixty:
> I understand that it may add some unneeded parentheses but honestly I feel like it's way more important to have the right result than the cleanest query possible.

Definitely agree; though let me see if I can fix the underlying issue

PR #3970, opened 2023-12-19
fix: Correct division precedence without paretheses everywhere

max-sixty:
Follow up to #3969.

I think this is a pretty reasonable fix. There are probably some instances where it has too many parentheses, but very few. And it uses our existing approach, doesn't require any big changes.


max-sixty:
I'll merge but extremely open to feedback / changes. 

If anyone can find cases where we use too few parentheses, that's an 🏆  bug report.

Cases where we use too many parentheses also welcome!

PR #3972, opened 2023-12-19
docs: Fix confusion: "celsius_to_fahrenhent" is more clear ...

richb-hanover:
I was massively confused because the documentation has two similarly-named functions that do the opposite things. Their names were:

* `fahrenheit_to_celsius` and
* `fahrenheit_from_celsius`

My poor brain was struggling to understand why the formulas were different. Renaming the latter (using F -> C) makes more sense.


max-sixty (approved):
Thanks!

Issue #2879, opened 2023-06-20
Improving code coverage


max-sixty:
### What's up?

One Good First Issue for folks who want to start writing rust — now we have [Code Coverage](https://app.codecov.io/github/prql/prql/tree/main?displayType=list) — is to find some code that isn't tested, and write a small test for it.

Our tests are really easy to write, and by mapping a test to the code that gets run, it'll give a good intro to the compiler.

max-sixty:
Edit: this specific example is done! Thanks to @nkicg6 in #3985.

I'm leaving the broader issue open since there's lots more out there...

---

As an example of the issue, replacing `Span`'s `Serialize` & `Deserialize` with standard implementations (diff below) doesn't break tests, so we don't know if this code is doing what's intended. 

I found that by browsing the `crates` path in codecov.io, and finding a file with very low coverage: https://app.codecov.io/github/prql/prql/commit/881e03fcc45bd7bd1859e864c55358bf68c652e0/blob/crates/prql-ast/src/span.rs.

<details>

```diff
diff --git a/crates/prql-ast/src/span.rs b/crates/prql-ast/src/span.rs
index 97c8319c..cd8b215e 100644
--- a/crates/prql-ast/src/span.rs
+++ b/crates/prql-ast/src/span.rs
@@ -3,7 +3,7 @@
 use std::fmt::{self, Debug, Formatter};
 use std::ops::Range;
 
-#[derive(Clone, PartialEq, Eq, Copy)]
+#[derive(Clone, PartialEq, Eq, Copy, Serialize, Deserialize)]
 pub struct Span {
     pub start: usize,
     pub end: usize,
@@ -22,69 +22,3 @@ fn fmt(&self, f: &mut Formatter<'_>) -> fmt::Result {
         write!(f, "{}:{}-{}", self.source_id, self.start, self.end)
     }
 }
-
-impl Serialize for Span {
-    fn serialize<S>(&self, serializer: S) -> std::result::Result<S::Ok, S::Error>
-    where
-        S: serde::Serializer,
-    {
-        let str = format!("{self:?}");
-        serializer.serialize_str(&str)
-    }
-}
-
-impl<'de> Deserialize<'de> for Span {
-    fn deserialize<D>(deserializer: D) -> std::result::Result<Self, D::Error>
-    where
-        D: serde::Deserializer<'de>,
-    {
-        struct SpanVisitor {}
-
-        impl<'de> Visitor<'de> for SpanVisitor {
[...]

PR #3985, opened 2023-12-20
test: add serde test for prqlc-ast span

nkicg6:
related to https://github.com/PRQL/prql/issues/2879#issuecomment-1712447738 

@max-sixty this addresses `Span`'s `Serialize` and `Deserialize`. 
Not sure if it was necessary to use the snapshot test for this unit test but it looked like that was what you were using throughout the rest of the repo so I tried to make it fit.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2984,8 +2984,10 @@ name = "prqlc-ast"
 version = "0.11.1"
 dependencies = [
  "enum-as-inner",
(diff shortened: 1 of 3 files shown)


max-sixty:
Super, thank you!

(For anyone else reviewing — I thought for a second we could use `assert_json_snapshot`, but actually we test it deserializes too, which is good)

PR #3988, opened 2023-12-21
fix: support group with take more than 1

### LINKED ISSUE #3986: Top with group on Postgres
### What happened?

We have a query builder in my company https://github.com/ToucanToco/weaverbird/, which is a click button interface that generates some config in our DSL that is then translated into Mongo, Pandas, many SQL dialects
We are rewriting our whole data pipeline in Rust and I wanted to use PRQL instead of generating many CTEs

I generated the code for the `top` step. The generated PRQL query works fine in the playground (without a dialect!) but our integration tests fail. And I force the dialect to Postgres in the playground, the query is indeed totally different

(In the MRE I want to keep the top 2 beers of each kind)

### PRQL input

```elm
prql target:sql.postgres

from_text """
price_per_l,alcohol_degree,name,cost,beer_kind,volume_ml,brewing_date,nullable_name
0.1599999964237213,13.5,Superstrong beer,2.890000104904175,Triple,330,2022-01-01,
0.05400000140070915,4.877377033233643,Ninkasi Ploploplop,2.890000104904175,India Pale Ale,330,2022-01-02,Ninkasi Ploploplop
0.004999999888241291,0.5699344873428345,Brewdog Nanny State Alcoholvrij,2.2899999618530273,Sans alcool,330,2022-01-03,Brewdog Nanny State Alcoholvrij
0.0560000017285347,6.994105339050293,Ardwen Blonde,2.0899999141693115,Best-sellers,330,2022-01-04,Ardwen Blonde
0.07000000029802322,8.930644989013672,Cuvée des Trolls,1.5499999523162842,Blonde,250,2022-01-05,
0.004999999888241291,1.2437106370925903,Weihenstephan Hefe Weizen Alcoholarm,1.590000033378601,Blanche & Weizen,500,2022-01-06,Weihenstephan Hefe Weizen Alcoholarm
0.04500000178813934,4.717409610748291,Bellfield Lawless Village IPA,2.490000009536743,India Pale Ale,330,2022-01-07,Bellfield Lawless Village IPA
0.08399999886751175,12.974271774291992,Pauwel Kwak,1.690000057220459,Belge blonde forte & Golden Ale,330,2022-01-08,Pauwel Kwak
0.06499999761581421,7.74746561050415,Brasserie De Sutter Brin de Folie,2.190000057220459,Blonde,330,2022-01-09,
0.05999999865889549,8.749608993530273,Brugse Zot blonde,1.7899999618530273,Blonde,330,2022-01-10,Brugse Zot blonde
"""
select {`beer_kind`,`price_per_l`}
group {`beer_kind`} (
  sort {+`price_per_l`}
  take 2
)
```

### SQL output

```SQL
SELECT
  DISTINCT ON (beer_kind) beer_kind,
  price_per_l
FROM
  table_0
ORDER BY
  beer_kind,
  price_per_l
```

### Expected SQL output

```SQL
table_1 AS (
  SELECT
    beer_kind,
    price_per_l,
    ROW_NUMBER() OVER (
      PARTITION BY beer_kind
      ORDER BY
        price_per_l
    ) AS _expr_0
  FROM
    table_0
)
SELECT
  beer_kind,
  price_per_l
FROM
  table_1
WHERE
  _expr_0 <= 2
```

### MVCE confirmation

- [X] Minimal example
- [X] New issue

### Anything else?

Loving PRQL. Thanks for your help and sorry if I didn't create the right query for our usecase 🙏

PrettyWood:
fix #3986


max-sixty (approved):
Awesome that you fixed this! Nice work!

prqlc/prql-compiler/src/sql/srq/preprocess.rs
```diff
@@ -173,8 +173,8 @@ pub(in crate::sql) fn distinct(
                     // DISTINCT
 
                     res.push(SqlTransform::Distinct);
-                } else if ctx.dialect.supports_distinct_on() {
-                    // DISTINCT ON
+                } else if ctx.dialect.supports_distinct_on() && range_int.end.unwrap_or(1) == 1 {
```
aljazerzen: This will not work correctly if you `take n..` (no upper bound in a range).

I've pushed a test that fails in this case.

prqlc/prql-compiler/tests/integration/snapshots/integration__queries__fmt__group_sort_limit_take_1.snap
```diff
@@ -0,0 +1,12 @@
+---
```
aljazerzen: This test seems unnecessary, as we already have `distinct_on.prql`

I'd prefer to be a bit conservative when adding integration tests, since they are slower to execute.

prqlc/prql-compiler/tests/integration/sql.rs
```diff
@@ -2034,6 +2034,37 @@ fn test_group_take_n() {
...
 
+#[test]
+fn test_group_take_n_02() {
+    assert_display_snapshot!((compile(r###"
+    prql target:sql.postgres
+
+    from employees
+    group department (
+      sort age
+      take 2..
```
max-sixty: Ah great spot @aljazerzen

aljazerzen (approved):
Looks good.

Was it hard to find where the bug is?

prqlc/prql-compiler/src/sql/srq/preprocess.rs
PrettyWood: Ah excellent catch! I messed up with my fallback lol... Thanks for the tests 🙏 
[1202f9a](https://github.com/PRQL/prql/pull/3988/commits/1202f9a16175986b03265f8199c7dd1894f97c70)

PrettyWood:
> Was it hard to find where the bug is?

Not at all :)

prqlc/prql-compiler/tests/integration/snapshots/integration__queries__fmt__group_sort_limit_take_1.snap
PrettyWood: [36eb293](https://github.com/PRQL/prql/pull/3988/commits/36eb2931761f275b750cbdc06c2468573a5b115d)

aljazerzen (approved):
> Not at all :)

That's nice to hear. I'm looking for indications that the codebase might not be ideally structured and should be made more approachable. But it looks like this is not the case.

prqlc/prql-compiler/tests/integration/snapshots/integration__queries__fmt__group_sort_limit_take_1.snap
max-sixty: > I'd prefer to be a bit conservative when adding integration tests, since they are slower to execute.

Yeah, I think we can be quite generous about unit tests; agree we should try and have fewer, more canonical integration tests if possible

PrettyWood: I tend to think the opposite. I'm working right now on the MSSQL issue and without integration tests, I feel like it's very very very hard to make sure that
- a dialect is truly supported
- you don't break anything by changing some logic
I don't mean that UT are useless or anything but looking at PRQL scope and its goal, integration tests should be IMHO way more used.
Those integration tests could run only once at night on `master` if we don't want to have them run everytime

max-sixty: Yes interesting. 

If they're helpful to you when developing, I would say to use them liberally in that case — we should do the thing that makes making contributions easier.

Personally I prefer to iterate over the very smallest possible thing — so in this case it would be making some SQL queries that work in MSSQL, and then working on getting unit tests to produce those queries. But others prefer a more "integrated" approach of getting PRQL that works on MSSQL in a single loop, even if the loop is slower, and that's OK in moderation imo...

PR #3989, opened 2023-12-21
feat: infer array to be of type of union of items

### LINKED ISSUE #2822: Type inference for arrays
Seems like we can't define a table that contains null?

```elm
from [
    {x = null},
    {x = '1'},
]
```

```
Error:
   ╭─[Root.prql:3:10]
   │
 3 │     {x = '1'},
   │          ─┬─
   │           ╰─── array expected type `null`, but found type `text`
───╯
```

aljazerzen:
Closes #2822

Behavior before this commit:
```
['hello', null] -> array of text, type error at second element
[false, 'hello'] -> array of bool, type error at second element
```

Behavior after:
```
['hello', null] -> array of text || null
[false, 'hello'] -> array of bool || text
```

### DIFF
--- a/prqlc/prql-compiler/src/semantic/resolver/types.rs
+++ b/prqlc/prql-compiler/src/semantic/resolver/types.rs
@@ -71,22 +71,9 @@ impl Resolver<'_> {
                 let mut intersection = None;
                 for item in items {
                     let item_ty = Resolver::infer_type(item)?;
-
-                    if let Some(item_ty) = item_ty {
-                        if let Some(intersection) = &intersection {
-                            if intersection != &item_ty {
-                                // TODO: compute type intersection instead
-                                return Ok(None);
-                            }
-                        } else {
-                            intersection = Some(item_ty);
-                        }
-                    }
+                    intersection = maybe_type_intersection(intersection, item_ty);
                 }
-                let Some(items_ty) = intersection else {
-                    // TODO: return Array(Infer) instead of Infer
-                    return Ok(None);
-                };
+                let items_ty = intersection.unwrap_or_else(|| Ty::new(TyKind::Any));
                 TyKind::Array(Box::new(items_ty))
             }
 
@@ -638,11 +625,14 @@ where
     F: Fn() -> Option<String>,
 {
     fn display_ty(ty: &Ty) -> String {
-        if ty.name.is_none() && ty.kind.is_tuple() {
-            "a tuple".to_string()
-        } else {
-            format!("type `{}`", write_ty(ty))
+        if ty.name.is_none() {
+            if let TyKind::Tuple(fields) = &ty.kind {
+                if fields.len() == 1 && fields[0].is_wildcard() {
+                    return "a tuple".to_string();
+                }
+            }
         }
+        format!("type `{}`", write_ty(ty))
     }
 
     let who = who();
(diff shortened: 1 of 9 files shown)


prqlc/prql-compiler/src/semantic/resolver/types.rs
```diff
@@ -72,22 +72,9 @@ impl Resolver<'_> {
...
-                        if let Some(intersection) = &intersection {
-                            if intersection != &item_ty {
-                                // TODO: compute type intersection instead
-                                return Ok(None);
-                            }
-                        } else {
-                            intersection = Some(item_ty);
-                        }
-                    }
+                    intersection = maybe_type_intersection(intersection, item_ty);
```
max-sixty: Wouldn't we generally call this a Union?

max-sixty (approved):

prqlc/prql-compiler/src/semantic/resolver/types.rs
snth: That was my immediate impression as well and I came to make the same comment.

aljazerzen: This call will create a union, but not in all cases.

If you have `intersection(int, int)`, then this is just `int`.

If you have `intersection({a = int}, {a = bool})`, then this is `{a = int || bool}`.

max-sixty: Aren't those unions? At least when I'm thinking about sets? :)

aljazerzen: They are not.

`int` is not an union. it's a primitive set.

`{a = int || bool}` is not an union. it's a type of tuples that contain a union.

snth: I thought the definition of types that we liked was that *types* are the union of all the possible values that instances of that type can take?

Another indicator is the `||` / or operator which is a logical disjunction which goes hand in hand with unions, e.g. see [Logical Disjunction > Set Theory](https://en.wikipedia.org/wiki/Logical_disjunction#:~:text=%5D%5B6%5D-,Set%20theory,-%5Bedit%5D)

aljazerzen: Ok, you are now talking about terminology here. `int` is a set of all integer numbers, which is technically a union of all integers, i.e. `0 || 1 || -1 || 2 || -2, ...`.

For this specific union, we have a "primitive" set in PRQL, called `int`. So `int` is not a union in PRQL type system.

I think we are not understanding each other here. Can both of you formulate your questions very verbosely?

snth: I'm not worried about the `int` part, it's the `int || bool` part. My understanding is that `int = {0, 1, -1, 2, -2, ...}` and `bool = {true, false}` and `int || bool = {true, false} U {0, 1, -1, 2, -2, ...} = {true, false, 0, 1, -1, 2, -2, ...}`.

Hence it seems natural to refer to these as unions.

aljazerzen: Of course, `int || bool` is a union.

Are you concerned by the naming of functions here?

aljazerzen: Oh, only now I see what you mean - it's not the naming. We should be using a union of types of the items, not their intersection.

prqlc/prql-compiler/src/semantic/resolver/types.rs
```diff
@@ -68,25 +68,15 @@ impl Resolver<'_> {
...
-                            intersection = Some(item_ty);
-                        }
+                        variants.push((None, item_ty));
                     }
                 }
-                let Some(items_ty) = intersection else {
-                    // TODO: return Array(Infer) instead of Infer
-                    return Ok(None);
-                };
+                let items_ty = Ty::new(TyKind::Union(variants));
```
PrettyWood, line 78: 👍

PrettyWood (approved):

PrettyWood:
Maybe update the PR title + add changelog

PR #3993, opened 2023-12-21
feat: type difference

aljazerzen:
For now this in an internal change without language changes.

It implements ability to express type `T - U` which contains all values of `T` that are not also in `U`.

This is needed to replace behavior provided by current lineage implementation with the type system.

Idea is that for a relation that we don't have knowledge of, we'll introduce a generic type argument:
```
from x
```
will infer:
```
module default_db {
  let x: [T]
}
```

In the case where we exclude a column from a relation, we will have to infer:
```
from x
select !{a}
```
... to have the type of `[T - {a = U}]`.

### DIFF
--- a/web/book/src/reference/spec/type-system.md
+++ b/web/book/src/reference/spec/type-system.md
@@ -236,7 +236,7 @@ A & B = (), where:
  - A and B are primitive, singleton, tuple or array
  - and A != B
 
-A - (A | B) = (), where
+A - (A | B) = ()
 
 A | () = A
 ```
--- a/web/book/src/reference/spec/type-system.md
+++ b/web/book/src/reference/spec/type-system.md
@@ -225,9 +225,9 @@ select [1.4, false, "foo"]
 
 ### Normalization
 
-Say we have a type expression E, composed of possibly many operators.
-To minimize this expression, we employ process of normalization,
-which is meant to take advantage of the following equalities:
+Say we have a type expression E, composed of possibly many operators. To
+minimize this expression, we employ process of normalization, which is meant to
+take advantage of the following equalities:
 
 ```
 A & A = A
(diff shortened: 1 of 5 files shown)


web/book/src/reference/spec/type-system.md
```diff
@@ -223,6 +223,62 @@ of entries, which can be of different types. Thus, it is a tuple.
...
+take advantage of the following equalities:
+
+```
+A & A = A
+
+A & B = (), where:
+ - A and B are primitive, singleton, tuple or array
+ - and A != B
+
+A - (A | B) = (), where
```
max-sixty:
```suggestion
A - (A | B) = ()
```

this is presumably always true, since `A - A = ()`?

web/book/src/reference/spec/type-system.md
```diff
@@ -223,6 +223,62 @@ of entries, which can be of different types. Thus, it is a tuple.
...
+
+```
+(A | B)      & C            = (A & C) | (B & C)
+A            & (B | C)      = (A & B) | (A & C)
+
+(A - B)      & C            = (A & C) - B
+A            & (B - C)      = (A & B) - C
+
+{A, B}       & {C, D}       = {A & C, B & D}
+[A]          & [B]          = [A & B]
```
max-sixty, line 258: Maybe obvious to others, but why is the type within `[A] & [B]` so different from the type from `A & B`?

web/book/src/reference/spec/type-system.md
```diff
@@ -223,6 +223,62 @@ of entries, which can be of different types. Thus, it is a tuple.
 select [1.4, false, "foo"]
 ```
 
+### Normalization
+
+Say we have a type expression E, composed of possibly many operators. To
```
max-sixty, line 228: I think it might be helpful to put the PR description here, giving the times where we do the `-` operation. I'm still thinking through it. It sounds like:
- The initial examples concern primitive types
- Then the subsequent examples also are about the type of tuples, where "type" also includes the containing columns

Is that the right way to think about them?

PR #3994, opened 2023-12-21
fix: take n rows for MSSQL

### LINKED ISSUE #2448: Problem with `take` with MsSQL
### What happened?

The generated SQL does not work for 3 reasons
* The keyword `ROWS` is missing after `OFFSET n`
* `OFFSET` and `TOP` cannot be used together
* `OFFSET` only works with `ORDER BY`
    - probably in the responsibility of the developer since offset does not make much sense without sorting

A way to fix this may be to replace `TOP` with `FETCH` ([link](https://www.sqlservertutorial.net/sql-server-basics/sql-server-offset-fetch/))

related to #1854

### PRQL input

```elm
prql target:sql.mssql

from tracks
take 3..5
```

### SQL output

```SQL
SELECT
  TOP (3) *
FROM
  tracks OFFSET 2
```

### Expected SQL output

_No response_

### MVCE confirmation

- [X] Minimal example
- [X] New issue

### Anything else?

_No response_

PrettyWood:
fix #2448


aljazerzen (approved):

PrettyWood:
There is a problem with one integration test. `OFFSET 0` always needs to be specified with `FETCH` but can't exist by itself
I'm looking at it

max-sixty:
IIUC, one downside of `FETCH` is that it [requires an `ORDER BY` clause](https://stackoverflow.com/questions/30981059/how-to-use-offset-and-fetch-without-order-by-in-sql-server).

We could try and add one in where it doesn't exist — e.g. `ORDER BY SELECT(NULL)`, or we could just raise an error where one doesn't exist. Given this is MSSQL, I would not push for something maximalist...

PrettyWood:
> [requires an ORDER BY clause](https://stackoverflow.com/questions/30981059/how-to-use-offset-and-fetch-without-order-by-in-sql-server).

Yup I planned to add an error message to just force the user to add a `sort` in PRQL. Would that be ok?

PrettyWood:
@max-sixty thanks for the `ORDER BY(SELECT NULL)` trick!!

prqlc/prql-compiler/src/sql/gen_query.rs
```diff
@@ -187,6 +187,33 @@ fn translate_select_pipeline(
...
+        if order_by.is_empty() {
+            order_by.push(sql_ast::OrderByExpr {
+                expr: sql_ast::Expr::Value(sql_ast::Value::Placeholder(
+                    "(SELECT NULL)".to_string(),
+                )),
+                asc: None,
+                nulls_first: None,
+            });
+        }
+    }
```
max-sixty, line 211: Awesome! 👏

max-sixty (approved):

PR #3995, opened 2023-12-21
test(bindings): fix after merge of #3994 (updates for the sql.mssql target)

PrettyWood:
Oopsy sorry 🙏 
https://github.com/PRQL/prql/actions/runs/7293662556


max-sixty (approved):
Thanks for updating!

I guess this was our canonical case for dialects! Otherwise very rare for these tests to fail...

PR #4006, opened 2023-12-25
feat: add `std.prql` module and rename the `std.prql_version` function to `std.prql.version`

### LINKED ISSUE #3996: `std.prql` module
Perhaps we can now move `std.prql_version` to `std.prql.version`?

https://github.com/PRQL/prql/blob/b19bd8077b9035d5bd88c9cfeddd56f6f4ab6c86/web/book/src/project/target.md?plain=1#L98-L107

(This is a breaking change, so we may need to deprecate the `prql_version` first and remove it later.)

I also imagine that it would be useful to have a function to get the current target, or to get some information as a relation.

eitsupi:
Close #3996


max-sixty (approved):
Great! Impressed you got around the keyword issue...

vanillajonathan:
Imagine if it was possible to do `std.prql.version.major` and `std.prql.version.minor`.

PR #4025, opened 2023-12-29
test: unit tests for jinja templates/dbt

nkicg6:
related to #2879


max-sixty (approved):
Very nice!

max-sixty:
FYI I just added #4026 so we can immediately see the impact on coverage for future PRs

PR #4028, opened 2023-12-31
test: Allow `prqlc fmt` to work on paths

max-sixty:
(I'm also doing this to see how easy the newer coverage approach is to see)


prqlc/prqlc/src/cli.rs
```diff
@@ -202,23 +203,24 @@ impl Command {
...
-                        .keys()
-                        .map(|x| format!("`{}`", x.display()))
-                        .sorted()
-                        .join(", ");
-                    bail!(
-                        "Currently `fmt` only works with a single source, but found multiple sources: {paths:?}"
-                    )
-                }
-                let (_, source) = sources.sources.into_iter().next().unwrap();
+                for (path, source) in sources.sources {
```
max-sixty, line 207: @aljazerzen do you think we should add the project root to the `SourceTree` struct? Currently it doesn't know the root and so can't write to the correct location.

Or is there a better data structure to use?

prqlc/prqlc/tests/test.rs
```diff
@@ -322,6 +323,24 @@ fn format() {
 
     ----- stderr -----
     "###);
+
+    // Single file
```
max-sixty, line 332:
```suggestion
    // FIXME: not good tests, since they don't actually test that the code was formatted...
    // Single file
```

prqlc/prqlc/src/cli.rs
aljazerzen: `SourceTree` is the data structure that is supposed to pass all needed sources to the compiler, along with their hierarchy. It would make sense to also information to obtain the absolute path to each of the files, so this can be displayed in error messages for example, or be used when formatting the files. 

So yes, an absolute path to the directory containing all the source files is a good idea. And it should be `Option<PathBuf>`, as this does not make sense when compiling a single snippet.

prqlc/prqlc/tests/test.rs
aljazerzen: Yeah, this would be better if we read the file afterwards.

aljazerzen (approved):

prqlc/prqlc/src/cli.rs
max-sixty: One alternative would be to store absolute paths in `sources` — WDYT?

max-sixty: (implemented with the additional `root` struct member, but can refactor later if needed)

aljazerzen: I'm uneasy with absolute paths in sources, because then the mapping looks like it has the root at `/`. Because of that the resolver would have to find the longest common prefix in all paths to determine the root.

So current state better.

PR #4030, opened 2023-12-31
devops: task to run coverage locally

max-sixty:
I couldn't get the VSCode extension to show coverage, but may return to this.

### DIFF
--- a/Taskfile.yml
+++ b/Taskfile.yml
@@ -320,8 +320,14 @@ tasks:
       - cargo insta test --accept --lib -- --quiet
 
   test-rust-coverage:
-    desc: Run tests with instrumentation for code coverage
+    desc: Run tests with instrumentation for code coverage.
+    summary:
+      Run tests with instrumentation for code coverage. This works with VS
+      Code's [Coverage
+      Gutters](https://marketplace.visualstudio.com/items?itemName=ryanluker.vscode-coverage-gutters)
+      extension.
     env:
+      # Use a different target dir so we don't poison the cache
       CARGO_TARGET_DIR: "./target-cov"
     cmds:
       - cargo llvm-cov  --lcov --output-path lcov.info --features=test-dbs
--- a/Taskfile.yml
+++ b/Taskfile.yml
@@ -28,9 +28,9 @@ vars:
     prql-lang.prql-vscode
     rust-lang.rust-analyzer
   cargo_crates: >
-    bacon cargo-audit cargo-insta cargo-release default-target mdbook
-    mdbook-admonish mdbook-footnote mdbook-toc nextest wasm-bindgen-cli wasm-opt
-    wasm-pack
+    bacon cargo-audit cargo-insta cargo-llvm-cov cargo-release default-target
+    mdbook mdbook-admonish mdbook-footnote mdbook-toc nextest wasm-bindgen-cli
+    wasm-opt wasm-pack
   # Excluding `elixir` atm given it's not enabled on Mac and currently unsupported
   brew_dependencies: |
     hugo
@@ -68,6 +68,7 @@ tasks:
       - task: install-pre-commit
       - task: install-npm-dependencies
       - pre-commit install-hooks
+      - rustup component add llvm-tools-preview
       - cmd: |
           echo "
           🟢 Setup complete! ✅🚀"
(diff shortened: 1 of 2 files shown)


Taskfile.yml
```diff
@@ -318,6 +318,25 @@ tasks:
...
+      # https://github.com/mozilla/grcov#generate-a-coverage-report-from-coverage-artifacts
+      CARGO_INCREMENTAL: 0
+      RUSTFLAGS: "-Cinstrument-coverage"
+      LLVM_PROFILE_FILE: "cargo-test-%p-%m.profraw"
+      # Use a different target dir so we don't poison the cache
+      CARGO_TARGET_DIR: "./target-cov"
+    cmds:
+      # Setup
+      - cargo install grcov
+      - rustup component add llvm-tools-preview
```
aljazerzen: I'd ask for these setup tasks not be present here, as this does not work on my machine.

aljazerzen (commented):
I remember using [tarpaulin](https://github.com/xd009642/tarpaulin) and it was quite pleasant.

aljazerzen:
I've pushed [coverage-llvm-cov](https://github.com/PRQL/prql/compare/main...coverage-llvm-cov), it adapts this PR and emits lcov.info file, which can be read by [a few different vscode extensions](https://marketplace.visualstudio.com/items?itemName=ryanluker.vscode-coverage-gutters).

max-sixty:
> I remember using [tarpaulin](https://github.com/xd009642/tarpaulin) and it was quite pleasant.

`-C instrument-coverage` & `grcov` additionally tracks the lines hit by calling `prqlc` in tests, which is much much better, since about half of our gaps weren't actually gaps. (This was done for Ci in https://github.com/PRQL/prql/pull/3987 though spreads across a few PRs so maybe wasn't easy to track)

max-sixty:
> I've pushed [coverage-llvm-cov](https://github.com/PRQL/prql/compare/main...coverage-llvm-cov),

Ah this seems better than using `-C instrument-coverage` & `grcov` alone; seems like it's a wrapper of that.

Annoying that I hadn't seen this before! 

Probably we should switch CI into using that too...

max-sixty:
This is changed to use `coverage-llvm-cov`! It works with VS Code's coverage gutters.

I'll plan to merge unless there are any objections

max-sixty:
@aljazerzen I didn't put this in the main test task:
- It's much slower — 54s vs 20s when running from an existing cache — partly because it has to recompile the existing binaries, because incremental caching is disabled
- It would not be cached with other builds

aljazerzen:
> incremental caching is disabled

I don't think rust supports incremental compilation at all - you mean it has to re-compile all of the dependencies? 

In my commit, I've set:
```
      # Use a different target dir so we don't poison the cache
      CARGO_LLVM_COV_TARGET_DIR: ../target-cov
```
... so first run takes ~1min and re-run takes ~20sec, even if I compile something else inbetween.

Are we using this task in the pipelines too?

max-sixty:
> I don't think rust supports incremental compilation at all - you mean it has to re-compile all of the dependencies?

https://blog.rust-lang.org/2016/09/08/incremental.html 

On my commit, it seems to be recompiling our own binaries on each run...

---

On your commit I get

```sh
task prqlc:test

error: failed to parse nextest config at `/Users/maximilian/workspace/prql/.config/nextest.toml`

Caused by:
  invalid type: map, expected a string for key `profile.default.slow-timeout`
error: process didn't exit successfully: `/Users/maximilian/.rustup/toolchains/1.74.0-aarch64-apple-darwin/bin/cargo nextest run --manifest-path /Users/maximilian/workspace/prql/Cargo.toml --target-dir /Users/maxim
ilian/workspace/prql/prqlc/../target-cov --features=test-dbs -p prqlc-ast -p prqlc-parser -p prql-compiler -p prqlc -p prql-compiler-macros -p compile-files -p prql -p prql-java -p prql-js -p prqlc-clib -p prql-pyt
hon` (exit status: 96)

aljazerzen:
> https://blog.rust-lang.org/2016/09/08/incremental.html

I was aware of experimental support for it, but I didn't know it's rolled out yet!
I just checked and does work very well! It is enabled by default for dev and test profiles, so you are correct that we'd not be taking advantage of that here.

---

But regardless of incremental compilation, when there was no changes to source, no crates should be rebuilt.

Ah, this is why:

> cargo llvm-cov/cargo llvm-cov run/cargo llvm-cov nextest cleans some build artifacts by default to avoid false positives/false negatives due to old build artifacts. This behavior is disabled when --no-clean

(from [here](https://github.com/taiki-e/cargo-llvm-cov#basic-usage))

---

About nextest config error: maybe try updating nextest? It works on my computer^tm with nextest 0.9.54.

max-sixty:
> About nextest config error: maybe try updating nextest? It works on my computer^tm with nextest 0.9.54.

Sorry, you're right!

---

OK, I'll merge this, then!

Issue #2930, opened 2023-06-29
Enums / named unions


vanillajonathan:
### What's up?

There could be a way to declare enums which could be used in queries.

## Example with implicit assignment:
```elm
enum InvoiceKind
{
    Paid,
    Unpaid
}

from invoices
filter invoice_status == InvoiceKind.Paid
```

Here the variant `Paid` is implicitly assigned the value 0, and `Unpaid` is implicitly assigned the value 1.

## Example with explicit assignment
```elm
enum InvoiceKind
{
    Paid = 10,
    Unpaid
}
```
Here the variant `Paid` is explicitly assigned the value 10, and `Unpaid` is implicitly assigned the value 11.

## Example with string enums
Here the variant is assigned a string.
```elm
enum InvoiceKind
{
    Paid = "PAID",
    Unpaid = "UNPAID"
}
```

aljazerzen:
This is something anticipated by the type system and something that will be useful for expanding our std module (for example for #2125).

At the moment this is not easily implementable, as our backend does not have capabilities for checking a type at runtime. I also don't have a plan for how will this be implemented.

I do know that enums should not be just enumerations of values, but a named union - a proper [sum-type](https://en.wikipedia.org/wiki/Tagged_union):

```
type InvoiceStatus = (
    Paid = (),
    Unpaid = { amount = float },
    Cancelled = { reason = str },
)
```
(I'm using made-up syntax that is most consistent with the rules for other type-system constructs.)

Related resource: https://www.parsonsmatt.org/2019/03/19/sum_types_in_sql.html

vanillajonathan:
But can this still be used together with `filter` to filter integers or strings?
Maybe the column `InvoiceKind` contains integers such as `1`, `2`, `3`, or a string such as `"PAID"` or `"UNPAID"`.

I can't remember what those integers mean so I want to look them via enum.

aljazerzen:
Yes it could be used in `filter`. I'm thinking about pattern matching very similar to Rust's:
```
from invoices
filter (match this.status {
  InvoiceStatus.Paid => true,
  InvoiceStatus.Unpaid { amount } => amount < 10.0,
  _ => false,
})
```

---

I didn't talk about it above, but the "discriminator" between the enum values (i.e. the thing that is actually written into the table) should be configurable, as you propose.

[...]

PR #4038, opened 2024-01-02
feat: named unions

aljazerzen:
Support for parsing and codegen of:

```elm
    type Status = (
      Paid = () ||
      Unpaid = float ||
      {reason = text, cancelled_at = timestamp} ||
    )
```

As we don't have any code for interaction with this, this does not change the behavior in any way, so I'm not adding a changelog.

### DIFF
--- a/prqlc/Taskfile.yaml
+++ b/prqlc/Taskfile.yaml
@@ -17,7 +17,7 @@ tasks:
     cmds:
       - cmd: |
           # remove trailing whitespace
-          rg '\s+$' --files-with-matches --glob '!*.rs' . \
+          rg '\s+$' --files-with-matches --glob '!*.{rs,snap}' . \
           | xargs -I _ sh -c "echo Removing trailing whitespace from _ && sd '[\t ]+$' '' _"
 
       - cmd: |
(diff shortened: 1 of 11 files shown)


vanillajonathan:
Seems related to #2930. This is cool but the syntax is quite jarring. I would have expected something much simpler and more akin to what is available in other languages.

max-sixty:
> This is cool but the syntax is quite jarring. I

In a way these are similar to Enums, since they're Sum Types. Though I don't see the immediate need to build that feature now. (and because we're early in this feature, and it is mostly internal, we shouldn't prioritize aesthetics highly)

One small change that might make it look better is to put the binary operators at the beginning of the next line. The standards for this changed to this in python a few years ago, which made pipelines of binary operators easier to read.

aljazerzen:
Like this?
```
    type Status = (
      || Paid = ()
      || Unpaid = float
      || {reason = text, cancelled_at = timestamp}
    )
```

We don't allow leading `||`, so this won't work. If you think about `||` as a comma, it makes sense that they are at the end of the line.

But again, these are just aesthetics.

max-sixty:
Almost, but there's no leading `||` (so the string of characters is still the same as the prior version)

```diff
    type Status = (
-      || Paid = ()
+      Paid = ()
      || Unpaid = float
      || {reason = text, cancelled_at = timestamp}
    )
```

PR #4040, opened 2024-01-02
feat: improve codegen for functions

aljazerzen:
Add types annotations.


prqlc/prql-compiler/src/codegen/ast.rs
```diff
@@ -136,10 +136,15 @@ impl WriteSource for ExprKind {
                 Some(r)
             }
             Func(c) => {
-                let mut r = String::new();
+                let mut r = "func ".to_string();
```
max-sixty, line 139: I guess this is going all-in on having `func`, no strong opposition for the moment

max-sixty (approved):

PR #4044, opened 2024-01-02
ci: Use `llvm-cov` in CI

max-sixty:
Based on #4030 (but still for CI)


max-sixty:
This also seems to allow coverage of the bindings, which is great.

I left a TODO re branch / region coverage

PR #4045, opened 2024-01-02
ci: Use branch coverage

max-sixty:
Trying this to assess whether the drop in measured coverage is accurate


max-sixty:
@PRQL/prql-team do we think this is more accurate? It's much stricter. 

For example these are only tagged as partially covered, I guess because the `Ok` and `Err` aren't both covered:

image

...which feels maybe too strict?

aljazerzen:
> For example these are only tagged as partially covered, I guess because the Ok and Err aren't both covered:

These cases will never be covered.

It is too strict.

max-sixty:
Yeah agree; closing.

Issue #4050, opened 2024-01-04
Zig tests started failing


max-sixty:
### What's up?

Any idea what caused this? I'm guessing it's an external change, maybe a zig version? I don't think anything has changed in that path, and the commit it failed on is 

https://github.com/PRQL/prql/actions/runs/7412719385/job/20172886376

CC @vanillajonathan as our resident zig expert! :)

Disabled in #4051

vanillajonathan:
Our CI uses the master branch of the Zig compiler. Commit https://github.com/ziglang/zig/commit/c20ad51c621ba18d2c90cc96d1b550831dc5d7a3 changes a bunch of stuff which breaks it because it removes the `linkSystemLibraryName` function.

We use the [goto-bus-stop/setup-zig](https://github.com/goto-bus-stop/setup-zig) GitHub Action which allows us to optionally specify the version of the Zig compiler to use. We could set it to version `0.11.0` for now.

max-sixty:
Yes, either pinning or adjusting for the newer zig sounds good!

max-sixty:
#4053

PR #4053, opened 2024-01-05
ci: Pin Zig compiler version

vanillajonathan:
Pin the Zig compiler version so the GitHub Action doesn't use the nightly version which has breaking changes.

This PR is related to issue #4050.

### DIFF
--- a/.github/workflows/test-clib.yaml
+++ b/.github/workflows/test-clib.yaml
@@ -34,7 +34,8 @@ jobs:
         uses: goto-bus-stop/setup-zig@v2
         with:
           cache: false
-      # TODO: Broken so disabling — https://github.com/PRQL/prql/issues/4050
-      # - name: Run example minimal-zig
-      #   working-directory: prqlc/bindings/clib/examples/minimal-zig
-      #   run: make run test
+          # Pinned version — https://github.com/PRQL/prql/issues/4050
+          version: 0.11.0
+      - name: Run example minimal-zig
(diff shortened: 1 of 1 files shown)


.github/workflows/test-clib.yaml
```diff
@@ -34,7 +34,8 @@ jobs:
         uses: goto-bus-stop/setup-zig@v2
         with:
           cache: false
-      # TODO: Broken so disabling — https://github.com/PRQL/prql/issues/4050
-      # - name: Run example minimal-zig
-      #   working-directory: prqlc/bindings/clib/examples/minimal-zig
-      #   run: make run test
+          # Pinned version — https://github.com/PRQL/prql/issues/4050
+          version: 0.11.0
```
eitsupi, line 38: Just a question: When is this version supposed to be updated and by whom?
For example, the version of Elixir is specified in the same way, but it has not been updated in two years and is probably currently inappropriate.

https://github.com/PRQL/prql/blob/eaf7839e57c827d10c85b4498de9f906c55a213b/.github/workflows/test-elixir.yaml#L27-L28

In my opinion it would be useful to be able to specify the latest release here. (This is hacky but I think we can use a script to set the latest version here)

vanillajonathan: Good question, I don't have any good answer.
I agree that it would be  useful to be able to specify the latest release but the `goto-bus-stop/setup-zig` GitHub Action does not have any support for that.

I files a issue: https://github.com/goto-bus-stop/setup-zig/issues/65

PR #4056, opened 2024-01-06
test: Add tests for Lexer output

max-sixty:
My overall goal is to output each stage of the compiler pipeline for a set of canonical queries.

I started with the lexer output. I had to make more changes than I expected — to get good `Display` & `Debug` on the tokens, we needed to add a couple of custom structs and implement `Display` & `Debug` on those.

Maybe there's a less invasive way?


max-sixty:
...and probably we pick a couple of extra-canonical queries, rather than do them all — there are a lot of lines of snapshots here; and they'll be many more if we add all the stages

prqlc/prqlc-parser/src/lexer.rs
```diff
@@ -506,71 +504,56 @@ impl std::fmt::Display for Token {
...
+        write!(f, "{} ", self.0)
+    }
+}
+impl std::fmt::Debug for TokenSpan {
+    fn fmt(&self, f: &mut std::fmt::Formatter) -> std::fmt::Result {
+        write!(f, "{}..{}: {:?}", self.1.start, self.1.end, self.0)
+    }
+}
+
+pub struct TokenStream(pub Vec<TokenSpan>);
```
aljazerzen: Technically, this is not a stream - it's should be named `TokenVec`.

prqlc/prqlc-parser/src/lexer.rs
```diff
@@ -462,17 +459,18 @@ impl std::cmp::Eq for Token {}
 impl std::fmt::Display for Token {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
         match self {
-            Token::NewLine => write!(f, "new line"),
```
aljazerzen, line 465: I think this will cause error messages to say "expected x found
", instead of "expected x found new line"

prqlc/prqlc-parser/src/lexer.rs
```diff
@@ -506,71 +504,56 @@ impl std::fmt::Display for Token {
...
 }
 
+#[derive(PartialEq, Eq, Hash)]
+pub struct TokenSpan(pub Token, pub std::ops::Range<usize>);
+
+impl std::fmt::Display for TokenSpan {
+    fn fmt(&self, f: &mut std::fmt::Formatter) -> std::fmt::Result {
+        write!(f, "{} ", self.0)
+    }
+}
```
aljazerzen: This is confusing, let's just stick with debug

aljazerzen: Confusing for new line tokes or tokens whose display contains a space.

aljazerzen (approved):
Looks good. Maybe we should be more explicit about compiler stages / AST passes and snapshot the queries/ in each of the stages.

It's also useful to have lexer exposed for highlighting in (maybe) lsp or somewhere else.

prqlc/prqlc-parser/src/lexer.rs
max-sixty: Thanks; missed that — I made some changes before realizing they hit error messages, and then forgot to undo this one...

PR #4060, opened 2024-01-08
refactor: Move `WithErrorInfo` to `prqlc-ast`, use in `prqlc-parser`

max-sixty:
A small refactor while working on other parts of the codebase this weekend

This led to some yak-shaving with coverage, and then the `anyhow` implementation, and then a default feature. Think it's probably better than not having it, even if it wasn't _that_ productive

### DIFF
--- a/.github/actions/build-prqlc/action.yaml
+++ b/.github/actions/build-prqlc/action.yaml
@@ -31,7 +31,7 @@ runs:
         save-if:
           ${{ (github.ref == 'refs/heads/main') && contains(inputs.target,
           'musl') }}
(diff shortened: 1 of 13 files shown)


aljazerzen (approved):
I agree, WithError would be in the same crate as Error. In the future that should be a separate "common" crate

max-sixty:
> In the future that should be a separate "common" crate

Ah and you're thinking that's different from `prql-ast`? I had been thinking that was some of the way there...

aljazerzen:
Yeah, it is, it's better than being in the prqlc-parser, but Error is not part of our AST, so it's a bit strange that it is in prqlc-ast.

Issue #4061, opened 2024-01-08
Python compiler needs a flag to return the full sequence of errors as json


brunorpinho:
Hello,

First of all, I would like to thank the devs for the amazing work! You have solved a huge huge huge problem across multiple industries by making this pipeline tool.

I have an use case where I want the full detailed json exception instead of a simple formatted SyntaxError. Is it possible to add a simple flag to the [`compile`](https://github.com/PRQL/prql/blob/aec243c0fe152062ccca8bb819e4ad3c717778ab/prqlc/bindings/python/src/lib.rs#L8) function called `on_exception_return_json` or something like that to get the result of [`e.to_json()`](https://github.com/PRQL/prql/blob/main/prqlc/prql-compiler/src/error_message.rs#L155) into a Python dict? It can even come as a variable of the Python `Exception` class.

I basically need the vector of `ErrorMessage` as json.

Cheers,
Bruno

max-sixty:
Yes, I could see a couple of approaches:

### Another function — `prql_to_sql`

`prql_to_sql` function that is similar to `compile`, but raises an error as pure JSON. It would operate similarly to `prql_to_pl`, which gives a long string of JSON:

```python
prql.prql_to_pl("^")


---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
Cell In[12], line 1
----> 1 prql.prql_to_pl("^")

ValueError: {"inner":[{"kind":"Error","code":null,"reason":"unexpected ^","hints":[],"span":"1:0-1","display":"Error: \n   ╭─[:1:1]\n   │\n 1 │ ^\n   │ ┬  \n   │ ╰── unexpected ^
\n───╯\n","location":{"start":[0,0],"end":[0,1]}}]}
```

...rather than a nice Error message:

```python

prql.compile("frm artists")


Traceback (most recent call last):

  File /opt/homebrew/lib/python3.9/site-packages/IPython/core/interactiveshell.py:3508 in run_code
    exec(code_obj, self.user_global_ns, self.user_ns)

  Cell In[3], line 1
    prql.compile("frm artists")

  File <string>
SyntaxError: Error:
   ╭─[:1:1]
   │
 1 │ frm artists
   │ ─┬─
   │  ╰─── Unknown name
───╯
```

I think this could be a quite small PR.

### Add the JSON onto the `Exception`

[...]

PR #4066, opened 2024-01-10
feat: Add function to python binding with JSON error

max-sixty:
This does the very basic version of #4061.

I tried to do the advanced version of https://github.com/PRQL/prql/issues/4061, where we add an attribute onto the exception, but couldn't manage. I think we need to wait for the next version of PyO3, where https://github.com/PyO3/pyo3/pull/3475 will work (using that exact code generates errors in 0.20.2)


eitsupi:
Why not use such functionality in `prql_to_pl`, `pl_to_rq`, and `rq_to_sql` instead of adding a new function `prql_to_sql`?
It seems logical to me that functions that provide a lower level API would provide more detailed errors.
(I think JS bindings can be updated in the same way)

max-sixty:
> Why not use such functionality in `prql_to_pl`, `pl_to_rq`, and `rq_to_sql` instead of adding a new function `prql_to_sql`? It seems logical to me that functions that provide a lower level API would provide more detailed errors. (I think JS bindings can be updated in the same way)

Yes, this is a good point! I think that would have been a better approach.

I think fairly soon we can just have an exception that has the full errors at attributes, rather than two different types of exception. Basically when we get the next PyO3 version.

Unless you feel strongly, I won't revert this — it labelled as being temporary, and it does seem marginally useful...

max-sixty:
Actually on further reflection, maybe this is worse and we should revert? And instead just suggest that folks use something like:

```python
try:
  x = prql_to_pl(query)
  x = pl_to_rq(x)
  x = rq_to_sql(x)
except Exception as e:
  json_exception = e
```

WDYT @eitsupi (/ others)?

eitsupi:
I feel it would be better to revert it for the following reasons:

1. Function name doesn't say what it does. (The name `prql_to_sql` does not indicate the nature of the error being a json.)
2. I think it's quite difficult to understand that there are multiple functions that do almost the same thing.

max-sixty:
OK agree, thanks @eitsupi . 

Reverted in #4074, and we can do the good thing after PyO3 is released

PR #4073, opened 2024-01-11
test: Reenable Windows stack overflow test

max-sixty:
Expecting this to fail, but opening it to track the issue, given its opacity.

ref https://github.com/PRQL/prql/issues/2857


max-sixty:
Superseeded by #4103

PR #4075, opened 2024-01-12
ci: Build `prqlc-clib` as artifacts

### LINKED ISSUE #3953: Publish the library files as an artifact after builds
### What's up?

Regarding ongoing discussions on Discord on the pros/cons of having a [mono-repo vs a repo dedicated for each language binding](https://discord.com/channels/936728116712316989/1177922526899351632), we came to a point where the lack of proper packaging of the library files (`.dll`, `.dylib` and `.so`) is a major impediment when executing the "repo by language bindings" strategy.

This issue is to track progress on packaging library files during builds and make them available for other repos.

max-sixty:
Closes https://github.com/PRQL/prql/issues/3953

Notes are inline about the exact names and structure being very open to change.

It's mostly a copy-paste from `build-prqlc` — though it's also different enough that I wouldn't suggest trying to make a generic action atm.

We had a discussion on crate naming on Discord, which might make names change: https://discord.com/channels/936728116712316989/1195504018882035752/1195513591302983731

### DIFF
--- a/prqlc/bindings/clib/Cargo.toml
+++ b/prqlc/bindings/clib/Cargo.toml
@@ -8,6 +8,10 @@ repository.workspace = true
 rust-version.workspace = true
 version.workspace = true
 
+# This means we can build with `--features=default`, which can make builds more generic
+[features]
+default = []
+
 [lib]
 crate_type = ["staticlib", "cdylib"]
 doctest = false
--- a/prqlc/bindings/clib/Cargo.toml
+++ b/prqlc/bindings/clib/Cargo.toml
@@ -13,8 +13,10 @@ version.workspace = true
 default = []
 
 [lib]
+# We produce both of these at the moment, but could consider refining this. ref
+# https://github.com/rust-lang/cargo/issues/8607 &
+# https://github.com/rust-lang/rust/issues/59302
 crate_type = ["staticlib", "cdylib"]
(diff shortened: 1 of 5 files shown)


prqlc/bindings/clib/Cargo.toml
```diff
@@ -8,8 +8,13 @@ repository.workspace = true
 rust-version.workspace = true
 version.workspace = true
 
+# This means we can build with `--features=default`, which can make builds more generic
+[features]
+default = []
+
 [lib]
 crate_type = ["staticlib", "cdylib"]
+# crate_type = ["cdylib"]
```
max-sixty: @eitsupi do you know what we should have here? We had had it just be `cdylib`, but then nothing is built under `musl`. Are we OK to have both?

eitsupi: I don't know but maybe rust-lang/cargo#8607, rust-lang/rust#59302?
Maybe it's better to just build gnu for now?

max-sixty: OK, I'll leave them both in for the moment, no harm in producing too much, and we can refine later.

We only have targets for `musl` for `prqlc` so I won't make new `gnu` ones unless we know there'll be a problem. But can move to `gnu` if anyone raises an issue...

PR #4077, opened 2024-01-13
refactor: Rename C bindings to `prqlc-c`

max-sixty:
As discussed on Discord (probably we should publish the proposed naming approach)


Taskfile.yml
```diff
@@ -413,12 +413,20 @@ tasks:
...
-    - cp target/release/libprqlc_lib.* prqlc/bindings/clib/libprqlc_lib.h
+    - cp target/release/libprqlc_c.* prqlc/bindings/prqlc-c/libprqlc_c.h
       prqlc/bindings/php/lib/
     - cd prqlc/bindings/php && composer install
 
+  build-prqlc-c-header:
+    desc: Build the C header for the C bindings.
+    dir: prqlc/bindings/prqlc-c
+    cmds:
+      - cargo install --locked cbindgen
```
max-sixty: @aljazerzen do I remember you preferred not to have these here? It does ensure the task works, without slowing much down, even if someone hasn't run `task setup-dev`...

(also not _that_ confusing if someone hasn't yet installed it, so I'm fine to remove it...)

aljazerzen: Yeah, sorry, this would fail on my box. The development shell is not allowed to install stuff globally.

aljazerzen (approved):

PR #4081, opened 2024-01-13
docs: Write down naming approach

max-sixty:
As discussed in https://discord.com/channels/936728116712316989/1195504018882035752

Do we all agree?

### DIFF
--- a/web/book/src/project/bindings/README.md
+++ b/web/book/src/project/bindings/README.md
@@ -52,4 +52,5 @@ Nascent bindings are in development, and may not yet fully work.
 Over time, we're trying to move to a consistent naming scheme:
 
 - Crates are named `prqlc-$lang`.
-- Packages are published to each language's package repository as `prqlc`.
+- Where possible, packages are published to each language's package repository
+  as `prqlc`.
--- a/web/book/src/project/bindings/README.md
+++ b/web/book/src/project/bindings/README.md
@@ -46,3 +46,10 @@ Nascent bindings are in development, and may not yet fully work.
 
 - [.NET](./dotnet.md)
 - [PHP](./php.md)
+
+## Naming
+
+Over time, we're trying to move to a consistent naming scheme:
+
+- Crates are named `prqlc-$lang`.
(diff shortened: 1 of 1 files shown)


aljazerzen (commented):
Beautiful :D

eitsupi:
I think this is a good approach, but as I think I've commented somewhere before, please note that changing prqlr to prqlc from now on is not allowed by CRAN so I can't change it.

max-sixty:
> I think this is a good approach, but as I think I've commented somewhere before, please note that changing prqlr to prqlc from now on is not allowed by CRAN so I can't change it.

OK, I can add.

Would we ever want to start a new package `prqlc`? (Or that's also not allowed)

@eitsupi FWIW I was less keen on making lots of changes, but I can see the benefit of unifying things. So from my perspective no rush or pressure to change the external interfaces...

eitsupi:
> Would we ever want to start a new package `prqlc`? (Or that's also not allowed)

Of course it is possible, but it would mean having to maintain two parallel GitHub repositories and switching between them.
It would make sense if, for example, this was close to version 1.0.0, etc, I don't want to do that at the moment because it would be too much trouble.

eitsupi:
Perhaps the most practical way is to publish the current prqlr as a new prqlc and publish a new version of the prqlr package that just re-exports prqlc.
https://stackoverflow.com/questions/22654421/how-to-rename-an-existing-r-package-on-cran

max-sixty:
> Perhaps the most practical way is to publish the current prqlr as a new prqlc and publish a new version of the prqlr package that just re-exports prqlc.
> https://stackoverflow.com/questions/22654421/how-to-rename-an-existing-r-package-on-cran

Yes, that's what we do for `prql-compiler` now.

---

TBC, no rush for these. The owner ultimately decides, so up to you on what we do with `prqlr`

eitsupi:
Of course, I am not sure if that is actually done. Aside from changing the package, changing the repository could invalidate the URLs referenced from CRAN, so at worst the package will be removed from CRAN in the process of being worked on.

PR #4083, opened 2024-01-13
internal: Combine `prql-compiler` into `prqlc`

max-sixty:
This is a big-in-size-but-small-in-impact PR which:
- Renames the `prql-compiler` crate as `prqlc` crate
- Moves the existing `prqlc` binary to `prqlc/cli` and puts it behind a feature `cli` enabled by default
- Adds an crate `prql-compiler` which just references `prqlc`, so that existing dependents don't need to change their code immediately


eitsupi:
How do you plan to support package managers that can be used to install the CLI?
Change the package name to something like prqlc-cli?
https://repology.org/project/prqlc/versions

max-sixty:
> How do you plan to support package managers that can be used to install the CLI?

Sorry if I'm being slow — do you mean that some package managers can't have both a library and a binary at the same name?

For example, the `cargo` command to install the CLI would be `cargo install prqlc`. But are there others like repology which can't?

max-sixty:
(because moving the files around means we break some links, I'm going to unblock the merge constraints, merge, add them back, and ensure everything passes)

eitsupi:
> Sorry if I'm being slow — do you mean that some package managers can't have both a library and a binary at the same name?

My point was whether the manifest on each package manager also needs to be updated.
In other words, if the CLI tool that was previously called prqlc is officially renamed to prqlc CLI, do we need to update the manifests and package names on each package manager as well?

max-sixty:
> In other words, if the CLI tool that was previously called prqlc is officially renamed to prqlc CLI, do we need to update the manifests and package names on each package manager as well?

I was thinking that we'd still call the CLI `prqlc`. So I hadn't foreseen updating any package managers. But not confident at all — are there pages that are possibly not correct now?

eitsupi:
If we continue to call it prqlc, I don't see a problem.
I have not yet fully grasped the name of the Rust crate.

Thanks for your explanation.

PR #4088, opened 2024-01-15
test: Move `prqlc-parser` tests into `prqlc-parser`

max-sixty:
Was there a good reason to have these in `prqlc`? Maybe we didn't want to create more binaries?

I think these are small enough to have as unit tests in the `prqlc-parser` crate, any objections?


aljazerzen:
The hope was to eliminate the need to run parser lib tests at all. But that effort is vain.

aljazerzen (approved):

max-sixty:
Yeah. It is nice to be able to only run the parser tests when working on the parser...

PR #4104, opened 2024-01-18
test: Demo of printing the stack

max-sixty:
A slicker version of https://github.com/PRQL/prql/issues/2857#issuecomment-1603190843 in case that's helpful.

Run with:

```
RUST_BACKTRACE=1 RUST_LOG=trace cargo +nightly test -p prqlc --lib -- long_query_inline --nocapture
```

...may want to redirect to a file since it prints a lot...


max-sixty:
Superseeded by #4103

PR #4105, opened 2024-01-18
internal: Add comments to the lexer

max-sixty:
This doesn't impact any external interfaces atm — it adds comments to the lexer and filters them out before they hit the parser.

While we're still iterating on how comments (+ other significant whitespace) will be part of the formatter, we'll need comments in the lexer.

### DIFF
--- a/prqlc/prqlc-parser/src/lexer.rs
+++ b/prqlc/prqlc-parser/src/lexer.rs
@@ -621,13 +621,13 @@ mod test {
 
     #[test]
     fn test_single_line_comment() {
-        assert_display_snapshot!(Token::Comment("This is a single-line comment".to_string()), @r###"
-        #This is a single-line comment
+        assert_display_snapshot!(Token::Comment(" This is a single-line comment".to_string()), @r###"
+        # This is a single-line comment
         "###);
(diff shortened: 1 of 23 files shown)


vanillajonathan:
It would useful to have docblock comments in the AST so it could be used for documentation generation.

aljazerzen (approved):

max-sixty:
> It would useful to have docblock comments in the AST so it could be used for documentation generation.

Yes v good point

vanillajonathan:
The lexer now creates a token for `DocComment` which is good but it is not available in PL. It would be good if the PL also contained that.

max-sixty:
> The lexer now creates a token for `DocComment` which is good but it is not available in PL. It would be good if the PL also contained that.

I agree but it might require some invasive work on the parser (check out the recent discussion about adding comments). Possibly we put this on the Statement struct and limit them doc comments to those...

aljazerzen:
> Possibly we put this on the Statement struct and limit them doc comments to those...

This seems reasonable - this would cover docs for variables (including functions), types, and modules. Is there anything else, we'd want to cover>?

PR #4110, opened 2024-01-19
internal: Add doc comments to lexer

max-sixty:


aljazerzen:
So the syntax for a doc comment is this:

```
#! This is a doc comment of statement a
let a
```

I would suggest the following:
- `## comment` to document the next statement,
- `#! comment` to document the parent module,
- `@expr` to annotate the next statement,
- `@!expr` to annotate the parent module.

max-sixty:
> * `## comment` to document the next statement,
> 
>     * `#! comment` to document the parent module,
> 
>     * `@expr` to annotate the next statement,
> 
>     * `@!expr` to annotate the parent module.

Yes good point that commenting on the parent module would be helpful.

I'm less keen on `##`, because we can get something very close by commenting out a set of lines that are partially already commented:

```
from artists
# derive foo = 5
derive foo = 10
```

Commenting all those lines out can make

```
# from artists
# # derive foo = 5
# derive foo = 10
```

...or even without the space if using a manual vim selection.

---

FWIW I used to forget exactly how the doc-comment vs. module-doc-comment worked in rust frequently.

One option would be to link these a bit closer with something like:

- `#. doc comment` for the next statement
- `#.. doc comment` for the parent module

...though it's not perfect (there's no `#...` for the parent module! `#.` might seem like "this" statement rather than the next statement...)

aljazerzen:
> because we can get something very close by commenting out a set of lines that are partially already commented:

Good point. Although this does not seem like a large problem to me.

---

For reference, this is Rust:
- `///` is a doc comment for the next statement
- `//!` is a doc comment for the parent module
- `#[ ... ]` is a proc macro for the next statement
- `#![ ... ]` is a proc macro for the parent statement

---

So you are proposing this:

- `#. doc comment` for the next statement
- `#.. doc comment` for the parent module
- `@annotation` for the next statement

What about this:

- `#. doc comment` for the next statement
- `#! doc comment` for the parent module
- `@annotation` for the next statement
- `@!annotation` for the parent statement

max-sixty:
> * 
> 
> 
> What about this:
> 
>     * `#. doc comment` for the next statement
> 
>     * `#! doc comment` for the parent module
> 
>     * `@annotation` for the next statement
> 
>     * `@!annotation` for the parent statement

Yes quite reasonable imo!

PR #4121, opened 2024-01-22
feat: parse annotations for all kinds of statements

aljazerzen:


max-sixty (approved):
Very nice! Much better this way

PR #4122, opened 2024-01-22
feat: generic type arguments

aljazerzen:


prqlc/prqlc/tests/integration/resolving.rs
```diff
@@ -95,3 +95,24 @@ fn resolve_types_04() {
...
+    "#,
+    )
+    .unwrap(), @r###"
+    let add_one = func <A: int | float> a <A> -> <A> (
+      std.add a 1
+    )
+
+    let my_float <float> = `(std.add ...)`
+
+    let my_int <int> = `(std.add ...)`
```
aljazerzen, line 116: This is the new behavior: we can specify type parameters of a function and then each of function calls gets it's types resolved separately, within the bounds specified on the parent.

prqlc/prqlc/src/semantic/resolver/functions.rs
```diff
@@ -131,19 +138,50 @@ impl Resolver<'_> {
...
-                named_params: vec![],
                 body: Box::new(Expr::new(ExprKind::Func(inner_closure))),
-                return_ty: None,
-                env: HashMap::new(),
+
+                // these don't matter
+                named_params: Default::default(),
+                return_ty: Default::default(),
+                env: Default::default(),
+                generic_type_params: Default::default(),
```
max-sixty, line 147:
```suggestion
                ..Default::default()
```

?

prqlc/prqlc/tests/integration/resolving.rs
```diff
@@ -95,3 +95,24 @@ fn resolve_types_04() {
     )
     "###);
 }
+
+#[test]
+fn resolve_generics_01() {
+    assert_snapshot!(resolve(
+        r#"
+    let add_one = func <A: int | float> a <A> -> <A> a + 1
```
max-sixty, line 103: What do the three `<A>`s mean here?
- First is defining `A` as int or float
- Second is typing `a` as `A`
- Is the third the return type of the function?

prqlc/prqlc/src/semantic/resolver/functions.rs
aljazerzen: `Func` does not implement `Default`. It would be nice if Rust knew not all fields need the Default trait for this syntax to work, but it does not.

prqlc/prqlc/tests/integration/resolving.rs
aljazerzen: Correct.

prqlc/prqlc/src/semantic/resolver/functions.rs
max-sixty: Ah yes sorry...

PR #4126, opened 2024-01-24
feat: table declarations

### LINKED ISSUE #2646: Table types
### What's up?

In https://github.com/PRQL/prql/issues/2622#issuecomment-1564616852, @aljazerzen makes an interesting suggestion re declaring tables. I'm moving here so we can keep that focused on the nulls question.

> Well, my current idea on how to declare tables looks like this:
> 
> ```
> # table is just a declaration without a value
> let foo <{
>    id = int,
>    name = str || null,
> }>
> ```

I quite like this!

Only one question — should it be an array of these? e.g.

> ```
> # table is just a declaration without a value
> let foo <[{
>    id = int,
>    name = str || null,
> }]>
> ```

aljazerzen:
Closes #2646 with adding support for:

```elm
    module my_schema {
      let my_table <[{ id = int, a = text }]>
    }

    let another_table <[{ id = int, b = text }]>

    my_schema.my_table
    join another_table (==id)
    take 10
```

The rule is that any variable declaration with relational type and without an assigned value will be treated as a table in the database.

Fully qualified name is used to refer to that table. This will be changed soon.

### DIFF
--- a/prqlc/prqlc-ast/src/error.rs
+++ b/prqlc/prqlc-ast/src/error.rs
@@ -112,29 +112,15 @@ impl std::fmt::Display for Errors {
 }
 
 pub trait WithErrorInfo: Sized {
-    fn span(&self) -> Option<&Span>;
-
     fn push_hint<S: Into<String>>(self, hint: S) -> Self;
 
     fn with_hints<S: Into<String>, I: IntoIterator<Item = S>>(self, hints: I) -> Self;
 
     fn with_span(self, span: Option<Span>) -> Self;
     fn with_code(self, code: &'static str) -> Self;
-
-    fn with_span_if_not_exists(self, span: Option<Span>) -> Self {
-        if self.span().is_none() {
-            self.with_span(span)
-        } else {
-            self
-        }
-    }
 }
 
 impl WithErrorInfo for Error {
-    fn span(&self) -> Option<&Span> {
-        self.span.as_ref()
-    }
-
     fn with_hints<S: Into<String>, I: IntoIterator<Item = S>>(mut self, hints: I) -> Self {
         self.hints = hints.into_iter().map(|x| x.into()).collect();
         self
@@ -158,10 +144,6 @@ impl WithErrorInfo for Error {
 
 #[cfg(feature = "anyhow")]
 impl WithErrorInfo for anyhow::Error {
-    fn span(&self) -> Option<&Span> {
-        self.downcast_ref::<Error>().and_then(|e| e.span.as_ref())
-    }
-
     fn push_hint<S: Into<String>>(self, hint: S) -> Self {
         self.downcast_ref::<Error>()
             .map(|e| e.clone().push_hint(hint).into())
@@ -191,10 +173,6 @@ impl WithErrorInfo for anyhow::Error {
 }
 
 impl<T, E: WithErrorInfo> WithErrorInfo for Result<T, E> {
-    fn span(&self) -> Option<&Span> {
-        self.as_ref().err().and_then(|x| x.span())
-    }
-
     fn with_hints<S: Into<String>, I: IntoIterator<Item = S>>(self, hints: I) -> Self {
         self.map_err(|e| e.with_hints(hints))
     }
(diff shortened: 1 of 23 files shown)


prqlc/prqlc-ast/src/expr/ident.rs
```diff
@@ -67,10 +67,18 @@ impl Ident {
             .all(|(prefix_component, self_component)| prefix_component == self_component)
     }
 
+    pub fn starts_with_path<S: AsRef<str>>(&self, prefix: &[S]) -> bool {
+        if prefix.len() > self.path.len() + 1 {
```
max-sixty: I'm probably being slow, but why the `+ 1`? If the prefix is longer than the path, isn't that the end of it?

prqlc/prqlc/src/codegen/ast.rs
```diff
@@ -433,14 +444,13 @@ impl WriteSource for SwitchCase {
 #[cfg(test)]
 mod test {
     use insta::assert_snapshot;
-    use similar_asserts::assert_eq;
 
     use super::*;
 
+    #[track_caller]
```
max-sixty, line 451: Interesting, TIL

prqlc/prqlc/src/semantic/lowering.rs
```diff
@@ -19,15 +19,22 @@ use crate::COMPILER_VERSION;
...
 /// - transforms are not nested,
 /// - transforms have correct partition, window and sort set,
 /// - make sure there are no unresolved expressions.
+///
+/// All table references must reside within module at `database_module_path`.
+/// They are compiled to table identifiers, using their path relative to the database module.
+/// For example, with `database_module_path=my_database`:
+/// - `my_database.my_table` will compile to `"my_table"`,
+/// - `my_database.my_schema.my_table` will compile to `"my_schema.my_table"`,
+/// - `my_table` will error out saying that this table does not reside in current database.
```
max-sixty, line 33: What's the utility of this? Relative to leaving them as they're specified in the query? (not claiming confidently that there isn't an advantage!)

prqlc/prqlc-ast/src/error.rs
```diff
@@ -112,15 +112,29 @@ impl std::fmt::Display for Errors {
...
+    fn span(&self) -> Option<&Span>;
+
     fn push_hint<S: Into<String>>(self, hint: S) -> Self;
 
     fn with_hints<S: Into<String>, I: IntoIterator<Item = S>>(self, hints: I) -> Self;
 
     fn with_span(self, span: Option<Span>) -> Self;
     fn with_code(self, code: &'static str) -> Self;
+
+    fn with_span_if_not_exists(self, span: Option<Span>) -> Self {
```
max-sixty: No great stress, but wonder if this is something that should be done at the call site unless it's extremely prevalent (I only count one instance atm!), to avoid having a proliferation of methods here

prqlc/prqlc/src/semantic/lowering.rs
max-sixty: or this will obviate the need for this?

> Fully qualified name is used to refer to that table. This will be changed soon.

prqlc/prqlc-ast/src/expr/ident.rs
aljazerzen: Prefix is a slice, self is an Ident. Ident has lentgh of its path, plus one for the name field.

Looks like a bug, but I don't think it is.

prqlc/prqlc-ast/src/error.rs
aljazerzen: I had it at the call site and it was a mess of low level error handling in a high level function. But you're right, it can go into a new function there.

prqlc/prqlc/src/semantic/lowering.rs
aljazerzen: This adds an ability to specify a module as a "database module":

```elm
module my_database {
    module my_schema {
        let my_table <[{}]>   # this will be `my_schema.my_table`
    }
}

module another_database {
    let another_table <[{}]>  # this will be `another_database`
}
```
The problem is that these two tables are defined in two different databases, so they cannot be present in the same query. That's why "current database" is a compiler parameter, so it knows for which database it is compiling for.

Idea being that in the future, we could support querying multiple databases, so PRQL source would compile into multiple queries that would be sent to each of the databases.

PR #4127, opened 2024-01-24
feat: Add Sublime Text grammar to README

vanillajonathan:
I have created a syntax highlight file for the macOS text editor Sublime Text.

The grammar is also supported by the Rust library [syntect](https://github.com/trishume/syntect) which is used by software such as [bat](https://github.com/sharkdp/bat) (`bat` is like `cat` but with syntax highlighting) and some other software such as documentation generators, static site generators, markdown software, etc (see [Projects using Syntect](https://github.com/trishume/syntect#projects-using-syntect)).

To be able to have it on [packagecontrol.io](https://packagecontrol.io/) we need to put it in an own repository, for example `sublime-prql`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@
 
 - Add syntax highlight file for GtkSourceView. (@vanillajonathan, #4062)
 - Add syntax highlight file for CotEditor. (@vanillajonathan)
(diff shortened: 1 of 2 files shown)


max-sixty:
Great — I created the repo and added you @vanillajonathan !

max-sixty:
I looked at the repo — looks great!

Issue #3825, opened 2023-11-19
PRQL query engine


aljazerzen:
I've proposed we build [Lutra](https://hackmd.io/vgAGlDSmR5KPSVhCvekxHg) a while back, with the intention of constructing a CLI for running PRQL.
The motivation was to make "something that is an execution platform for PRQL".
CLI seemed most useful way of providing that to a user like me.
But the CLI ended up being the major part of the proposal, leaving the core idea a bit obfuscated.

So now, I'll revive the core idea and propose an execution engine, that would:
- take PRQL an in input and
- return the value of the input expression as the output.

This is different from the compiler in that the compiler returns an SQL query *that can be executed* to obtain the value of the input expression.

So make this happen, it would:
- compile PRQL to appropriate SQL dialect,
- use an appropriate adapter to make a database query,
- convert the result to a common format (probably arrow),
- print/write the output.

Why would this be useful?
- the engine can do stuff that should not be done by the compiler (mainly network calls),
- we already have something similar, scatterted between the playground, integration tests and prql-query,
- current implementations are not fully correct, as they convert many types to strings,
- a standard way to connect to a database would be useful for a Language Server Protocol, for retrieving database schema, which would be used for generating table type definitions,
- this would be how you can "try out PRQL" the easy way.

I've arrived to the idea that we need something similar a few times now.
This is probably because I see that we are constantly working on "PRQL -> SQL" instead of "PRQL -> relations", as I tried to put into [words here](https://prql-lang.org/book/project/contributing/language-design.html).

Future expansions:
- support different backends, including local computation engine,
- pull data from multiple sources and apply additional computation on the joined datasets,
- download external PRQL sources (packages)

---

My question now is: **do you think this is worth developing?**

max-sixty:
I thought the idea had lots of promised when we discussed it before! IIRC it would also manage "a PRQL project" — which to some extent has morphed into our nascent module system.

I think my main concern is how PRQL-specific some of it is. There are lots of attempts to build query engines / database-agnostic middleware / etc — https://github.com/GlareDB/glaredb is another recent one, as an example.

The parts that are PRQL-specific would be:
> - a standard way to connect to a database would be useful for a Language Server Protocol, for retrieving database schema, which would be used for generating table type definitions
> - this would be how you can "try out PRQL" the easy way.

[...]

PR #4134, opened 2024-01-25
feat: lutra query runner

aljazerzen:
Initial take on #3825. Inspired by prql-query.

Adds a query runner named Lutra that is basically a wrapper for `prqlc compile`, `sqlite3 run` and `print(pl.DataFrame)`.

Connection parameters are defined within PRQL source with `@lutra.sqlite` annotation.

Uses connectorx for executing the query and converting to Apache Arrow. Uses polars to print the dataframe.

Next step is to create Python bindings so the resulting dataframe can be used with other existing tooling.


hulxv:
Before a few months, I built a tool called [prqlite](https://github.com/hulxv/prqlite) (I think you took a look at it). Is there any difference between it and Lutra?

eitsupi:
Oops, this seems to depend on a very old version of polars.......

aljazerzen:
> Before a few months, I built a tool called [prqlite](https://github.com/hulxv/prqlite) (I think you took a look at it). Is there any difference between it and Lutra?

There are. prqlite is primarily a TUI and it does that well. My intention with Lutra is to be able to call it from Python (and other languages) while providing it with only PRQL source.

aljazerzen:
> Oops, this seems to depend on a very old version of polars

Yeah. It also won't compile for wasm32-unknown-unknown without a serious effort, as we'd have to compile polars and rusqlite too. That does not seem like a battle worth taking right now.

hulxv:
> My intention with Lutra is to be able to call it from Python (and other languages) while providing it with only PRQL source.

Then, Lutra is something like a library or do I misunderstand it?

aljazerzen:
> Then, Lutra is something like a library or do I misunderstand it?

Yes, but also with an option to invoke it from the command line.

max-sixty:
> Before a few months, I built a tool called [prqlite](https://github.com/hulxv/prqlite) (I think you took a look at it). Is there any difference between it and Lutra?

@Hulxv this is cool! Nice work. I don't think we knew about this...

(the goals seem to be slightly different to Lutra fwiw)

lutra/Taskfile.yml
```diff
@@ -0,0 +1,57 @@
...
+    cmds:
+      - cmd: cargo run -- print example-project
+
+  fmt:
+    desc: Format prqlc source files
+    cmds:
+      - cmd: |
+          # remove trailing whitespace
+          rg '\s+$' --files-with-matches --glob '!*.{rs,snap}' . \
+          | xargs -I _ sh -c "echo Removing trailing whitespace from _ && sd '[\t ]+$' '' _"
```
max-sixty, line 17: (FYI if you're in the mood for new tools, I really like `rargs` rather than `xargs` for this sort of thing. I almost never use `xargs` now)

max-sixty: ...even if I still maintain that `pre-commit` is better for these sorts of lints! 😄

lutra/Taskfile.yml
```diff
@@ -0,0 +1,57 @@
...
+    desc: The full test. Generates coverage report.
+    env:
+      # Use a different target dir so we don't poison the cache
+      CARGO_LLVM_COV_TARGET_DIR: ../target-cov
+    cmds:
+      - cmd: |
+          cargo \
+            llvm-cov --lcov --output-path lcov.info \
+            nextest \
+            {{.packages}}
```
max-sixty, line 48: FYI if you want to put `nextest` & `{{.packages}}` into the [base task](https://github.com/PRQL/prql/blob/d52312db53c6f8426594b09a1c90bc8c1b683a83/Taskfile.yml#L339) totally fine w me. Then we have a single task and can pass `packages`.

(but also zero stress if you prefer to have them separate)

lutra/example-project/Project.prql
```diff
@@ -0,0 +1,19 @@
+
+## This module is configured to pull data from an SQLite chinook.db database
+@(lutra.sqlite {file="chinook.db"})
```
max-sixty, line 3: How do we want to think about coupling the target and the query?

I would think we want to use some form of composition, where we say "associate this `chinook.db` file with this module `chinook`" somewhere, but we don't couple their definitions, so it's possible to change.

At a more conceptual level, it's could be similar to applying a functor in OCaml (or somewhat implementing a trait in Rust) — specifying a mapping of conceptual queries/functions to concrete queries/functions. Another reference is dbt's approach for templating the source of a query, so it's possible to run the same queries on a production vs dev vs personal database / schema.

I don't have a really good idea for how to implement this, and obv fine to experiment with things. But I do think that a really good design would be to make decouple the query and the target.

max-sixty (approved):
Exciting!

Maybe we put a readme of a couple of lines saying what it is and that it's still experimental?

lutra/example-project/Project.prql
aljazerzen: Yeah, I was thinking about not having actual connection params within the PRQL source, similar to how dbt does it.

I've opted for this approach because it is way more convenient and it does not actually rule out composition - the queries are not defined within the database module. This means that one could swap out the database module and run the queries on a different database.

This could be done by defining the database module in a separate file, which is not committed to VCS or is overridden in production.

lutra/Taskfile.yml
aljazerzen: I like to have a separate task that builds and tests only the thing that I'm currently developing.

What do you mean "pass packages"? To have the base task take packages as argument?

Cargo.toml
```diff
@@ -11,6 +11,7 @@ members = [
   "prqlc/prqlc-parser",
   "prqlc/prqlc",
   "prqlc/prqlc/examples/compile-files", # An example
+  "lutra",
```
aljazerzen: @max-sixty lutra currnently does not compile on `wasm32-unknown-unknown`. That is partially because:
- there are (transitive) dependencies of connectorx that try to work with a terminal,
- rusqlite does not support `wasm32-unknown-unknown`, https://github.com/rusqlite/rusqlite/pull/1010, although it looks like it will soon

Is it feasible to disable compilation and testing of lutra on `wasm32-unknown-unknown`?
It is a crate that currently does not support WASM, so hiding these features behind conditionals feels dirty, as the resulting pruned crate would not have any functionality left.

aljazerzen: I'm talking about this: https://github.com/PRQL/prql/actions/runs/7654921007/job/20859910479?pr=4134

max-sixty: > It is a crate that currently does not support WASM, so hiding these features behind conditionals feels dirty, as the resulting pruned crate would not have any functionality left.

I think that's fine; similar to how the `prqlc` CLI didn't compile on wasm — it instead had a main that immediately panics: https://github.com/PRQL/prql/blob/7c510444ce8fdbe4a53b6102095276f9ecfb3c1d/prqlc/prqlc/src/main.rs#L8-L9 (it's now part of the compiler crate, but the comment still applies to the binary from that crate)

aljazerzen: Remind me, why do we want to compile all binaries and libraries for WASM? Is it not ok to fail compilation, if what you are compiling cannot be used on WASM?

max-sixty: We want to be able to run `cargo build` on WASM, without some modules failing — we'd prefer to specify what should be compiled in the rust code than in CLI args...

lutra/example-project/Project.prql
max-sixty: Yes, we can think about this.

I think there's some nice design that's possible, where we map the relations to concrete tables. dbt do a nice job at the moment even if it needs a decent amount of jinja.

I'll meditate on it...

Cargo.toml
aljazerzen: But then we are building things that don't make sense? Compiling and linking binaries that are useless?

I don't think this is a good idea, so I won't do this myself. If you feel strongly about it and think that this repo is the best place for lutra to reside in, please do it yourself. Otherwise, this will go into a separate repo.

max-sixty: But what's the alternative? That `cargo build` fails under some targets? That's going to need a lot of special casing in tests.

IIRC there's lots of discussion in rust issues on making crates target-dependent — but it's not implemented yet, so people do this.

max-sixty: @aljazerzen & I chatted briefly — I'm going to add onto this PR the "standard thing" of ignoring code under targets it can't compile to, to move this along.

If we have other approaches that work better, happy to change to those later! But let's not let devops get in the way of important work like this...

lutra/Taskfile.yml
max-sixty: > What do you mean "pass packages"? To have the base task take packages as argument?

Currently this task has a `{{.packages}}` arg, so it's possible to pass a subset of packages

> I like to have a separate task that builds and tests only the thing that I'm currently developing.

Very much agree! TBC, this task is almost identical to the task in the parent task, but passes `-p lutra` to `.packages`. So they could be combined.

(I can make a PR later, maybe that's an easier way of explaining)

max-sixty (approved):

Cargo.toml
max-sixty: Done

max-sixty:
FYI can merge despite the failures — one is a `cargo-audit` which is fine and we can ignore; the other is the drop in test coverage

aljazerzen:
I'd like to replace connector-x with connector-arrow before merging. ConnectorX has stale dependencies and does a lot of things that are not needed in for lutra.

PR #4135, opened 2024-01-25
feat(playground): add new query button

### LINKED ISSUE #3631: playground could have a "blank" template as well
### What's up?

Playground is filled with nice examples, but if user just wants to paste a bug report to try out, then it is not that obvious how to use playground, especially from mobile where CTRL+A / select all doesn't seem to be easily available.
Therefore make sense to give a "blank" template as well, just ready for paste and run.

Also, that should probably go into new issue, but all tables used in examples used in the book could be available in playground. That would be make it easier to run what is in the book, and see the results. Example of missing one `orders`.

hulxv:
## Preview

InShot_20240125_180706597

Close #3631

### DIFF
--- a/web/playground/src/workbench/Workbench.jsx
+++ b/web/playground/src/workbench/Workbench.jsx
@@ -95,7 +95,7 @@ class Workbench extends React.Component {
       this.monaco.editor.setModelMarkers(
         this.editor.getModel(),
         "prql",
-        monacoErrors,
+        monacoErrors
       );
       return;
     }
@@ -149,6 +149,27 @@ class Workbench extends React.Component {
     ]);
   }
 
+  new_query() {
+    let filename = prompt(`Enter a name for your new query`);
+    if (filename) {
+      if (!filename.endsWith(".prql")) {
+        filename += ".prql";
+      }
+      
+      this.setState({
+        filename,
+        prql: "",
+        output: null,
+        outputTab: "arrow",
+        prqlError: null,
+        duckdbError: null,
+      });
+      if (this.editor) {
+        this.editor.setValue("");
+      }
+    }
+  }
+
   rename() {
     let filename = prompt(`New name for ${this.state.filename}`);
     if (filename) {
@@ -174,6 +195,9 @@ class Workbench extends React.Component {
               </button>
               <button className="action" onClick={() => this.save()}>
                 Save
+              </button>
+              <button className="action" onClick={() => this.new_query()}>
+                New Query
               </button>
             </div>
             <Editor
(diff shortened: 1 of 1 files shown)


max-sixty:
Thanks — I think this sort of thing is good.

Rather than interrupting the flow with a popup asking for a query name, how about a button for "Clear query" which clears the query and sets the name to `new-N.prql`, where `N` is a number that works.

"Clear query" (or something similar) also is clear that it's a destructive event — I can imagine clicking "New query" and not realizing my existing query is about to be deleted if I hadn't saved it...

Hulxv:
> "Clear query" (or something similar) also is clear that it's a destructive event — I can imagine clicking "New query" and not realizing my existing query is about to be deleted if I hadn't saved it...

Exactly! The "Clear query" button that removes your query directly is a bad experience for the user.

hulxv:
> Rather than interrupting the flow with a popup asking for a query name, how about a button for "Clear query" which clears the query and sets the name to `new-N.prql`, where `N` is a number that works.
> 
> "Clear query" (or something similar) also is clear that it's a destructive event — I can imagine clicking "New query" and not realizing my existing query is about to be deleted if I hadn't saved it...

What about saving the query before clearing the query and notifying the user before that? I think it's a better experience for the user

max-sixty:
> What about saving the query before clearing the query and notifying the user before that? I think it's a better experience for the user

I don't think we want to get in the way of the user — so we don't want a box popping up asking "Do you want to save??" — instead if the button's actions are obvious, the user can decide whether to save before clicking the button.

So I do think that "Clear query" is better than "New query" in that regard...

hulxv:
I'm sorry for the late response. I'm away from my machine these days. So I hope you'll excuse me.

Anyway. Is "Clear query" really useful? I don't know why we would just put a button to remove queries only. The user can use `ctrl` + `a` then click `backspace` and all is done. But, The "New Query" button doesn't remove queries only. It's something like creating a new file. In future, We can add something like taps to switch between files to improve the user experience.

max-sixty:
> Is "Clear query" really useful?

I think it's somewhat useful... :)

Isn't "Clear query" + "Rename" the same as "New Query"?

Minimum surprise is important — and the current "New Query" clears the existing query, which might be surprising...

hulxv:
> Isn't \"Clear query\" + \"Rename\" the same as \"New Query\"?

Technically, yes. but the difference is that `New Query` will be useful if we add `taps` so that the user can switch between them. 

> Minimum surprise is important — and the current "New Query" clears the existing query, which might be surprising...

To avoid that, I think we should work on `taps`

max-sixty:
Yes, tabs could be good! 

(It does seem like a much bigger project, though...)

hulxv:
> (It does seem like a much bigger project, though...)

Yeah sure. Next week, I will try to work on it and maybe cancel this PR

PR #4152, opened 2024-01-30
feat: Add documentation generator

vanillajonathan:
This PR adds a HTML documentation generator as a standalone binary, but I would like to see it part of the `prqlc` CLI through a `doc` option, like `cargo doc`.

It doesn't yet include any docblock comments since they're not yet exposed in the PL/AST.

It is a start but there is plenty of room for improvement. It doe not include the content of modules.


max-sixty:
I think this could be really cool, great idea!

Is it possible to share what the output looks like? Maybe as an insta snapshot test?

I guess this could go in a couple of directions — it could be something like "HTML compiler output" or it could be more like "documentation". "Documentation" would be more like https://www.getdbt.com/blog/using-dbt-docs; i.e. for the users of the PRQL project that someone built; it would need more focus on doc-comments and the content of queries.

Producing HTML by hand is totally fine at the moment. If we decide grow this — and I think there's lots of room to grow — we might want to use some templating. I don't know the best frameworks for that; we can push off making a decision for a bit.

---

To integrate into `prqlc`, we'll want to add a member to the `DebugCommand` enum and add an entry [here](https://github.com/PRQL/prql/blob/461741a2562b2fb01868b136818fe4847224d7fc/prqlc/prqlc/src/cli/mod.rs#L300). I think then just returning the HTML as bytes will make it all flow through. (does that make sense? possibly I'm not being explicit enough...)

---

In terms of space to grow, I would _love_ to have something which helped understand how queries are flowing through the compilation stages — I could imagine something which used the spans to link the PRQL->PL->RQ; possibly even gave information when hovering over items. That would be more of "HTML compiler output". Currently our compiler output is quite verbose and not that easy to visualize across stages.

vanillajonathan:
> Is it possible to share what the output looks like? Maybe as an insta snapshot test?

Yes, it would be possible, but I don't know if it provides much of a value.
 
> I guess this could go in a couple of directions — it could be something like "HTML compiler output" or it could be more like "documentation". "Documentation" would be more like https://www.getdbt.com/blog/using-dbt-docs; i.e. for the users of the PRQL project that someone built; it would need more focus on doc-comments and the content of queries.

My intention was to make a documentation generator used by data analysts to generate documentation for their modules and for that documentation to be read by data analysts who use those modules. Something similar to ESDoc, TSDoc, phpDocumentor, Sphinx, and `cargo doc`.

> To integrate into `prqlc`, we'll want to add a member to the `DebugCommand` enum and add an entry [here](https://github.com/PRQL/prql/blob/461741a2562b2fb01868b136818fe4847224d7fc/prqlc/prqlc/src/cli/mod.rs#L300). I think then just returning the HTML as bytes will make it all flow through. (does that make sense? possibly I'm not being explicit enough...)

I don't envision it as something under the debug command, I want it as its own command to be ran as `prqlc doc` (like `cargo doc`).

> In terms of space to grow, I would _love_ to have something which helped understand how queries are flowing through the compilation stages — I could imagine something which used the spans to link the PRQL->PL->RQ; possibly even gave information when hovering over items. That would be more of "HTML compiler output". Currently our compiler output is quite verbose and not that easy to visualize across stages.

That is an interesting idea but not what I had envisioned.

max-sixty:
> That is an interesting idea but not what I had envisioned.

Yes totally fine, the `prql doc` approach is good too.

In that case, we should definitely include doc comments, that's probably the most important thing.

> > Is it possible to share what the output looks like? Maybe as an insta snapshot test?
> 
> Yes, it would be possible, but I don't know if it provides much of a value.

I definitely think we need some tests! It doesn't have to be that thorough at this stage (one benefit of insta tests is that they are immediately helpful and easy to write...)

eitsupi:
This sounds great, but I suspect that if possible, it would be better to generate Markdown rather than HTML here and use a different site generator for HTML generation.

max-sixty:
> This sounds great, but I suspect that if possible, it would be better to generate Markdown rather than HTML here and use a different site generator for HTML generation.

Yes, good point — if we're not using HTML features — e.g. colors, hovers, etc — then markdown is likely better...

vanillajonathan:
Markdown is also an interesting approach for documentation but maybe it would be nice to get straight HTML too like `cargo doc`. Maybe there could one function  for HTML format and one function for Markdown?

max-sixty:
> Markdown is also an interesting approach for documentation but maybe it would be nice to get straight HTML too like `cargo doc`. Maybe there could one function for HTML format and one function for Markdown?

As long as we're only using simple text features, markdown does seem better — it's easy to convert markdown into HTML later, but HTML is difficult to convert into markdown. And markdown is more flexible for display, much simpler...

vanillajonathan:
> > Markdown is also an interesting approach for documentation but maybe it would be nice to get straight HTML too like `cargo doc`. Maybe there could one function for HTML format and one function for Markdown?
> 
> As long as we're only using simple text features, markdown does seem better — it's easy to convert markdown into HTML later, but HTML is difficult to convert into markdown. And markdown is more flexible for display, much simpler...

Many documentation generation tools such as Rust's `cargo doc`, [phpDocumentor](https://www.phpdoc.org/), [Sphinx](https://www.sphinx-doc.org/), [JSDoc](https://jsdoc.app/), [ESDoc](https://esdoc.org/), [docfx](https://dotnet.github.io/docfx/), etc. generates HTML. I think HTML is good because you don't have to use some Markdown converter or static site generator.

We could also have a `doc-markdown` option so we can generate both HTML and Markdown, this would likely involve duplicating the function and changing the HTML to Markdown.

I don't know how to make a snapshot test. By the way, currently the output includes the PRQL compiler version and I've been wanting to add the date of generation too, so I don't know if it possible to use a snapshot the output is dynamic.

max-sixty:
> I don't know how to make a snapshot test.

Check out some other examples in the repo — `assert_snapshot` or `assert_display_snapshot`. It's a really nice way of testing...

TBC, tests are extremely important, we have 90% code coverage and we'd like to increase that to 100% — it's not possible to merge code without tests.

> By the way, currently the output includes the PRQL compiler version and I've been wanting to add the date of generation too

Yes, this is a downside. We have some tests that need to re-run when the version changes. We would need an option to exclude the date in order to add that.

---

> Many documentation generation tools such as Rust's `cargo doc`, [phpDocumentor](https://www.phpdoc.org/), [Sphinx](https://www.sphinx-doc.org/), [JSDoc](https://jsdoc.app/), [ESDoc](https://esdoc.org/), [docfx](https://dotnet.github.io/docfx/), etc. generates HTML. I think HTML is good because you don't have to use some Markdown converter or static site generator.

I agree, but I'm not sure it's the best place to _start_, because we're not using any HTML features at the moment. It would be easy to grow this into a full HTML output if we grow it, but it would be much harder to reduce this to a markdown output if we decide not to add lots more features.

That said, I'd be OK to put this in a `prqlc experimental doc` command if you want to merge this in (with tests) — that way we can start merging things in, and evolve. With a `prqlc doc` command, the bar is much higher for forward compatibility.

max-sixty:
@vanillajonathan looking good!

Can we add a test like [this one](https://github.com/PRQL/prql/blob/2d9e8fccdcf58a562b37f025c7ac28a47dfc8c8b/prqlc/prqlc/tests/integration/cli.rs#L383-L426)? If you copy & paste and then change what's passed in, then insta will fill the result in. Does that make sense?

vanillajonathan:
I added a test.

max-sixty:
Thanks for adding the test!

I think this will be useful and we can build on it. I don't think it's that useful *yet* (which is often the case for initial features). Having docstrings there will be a big change, though I need to work on the AST for that to work.

So for the moment — can we put behind a `prqlc beta doc` or `prqlc experimental doc`? I'd like us to be able to ship cool ideas like this while also setting expectations for users on what they should start using immediately...

(If you need a hand on how to add a subcommand with clap, let me know, happy to help)

vanillajonathan:
Okay, I moved the `doc` command to under the "experimental" subcommand.

max-sixty:
Great — looking good! Thanks @vanillajonathan 

Could we try and get some more test coverage? Looks like this is at 70%. If it's as easy as extending the example in the test then it's worthwhile, both for the test coverage and so we have examples that show what the command can do. 

(If it would require changing the code itself, then it's fine to leave it for the moment)

max-sixty (approved):
Looking good @vanillajonathan ! 

Feel free to merge!

prqlc/prqlc/src/cli/docs_generator.rs
```diff
@@ -0,0 +1,430 @@
...
+//                 }
+//                 _ => (),
+//             }
+//         }
+
+//         docs.push_str("</section>\n");
+//     }
+
+//     html.replacen("{{ content }}", &docs, 1)
+// }
```
max-sixty, line 173: @vanillajonathan we should probably remove this big comment block!

vanillajonathan: Or maybe we should add a `doc-html` command?

max-sixty: We could add `--format=html`, but:
- We would want some reason to use HTML rather than Markdown — something that an html markdown generator couldn't give us.
- If there are ways of improving the current one, I think that's more productive than adding more formats at the moment. It's a cool idea at the moment, but not _that_ helpful. For example, how does it handle multi-file projects?

vanillajonathan: HTML is directly usable and easy to use and publish without having any Markdown viewer installed or use any Markdown-to-HTML processor or static site generator.

HTML allows us rich semantics through the use the `<var>` element to define variables, and the `<time>` element for when the documentation was generated.

It does not yet handle multi-file projects. I created a simple implementation which can hopefully be extended in the future if the need should arise.

max-sixty: OK. I really would encourage trying to make something that will be useful for users — that's the best way to get feedback on what's working, what else would be helpful, etc.

Adding multi-file projects could be great. There is already a good implementation of something that collects together multiple files in `prqlc collect`; it shouldn't need to collect the files itself. I could see that being the most useful part of something like `prqlc doc` — you have a large project and want to have a single page on what it contains.

> HTML is directly usable and easy to use and publish without having any Markdown viewer installed or use any Markdown-to-HTML processor or static site generator.

But we should try and have our tools be orthogonal — we don't need to make HTML directly unless there's a good reason to not just use a standard converter...

> HTML allows us rich semantics through the use the `<var>` element to define variables, and the `<time>` element for when the documentation was generated.

Yes, true.

But then when we need to make the tool better, we need to update the HTML generator too! Which then makes it harder to make the tool better...

---

If you're looking for other things to do — we very much appreciate your contributions — then improving code coverage and improving error messages would both be really helpful.

PR #4154, opened 2024-01-30
docs: Add link from `from` page to reading files

max-sixty:
I notice we don't have the `read_parquet` functions in our reference, this at least links to them


max-sixty:
@aljazerzen I notice the "How do I" section in the book is a bit sparse. I wonder whether we roll that into the reference, and I do some work to improve the tutorial. Or do you think it's worth investing in the "How do I" section?

aljazerzen:
"How do I" is generally a different type of documentation that we'd also need, but yea, the three current pages are few.

This is supposed to be the section we fill out when people ask "how do I" and might touch a few different topics from the reference.

max-sixty:
This was subsumed by #4293

Issue #725, opened 2022-06-28
Make prql a PostgreSQL stored Procedure Language?


michelp:
https://github.com/tcdi/pgx allows making PostgreSQL extensions with Rust, it seems like a logical connection would be to use pgx to make prql a stored procedure language in the database.

```
CREATE FUNCTION foo() RETURNS bar LANGUAGE plprql AS $$
... prql query here
$$;
```

Thoughts?

max-sixty:
That looks very exciting!

So IIUC this would intercept the query, written in PRQL, compile it to SQL, and then send than on to postgres?

IIUC, extensions can do much more than we'd need — e.g. be involved during the query — but even if we don't use that, this would let people send PRQL directly to postgres. Is that correct?

Thanks for raising this @michelp .

michelp:
> So IIUC this would intercept the query, written in PRQL, compile it to SQL, and then send than on to postgres?

> IIUC, extensions can do much more than we'd need — e.g. be involved during the query — but even if we don't use that, this would let people send PRQL directly to postgres. Is that correct?

Yes to both, and you can cache the execution plan itself and only pay for the compilation once.  In addition prql could call other routines, be called *by* other routines, accept arguments, return simple and composite types, set returning functions: it would make prql a first class stored procedure language just like sql, plpgsql, plv8, plpythonu, etc.

I'm not a Rust programmer so unfortunately that's as specific as I can get, an extension can register a new language callback and be instantiated into a database with `CREATE LANGUAGE`

https://www.postgresql.org/docs/current/sql-createlanguage.html

Here's plpython3's callback as a starting point:

https://doxygen.postgresql.org/plpy__main_8c.html#afe79221c59a91b3b176c75a28a28709b

> 
> Thanks for raising this @michelp .

michelp:
Looks like there's also a "sample" pl language for tests https://doxygen.postgresql.org/plsample_8c.html

max-sixty:
Great, thanks!

If anyone would like to explore this more, please do! I'm guessing it's a big project, but it could be very cool indeed.

snth:
This would be very cool indeed!

I just want to clarify one point as I think there might be a slight misunderstanding (possibly on my part). This only applies to **Stored Procedures** and **Functions**. This doesn't actually allow Postgres to understand general queries written in PRQL, unless those are wrapped in a Stored Procedure.

[...]

PR #4162, opened 2024-02-01
docs: Add example for running prql code directly in PostgreSQL without cursor

kaspermarstal:
This PR adds an example to the PostgreSQL extension docs for running PRQL code directly without the use of curser. The new function in the extension returns `setof record`. We previously discussed this issue in #725. The solution was to declare the record's type using e.g. `as (name text, age int)`.

### DIFF
--- a/web/book/src/project/integrations/postgresql.md
+++ b/web/book/src/project/integrations/postgresql.md
@@ -44,7 +44,7 @@ limit 2;
   3 |     1001 |     2 | Player1 |     1 |      7
 (2 rows)
 
--- Same as above, but returns cursor
+-- Same as above without the need for the static types, but returns cursor
 select prql('from matches | filter player == ''Player1''', 'player1_cursor');
 fetch 2 from player1_cursor;
 ```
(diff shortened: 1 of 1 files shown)


web/book/src/project/integrations/postgresql.md
```diff
@@ -31,18 +31,22 @@ select * from match_stats(1001)
...
+as (id int, match_id int, round int, player text, kills int, deaths int)
+limit 2;
 
  id | match_id | round | player  | kills | deaths
 ----+----------+-------+---------+-------+--------
   1 |     1001 |     1 | Player1 |     4 |      1
   3 |     1001 |     2 | Player1 |     1 |      7
 (2 rows)
+
+-- Same as above, but returns cursor
```
max-sixty:
```suggestion
-- Same as above without the need for the static types, but returns cursor
```

WDYT?

max-sixty (approved):

web/book/src/project/integrations/postgresql.md
kaspermarstal: Yes, good clarification

PR #4166, opened 2024-02-02
ci: Add macos-14 to CI

max-sixty:


max-sixty:
Relies on https://github.com/rhysd/actionlint/issues/393, or we add a config file in the meantime

eitsupi:
It has been announced that macos-latest will switch from 12 to 14 in a few months, so it may be better to specify the version without using latest.

max-sixty:
> It has been announced that macos-latest will switch from 12 to 14 in a few months, so it may be better to specify the version without using latest.

Should we keep 12 after that change, though?

eitsupi:
> Should we keep 12 after that change, though?

I suspect that even at this stage there is no need to keep 12, since macOS cross-compilation seems to work fine, wouldn't it be enough to just switch from latest to 14? Due to differences in machine performance, 14 tends to finish processing faster.
The smaller size of the matrix will also ease CI congestion.

max-sixty:
I left the old mac-os ones there for the moment; we can remove when this becomes `latest`, or I'm fine with someone doing it sooner if they think that's better

PR #4174, opened 2024-02-06
feat: lutra python bindings

aljazerzen:
Followup for #4134


lutra/bindings/python/src/lib.rs
```diff
@@ -0,0 +1,38 @@
...
+    // From https://github.com/PyO3/maturin/issues/100
+    m.add("__version__", env!("CARGO_PKG_VERSION"))?;
+    Ok(())
+}
+
+#[pyfunction]
+pub fn execute_one(
+    project_path: &str,
+    expression_path: &str,
+) -> PyResult<PyArrowType<RecordBatch>> {
```
eitsupi, line 21: As you may know, this package probably does not need to have `pyarrow` as a dependency for this function.
It should be sufficient to implement The Arrow PyCapsule Interface.
https://arrow.apache.org/docs/format/CDataInterface/PyCapsuleInterface.html

aljazerzen: I noticed errors about missing pyarrow, yes. I've added it now as a dependency of the bindings.

Do you mean to use PyCapsule instead of adding the dep? It seems to be experimental.

lutra/bindings/python/noxfile.py
```diff
@@ -0,0 +1,42 @@
+"""Nox session configuration."""
+import os
+from pathlib import Path
+from typing import List
```
aljazerzen: To be honest, I don't know what does this file do.

Installs different versions of Python and runs the tests? But implemented in Python?

precommit mypy seems to have a problem with there being two noxfile.py, but if my understanding is correct, it would make sense to have two, as these are two separate packages.

It would make more sense to run mypy separately on prqlc-python and lutra-python

lutra/bindings/python/src/lib.rs
eitsupi: This is a relatively recent addition, so it's marked as experimental, but I think it's slowly spreading.
arrow-rs is also using it.
https://github.com/apache/arrow-rs/blob/bd19a624469df652f24ee37b65056535ff665769/arrow/src/pyarrow.rs#L177-L183

aljazerzen: Oh, I see - it's a wrapper for a pointer to a PyArrow object, that is supposed to be arrow version independent.

And that makes it able to pass it to Python as a PyCapsule, so pyarrow dep is not needed.

aljazerzen: Wait, but arrow-rs does not support it yet for `ToPyArrow` trait?

https://github.com/apache/arrow-rs/blob/bd19a624469df652f24ee37b65056535ff665769/arrow/src/pyarrow.rs#L201-L210

aljazerzen: https://github.com/apache/arrow-rs/issues/5079

https://github.com/apache/arrow-rs/pull/5070

> It's not possible (I don't think) to implement exporting to the PyCapsule interface because no Python classes are exposed.

It looks like we cannot do this yet. When arrow-rs changes its implementation, I think we'll automatically start producing PyCapsules, so that's nice.

eitsupi: This interface must be implemented in a Python class and probably will not be provided by arrow-rs itself?

In any case, we can take a look at it later.

lutra/bindings/python/noxfile.py
max-sixty: Some people really like `nox`. I haven't found it particularly helpful in my work.

So unless someone remembers that we really needed it in `prql_python`, I would suggest deleting it in Lutra...

lutra/bindings/python/pyproject.toml
```diff
@@ -0,0 +1,50 @@
...
+
+# When https://github.com/PyO3/pyo3/pull/2135 merges, we can remove this config.
+
+features = ["pyo3/extension-module"]
+
+[project.optional-dependencies]
+test = [
+  "pytest >= 7",
+  "mypy == 1.8.0",
+  "pyarrow ~= 15.0.0",
```
max-sixty:
```suggestion
```

Already covered above?

PR #4187, opened 2024-02-08
build(lutra): Resolve naming issues

max-sixty:
This fixes the naming conflicts while letting us name the python module `lutra`.

It also renames the python package name for PyPI to `lutra`, from `lutra-python`


max-sixty:
We can also use this approach to rename `pyprql` to `prqlc`

aljazerzen (approved):
Nice

PR #4202, opened 2024-02-09
docs: Add resolution approach to Code of Conduct

max-sixty:
We need this for NumFocus. I added a really short one.

Does this seem reasonable to everyone?


aljazerzen (approved):
This looks reasonable.

But we do have COC already: https://github.com/PRQL/prql/blob/main/.github/CODE_OF_CONDUCT.md

Not sure why it is in .github though.

max-sixty:
Oh, I didn't even know we had one! I guess we've had such good conduct.

Shall we keep the old one? Not sure whether that was well thought through or just a copy-paste.

NumFocus wanted us to add a way of reporting violations, so either way we should add that at the bottom.

Any votes for the existing long one vs this proposed short one? I don't have a strong view; I think it's unlikely we use this that much...

eitsupi:
> Not sure whether that was well thought through or just a copy-paste.

Maybe this?
https://www.contributor-covenant.org/version/2/0/code_of_conduct/

aljazerzen:
I remember now: someone was asking on Discord if have code of conduct and recommended contributor covenant to be used. So that's what I did.

Reading both of them now I see no practical difference so I'm fine with using any of them.

max-sixty:
OK merging!

PR #4219, opened 2024-02-12
refactor: convert source tree to module right after parsing

aljazerzen:
- refactor: convert source tree to module def during parsing
- remove generics from SourceTree
- refactor: clean imports


prqlc/prqlc-parser/src/file_tree.rs
max-sixty, line 1: Plausibly remove?

prqlc/prqlc/src/ir/pl/fold.rs
```diff
@@ -1,9 +1,9 @@
+use crate::ast::{TupleField, Ty, TyFunc, TyKind};
```
max-sixty: Will this nullify the doc-comment?

prqlc/prqlc/src/parser.rs
```diff
@@ -1,34 +1,159 @@
...
     } else {
         Err(Errors(errors).into())
     }
 }
+
+struct SourceFile<'a> {
+    file_path: &'a PathBuf,
+    module_path: Vec<String>,
+    content: &'a String,
+}
```
max-sixty, line 47: Is it idiomatic to have refs to traditionally owned types like `Pathbuf` & `String`? I would have generally expected to see `&'a Path` xor `PathBuf`; same for string 

(but no _great_ stress obv...)

max-sixty (approved):
Makes a lot of sense re the `ModuleDef` vs `<SourceTree<Vec<Stmt>>>` change!

prqlc/prqlc/src/ir/pl/fold.rs
aljazerzen: No, because this doc comment is commenting the `use anyhow::Result`. Which is a typo in itself. :D

prqlc/prqlc/src/parser.rs
aljazerzen: This makes sense still, because these are references into the SourceTree. This object exists only during `linearize_tree` and `parse`.

max-sixty: https://github.com/PRQL/prql/pull/4222

PR #4221, opened 2024-02-13
feat: `select {alias = this}`

aljazerzen:


prqlc/prqlc/tests/integration/sql.rs
```diff
@@ -4977,3 +4977,18 @@ fn test_param_declarations() {
       b = $a
     "###);
 }
+
+#[test]
+fn test_relation_aliasing() {
+    assert_display_snapshot!(compile(
+        r###"
+    from x | select {y = this} | select {y.hello}
```
max-sixty, line 4985: Cool! So `y` is a tuple / relation

max-sixty (approved):

prqlc/prqlc/tests/integration/sql.rs
aljazerzen: Idea is that if `x` is of type `[{a = int, b = bool}]`, then
```
x | select {y = this}
```
... will have the type of `[{y = {a = int, b = bool}}]`.

... which then gets flattened when converting to an SQL relation.

max-sixty: Yes perfect.

Will be great to have an LSP one day, would be so cool to hover over `y` and get that type...

vanillajonathan:
Maybe you can add something to the changelog, or some release notes or to the documentation something about this.

aljazerzen:
Maybe, but it is a minor change, implemented with a hack. I plan to strip the "`Lineage`" part of our resolver out, and the new implementation will have this feature. So I've just added this hack to the lineage so the breakage between the new and the old implementations will be as small as possible.

So let's add this change to the changelog when the new type-based resolver is merged.

PR #4227, opened 2024-02-14
test: Pin mysql image

max-sixty:
Notes inline


eitsupi:
Sorry but it is not clear to me what you wanted to do here or what was going on, the title of the PR indicates MySQL but only ClickHouse was actually changed.

max-sixty:
Sorry, the PR changed a lot as I was trying to figure out what the problem was; it ended up being the ClickHouse DB, but in the MySQL driver. I know you did some work on the external DBs.

No action needed! (At some point we can figure out how to unpin but nothing needed for the moment)

eitsupi:
Ok, thanks for explaining.

PR #4238, opened 2024-02-16
refactor: phase out anyhow in prqlc core

### LINKED ISSUE #4220: Replace `anyhow::Result` with our own `prqlc::Result` alias
We use `anyhow::Result` for return type of most function calls within prqlc, even though it 99% of cases, it it can only return error of type `prqlc::Error`.

I suggest we do the following:

```
type prqlc::Result<T> = Result<T, prqlc::Error>
```
... and use that instead of `anyhow::Result`.

It is probably good to be more explicit about what the error is.

aljazerzen:
- remove anyhow from the prqlc
- clippy
- refactor: remove anyhow from prqlc

Closes #4220


max-sixty (approved):
I'm surprised everything is fine without it... No really awkward constructs etc...

aljazerzen:
Well yes - we were only producing `Error` or `String` as the error types in the core compiler.
There were a few other errors which I converted to strings, all coming from `from_text` for parsing CSV and JSON. I've converted those to strings and wrapped into `Error::new_simple()`.

But this makes sense: output of the compiler must be either successful compilation or an error message that points to a span of the source and maybe has hints.

PR #4241, opened 2024-02-16
fix: Fixup bindings from `db.` change

max-sixty:
Initial push will fail but will keep pushing while I work


aljazerzen:
Thank you for cleanup the mess after me!

Issue #4254, opened 2024-02-25
prqlc CLI not built with cli feature enabled (version 0.11.2 and 0.11.3)


tryan-dev:
### What happened?

This is my first time attempting to use prql.

Most of the downloads I have tried, for multiple platforms, do not work from the cli. Invoking the executable from a shell prompt results in the following message:

```
PS C:\Users\TimRyan\code\prql\prqlc-v0.11.3-x86_64-pc-windows-msvc> .\prqlc.exe --help
thread 'main' panicked at prqlc\prqlc\src\main.rs:11:5:
Crate is not built with the `cli` feature enabled, or was built for a wasm target.
```

I installed prql with scoop on Windows 11 (23H2), which I use as a package manager for my work Windows OS. The scoop-installed prql failed to work from PowerShell 7, as well as good old fashioned cmd.exe.

Next, I tried downloading the Windows executable directly from Github and got the same results (displayed above). I also tried the binaries for macOS arm64, and for Linux running under WSL with the same results.

The only binary that has worked for me so far is from Homebrew on macOS.

I'd really like to try prql for work, which will require having prql on Windows, unfortunately. Hoping I'm either missing something obvious, or that there's a simple issue with your CI pipeline.

### PRQL input

```elm
prqlc.exe --help
```

### SQL output

```SQL
thread 'main' panicked at prqlc\prqlc\src\main.rs:11:5:
Crate is not built with the `cli` feature enabled, or was built for a wasm target.
```

### MVCE confirmation

- [x] Minimal example
- [X] New issue

eitsupi:
I can reproduce this.
The last version that worked seems to be 0.11.1.
https://github.com/PRQL/prql/releases/tag/0.11.1

You can install with `winget install prqlc --version 0.11.1`

I'm surprised no one has noticed this in a month (i.e. no one is using prqlc!)

eitsupi:
I think this line causes this. `--no-default-features` should be removed.
https://github.com/PRQL/prql/blob/26940ddef107bcd81b89876b788b082fb8974ced/.github/actions/build-prqlc/action.yaml#L71

@max-sixty I think we should create a patch release as sooon as possible.

max-sixty:
Sorry, this was my fault. Thanks for spotting @tryan-dev

max-sixty:
Released 0.11.4

tryan-dev:
I logged in this morning and scoop had the patch release already. Thanks everyone!

[...]

PR #4255, opened 2024-02-26
ci(release): fix prqlc CLI build

eitsupi:
Adresses #4254

If this works well, I would like to backport the branch from the 0.11.3 tag.

### DIFF
--- a/.github/actions/build-prqlc/action.yaml
+++ b/.github/actions/build-prqlc/action.yaml
@@ -68,9 +68,8 @@ runs:
         # even at the cost of slighly less efficiency.)
         args:
           --profile=${{ inputs.profile }} --locked --target=${{ inputs.target }}
-          --no-default-features --features=${{ inputs.features }} ${{
-          contains(inputs.target, 'musl') && '--package=prqlc' ||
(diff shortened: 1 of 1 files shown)


eitsupi:
The artifact seems working.
I will wait for the review for the next half day or so, and if there is nothing, I would like to merge and backport.

max-sixty:
Thanks a lot @eitsupi !

On a flight but let's merge and I'll try and do a hotfix release from here

max-sixty:
Going to use the `web` branch as a "stable" branch that we can release from without the `db.` change, let me know if that's mistaken in some way...

eitsupi:
> Going to use the `web` branch as a "stable"

That's smart! Perhaps we need to be careful only about bumping the version?

max-sixty:
> That's smart!

Yeah generally I think we should be quite liberal about releasing from `main`, not maintaining lots of versions. But here the `db.` change is so big and this is a bad mistake (sorry!).



> Perhaps we need to be careful only about bumping the version?

Yes, good idea if you're referring to the process that bumps the changelog etc, let me do that now

PR #4260, opened 2024-02-26
ci: Align `prqlc` cli build

max-sixty:
This restores the `--no-default-features`, which we use in all builds, and adds the `cli` feature explicitly in the release workflow

### DIFF
--- a/.github/actions/build-prqlc/action.yaml
+++ b/.github/actions/build-prqlc/action.yaml
@@ -45,8 +45,7 @@ runs:
 
     - if: runner.os == 'Windows' && inputs.profile == 'release'
       shell: bash
-      run: |
-        echo 'RUSTFLAGS=-Ctarget-feature=+crt-static' >>"$GITHUB_ENV"
+      run: echo 'RUSTFLAGS=-Ctarget-feature=+crt-static' >>"$GITHUB_ENV"
 
     - if: inputs.target == 'aarch64-unknown-linux-musl'
       shell: bash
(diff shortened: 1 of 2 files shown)


.github/actions/build-prqlc/action.yaml
```diff
@@ -100,3 +100,7 @@ runs:
     - id: echo-artifact-name
       shell: bash
       run: echo "artifact-name=${{ env.ARTIFACT_NAME }}" >>"$GITHUB_OUTPUT"
+
+    - name: test-cli-works
+      shell: bash
+      run: ${{ env.ARTIFACT_NAME }} --help
```
eitsupi: Since they may be doing cross-compilation, I think it would be better to run the job separately in the workflow instead of doing it here.
Like this:
https://github.com/PRQL/prqlc-r/blob/aa80d5299a107f32d324bdd4646c1dfb22e8e5bd/.github/workflows/release-lib.yml

max-sixty: I agreed at first... But given the artifact is bundled, doesn't this require un-tarring / un-zipping? (not impossible but more than a line...)

max-sixty: OK I'm trying for one target, which seems like it'll cover most failures

eitsupi: Wouldn't it be better to test on all platforms to prevent something like #4268?
I was the one who added this in the first place, so I obviously should add tests for this......

eitsupi: (By the way, it seems that prqlc CLI archives for macOS were broken from 0.11.2, same as #4254.)

max-sixty: I added for mac and left a TODO for Windows!

eitsupi: Wouldn't it be better to have a separate job for build, test and release and only release if the test is successful?

max-sixty: I've not found "test first, then release if it passes" to be a good pattern — breaking a release isn't useful — the version is already tagged, we shouldn't be changing the code on a tag. Testing should be in previous runs, like we have here.

Open-minded on breaking up the job, I agree it's a bit of a bolt-on atm...

PR #4274, opened 2024-02-27
web: switch browser build to '--target web' for ESM module usage

### LINKED ISSUE #4273: make PRQL usable from observable/framework
### What's up?

The current setup of the JS bindings won't allow for using them from https://observablehq.com/framework/. I've tried many things (😅) and finally came across [this page](https://observablehq.com/@huw/prql), which explains the conundrum [here](https://observablehq.com/@huw/prql#cell-26):

> [...] PRQL publishes 3 entrypoints to npm—bundler, node, and web—but unfortunately the web entrypoint is actually compiled with the no-modules flag, which sets the wasm_bindgen variable globally rather than exporting anything from the package. None of these entrypoints work in Observable using any methods.

The page goes on explaining a workaround, which works, but it would be so nice if you could just say
```js
import init, { compile } from "npm:prql-js";
await init()
// ... etc ...
```
instead of manually building prql and putting the files into the repository.

(ℹ️ The linked page predates Observable Framework, but the reconstruction of the problem is still valid.)

The gist is that all it would take would be to switch the `web` target to `web` instead of `no-modules`. I'd be happy to push a PR, but I don't know about the impact of replacing `no-modules`. To play extra nice with JS CDNs (like jsdelivr, esm.sh, unpkg.com), we'd need to keep the package.json as-is, I think. That _could_ break existing usage? 🤔 

Related https://github.com/PRQL/prql/issues/1836

srenatus:
If the updated README.md snippet works remains to be shown. I've tried emulating it via

    import init, { compile } from "https://cdn.jsdelivr.net/gh/srenatus/prql@sr-testing/prqlc/bindings/js/dist/web/prql_js.js";

but it's not entirely the same thing. 😬 

image

Fixes https://github.com/PRQL/prql/issues/4273.

### DIFF
--- a/prqlc/bindings/js/README.md
+++ b/prqlc/bindings/js/README.md
@@ -73,18 +73,12 @@ console.log(sql);
 ```html
 <html>
   <head>
-    <script src="./node_modules/prql-js/dist/web/prql_js.js"></script>
-    <script>
-      const { compile } = wasm_bindgen;
+    <script type="module">
+      import init, { compile } from './dist/web/prql_js.js';
+      await init();
 
-      async function run() {
-        await wasm_bindgen("./node_modules/prql-js/dist/web/prql_js_bg.wasm");
-        const sql = compile("from db.employees | select first_name");
-
-        console.log(sql);
-      }
-
-      run();
+      const sql = compile("from db.employees | select first_name");
+      console.log(sql);
     </script>
   </head>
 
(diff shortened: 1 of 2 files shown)


max-sixty:
Thanks @srenatus ! That seems promising!

Can we avoid using a CDN in the example? Ideally it would be very self-contained...

srenatus:
Updated! And validated with the README's example as index.html and `python -m http.server`:
image

max-sixty:
Awesome! Thank you @srenatus !

max-sixty:
To the extent you're up for adding a line in the changelog, that's helpful, otherwise I'll try and add when we do a release...

PR #4278, opened 2024-02-28
fix: variable docs

### LINKED ISSUE #4275: The `let main =` example in the book doesn't work
### What happened?

In the [variables](https://prql-lang.org/book/reference/declarations/variables.html) page of the PRQL book, it states that `from x` is equivalent to `let main = x`.  When I try the example in the playground, it doesn't work unless I use `let main = from x`.

I think this is just a documentation issue and not a bug in PRQL itself.

### PRQL input

```elm
let main = albums
```

### SQL output

```SQL
Error: 
   ╭─[:1:12]
   │
 1 │ let main = albums
   │            ───┬──  
   │               ╰──── Unknown name `albums`
───╯
```

### Expected SQL output

```SQL
SELECT
  *
FROM
  albums
```

### MVCE confirmation

- [X] Minimal example
- [X] New issue

### Anything else?

_No response_

aljazerzen:
Closes #4275

### DIFF
--- a/web/book/src/reference/declarations/variables.md
+++ b/web/book/src/reference/declarations/variables.md
@@ -24,7 +24,7 @@ Syntactically, variables can take 3 forms.
 - The final expression of a pipeline defaults to taking the name `main`.
 
   ```prql no-eval
-  from db.x
+  x
   ```
 
   ... is equivalent to:


max-sixty (approved):
Ah perfect! I should have realized. Thanks

PR #4318, opened 2024-03-08
refactor: rename `types::TupleField` to `types::TyTupleField`

aljazerzen:
This PR is the first commit in series of trying to formalize what expression aliases are (i.e. `x` in `select x = my_col`).

The idea was that these are all just tuple field names, but it turns out this is not the case, so I'm abandoning that idea.

This PR is just a refactor that makes sense on its own, remaining commits are on branch `feat-tuple-field`.

### DIFF
--- a/lutra/lutra/src/pull_schema.rs
+++ b/lutra/lutra/src/pull_schema.rs
@@ -2,7 +2,7 @@ use anyhow::Result;
 use arrow::datatypes::{DataType, SchemaRef};
 use connector_arrow::api::SchemaGet;
 use itertools::Itertools;
-use prqlc::ast::{PrimitiveSet, Stmt, TupleField, Ty, TyKind, VarDef};
+use prqlc::ast::{PrimitiveSet, Stmt, TyTupleField, Ty, TyKind, VarDef};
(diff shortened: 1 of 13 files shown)


vanillajonathan:
What does "Ty" mean?

aljazerzen:
type

We have a bunch of other things named this way.

vanillajonathan:
It is far from obvious that "Ty" means "Type". It's an abbreviation I've never heard of. Seems not so helpful to short down a four-letter word to a two-letter abbreviation. Rust has a convention of appending "Kind" as a suffix on enums so maybe this would be better named `TupleFieldKind`.

aljazerzen:
Well, it is a compiler abbreviation, I guess. Here I wanted to distinguish between the "TupleField expression" and "TupleField type"

Rust's ast does use it:

https://doc.rust-lang.org/beta/nightly-rustc/rustc_ast/ast/struct.Ty.html

We use it:

https://github.com/PRQL/prql/blob/main/prqlc/prqlc-ast/src/types.rs#L10

https://github.com/PRQL/prql/blob/main/prqlc/prqlc-ast/src/types.rs#L86

PR #4325, opened 2024-03-09
feat: import

aljazerzen:
this is basically let, but:
- not supporting any expressions by idents,
- supporting all kinds of declarations (such as modules and types)


snth:
Interesting. I have some questions but I need to first familiarise myself with how the rest of the module system works.

max-sixty:
Good to move things along on `main`. 

FWIW I thought we were going to do implicit imports. But easy to change down the line if we need to...

PR #4332, opened 2024-03-13
feat: Add pl_to_prql to JS bindings

jacquayj:
Adds `pl_to_prql` to JS bindings


max-sixty:
Thanks a lot @jacquayj ! And welcome to PRQL!

PR #4333, opened 2024-03-16
feat: Add `display` option to `compile`, allow specifying colors

max-sixty:


prqlc/bindings/prqlc-python/src/lib.rs
```diff
@@ -1,44 +1,43 @@
...
+        ));
+    };
+
+    Ok(prql_query)
+        .and_then(prqlc_lib::prql_to_pl)
+        .and_then(prqlc_lib::pl_to_rq)
+        .and_then(|rq| prqlc_lib::rq_to_sql(rq, &options.unwrap_or_default()))
         .map_err(|e| e.composed(&prql_query.into()))
-        .map_err(|e| (PyErr::new::<exceptions::PySyntaxError, _>(e.to_string())))
+        .map_err(|e| (PyErr::new::<exceptions::PyValueError, _>(e.to_string())))
```
eitsupi, line 20: Sorry to ask this now, but is the `display` argument really effective here?

I had trouble getting this option to work when I ported it to prqlc-r (PRQL/prqlc-r#317), but after looking at the source code I realized that this only works within the `prql::compile` function.
IIUC, this function should be updated to something like:

```rust
    prqlc_lib::compile(prql_query, &options)
        .map_err(|e| PyErr::new::<exceptions::PyValueError, _>(e.to_string())))
```

(Sorry don't have time to try today, but if no one sees this I guess I can take a look in a few days)

max-sixty: Yes, I'm not sure why I did it that way, I guess I was adjusting from the previous version. Please feel free to simplify!

eitsupi: #4919

Issue #4338, opened 2024-03-17
Move `db.` changes to a new branch


max-sixty:
### What's up?

As discussed on the dev call, we're going to revert this from the main branch, and then work on the new resolver on a separate branch, which will let us experiment there.

PR #4347, opened 2024-03-18
docs: Add back some docs we lost

max-sixty:
I must have mistakenly lost these when reorganizing the `from` docs


max-sixty:
Ah, sorry, this was actually us losing them for good reason. 

But coincidentally they'll come back with #4338

PR #4349, opened 2024-03-20
chore: Move `db.` to a new branch

max-sixty:
As discussed in https://github.com/PRQL/prql/issues/4338, this takes out the `db.` change from `main`, so we can release as normal while experimenting with a new branch.

I _almost_ got the revert working. I got a bit stuck at the top of `names.rs` — possibly listing the columns in the error message incompatible with the old code? If @aljazerzen is around and could take a look, that would be great...


prqlc/prqlc/src/semantic/resolver/names.rs
```diff
@@ -16,16 +16,21 @@ use super::Resolver;
...
+                    break;
+                }
+                ident = ident.pop_front().1.unwrap();
+                res = self.resolve_ident_core(&ident, None);
             }
-            ident = ident.pop_front().1.unwrap();
-            res = self.resolve_ident_core(&ident);
-        }
+            res
+        };
```
aljazerzen, line 33: You mean this?

What's the problem here?

max-sixty: We get lots of errors — for example: https://github.com/PRQL/prql/actions/runs/8363115491/job/22895344620?pr=4349#step:14:615

aljazerzen: I was the tests that were wrong.

PR #4356, opened 2024-03-24
feat: indirection parsing

aljazerzen:
Parsing for these:
```
"hello".length
{a = 1}.1
123.length
```

### DIFF
--- a/prqlc/prqlc-ast/src/expr.rs
+++ b/prqlc/prqlc-ast/src/expr.rs
@@ -42,7 +42,11 @@ pub struct Expr {
 
 #[derive(Debug, EnumAsInner, PartialEq, Clone, Serialize, Deserialize, strum::AsRefStr)]
 pub enum ExprKind {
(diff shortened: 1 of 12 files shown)


vanillajonathan:
Seems really cool!

I brought up a discussion earlier in #1767 which sounds related.

max-sixty:
Is the intention that in `foo.bar`, `bar` can be a function which takes `foo` as an input?

If so, how would we discriminate between the column `bar` and a function `bar` in `foo.bar`?

aljazerzen:
No, i didnt mean to use this for associated functions/methods - i mean to use it for accessing fields of a tuple.

max-sixty:
OK yes great. So then are the top and bottom examples right?

> "hello".length
> {a = 1}.1
> 123.length

aljazerzen:
Syntactically, yes. Semantically, no.

max-sixty:
Cool, thanks

PR #4376, opened 2024-04-07
docs: improve `join` docs

### LINKED ISSUE #4341: CROSS JOIN or multi from condition in postgres
# syntax for cross join

It might be just me, but I cant find a way to write this in prql. I know its basic, but I can't find the example anywhere. I also could not see any test cases with prql -> sql checks where I could search for my use case - I only found https://github.com/PRQL/prql/tree/main/prqlc/prqlc/tests/integration/queries

The use case is simple - put query params in a table and reuse that when building your data up through multiple steps. Example below is similar.

```
with firstnames as (
  select 'First' as first
),

lastnames as (
  select 'Last' as last
)

select * from firstnames, lastnames
-- OR
select * from firstnames CROSS JOIN lastnames;
```

So its basically a join without a condition ... can it be done in prql?

aljazerzen:
Closes #4341


max-sixty:
(never responded to this, thanks, much better!)

Issue #4421, opened 2024-04-25
Compiler crash on `from_text` with zero rows


kgutwin:
### What happened?

It should be possible (not very useful, perhaps, but still possible) to define a relation using `from_text` where it has known columns but no rows. At a minimum, such a definition ought not to crash...

### PRQL input

```elm
from_text "col_a,col_b"
# or, in JSON format
from_text format:json '{"columns": ["col_a","col_b"], "data": []}'
```

### SQL output

```SQL
n/a, crashes
```

### Expected SQL output

```SQL
-- could be something like this
WITH table_0 AS (
  SELECT
    NULL AS col_a,
    NULL AS col_b
  WHERE 1 = 0
)
SELECT
  col_a,
  col_b
FROM
  table_0
```

### MVCE confirmation

- [X] Minimal example
- [X] New issue

### Anything else?

The message on crash is:

```
The application panicked (crashed).
Message:  removal index (is 0) should be < len (is 0)
Location: prqlc/prqlc/src/sql/gen_query.rs:446
```

Tested with latest code, commit d642a30

m-span:
I'm not sure that the provided SQL output is the best choice moving forward, but I'm making a PR that at least provides an error message for this situation and prevents a crash

snth:
@kgutwin What is your use case for this?

1. Is this as a kind of schema definition/declaration, or
2. you are processing files which sometimes don't have any data?

For case 2. @KaeporaGaebora should help (and is greatly appreciated 🙏).

For case 1., this kind of thing should be handled by the type system once implemented.

Please note that the following works:
```elm
from_text "col_a,col_b\n,"
```

You can produce the SQL you provided with the following:
```elm
from [{col_a=null, col_b=null}]
```

HTH

kgutwin:
Thanks - the use case might be a little uncommon, but our team is working on code that takes user input and generates PRQL documents. As part of that process, it's typical for users to generate empty relations with known columns. It's not very useful to them -- they will typically immediately begin adding rows -- but since our app generates and executes the PRQL "live", having a compiler error on an empty relation isn't something we want the users to need to deal with.

We can, of course, catch the compiler error and smoothly handle it, so the solution proposed by @KaeporaGaebora is indeed plausible; but I do wonder whether an error is the right solution, given that it is definitely not a SQL error to have an empty relation. As of now, we don't have a specific need to be able to create an empty relation, but I suppose at some point in the future it might come up.

[...]

PR #4422, opened 2024-04-26
fix: add error msg for no rows in from_text (#4421)

m-span:
addresses #4421 by adding an error message if no data rows are provided with `from_text`

### DIFF
--- a/prqlc/prqlc/src/sql/gen_query.rs
+++ b/prqlc/prqlc/src/sql/gen_query.rs
@@ -447,7 +447,7 @@ fn translate_relation_literal(data: RelationLiteral, ctx: &Context) -> Result<sq
         return Err(
             Error::new_simple("No rows provided for `from_text`".to_string()).push_hint(
                 "add a newline, then a row of data following the column. If using \
-                the json format, ensure `data` isnt empty",
+                the json format, ensure `data` isn't empty",
             ),
         );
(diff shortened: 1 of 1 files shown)


prqlc/prqlc/src/sql/gen_query.rs
```diff
@@ -443,6 +443,15 @@ fn translate_relation_literal(data: RelationLiteral, ctx: &Context) -> Result<sq
         selects.push(body)
     }
 
+    if selects.is_empty() {
+        return Err(
+            Error::new_simple("No rows provided for `from_text`".to_string()).push_hint(
+                "add a newline, then a row of data following the column. If using \
+                the json format, ensure `data` isnt empty",
```
vanillajonathan: Maybe "JSON" instead of "json".

max-sixty:
```suggestion
                the json format, ensure `data` isn't empty",
```

max-sixty:
Nice, thanks @KaeporaGaebora ! 

A test would be great too, we have a file for them here: https://github.com/PRQL/prql/blob/76c4f767c4312976dba0fbadfa6c27b84e31faa9/prqlc/prqlc/tests/integration/error_messages.rs

I won't block merging this though

PR #4430, opened 2024-04-29
build: Replace `atty` with `is-terminal`

max-sixty:
Continuation of https://github.com/PRQL/prql/pull/3240, replacing a deprecated dependency

FYI I tried using the clio approach, but it still has the same existing issue re printing the help message even when not in a terminal.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1809,12 +1809,12 @@ dependencies = [
 
 [[package]]
 name = "is-terminal"
-version = "0.4.10"
+version = "0.4.12"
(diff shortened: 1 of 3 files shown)


max-sixty:
FYI @aj-bagwell tell me if the @ isn't helpful — but [this](https://github.com/PRQL/prql/pull/4430/files#diff-0ada0ef3423576d28b7c2d902ecdd1775344d03c39e01bf3eb94b43f8cab6a89L467) seems not to work — it returns `true` even when stdin is being piped to...

PR #4443, opened 2024-05-01
ci: Update megalinter workflow

max-sixty:
Copies from https://raw.githubusercontent.com/oxsecurity/megalinter/main/TEMPLATES/mega-linter.yml

(still wish this would be an action we can reference, rather than copying & pasting 150 lines...)

### DIFF
--- a/.github/workflows/lint-megalinter.yaml
+++ b/.github/workflows/lint-megalinter.yaml
@@ -1,6 +1,3 @@
-# This MegaLinter workflow is very experimental stage. Open to changes and updates.
-# https://github.com/PRQL/prql/pull/1974
-
 # MegaLinter GitHub Action configuration file
 # More info at https://megalinter.io
 name: lint-megalinter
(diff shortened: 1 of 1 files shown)


.github/workflows/lint-megalinter.yaml
```diff
@@ -9,88 +6,159 @@ on:
...
+# Comment env block if you do not want to apply fixes
+env:
   # Apply linter fixes configuration
-  # APPLY_FIXES: all # When active, APPLY_FIXES must also be defined as environment variable (in github/workflows/mega-linter.yml or other CI tool)
-  APPLY_FIXES_EVENT: pull_request # Decide which event triggers application of fixes in a commit or a PR (pull_request, push, all)
-  APPLY_FIXES_MODE: commit # If APPLY_FIXES is used, defines if the fixes are directly committed (commit) or posted in a PR (pull_request)
+  #
+  # When active, APPLY_FIXES must also be defined as environment variable
+  # (in github/workflows/mega-linter.yml or other CI tool)
+  APPLY_FIXES: all
```
eitsupi: I think there is a risk of conflict with modification by the pre-commit workflow if automatic modification is allowed.

.github/workflows/lint-megalinter.yaml
```diff
@@ -9,88 +6,159 @@ on:
...
 
       # MegaLinter
-      - name: lint-megalinter
-        id: ml
+      - name: MegaLinter
+
         # You can override MegaLinter flavor used to have faster performances
         # More info at https://megalinter.io/flavors/
-        uses: oxsecurity/megalinter@v7.11.1
+        uses: oxsecurity/megalinter@v7
```
eitsupi: It implicitly changes the linter settings so that one day CI will start failing. We've had that happen in the past, which is why I changed it to pin a version (#3056).

eitsupi (changes requested):

.github/workflows/lint-megalinter.yaml
max-sixty: Thanks, fixed

.github/workflows/lint-megalinter.yaml
max-sixty: Thanks, fixed

eitsupi (approved):
Looks good, thanks.

PR #4444, opened 2024-05-01
ci: Upgrade artifact actions to v4

max-sixty:


eitsupi (commented):
If I understand correctly, I think this is still not enough. A step is required to merge multiple artifacts.
https://github.com/actions/upload-artifact/blob/main/docs/MIGRATION.md#merging-multiple-artifacts

I'll take a look when I have time today.

max-sixty:
> If I understand correctly, I think this is still not enough. A step is required to merge multiple artifacts.
> https://github.com/actions/upload-artifact/blob/main/docs/MIGRATION.md#merging-multiple-artifacts

I don't think so! 

From: https://github.com/actions/upload-artifact/blob/main/merge/README.md

> For most cases, this may not be the most efficient solution. See [the migration docs](https://github.com/actions/upload-artifact/blob/main/docs/MIGRATION.md#multiple-uploads-to-the-same-named-artifact) on how to download multiple artifacts to the same directory on a runner. This action should only be necessary for cases where multiple artifacts will need to be downloaded outside the runner environment, like downloads via the UI or REST API.

eitsupi:
I suppose python release will fail.

max-sixty:
I think everything is fixed now! If you spot anything let me know and I'll come back to it...

PR #4454, opened 2024-05-05
build: Bump duckdb to 0.10.2

max-sixty:


max-sixty:
@aljazerzen FYI we have a couple of version of arrow now, because of connectorX

```
cargo tree --all-features -i arrow@49
arrow v49.0.0
├── connector_arrow v0.4.1
│   └── lutra v0.11.5 (/Users/maximilian/workspace/prql/lutra/lutra)
│       └── lutra-python v0.11.5 (/Users/maximilian/workspace/prql/lutra/bindings/python)
├── lutra v0.11.5 (/Users/maximilian/workspace/prql/lutra/lutra) (*)
└── lutra-python v0.11.5 (/Users/maximilian/workspace/prql/lutra/bindings/python)
```

aljazerzen:
We do, yes, but that should not be a problem, right?

A bit longer compilation times, but nothing else.

max-sixty:
> We do, yes, but that should not be a problem, right?
> 
> A bit longer compilation times, but nothing else.

Yes totally, only longer compilations, v low priority

(Have been doing some low priority items myself while thinking about whether we need to rewrite the codegen completely!)

PR #4463, opened 2024-05-11
[draft] Add Span to our Lexer error

max-sixty:
@m-span re your message on Discord — FYI a few weeks ago I tried to add `Span` to the lexer error, to try and unify the error types (so we're thinking similar things!).

But I got arcane chumsky errors and didn't manage to figure it out. If you can, great, but also this may be a warning beacon that it's not so easy...


max-sixty:
Done better by @m-span !

PR #4470, opened 2024-05-13
refactor: more error info, span merge

m-span:
- ~~merges ast into parser~~ moves error from span to parser
- merges span, error objects into single, custom objects
- handles error struct conversion better so errors are more informative
- fixes existing bad error message for interpolated strings
- adds new bad error message for interpolated strings


prqlc/prqlc/src/cli/mod.rs
```diff
@@ -199,6 +211,37 @@ pub struct IoArgs {
     main_path: Option<String>,
 }
 
+#[derive(Copy, Clone, Debug, Default)]
+struct LoggingHelp;
+
+impl LogLevel for LoggingHelp {
```
max-sixty: This is useful!

max-sixty (commented):
Nice, great to see you've made progress!

It would be easier to review if we can clean up some of the diffs — looks like we have some merge conflicts and some formatting-only changes. (I don't think we need to be sticklers about splitting things into commits within a PR.)

It also would be easier to review if we can split things into smaller PRs; for example there's a nice `-v` option that we could merge immediately. Review difficulty seems to increase super-linearly with size, so breaking things up is easier. But it's also not fun to do ex-post, so it may not be worthwhile at this stage.

Does the `prqlc-ast` / `prqlc-parser` crate merging give us anything at the moment? I'm hesitant to go through another refactor without either a clear reduction in code complexity or a win on language functionality...

Nice work on the error message improvements — that part sounds great.

prqlc/prqlc/src/cli/mod.rs
m-span: I refactored this out into a separate PR #4471

m-span:
Dupilicate code and concepts in `span` and `error` were merged. I moved ast files back to prqlc-ast, *except* for error, which had to stay in the `parser` to avoid circular deps.

prqlc/prqlc/tests/integration/bad_error_messages.rs
```diff
@@ -78,17 +78,14 @@ fn test_bad_error_messages() {
...
     Error:
-       ╭─[:3:14]
+       ╭─[:1:21]
        │
-     3 │     select f"{}"
-       │              ┬
-       │              ╰── unexpected end of input while parsing interpolated string
+     1 │ from x | select f"{}
+       │                     │
+       │                     ╰─ unexpected
```
max-sixty, line 89: Is this correct? We get a much nicer error just below this in the next file, I think for the same input...

prqlc/prqlc/tests/integration/error_messages.rs
```diff
@@ -364,3 +364,16 @@ fn available_columns() {
...
 }
+
+#[test]
+fn empty_interpolations() {
+    assert_snapshot!(compile(r#"from x | select f"{}" "#).unwrap_err(), @r###"
+    Error:
+       ╭─[:1:20]
+       │
+     1 │ from x | select f"{}"
+       │                    ┬
```
max-sixty, line 376: (here)

max-sixty (approved):
Superb, thanks!

prqlc/prqlc/tests/integration/bad_error_messages.rs
m-span: its missing the terminating `"` to the interpolated string

prqlc/prqlc/tests/integration/error_messages.rs
m-span: This was moved from `bad_error_messages` to `error_messages` because I updated how it responds, and I think it's better

max-sixty: Yes much better!

prqlc/prqlc-parser/Cargo.toml
```diff
@@ -12,10 +12,13 @@ version.workspace = true
 doctest = false
 
 [dependencies]
-itertools = {workspace = true}
-prqlc-ast = {path = "../prqlc-ast", version = "0.11.5" }
-semver = {version = "1.0.22"}
+prqlc-ast = {path = "../prqlc-ast"}
```
max-sixty:
```suggestion
prqlc-ast = {path = "../prqlc-ast", version = "0.11.5" }
```
otherwise we can't publish

prqlc/prqlc-parser/src/lib.rs
```diff
@@ -1,29 +1,37 @@
+extern crate core;
```
max-sixty: nit: I think unneeded these days?

max-sixty (commented):
This is looking great!

What's the overall approach for the dependencies running through the crates? i.e. I see that `chumsky` is now a dependency of `prqlc-ast`, but also that we've move the errors to `prqlc-parser`. Smallish changes are fine as we simplify things; we can hopefully do those while maintaining coherence though.

max-sixty:
FYI I merged main in, as merging the import sorting wasn't as trivial as I had hoped; and keen to not slow this down. 

Also added a reference to this in #4521

prqlc/prqlc-parser/Cargo.toml
max-sixty: FYI @m-span we missed this and it broke the publishing! (No great stress tbc, wanted to let you know out of respect rather than blame...)

m-span: 🫨 whoops!

PR #4471, opened 2024-05-14
feat: add logging verbosity flags to cli

m-span:
Makes things a little easier to debug and trace

### DIFF
--- a/prqlc/prqlc/Cargo.toml
+++ b/prqlc/prqlc/Cargo.toml
@@ -61,6 +61,7 @@ sqlformat = "0.2.3"
 sqlparser = {version = "0.46.0", features = ["serde"]}
 strum = {version = "0.26.2", features = ["std", "derive"]}
 strum_macros = "0.26.2"
+clap-verbosity-flag = "2.2.0"
 
 [target.'cfg(not(target_family="wasm"))'.dependencies]
 
--- a/prqlc/prqlc/Cargo.toml
+++ b/prqlc/prqlc/Cargo.toml
@@ -71,7 +71,7 @@ anyhow = {version = "1.0.86", features = ["backtrace"], optional = true}
 clap = {version = "4.4.18", features = ["derive", "env", "wrap_help"], optional = true}
 clap_complete_command = {version = "0.5.1", optional = true}
 # Pinned for MSRV
-clap-verbosity-flag = { version = "=2.1.2", optional = true }
+clap-verbosity-flag = {version = "=2.1.2", optional = true}
 clio = {version = "0.3.3", features = ['clap-parse'], optional = true}
 color-eyre = {version = "0.6.3", optional = true}
 colorchoice-clap = {version = "1.0.0", optional = true}
@@ -82,7 +82,9 @@ walkdir = {version = "2.5.0", optional = true}
 
 # Not direct dependencies, but pinning because of bugs in previous versions. Can
 # remove when dependencies no longer use it. (If CI passes, it can be removed.)
-clap_complete = {version = "4.4.9", optional = true}
+clap_builder = {version = "=4.4.18", optional = true}
+clap_complete = {version = "=4.4.10", optional = true}
+clap_complete_fig = {version = "=4.4.2", optional = true}
 
 # We use minijinja just for the Jinja lexer, which is not part of the
 # public interface which is covered by semver guarantees.
--- a/prqlc/prqlc/Cargo.toml
+++ b/prqlc/prqlc/Cargo.toml
@@ -70,7 +70,8 @@ strum_macros = "0.26.2"
 anyhow = {version = "1.0.86", features = ["backtrace"], optional = true}
 clap = {version = "4.4.18", features = ["derive", "env", "wrap_help"], optional = true}
 clap_complete_command = {version = "0.5.1", optional = true}
-clap-verbosity-flag = { version = "2.1.2", optional = true }
+# Pinned for MSRV
+clap-verbosity-flag = { version = "=2.1.2", optional = true }
 clio = {version = "0.3.3", features = ['clap-parse'], optional = true}
 color-eyre = {version = "0.6.3", optional = true}
 colorchoice-clap = {version = "1.0.0", optional = true}
--- a/prqlc/prqlc/Cargo.toml
+++ b/prqlc/prqlc/Cargo.toml
@@ -70,7 +70,7 @@ strum_macros = "0.26.2"
 anyhow = {version = "1.0.86", features = ["backtrace"], optional = true}
 clap = {version = "4.4.18", features = ["derive", "env", "wrap_help"], optional = true}
 clap_complete_command = {version = "0.5.1", optional = true}
-clap-verbosity-flag = { version = "2.2.0", optional = true }
+clap-verbosity-flag = { version = "2.1.2", optional = true }
 clio = {version = "0.3.3", features = ['clap-parse'], optional = true}
 color-eyre = {version = "0.6.3", optional = true}
 colorchoice-clap = {version = "1.0.0", optional = true}
--- a/prqlc/prqlc/Cargo.toml
+++ b/prqlc/prqlc/Cargo.toml
@@ -16,6 +16,7 @@ cli = [
   "anyhow",
   "clap_complete_command",
   "clap_complete",
+  "clap-verbosity-flag",
   "clap",
   "clio",
   "color-eyre",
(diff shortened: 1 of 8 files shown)


prqlc/prqlc/src/cli/mod.rs
```diff
@@ -1,42 +1,51 @@
...
     let mut cli = Cli::parse();
+    env_logger::builder()
+        .filter_level(cli.verbose.log_level_filter())
+        .format_timestamp(None)
+        .init();
+    color_eyre::install()?;
     cli.color.write_global();
 
+    if Some(log::Level::Trace) == cli.verbose.log_level() {
+        env::set_var("RUST_BACKTRACE", "1");
```
max-sixty: Possibly only do this if it's not already set? Otherwise `RUST_BACKTRACE=full` will be overwritten

max-sixty (approved):
Looks good!

I left one small comment, then good to merge

(need to run `cargo fmt` too)

prqlc/prqlc/Cargo.toml
```diff
@@ -54,6 +54,7 @@ sqlformat = "0.2.3"
 sqlparser = {version = "0.46.0", features = ["serde"]}
 strum = {version = "0.26.2", features = ["std", "derive"]}
 strum_macros = "0.26.2"
+clap-verbosity-flag = "2.2.0"
```
max-sixty: Ah, and we should move this down to the next section, allow it to be optional, and add it to the cli feature; otherwise it'll interfere with the wasm compilation...

prqlc/prqlc/src/cli/mod.rs
m-span: I'm starting to rethink this. Perhaps I should remove this part altogether - I'll leave backtracing to the env var, and only have this PR affect logging?

max-sixty: That's fine!

prqlc/prqlc/Cargo.toml
m-span: fixed 👍

max-sixty (approved):

max-sixty:
Looks good! We can def merge pending tests + formatting...

max-sixty:
Would be great to merge this! Let me know if you want a hand with the formatting?

prqlc/prqlc/Cargo.toml
```diff
@@ -62,6 +63,7 @@ strum_macros = "0.26.2"
 anyhow = {version = "1.0.82", features = ["backtrace"], optional = true}
 clap = {version = "4.4.18", features = ["derive", "env", "wrap_help"], optional = true}
 clap_complete_command = {version = "0.5.1", optional = true}
+clap-verbosity-flag = { version = "2.2.0", optional = true }
```
eitsupi: The addition of this seems to bring the MSRV to 1.74.
My question is that changing the MSRV of prqlc here will even affect its functionality as a built-in that does not require a
cli feature.
Why are cli features included in default? I think it would be better to exclude them from default.

(As a side note, I think the MSRV could be 1.74 for now. I just don't think it should be pulled to other parts of the cli feature about the future.)

max-sixty: CLI is in default otherwise `cargo install prqlc` doesn't work, and that is likely to be done by many more people than adding prqlc as a library dependency.

We have changed this around a couple of times — having a separate crate etc — I do think this is the right tradeoff at the moment, even though it's indeed imperfect.

It does present an issue with defining the MSRV though, since we need to have a value in the crate and it can't be dependent on features. @eitsupi IIUC the R ecosystem & debian were the main impediments to increasing the MSRV; is that still the case?

prqlc/prqlc/src/cli/mod.rs
```diff
@@ -1,26 +1,24 @@
...
-mod docs_generator;
-mod jinja;
-mod watch;
+use std::collections::HashMap;
+use std::env;
+use std::io::{Read, Write};
+use std::ops::Range;
+use std::path::Path;
+use std::process::exit;
+use std::str::FromStr;
```
max-sixty, line 9: This sorting makes sense, thanks

prqlc/prqlc/Cargo.toml
max-sixty: @m-span if we downgrade to 2.1.2 then we get the 1.70.0 MSRV; let's do that I think. Then we'll upgrade when MSRVs match in the near future.

eitsupi: > IIUC the R ecosystem & debian were the main impediments to increasing the MSRV; is that still the case?

Over the past month, rustc on Debian has migrated rapidly from 1.70 to 1.74, and Ubuntu is currently at 1.75, so I consider it quite unlikely that setting 1.74 or 1.75 will cause problems.

image

The problem with CRAN is that it continues to use Fedora 36 (time stopped at rustc 1.69), which reached EOL a year ago, so it is not worth considering.
They do not have a roadmap for how long they will continue to use Fedora 36 and cannot predict the future.

Since they are volunteers and do not seem to have the resources to update Fedora under the current system, I suspect that this situation may continue until the infrastructure is completely rebuilt on an IaC basis.

Either way I will either patch to lower the MSRV (which may involve manual modification of the lock file and is a tremendous pain) or simply give up submitting to CRAN.

max-sixty: Ah OK, that's a shame re CRAN, hope they figure out a way to go forwards.

Given they're at 1.69, I guess staying on 1.70 doesn't help much? And we can bump to 1.74? I think it's quite easy for us to stay at 1.70 (in this case a tiny downgrade), but sounds like that's not actually that helpful for CRAN?

eitsupi: Thanks, I think the process of lowering from 1.70 to 1.69 is much easier than the process of lowering from 1.74 to 1.69.
As far as I am aware there is no new syntax introduced in 1.70 in prqlc at the moment, just some dependencies.
If the checks in 1.70 were to be stopped, it would include many more dependencies, which would make it more difficult to identify the problem.

max-sixty: OK, let's leave it at 1.70 for the moment then.

Let's do the downgrade and merge this PR!

m-span: Apparently we aren't even on MSRV 1.70 right now:

```
~/projects/prql git:[main]
cargo msrv --linear
Fetching index
Determining the Minimum Supported Rust Version (MSRV) for toolchain aarch64-apple-darwin
Using check command cargo check
Check for toolchain '1.78.0-aarch64-apple-darwin' succeeded
Check for toolchain '1.77.2-aarch64-apple-darwin' succeeded
Check for toolchain '1.76.0-aarch64-apple-darwin' succeeded
Check for toolchain '1.75.0-aarch64-apple-darwin' succeeded
Check for toolchain '1.74.1-aarch64-apple-darwin' succeeded

Check for toolchain '1.73.0-aarch64-apple-darwin' failed with:
┌───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┐
│ error: package `pulldown-cmark v0.10.3` cannot be built because it requires rustc 1.74 or newer, while the currently active rustc version is 1.73.0                           │
│ Either upgrade to rustc 1.74 or newer, or use                                                                                                                                 │
│ cargo update -p pulldown-cmark@0.10.3 --precise ver                                                                                                                           │
│ where `ver` is the latest version of `pulldown-cmark` supporting rustc 1.73.0                                                                                                 │
└───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┘
   Finished The MSRV is: 1.74.1   ██████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████ 00:00:03
   ```

m-span: but yes, I did downgrade that package to `2.1.2`

eitsupi: > Apparently we aren't even on MSRV 1.70 right now:

Please check CI, only prqlc is confirming MSRV.
https://github.com/PRQL/prql/blob/045c03dfc4f28a0e484e50f0ad6ba0d8e7c09689/.github/workflows/tests.yaml#L495-L498

> but yes, I did downgrade that package to `2.1.2`

The lock file does not seem updated.

eitsupi (changes requested):
The test still fails. Perhaps a patch version update of one of the dependencies has changed the MSRV.
I think you need to undo all the changes to the lock file and update it again.

max-sixty:
Cargo has a slightly odd feature where it presumes `~` in requirements. So I pinned and pushed. Sorry we got bogged down a bit in the overhead on this one. The feature is nice!

Edit: @eitsupi looks like you need to give the "OK" on the changes[^1]

[^1]: maybe we shouldn't use this feature unless something is really important and won't get checked in tests (which this will), since it blocks until the original individual approves...

max-sixty (approved):

eitsupi (approved):
Thanks!

m-span:
@max-sixty thank you for your help with this

PR #4477, opened 2024-05-16
devops: rustfmt only runs on staged files

m-span:
files were starting to deviate from rustfmt, making diffs of other branches more complex

fmt stage is now written within yaml file, so no more network issues

by using `rustfmt` instead of `cargo fmt`, the pre-commit fmt only runs on staged files

all files have also been run against `rustfmt`


max-sixty:
What's the advantage of running `rustfmt` over `cargo fmt`? I haven't seen projects do the former...

Why would they give different outputs?

max-sixty:
@m-span is there any chance you might have local configs which cause different outputs? If I pull this branch and then run `pre-commit run -a`, the rust code all resets back to that of `main`

m-span:
`cargo fmt` always formats all rust files in the full project. `rustfmt` can be fed a list of files (from pre-commit) so that it only formats the files that have changed. This is a known issue/difference for pre-commit.

What is the intended developer experience? Every PR must have all rust files formatted, or each PR only expects changed files to be formatted?

m-span:
I ask because in order for other branches to 'pass', they have to cargo fmt files that weren't even affected? If this is a hard requirement, how did it get desynced?

max-sixty:
> I ask because in order for other branches to 'pass', they have to cargo fmt files that weren't even affected? If this is a hard requirement, how did it get desynced?

I don't think it is desynced! 

Expressed as humbly as possible — yours is the first system to have these formatting issues — are you sure it's not something up with your end?

From above:

> @m-span is there any chance you might have local configs which cause different outputs? If I pull this branch and then run `pre-commit run -a`, the rust code all resets back to that of `main`

For a hard test — CI checks `cargo fmt` passes in CI on `main`...

max-sixty:
Is there possibly a config file in one of these locations:

> If none of these directories contain such a file, both your home directory and a directory called rustfmt in your global config directory (e.g. .config/rustfmt/) are checked as well.

From https://rust-lang.github.io/rustfmt/?version=v1.6.0&search=

m-span:
problem was with my environment

PR #4492, opened 2024-05-20
test: Skip `glaredb` test, which starting raising an error

max-sixty:


max-sixty:
@scsmithr we got some errors from glaredb — hope it's reasonable to @ you in case this is unexpected.

I'm guessing CI broke because we don't pin versions and a new version was released. The two SQL queries which raise errors are:

https://github.com/PRQL/prql/blob/59b2294eed4022291a225f0ac3cca449b82f67b4/prqlc/prqlc/tests/integration/snapshots/integration__queries__compile__text_module.snap
https://github.com/PRQL/prql/blob/59b2294eed4022291a225f0ac3cca449b82f67b4/prqlc/prqlc/tests/integration/snapshots/integration__queries__compile__window.snap

We're skipping those for the moment to get CI back to greeen

scsmithr:
Definitely unexpected, will poke around.

PR #4494, opened 2024-05-21
web: Add link for Databend integration

max-sixty:


max-sixty:
Let's do a release soon and this will go out then...

PR #4495, opened 2024-05-21
docs(book): Give the book a home page

max-sixty:
Currently it starts on `Relations`, not sure how that happened or how I hadn't noticed (assuming it's not intentional).

This changes it to start on the Summary page. Previously the Summary text was never shown; only the links that were grabbed and shown on the left bar.

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,7 +35,7 @@ aggregate {                                   # `aggregate` reduces each column 
 }
 ```
 
-Here's a fuller example of the language;
+Here's a fuller example of the language:
 
 ```elm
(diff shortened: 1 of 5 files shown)


max-sixty:
i.e. this is the existing home page:

image


The new one isn't _great_ but it's less confusing. Very open to improvements...

(edited since the links didn't seem to work when the summary was used as an index, and worth improved it a bit)

image

richb-hanover:
+1 This "cover page" a significant improvement. Thanks!

PR #4498, opened 2024-05-21
fix: Add support for expressions on left hand side of std.in operator

### LINKED ISSUE #4497: std.in operator does not support expression on left-hand-side
### What happened?

I'm trying to make a generic "boolean" conversion function, and this is where I am starting:

```
let coalesce_bool = func
  val <scalar || null>
  -> case [
    val == null => null,
    val == 0 => false,
    val == 1 => true,
    (val | in ["no", "n", "N", "false"]) => false,
    (val | in ["yes", "y", "Y", "true"]) => true,
  ]
```

This is working well, for example in the following:

```
from [{foo = 1}, {foo = null}]
derive { foo_b = coalesce_bool foo }
```

However, feeding a numeric value other than 0 or 1 into the function is causing DuckDB to raise a Conversion Error ("Could not convert string 'no' to INT32). This is because DuckDB gets to the corresponding line in the CASE statement

```
    WHEN foo IN ('no', 'n', 'N', 'false') THEN false
```

and tries to cast the strings in the array to numbers.

The solution for me seemed simple -- to cast `val` to VARCHAR to do the comparison by strings:

```
let coalesce_bool = func
  val <scalar || null>
  -> case [
    val == null => null,
    val == 0 => false,
    val == 1 => true,
    (val | as VARCHAR | in ["no", "n", "N", "false"]) => false,
    (val | as VARCHAR | in ["yes", "y", "Y", "true"]) => true,
  ]
```

but this causes a PRQL compiler panic:

```
Message:  args to `std.array_in` must be a column ref and an array
Location: prqlc/prqlc/src/sql/gen_expr.rs:176

  ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ BACKTRACE ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
                                ⋮ 13 frames hidden ⋮                              
  14: prqlc::sql::gen_expr::process_array_in::h96ab7dc44c311876
      at /Users/kgutwin/github/prql/prqlc/prqlc/src/sql/gen_expr.rs:176
  15: prqlc::sql::gen_expr::translate_expr::h392231bd9a27822a
      at /Users/kgutwin/github/prql/prqlc/prqlc/src/sql/gen_expr.rs:106
[ ... snip ... ]
```

I looked at [prqlc/prqlc/src/sql/gen_expr.rs](https://github.com/PRQL/prql/blob/5ccb89c430721e68bb9db0475ef94266d83d7cf1/prqlc/prqlc/src/sql/gen_expr.rs#L176) and it's not obvious to me why there's a panic there, especially as the other `process_*()` functions seem to handle left-hand-side expressions easily. Would it be possible to remove this panic in order to support more complex expressions for the `std.in` function? Or is there an underlying reason why the IN operator needs to have a column reference only on the left-hand side?

### PRQL input

```elm
from [{ a = 1 }]
derive { b = ( a+1 | in [2, 3] ) }
```

### SQL output

```SQL
n/a (compiler crashes)
```

### Expected SQL output

```SQL
WITH table_0 AS (
  SELECT
    1 AS a
)
SELECT
  a,
  a + 1 IN (2, 3) AS b
FROM
  table_0
```

### MVCE confirmation

- [X] Minimal example
- [X] New issue

### Anything else?

_No response_

kgutwin:
Resolves #4497.

I'm hoping this is as simple as it needs to be to resolve this... if there's a more complex reason why `std.in` needs to not support expressions on the left-hand side, more discussion is appreciated!

Thanks for this package - it's making a big impact on our project, which we hope to release relatively soon!

### DIFF
--- a/prqlc/prqlc/src/sql/gen_expr.rs
+++ b/prqlc/prqlc/src/sql/gen_expr.rs
@@ -103,7 +103,7 @@ pub(super) fn translate_expr(expr: Expr, ctx: &mut Context) -> Result<ExprOrSour
                     }
                 }
                 "std.concat" => return Ok(process_concat(&expr, ctx)?.into()),
-                "std.array_in" => return Ok(process_array_in(args, ctx)?.into()),
+                "std.array_in" => return Ok(process_array_in(&expr, args, ctx)?.into()),
                 "std.date.to_text" => {
                     return Ok(process_date_to_text(&expr, name, args, ctx)?.into())
                 }
@@ -157,10 +157,14 @@ fn process_null(name: &str, args: &[Expr], ctx: &mut Context) -> Result<sql_ast:
 }
 
 /// Translates into IN (v1, v2, ...) if possible
-fn process_array_in(args: &[Expr], ctx: &mut Context) -> Result<sql_ast::Expr> {
+fn process_array_in(expr: &Expr, args: &[Expr], ctx: &mut Context) -> Result<sql_ast::Expr> {
     match args {
         [col_expr @ Expr {
-            kind: ExprKind::ColumnRef(_),
+            kind: ExprKind::ColumnRef(_)
+                | ExprKind::Literal(_)
+                | ExprKind::SString(_)
+                | ExprKind::Param(_)
+                | ExprKind::Operator { name: _, args: _ },
             ..
         }, Expr {
             kind: ExprKind::Array(in_values),
@@ -173,7 +177,10 @@ fn process_array_in(args: &[Expr], ctx: &mut Context) -> Result<sql_ast::Expr> {
                 .collect::<Result<Vec<sql_ast::Expr>>>()?,
             negated: false,
         }),
-        _ => panic!("args to `std.array_in` must be a column ref and an array"),
+        _ => Err(
+            Error::new_simple("args to `std.array_in` must be an expression and an array")
+                .with_span(expr.span)
+        ),
     }
 }
 
--- a/prqlc/prqlc/src/sql/gen_expr.rs
+++ b/prqlc/prqlc/src/sql/gen_expr.rs
@@ -160,7 +160,8 @@ fn process_null(name: &str, args: &[Expr], ctx: &mut Context) -> Result<sql_ast:
 fn process_array_in(expr: &Expr, args: &[Expr], ctx: &mut Context) -> Result<sql_ast::Expr> {
     match args {
         [col_expr @ Expr {
-            kind: ExprKind::ColumnRef(_)
+            kind:
+                ExprKind::ColumnRef(_)
                 | ExprKind::Literal(_)
                 | ExprKind::SString(_)
                 | ExprKind::Param(_)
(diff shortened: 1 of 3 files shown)


max-sixty:
@kgutwin very nice fix!

Merging now — lmk if you want a release soon, was planning to do one in a few days, can push up if needed.

Great if you can add a changelog in another PR, no stress tho

max-sixty:
(+ if you can run `cargo fmt` then great, otherwise I can)

prqlc/prqlc/tests/integration/sql.rs
```diff
@@ -1155,6 +1155,8 @@ fn test_in_values_01() {
     from employees
     filter (title | in ["Sales Manager", "Sales Support Agent"])
     filter (employee_id | in [1, 2, 5])
+    filter (f"{emp_group}.{role}" | in ["sales_ne.mgr", "sales_mw.mgr"])
+    filter (s"{metadata} ->> '$.location'" | in ["Northeast", "Midwest"])
```
PrettyWood, line 1159:
```suggestion
    filter (s"{metadata} -> '$.location'" | in ["Northeast", "Midwest"])
```
?

max-sixty: Ah, good point. 

I thought the tests had passed, but actually the `cargo fmt` test failed before these had a chance to run. So when we fix `fmt`, we'll get to test this properly.

It will need a run of `task test-rust` to update the snapshots (or `cargo insta test --accept --features=default,test-dbs`)

kgutwin: @PrettyWood I think `->>` is the right operator here, since `in ["", ...]` is comparing to a VARCHAR which is what `->>` returns. Reference: https://duckdb.org/docs/extensions/json.html#json-extraction-functions

max-sixty: Don't need to delay merging, but FYI adding this to a query in this path https://github.com/PRQL/prql/blob/76c4f767c4312976dba0fbadfa6c27b84e31faa9/prqlc/prqlc/tests/integration/queries/cast.prql would test on duckdb & postgres DBs in CI

prqlc/prqlc/src/sql/gen_expr.rs
```diff
@@ -173,7 +178,10 @@ fn process_array_in(args: &[Expr], ctx: &mut Context) -> Result<sql_ast::Expr> {
                 .collect::<Result<Vec<sql_ast::Expr>>>()?,
             negated: false,
         }),
-        _ => panic!("args to `std.array_in` must be a column ref and an array"),
+        _ => Err(
+            Error::new_simple("args to `std.array_in` must be an expression and an array")
```
max-sixty, line 182:
```suggestion
            Error::new_simple("args to `std.array_in` must be an expression and an array respectively")
```

max-sixty:
Thanks @kgutwin !

PR #4513, opened 2024-05-29
devops: Switch `setup-dev` to using `brew` for `pre-commit`

max-sixty:
As discussed on Discord: https://discord.com/channels/936728116712316989/1245267066823508041

### DIFF
--- a/Taskfile.yml
+++ b/Taskfile.yml
@@ -115,7 +115,8 @@ tasks:
     status:
       - which maturin
     cmds:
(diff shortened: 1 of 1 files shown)


max-sixty:
Actually `brew install maturin` installs rust outside of `rustup` which causes the build to break! So instead this installs `brew install pipx` and then uses pipx, which is the recommended installer for maturin. It will check if maturin exists elsewhere and not attempt to install if it already exists.

PR #4517, opened 2024-05-31
fix: Can't run playground or book from DevContainer

### LINKED ISSUE #4482: Dev Container: can't run playground or book (can run website)
### What happened?

Using PRQL commit 9095495 in a Docker container on macOS on Intel (latest Docker: 4.30.0):

With the update to MSRV, I decided to rebuild. I removed all images, volumes, and containers related to PRQL. I then rebuilt the container. `task web:run-website` works exactly as expected. However both these commands give problems:

* `task web:run-playground` and
* `task web:run-book`

Both tasks seem to run in the terminal window as usual, however, going to the associated web page only displays a blank page. The browser looks as if it's waiting indefinitely. 

Using `curl` shows that no response comes from the web server (see below). What other diagnostic info could I provide?

**Update:** Using `curl` from a terminal window _within_ VSCode works as expected for both the playground and book. 

```
?130 ExceLint-addin % curl -v  http://localhost:5173/playground/playground
*   Trying [::1]:5173...
* connect to ::1 port 5173 failed: Connection refused
*   Trying 127.0.0.1:5173...
* Connected to localhost (127.0.0.1) port 5173
> GET /playground/playground HTTP/1.1
> Host: localhost:5173
> User-Agent: curl/8.4.0
> Accept: */*
>
```

### PRQL input

```elm
N/A
```

### SQL output

```SQL
N/A
```

### Expected SQL output

```SQL
N/A
```

### MVCE confirmation

- [ ] Minimal example
- [X] New issue

### Anything else?

_No response_

richb-hanover:
The problem was that the DevContainer became sensitive to the difference between 'localhost', '127.0.0.1' and '0.0.0.0'
The fix is to force the components to bind to 0.0.0.0

### DIFF
--- a/web/Taskfile.yaml
+++ b/web/Taskfile.yaml
@@ -24,13 +24,13 @@ tasks:
     desc: Build & serve the static website for interactive development.
     dir: website
     cmds:
-      - hugo server
+      - hugo server --bind 0.0.0.0
 
   run-book:
     desc: Build & serve the book for interactive development.
     dir: book
     cmds:
-      - mdbook serve --port=3000
+      - mdbook serve --port=3000 -n 0.0.0.0
 
   run-playground:
     desc: Build & serve the playground for interactive development.
--- a/web/playground/package.json
+++ b/web/playground/package.json
@@ -42,7 +42,7 @@
   "type": "module",
   "private": true,
   "scripts": {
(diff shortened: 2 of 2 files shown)


max-sixty:
Nice, thanks!

PR #4518, opened 2024-06-01
feat: PRQL codegen associativity

aljazerzen:


max-sixty (approved):
...and nice code comments, easy to follow with examples — thanks!

PR #4522, opened 2024-06-03
style: Sort imports

max-sixty:
Currently our imports are kinda all over. This uses an unstable feature to sort them into groups. While it's not possible to run this on every commit, hopefully having them in groups makes it easier to maintain.

To re-run:

```sh
echo 'group_imports = "StdExternalCrate"' > rustfmt.toml
cargo +nightly fmt
```

...and when the feature is made stable, then we can enable it...

### DIFF
--- a/lutra/bindings/python/src/lib.rs
+++ b/lutra/bindings/python/src/lib.rs
@@ -1,9 +1,9 @@
 #![cfg(not(target_family = "wasm"))]
 
-use arrow::{pyarrow::PyArrowType, record_batch::RecordBatch};
-use itertools::Itertools;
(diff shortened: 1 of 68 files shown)


max-sixty:
(inspired by @m-span's sorting, feedback welcome)

max-sixty:
I'll merge since I don't think there's much downside. If there are merge conflicts, accepting both and running the command above + `cargo clippy` should sort everything out (let me know if not, happy to help)

m-span:
I occasionally sort imports in RustRover using a built in tool (`Code` > `Optimize Imports`), but I prefer your approach here to use rustfmt

PR #4533, opened 2024-06-03
feat: Output detailed lineage graph

kgutwin:
This PR adds a new debug CLI command: `prqlc debug lineage`.

When provided PRQL input, it produces JSON or YAML formatted output with:

* `frames`: A list of `Span` and `Lineage` records corresponding to each "frame" in the main pipeline.
    * Note: This is unchanged from the original `FrameCollector` implementation.

* `nodes`: A list of expression graph nodes. Each node corresponds to an expression generated through the resolve process, and can be linked to other nodes through both `target` relationships (which denote data flow) and `parent`/`children` relationships (which reflect the original query structure).

* `ast`: The parsed PL AST.

The `frames` columns can be linked to target nodes in the `nodes` set. Those nodes can be then linked to other nodes, ultimately to source tables, literals, or other expressions. This allows algorithms to walk this graph and determine the source of each column in a PRQL output.

---

Example output for the simple PRQL query `from invoices | select { id, fixed = 15 }`:

<details><summary>Example</summary>
<p>

```yaml
frames:
- - 1:16-41
  - columns:
    - !Single
      name:
      - invoices
      - id
      target_id: 120
      target_name: null
    - !Single
      name:
      - fixed
      target_id: 121
      target_name: null
    inputs:
    - id: 118
      name: invoices
      table:
      - default_db
      - invoices
nodes:
- id: 118
  kind: Ident
  span: 1:0-13
  ident: !Ident
  - default_db
  - invoices
  parent: 123
- id: 120
  kind: Ident
  span: 1:25-27
  ident: !Ident
  - this
  - invoices
  - id
  targets:
  - 118
  parent: 122
- id: 121
  kind: Literal
  span: 1:37-39
  alias: fixed
  parent: 122
- id: 122
  kind: Tuple
  span: 1:23-41
  children:
  - 120
  - 121
  parent: 123
- id: 123
  kind: 'TransformCall: Select'
  span: 1:16-41
  children:
  - 118
  - 122
ast:
  name: Project
  stmts:
  - VarDef:
      kind: Main
      name: main
      value:
        Pipeline:
          exprs:
          - FuncCall:
              name:
                Ident: from
              args:
              - Ident: invoices
          - FuncCall:
              name:
                Ident: select
              args:
              - Tuple:
                - Ident: id
                - Literal:
                    Integer: 15
                  alias: fixed
    span: 1:0-42
```

</p>
</details> 

---

I'm trying to put this feature together because our current project requires tracking custom column-level metadata through PRQL transformations. Because the metadata can change as a result of specific transformation functions, we need the ability to track the source of columns to specific nodes within the AST. While the resolving process does generate a new set of expressions that are simplified from the original AST, it's possible to use the span information to link the resolved expression graph back to the original source and achieve our goal.


kgutwin:
Since the output of `debug lineage` is a bit hard to read, I put together a Python script that converts the output into a GraphViz visualization. For the example PRQL input:

```prql
from employees
filter start_date > @2021-01-01
derive {
  gross_salary = salary + (tax ?? 0),
  gross_cost = gross_salary + benefits_cost,
}
filter gross_cost > 0
group {title, country} (                      # `group` runs a pipeline over each group
  aggregate {                                 # `aggregate` reduces each group to a value
    avg_gross_salary = average gross_salary,
    sum_gross_cost = sum gross_cost,          # `=` sets a column name
  }
)
filter sum_gross_cost > 100_000
derive id = f"{title}_{country}"
derive country_code = s"LEFT({country}, 2)"
sort {sum_gross_cost, -country}
take 1..20
```

you can get the visualization:

trace

The Python script is pretty rough now but I could clean it up if posting it somewhere would be appropriate. I thought it would be ideal to roll that functionality into the `debug lineage` command, but that's pushing my Rust skills a bit farther than I have time for right now 😄

kgutwin:
I have a couple of questions before I think this PR is ready:

1. Do you agree that this is a "debug" feature? I think so, because I still think it's a bit rough around the edges, but I wasn't sure if this was on the roadmap in the way that it should be promoted to a more user-visible feature.
2. Related to question 1, can this be added to the Python/JS bindings? This is actually very important to my use case. If it's still at the "debug" feature stage, it may be a little confusing to have an API call for this feature to appear next to the main APIs, so should it be separated off with a name prefix or something similar?
3. How do you recommend this be documented? I'm happy to add docs to the PR but I don't see where the debug CLI commands are documented. Any recommendations?

max-sixty:
The graph looks extremely cool!

Overall the code looks good. Addressing the specific questions:

> Do you agree that this is a "debug" feature? I think so, because I still think it's a bit rough around the edges, but I wasn't sure if this was on the roadmap in the way that it should be promoted to a more user-visible feature.

Yes I think that's right. It seems robust enough we can put it under `debug` rather than `beta`. Easy to promote to top-level at some point.

> Related to question 1, can this be added to the Python/JS bindings? This is actually very important to my use case. If it's still at the "debug" feature stage, it may be a little confusing to have an API call for this feature to appear next to the main APIs, so should it be separated off with a name prefix or something similar?

I do think having a small API surface for the bindings is important, since those changing has a big impact. 

We could have a `beta` module in the API? Or just add a label to the functions stating that it's experimental? I don't think we need to be too precious if we clearly signpost it.

If you want to get something working for yourself quickly, obv reasonable to shell out `subprocess.run(f"prqlc debug ...")`. But much less robust than actual bindings, and can't be distributed in the same way.

> How do you recommend this be documented? I'm happy to add docs to the PR but I don't see where the debug CLI commands are documented. Any recommendations?

Adding to the docstring of the CLI structs is best, and then it appears in `prqlc foo --help`. Those are inconsistently documented at the moment. Having the docs "dense" with the code & its usage is good, so that's better than a separate page somewhere.

Would it be possible to put the command which generates the graphviz in there? Would be great for others to be able to do that. I could also imagine us building that out even more.

vanillajonathan:
Cool graph, you could add some colors to it.

kgutwin:
Ready for review and consideration for merging!

I cleaned up the GraphViz script and posted it as a gist here: https://gist.github.com/kgutwin/efe5f03df5ff930d899249018a0a551b

I'm not sure if you would prefer this script to be added to the repository itself... it seems more like a "contrib" item than anything else, and I don't know if you have a policy around "contrib" content. If you'd like it to go into the main repo, just let me know where you would recommend.

I did also add color to the output, as well as a small "legend" to explain some of the node shapes and line types. Here's the new output:

trace

prqlc/prqlc/src/cli/mod.rs
```diff
@@ -189,7 +190,41 @@ enum DebugCommand {
...
+    /// * "targets": Any upstream sources of data for this expression, as a list
+    ///   of node IDs (optional).
+    ///
+    /// * "children": A list of expression IDs contained within this expression
+    ///   (optional).
+    ///
+    /// * "parent": The expression ID that contains this expression (optional).
+    ///
+    /// A Python script for rendering this output as a GraphViz visualization is
+    /// available at https://gist.github.com/kgutwin/efe5f03df5ff930d899249018a0a551b.
```
max-sixty, line 227: Excellent docs!

And thanks for the gist link. I hadn't realized it was so much code, think the gist is v reasonable

prqlc/prqlc/src/lib.rs
```diff
@@ -448,6 +448,42 @@ impl<S: ToString> From<S> for SourceTree {
...
+        fc.ast = ast;
+
+        Ok(fc)
+    }
+
+    pub mod json {
+        use super::*;
+
+        /// JSON serialization of FrameCollector lineage
+        pub fn from_lineage(
```
max-sixty, line 475: Given these are public functions, could we add a very quick test to them? It can just be a few lines. I know they're tested in the python tests, but those are run less often, and we should really try not to break these given they're exported...

max-sixty (approved):
Really nice, thanks a lot.

I added one very small request, then good to merge!

max-sixty:
> I'm not sure if you would prefer this script to be added to the repository itself... it seems more like a "contrib" item than anything else, and I don't know if you have a policy around "contrib" content. If you'd like it to go into the main repo, just let me know where you would recommend.
> 
> I did also add color to the output, as well as a small "legend" to explain some of the node shapes and line types. Here's the new output:

No strong view — works fine as a gist atm, and then could incorporate if it gets wider use. (Could even imagine having a graphviz output from rust at some point...)

prqlc/bindings/prqlc-python/python/prqlc/__init__.py
```diff
@@ -0,0 +1,6 @@
+# ruff: noqa: F403, F405
+from .prqlc import *
+
+__doc__ = prqlc.__doc__
+if hasattr(prqlc, "__all__"):
+    __all__ = prqlc.__all__
```
max-sixty, line 8: (FYI this file seems a bit inelegant, but am guessing as a result of the python bindings. We could add a TODO to clean up if we're not confident we need this indirection)

prqlc/bindings/prqlc-python/src/lib.rs
```diff
@@ -50,17 +50,45 @@ pub fn rq_to_sql(rq_json: &str, options: Option<CompileOptions>) -> PyResult<Str
         .map_err(|err| (PyErr::new::<exceptions::PyValueError, _>(err.to_json())))
 }
 
+mod debug {
+    use super::*;
+
+    #[pyfunction]
+    pub fn prql_lineage(prql_query: &str) -> PyResult<String> {
```
max-sixty, line 57: If you have the patience we could also add a quick snapshot tests to these; there's an example with `compile` at the bottom. But not strictly necessary.

prqlc/prqlc/tests/integration/snapshots/integration__queries__debug_lineage__window.snap
```diff
@@ -0,0 +1,6 @@
...
...span":"1:999-1020","children":[195,196]}],"ast":{"name":"Project","stmts":[{"VarDef":{"kind":"Main","name":"main","value":{"Pipeline":{"exprs":[{"FuncCall":{"name":{"Ident":"from"},"args":[{"Ident":"tracks"}]}},{"FuncCall":{"name":{"Ident":"group"},"args":[{"Ident":"genre_id"},{"Pipeline":{"exprs":[{"FuncCall":{"name":{"Ident":"sort"},"args":[{"Ident":"milliseconds"}]}},{"FuncCall":{"name":{"Ident":"derive"},"args":[{"Tuple":[{"FuncCall":{"name":{"Ident":"row_number"},"args":[{"Ident":"this"}]},"alias":"num"},{"FuncCall":{"name":{"Ident":"count"},"args":[{"Ident":"this"}]},"alias":"total"},{"FuncCall":{"name":{"Ident":"last"},"args":[{"Ident":"track_id"}]},"alias":"last_val"}]}]}},{"FuncCall":{"name":{"Ident":"take"},"args":[{"Literal":{"Integer":10}}]}}]}}]}},{"FuncCall":{"name":{"Ident":"sort"},"args":[{"Tuple":[{"Ident":"genre_id"},{"Ident":"milliseconds"}]}]}},{"FuncCall":{"name":{"Ident":"select"},"args":[{"Tuple":[{"Ident":"track_id"},{"Ident":"genre_id"},{"Ident":"num"},{"Ident":"total"},{"Ident":"last_val"}]}]}},{"FuncCall":{"name":{"Ident":"filter"},"args":[{"Binary":{"left":{"Ident":"genre_id"},"op":"Gte","right":{"Literal":{"Integer":22}}}}]}}]}}},"span":"1:762-1021"}]}}
```
max-sixty, line 6: Possibly we should be using `to_string_pretty` so these are on multiple lines, but that's for another PR...

max-sixty:
Bah, would you mind running `cargo fmt`? 

Then please merge. Great PR!

(I added you with merge permissions...)

kgutwin:
Thank you so much for your review and helpful comments! I did the last `cargo fmt` and added a changelog entry. However, I don't see a way to merge, I'm not sure I have permissions. Can you merge for me? Thanks!

prqlc/bindings/prqlc-python/python/prqlc/__init__.py
kgutwin: When I added the `debug` submodule to the python bindings, I needed to update the type declarations as well... which led me towards needing to use Maturin's [mixed Rust/Python project](https://www.maturin.rs/project_layout#mixed-rustpython-project) layout. I then just copied Maturin's default `__init__.py` (that they provide when you have a pure Rust project) into that directory.

I am sure we could be more precise about the module init here, but given that it was working fine before (plus the type stubs are there, which are more descriptive anyway) I decided this was good for now. The TODO makes sense especially if more Python-native code will be added later.

Issue #3393, opened 2023-08-24
Add build information into build


max-sixty:
### What's up?

It could be useful to have the commit & date of a build in the comment, rather than just the version number.

IIUC [vergen](https://docs.rs/vergen/latest/vergen/index.html#vergen---emit-cargo-instructions-from-a-build-script) enables that, with a `build.rs`

If anyone would like an interesting addition which involves cargo, this could be a good one. It looks like it's not that much code. 

Probably we want lots of info on `prqlc --version` like [this example](https://docs.rs/vergen/latest/vergen/index.html#use-cases), and then in the comment in the SQL we want the commit hash in addition to the latest tag iff it's not on a tag.

eitsupi:
> IIUC [vergen](https://docs.rs/vergen/latest/vergen/index.html#vergen---emit-cargo-instructions-from-a-build-script) enables that, with a `build.rs`

I think I looked into this a long time ago, but I suspect that if we have cache enabled this may not work as well as we think.

max-sixty:
> I suspect that if we have cache enabled this may not work as well as we think.

Interesting.

It looks like it'll rerun the build (I think just for the specific crate) if the current `HEAD` changes:

> Will emit [cargo:rerun-if-changed=.git/HEAD](https://doc.rust-lang.org/cargo/reference/build-scripts.html#rerun-if-changed) if the git feature is enabled. This is done to ensure any git instructions are regenerated when commits are made.

So committing will rerun the build. That's not _too_ bad, I guess depending on often folks commit code locally. I think it's equivalent to making a no-op change to a file, which recompiling from takes 6.48s locally. (NB: compiling all our crates takes [40 seconds in CI](https://github.com/PRQL/prql/actions/runs/5978669246/job/16221209785#step:9:1), but doesn't take advantage of incremental builds)

If that's correct, and that's all the costs we'd bear (or am I missing others?), I'd say it's worthwhile, but not a particularly strong view.

haydenflinner:
Closed by #4804 ? Found via the linked closed PR above, #4546

PR #4546, opened 2024-06-06
feat: add build information to build comment

m-span:
Adds feat. requested in #3393 

example output:

```elm
SELECT
  b
FROM
  a

-- Generated by PRQL compiler version:0.11.3-230-ga6218a6f (https://prql-lang.org)
```

On my M1 pro, it appears to add ~0.5s to the re-build process, bringing the total to somewhere between 3.6 and 5.7 seconds


m-span:
hmm CI tests are failing in an interesting way:

```
thread 'main' panicked at prqlc/prqlc/src/lib.rs:124:26:
Invalid prqlc version number: c91cacc: Error("unexpected character 'c' while parsing major version number")
```

I don't know how `git describe --tags` is only outputting `c91cacc`, especially since that hash commit doesn't seem to exist.

Maybe there's something about how the CI is run that I don't understand? Is `git` available via command line within the CI?

max-sixty:
Nice!

Interesting re the git issue. I agree the commit doesn't seem to exist. But OTOH it does state here that that does seem to be the commit... https://github.com/PRQL/prql/actions/runs/9395042476/job/25873760775?pr=4546#step:2:56

I think it's probably creating a merge commit to simulate this branch being merged into main. And it's possibly doing a shallow clone so the tags aren't there.

Do you know if we can change the clone to be less shallow?

Either way we should make this robust to not having tags there — e.g. fall back to an empty string, I think.

max-sixty:
Yes seems like we need something like: https://github.com/actions/checkout/issues/217#issuecomment-900923602

max-sixty:
No obligation to use vergen, but:

- when I run this I don't get the version — is there a different check you're running?

```
cargo run -p prqlc -- --version
    Finished dev [unoptimized + debuginfo] target(s) in 0.07s
     Running `target/debug/prqlc --version`
prqlc 0.11.5
```

- are we sure about the `rerun-if-changed` approach is correct? I think vergen tries to handle that correctly (but don't know detail)

prqlc/prqlc/build.rs
```diff
@@ -0,0 +1,14 @@
+use std::fmt::Error;
+use std::process::Command;
+// use vergen::EmitBuilder;
+
+pub fn main() -> Result<(), Error> {
+    let output = Command::new("git")
+        .args(["describe", "--tags"])
+        .output()
+        .unwrap();
```
eitsupi, line 9: Doesn't this mean that the build will fail if the git command is not available?

eitsupi:
I'm not sure how this works with downstream packages. What happens if I install from crates.io?

max-sixty:
> I'm not sure how this works with downstream packages. What happens if I install from crates.io?

It would need to fall back to just the version number

prqlc/prqlc/build.rs
aljazerzen: Yes, this will be a problem on some build platforms without git or if the build process strips git repo out of source code.

aljazerzen (commented):
I think this is unnecessary for released versions of prqlc, as version number already describes git tag, which gets us the git rev.

It is useful for development builds. If you also think so, I'd suggest we:
- add git commit hash only if current commit does not have a git tag,
- silently skip this if git is not installed,
- silently skip this if git repo is not found.

max-sixty:
No urgency, but just FYI I thought this was cool and would be nice to have!

max-sixty:
Closed by #4804 (added us both onto changelog @m-span , hope that's reasonable)

PR #4560, opened 2024-06-08
feat: Add syntax highlight file for GNU nano

vanillajonathan:
This PR adds a syntax highlight file for the [GNU nano](https://nano-editor.org/) text editor.


max-sixty:
Nice!

PR #4568, opened 2024-06-09
ci: Add test for file listing in release-python

max-sixty:


max-sixty:
I think finally fixed by #4577

PR #4569, opened 2024-06-09
build: Set version on Snap

vanillajonathan:
This replaces the `version` field in the `snapcraft.yaml` manifest file with the `github.ref_name` variable from the GitHub Action context.

This uses the name of the tagged release instead of the Git commit hash as version.

### DIFF
--- a/prqlc/prqlc/Cargo.toml
+++ b/prqlc/prqlc/Cargo.toml
@@ -126,8 +126,8 @@ name = "bench"
 [[package.metadata.release.pre-release-replacements]]
 exactly = 1
 file = "../packages/snap/snapcraft.yaml"
-search = "^version: '[\\d.]+'$"
-replace = "version: '{{version}}'"
+search = '^version: "[\d.]+"$'
+replace = 'version: "{{version}}"'
 
 [[package.metadata.release.pre-release-replacements]]
 exactly = 1
--- a/prqlc/prqlc/Cargo.toml
+++ b/prqlc/prqlc/Cargo.toml
@@ -126,7 +126,7 @@ name = "bench"
 [[package.metadata.release.pre-release-replacements]]
 exactly = 1
 file = "../packages/snap/snapcraft.yaml"
-search = "^version: (git|'[\\w.\\-]+')$"
+search = "^version: '[\\d.]+'$"
 replace = "version: '{{version}}'"
 
 [[package.metadata.release.pre-release-replacements]]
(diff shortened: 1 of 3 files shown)


max-sixty:
Thanks!

Though should we instead use the replacement approach from here: https://github.com/PRQL/prql/blob/6e5db30646e3255a3da354ae2c997f0048c540a6/prqlc/prqlc/Cargo.toml#L126-L130

Then we can observe the version number and everything uses the same approach.

You can test locally with `cargo release replace -x --no-confirm`, from https://github.com/PRQL/prql/blob/ff492a5a98dff52c429a39607eaa185111fdbead/web/book/src/project/contributing/development.md#L489

vanillajonathan:
I think that would commit the `snapcraft.yaml` file though, but maybe that is okay?

max-sixty:
> I think that would commit the `snapcraft.yaml` file though, but maybe that is okay?

Yes! We have the version numbers in lots of files — `Cargo.toml`, `package.json` , etc

vanillajonathan:
Okay, I now do the replacement in `Cargo.toml`.

prqlc/prqlc/Cargo.toml
```diff
@@ -123,6 +123,12 @@ name = "bench"
 # refer to the path relative to `prqlc`, so we choose this crate out of
 # default.
 
+[[package.metadata.release.pre-release-replacements]]
+exactly = 1
+file = "../packages/snap/snapcraft.yaml"
+search = "^version: (git|'[\\w.\\-]+')$"
```
max-sixty: It'll always be the version number; never `git`, so we could adjust this.

And we should set the current version in the snap file!

So same approach as we follow in all the other version definitions, hope that makes sense. Thanks

vanillajonathan: Yeah, we can do that.

In theory it should be possible to use `git` as  version, I think the reason the version is `0` is because the GitHub Action [`action/checkout`](https://github.com/actions/checkout) only fetches the last commit and does not fetch any tags. It has a `fetch-tags` option that can be set to `true` but the `fetch-depth` option has the default value `1` but if set to `0` it will fetch all history for all branches and tags but that may take time on larger repositories.

max-sixty: Do you see what we have in Cargo files now? We have the actual version. So no replacement needs to happen in CI, it's all done locally and pushed.

So we need to have the current version in the snap file, as part of this PR.

Does that make sense?

max-sixty: (zero rush but if what I'm saying is confusing, let me know (sometimes people say I can be confusing)! I'm happy to do the final change to this if that would be helpful. TY)

vanillajonathan: No, you're not confusing, its just me who was hesitating on which approach to go with.

max-sixty:
It's great, thank you!

PR #4571, opened 2024-06-09
ci: Update to Node20 actions

richb-hanover:
Rather than use my `richb-hanover/cargo` (which will forever lag the proper versions), switch to `clechasseur/rs-cargo` which seems to be in active maintenance. This clears up the lion's share of "Node 16 is deprecated..." warnings during builds.

@clechasseur was also kind enough to update `rustsec/audit-check@v1` to use Node 20 (it's `v1.4.1` now). However, it appears that the upstream hasn't published that new action. There's an [Issue in the repo](https://github.com/rustsec/audit-check/issues/22) requesting this.

Once that's complete, it looks as if all the warnings during nightly builds will have been eliminated.


max-sixty:
Super, thanks a lot @richb-hanover !

PR #4578, opened 2024-06-10
build: Rename `prql-js` to `prqlc-js`

max-sixty:
But I'd like to find a way of getting the npm upload to be `prqlc`, rather than `prqlc-js`. We manage this in python...

### DIFF
--- a/prqlc/bindings/js/package.json
+++ b/prqlc/bindings/js/package.json
@@ -1,5 +1,5 @@
 {
-  "browser": "dist/web/prql_js.js",
+  "browser": "dist/web/prqlc_js.js",
   "description": "JavaScript bindings for prqlc",
   "devDependencies": {
     "chai": "^5.0.0",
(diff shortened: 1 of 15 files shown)


prqlc/bindings/js/package.json
```diff
@@ -12,8 +12,8 @@
   ],
   "license": "Apache-2.0",
   "author": "PRQL team",
-  "main": "dist/node/prql_js.js",
-  "name": "prql-js",
+  "main": "dist/node/prqlc_js.js",
+  "name": "prqlc-js",
```
eitsupi: > But I'd like to find a way of getting the npm upload to be prqlc

I suppose this line should be `"name": "prqlc"`
<https://docs.npmjs.com/cli/v6/configuring-npm/package-json#name>

max-sixty: Oh nice, thanks. Do you think this takes care of:

> But I'd like to find a way of getting the npm upload to be `prqlc`, rather than `prqlc-js`. We manage this in python...

max-sixty: Updated — let's see!

max-sixty:
Looks like it works — let's see, I'm 30% chance it doesn't work, but fastest way to fix is to try...

PR #4582, opened 2024-06-10
fix: better testing of lineage, and fix small bug with lineage tracking

kgutwin:
This improves the lineage tests (makes the snapshots more legible and has more comprehensive Python binding tests).

As part of this, a small bug for span assignments was discovered and fixed regarding assigning spans for nested pipelines. 

This has the curious side effect of changing the "bad error message" for #3870 to be more specifically located to the `aggregate` portion of the failing query. I hope that this might help track down the underlying bug.

### DIFF
--- a/prqlc/bindings/prqlc-python/src/lib.rs
+++ b/prqlc/bindings/prqlc-python/src/lib.rs
@@ -215,16 +215,16 @@ mod test {
     #[test]
     fn debug_prql_lineage() {
         assert_snapshot!(
-            debug::prql_lineage(r#"from a"#).unwrap(),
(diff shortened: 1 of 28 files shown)


max-sixty (approved):
This is indeed better, thanks!

I also wondered whether we should use `to_json_pretty` in our json methods, so the result had nicer line breaks. This solves some of that problem by using YAML.

Separate point but — it's a lot of lines; possible we should try and consolidate some of our integration tests, or run just a couple for anything apart from SQL — and low priority.

kgutwin:
I know it's a lot of snapshot lines, but I think that if lineage ever becomes a "real feature", we're going to want to test it consistently against the full span of queries. In fact, having this full test suite was already really useful in tracking down this little lineage bug, because it only occurred in rare cases; because it showed up in a couple of queries, it made it much more feasible to pin it down.

Thanks for the review!

max-sixty:
> I know it's a lot of snapshot lines, but I think that if lineage ever becomes a "real feature", we're going to want to test it consistently against the full span of queries. In fact, having this full test suite was already really useful in tracking down this little lineage bug, because it only occurred in rare cases; because it showed up in a couple of queries, it made it much more feasible to pin it down.

OK great, that does make sense

PR #4588, opened 2024-06-10
fix(prqlc-python): add missing type annotation for pl_to_prql

kgutwin:
Forgot the type annotation for the latest binding. I wonder if it's possible to do an automated check on this...


max-sixty:
Yes we could add a mypy strict confirm...

kgutwin:
That would be a good first step -- although what would have been needed to have caught this flub is cross-checking between the pyo3 output and the `.pyi` type stub file. Mypy can't check that because it trusts the `.pyi` type stubs. 

It looks like pyo3 is already tracking this over at https://github.com/PyO3/pyo3/issues/2454 ...

max-sixty:
OIC — good point — so mypy checks the static python code. Makes sense.

And I notice we [already have `strict = true`](https://github.com/PRQL/prql/blob/1a3fc6fbee3b1d80e5a9e0210a4945ebdfa7611f/prqlc/bindings/prqlc-python/pyproject.toml#L45-L46)...

Issue #4595, opened 2024-06-11
using `in` with an empty array pattern produces invalid SQL


Globidev:
### What happened?

`in [] value` transpiles to `value in ()` which is not valid syntax (at least in postgresql)

### PRQL input

```elm
prql target:sql.postgres
from table
filter (field | in [])
```

### SQL output

```SQL
SELECT
  *
FROM
  "table"
WHERE
  field IN ()
```

### Expected SQL output

```SQL
SELECT
  *
FROM
  "table"
WHERE
  FALSE -- Or maybe some other smart alternative I don't know about
```

### MVCE confirmation

- [X] Minimal example
- [X] New issue

### Anything else?

Thanks for this wonderful project ❤️

max-sixty:
Thanks @Globidev . Indeed, seems like we should we working around that postgres issue. (Ideally postgres would allow an empty list!).

This would likely be a fairly small PR into https://github.com/PRQL/prql/blob/2f2c08b168e9f6894ff65e315da44c6a86e6237d/prqlc/prqlc/src/sql/gen_expr.rs#L159-L186 if anyone wants to give it a go.

PrettyWood:
Closed by #4598

PR #4598, opened 2024-06-11
fix: generate `FALSE` instead of `expr in ()` conditions

Globidev:
This is an attempt at fixing #4595

Disclaimer: Pretty fluent in Rust but I just discovered the codebase, so I probably missed a bunch of things. 
I guess we'll want to add tests, and I'll likely need some directions as to where exactly to put them :)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,8 @@
 
 **Fixes**:
 
+ - using `in` with an empty array pattern (e.g. `expr | in []`) will now output a constant `false` condition instead of an `expr IN ()`, which is syntactically invalid in some SQL dialects (@Globidev, #4598)
+
 **Documentation**:
 
 **Web**:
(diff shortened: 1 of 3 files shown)


max-sixty:
Awesome @Globidev ! Nice work.

We can add a test next to https://github.com/PRQL/prql/blob/6b4c4389cff35b04296f234e5639cd45be401087/prqlc/prqlc/tests/integration/sql.rs#L1213-L1226

Then running `cargo insta test --accept` (or `task test-rust`) will fill in the generated SQL.

max-sixty:
A changelog entry would be nice but not strictly required

Globidev:
Thanks a bunch for the pointers!

Feel free to point out anything that would be missing or need tweaking

max-sixty:
Super, thanks @Globidev !

PR #4609, opened 2024-06-14
ci: Enable automatic bumping of rust toolchain

max-sixty:


max-sixty:
Going to see whether this works, probably 40% it runs when it should and not when it shoudn't. Fastest way is to just merge though.

PR #4622, opened 2024-06-16
docs: Add link to twitter badge

max-sixty:


vanillajonathan:
Should this use the x.com domain?

max-sixty:
Would be fine too change it!

Issue #4603, opened 2024-06-12
Gradually move `prlqc-ast` code into other crates


max-sixty:
### What's up?

As discussed with @aljazerzen & @m-span (though not necessarily agreed yet, feedback welcome), `prqlc-ast` isn't quite pulling its weight as a "cross-compiler external interface".

I'm very keen to continue to make the compiler more modular _by stage_. But this crate doesn't do that — instead it adds context to understanding a particular stage.

So I think it makes sense to:
- At least move into `prqlc-parser` the items which only have `prqlc-parser` as a dependent. For example [`span.rs`](https://github.com/PRQL/prql/blob/28fe4a5c16c5bcef19c07f1d5033326c7fe34fd8/prqlc/prqlc-ast/src/span.rs) (or `token.rs`, which I recently combined with similar code)
  - (technically `prqlc` also depends on `Span` — but only because it's used in `prqlc-parser`, I'm counting that as "only have `prqlc-parser` as a dependent", hope that's not too confusing...)
- Possibly do more — e.g. we have some code in [`generic.rs`](https://github.com/PRQL/prql/blob/86527e28fd01f9d7bf6abcc2894a576f7f03176d/prqlc/prqlc-ast/src/expr/generic.rs) which lots of crates depend on — we could consider also dissolving that. But we can wait to see what's left

Keen not to just continually refactor the library (meme below), but in this case I do think it would make it simpler and is worthwhile.

<img src="https://github.com/PRQL/prql/assets/5635139/f1a2dd65-ea7a-4e41-aa7f-84cd99686d91" width="300" />

m-span:
I have some ideas on how to handle this, but as I understand, @aljazerzen is in the middle of a refactor. I've generally found it bad practice to have more than one refactor being worked on at the same time, as the second to be merged in will have a hell of a bad time with it. Ironically, if the organization @max-sixty proposes is in place, this wouldn't be much of an issue.

@aljazerzen what is the state of your refactor, and how far does it encompass? Can you give me an idea of what you are/are not touching so I can respect the work you've done so far?

m-span:
Another related question:
 
```ascii
            PRQL

   (parser) │ ▲
 prql_to_pl │ │ pl_to_prql
            │ │
            ▼ │      json::from_pl
                   ────────►
           PL AST            PL JSON
                   ◄────────
            │        json::to_pl
            │
 (resolver) │
   pl_to_rq │
            │
            │
            ▼        json::from_rq
                   ────────►
[...]

PR #4634, opened 2024-06-18
refactor!: prqlc-parser major reorg changes, remove prqlc-ast

m-span:
From #4603 , we have the following new terminology:

| stage	|sub-stage	|AST
|----|----|----|
 |parse	|lexer	|string ->  _LR — Lexer Representation_	
 |parse	|parser	|tokens -> _PR — Parser Representation_	
 |semantic|	ast_expand	|PR -> _PL — Pipelined Language_	
 |semantic	|resolver	|PL	
 |semantic	|flatten	|PL	
 |semantic	|lowering	|PL -> _RQ — Resolved Query_	
 |sql	|preprocess	|RQ	
 |sql	|srq-compiler	|RQ ->  _PQ — Partitioned Query_	
 |sql	|postprocess	|PQ
 |sql	|sql-compiler	|PQ -> sqlparser::ast
 |sql|	codegen|	sqlparser::ast -> string

Goals:

- [x] Move all files from `prqlc-ast` to `prqlc-parser`. `prqlc-ast` is to be deprecated
- [x] Within parser stage, rename objects, modules and methods to meet new spec. Organize lexer and parser stage into separate modules.


m-span:
still a wip, need to make sure all references are updated correctly, but feel free to take a look and comment at how `prqlc-parser` is now organized

max-sixty:
Looking great at first glance!

(Unfortunately it looks like it may not be based off the latest `main` and so have some conflicts; worth pulling the latest main... really hope that's not a painful merge. To the extent it's caused by https://github.com/PRQL/prql/pull/4522, there are instructions there for how to reproduce that; so running the same commands on this branch should produce a very similar, easily-mergable result)

max-sixty:
Super minor but possibly standards are for `pr` to be lowercase; no view from me beyond following the rust naming standards

m-span:
There's more work to be done, but this is a good first step. We might need to talk about how we want to be re-exporting our imports.
These changes are absolutely Major version number update-API-breaking changes, since we are re-naming our objects.

m-span:
I tried to do the least amount of changes that accomplishes the goal, but its still a lot. Requesting many eyes on this. Also, some test are suddenly failing - trying to figure it out, but if anyone has any solutions, feel free to suggest or push.

prqlc/prqlc-parser/src/lib.rs
```diff
@@ -1,35 +1,25 @@
...
 pub use self::lexer::TokenVec;
-use crate::err::error::ErrorSource;
+
+pub mod error;
+pub mod lexer;
+pub mod parser;
+pub mod span;
+
+mod test;
+// pub use crate::err::*;
```
max-sixty: Super minor and let's not delay merging — but IIUC the rust standard is putting `mod`s above `use`s; the PR seems to reverse those.

(another one where I don't have a view apart from doing the standard thing — so if I'm wrong on the standard thing then let's do the correct standard thing :) )

max-sixty:
> I tried to do the least amount of changes that accomplishes the goal, but its still a lot. Requesting many eyes on this. Also, some test are suddenly failing - trying to figure it out, but if anyone has any solutions, feel free to suggest or push.

Yes perfect. I have a few small things to merge, and possibly a fairly big thing in #4639, so great if we can do these in smallish PRs with smaller conflicts.

A couple test failures indeed look surprising. Sometimes I realize I'm importing the wrong `Expr` etc; so I guess that's possible here?!

(a couple are easy clippy ones, will probably auto-fix with `clippy --fix`)

m-span:
there seems to be issues with the `ast_code_matches.rs` tests. What is the purpose of these tests? They don't seem to tolerate refactors well...

max-sixty:
> there seems to be issues with the `ast_code_matches.rs` tests. What is the purpose of these tests? They don't seem to tolerate refactors well...

Funny you asked, I actually just pushed a change removing them. (They existed as a compromise to having multiple copies of similar AST structs around...)

prqlc/prqlc-parser/src/parser/pr/mod.rs
```diff
@@ -0,0 +1,21 @@
+//this exports everything within as if it was at the root.
+//eg: PR::expr::Expr becomes PR::Expr
+pub use expr::*;
+pub use ident::*;
+pub use ops::*;
+pub use stmt::*;
+pub use types::*;
```
max-sixty, line 15: This is pretty aggressive exporting. No need to delay merge, but was this the case before? If not, would be good to trim it down at some point...

max-sixty: OK it was like this prior! Closing

prqlc/prqlc-parser/src/parser/pr/mod.rs
max-sixty, line 1: Not for this PR (this existed [before](https://github.com/PRQL/prql/pull/4634/files#diff-77aa79b7ef4fed0c42b75da58df055402450091a93c1ca7a8b5d96f92c34e5bc) — but exporting `foo::*` _and_ exporting the module `foo` seems prone to confusion and IIUC serves no purpose — guarantees there are two ways of accessing the same variable (I think?)

prqlc/prqlc-parser/src/error/mod.rs
```diff
@@ -227,121 +229,3 @@ impl<T, E: WithErrorInfo> WithErrorInfo for Result<T, E> {
         self.map_err(|e| e.with_source(source))
     }
 }
-
-#[cfg(test)]
-mod tests {
```
max-sixty, line 232: @m-span did we lose this? A quick ctrl-f doesn't get any hits, am I missing something?

prqlc/prqlc-parser/src/lexer/mod.rs
```diff
@@ -428,195 +473,3 @@ fn end_expr() -> impl Parser<char, (), Error = Cheap<char>> {
 
 #[derive(Serialize, Deserialize, Debug, Clone, PartialEq, Eq)]
 pub struct TokenVec(pub Vec<Token>);
-
-#[cfg(test)]
-mod test {
```
max-sixty, line 433: (checked this and this moved to `test.rs`)

prqlc/prqlc-parser/src/parser/pr/mod.rs
m-span: My first stab at this code involved rewriting all the imports and exports. However, this made the refactor twice as hard, so I tried to export things in a way that `prqlc-compiler` is expecting. That said, we could definitely optimize things.

One such question, for example, is `lexer::lr::Literal`. It is used in every subsequent AST. Right now, they are (almost) all referencing `lexer::lr::Literal`.
The other option is that each AST re-exports all the objects it uses. So, even though only one `Literal` exists, there will also be a `pr::Literal`, `pl::Literal,` `q::literal`

prqlc/prqlc-parser/src/error/mod.rs
m-span: i did some other minor refactors in parser while restructuring. Large file with tests -> tests in separate file

prqlc/prqlc-parser/src/lexer/mod.rs
m-span: 👍

prqlc/prqlc-parser/src/parser/pr/mod.rs
m-span: PR will reference `lr::Literal`... PL will only have to import PR and its `pr::Literal`, etc etc

prqlc/prqlc-parser/src/error/mod.rs
max-sixty: Sorry, GitHub wasn't showing the diff, found it when searching locally

prqlc/prqlc-parser/src/parser/pr/mod.rs
max-sixty: Yes good idea to push into another PR.

No strong view on exactly how that's done. I think the existing state is strictly worse than some other alternatives since it's duplicating the locations of everything without providing convenience, so definitely some improvements possible.

prqlc/prqlc-parser/src/lexer/lr/mod.rs
max-sixty, line 1: Same here — IIUC it would be strictly better to have only one of these — no strong view on which

prqlc/prqlc-parser/src/parser/pr/mod.rs
m-span: yep, another place for export optimization. Saving that for the next 'pass'

max-sixty:
I'm sure we'll find some more refinements but looking really good for the moment — will hit the big button.

I made some small-ish refinements, feel free to object to anything if I went against your intention...

prqlc/prqlc-parser/src/lib.rs
max-sixty: I made some changes for these

max-sixty:
(could definitely do with a changelog entry, doesn't need to be part of this PR tho)

m-span:
Looking good from this side. Cleanups were great. Just a small shame I coudln't get the refactor on `prqlc_parser::parse_source()` working that I had to undo it via [this commit](https://github.com/PRQL/prql/pull/4634/commits/3dcde938489289adb6c91d26722d8c39bfbbbb53).
It did some nice things like hide stuff that didn't need to be exposed

m-span:
> (could definitely do with a changelog entry, doesn't need to be part of this PR tho)

I'll add it.
Also, this is a breaking api change...should we increment the working version number to a major increment?

max-sixty:
> Also, this is a breaking api change...should we increment the working version number to a major increment?

Let's add `!` in the PR description and then we'll do it at release time

(your approach is reasonable but it may not generalize well to multiple breaking changes in a release without another mechanism...)

CHANGELOG.md
```diff
@@ -20,6 +20,11 @@
 
 **Internal changes**:
 
+- Major reorganization of `prqlc-parser`. Files have been re-arranged and
+  exports have changed
+
+- Deprecated `prqlc-ast`. Functionality and objects were moved to `prqlc-parser`
```
```suggestion
- Major reorganization of `prqlc-parser` — `prqlc-ast` is merged into `prqlc-parser`, and `prqlc-parser`'s files are rearranged, including its exports. This is part of an effort to modularize the compiler by stage, reducing the amount of context that's required to understand a single stage. There will likely be some further changes (more detail in the PR description). (@m-span, #4634)
  -  This is a breaking change for any libraries that use `prqlc-parser` which should be fairly rare.
```

(I actually wiped it `prqlc-ast`; don't think there's any need to keep it around...)

CHANGELOG.md
```diff
@@ -20,6 +20,16 @@
...
 
+- Major reorganization of `prqlc-parser` — `prqlc-ast` is merged into
+  `prqlc-parser`, and `prqlc-parser`'s files are rearranged, including its
+  exports. This is part of an effort to modularize the compiler by stage,
+  reducing the amount of context that's required to understand a single stage.
+  There will likely be some further changes (more detail in the PR description).
+  (@m-span, #4634)
+
+  - This is a breaking change for any libraries that use `prqlc-parser` which
+    should be fairly rare.
```
```suggestion
  - This is a breaking change for any libraries that depend on `prqlc-parser` (which
    should be fairly rare).
```

max-sixty:
Thank you @m-span !

PR #4638, opened 2024-06-20
docs: Add comment re `expr_call` duplication

max-sixty:
Not sure why this doesn't work. We don't necessarily need to merge this but pushing in case anyone knows offhand...


prqlc/prqlc-parser/src/expr.rs
```diff
@@ -27,6 +27,9 @@ pub fn expr() -> impl Parser<TokenKind, Expr, Error = PError> + Clone {
             .map(|x| x.to_string())
             .map(ExprKind::Internal);
 
+        // This simply inlines `expr_call`. But if we call `expr_call` here, we get a
+        // stack overflow — possibly because this is a recursive parser?
+        // let nested_expr = pipeline(expr_call().boxed()).boxed();
```
aljazerzen: It's not exactly an inline:
- the `expr` in this function is the reference to the "recursive" node,
- the `expr` in `expr_call()` is the whole expr parser, reconstructed.

aljazerzen: To be fair the is the most confusing part of the parser (and prql cfg).

It stems from the fact that some locations is PRQL allow "expression that can have function calls at top-level" and other locations allow only "expression that don't have function calls at top-level".

For example:
- `let a = ____` is the first case,
- `join ____ (owner==id)` is the second case. Here, the `___` cannot be expr with function calls at the top-level (for example `from db.mytable`) because that would also consume the last argument to `join`. It would parse as `(join (from db.mytable (owner==id))` instead of `(join (from db.mytable) (owner==id)`.

This stems from the fact that we don't require parenthesis for func calls or even commas.

PR #4645, opened 2024-06-21
refactor: clean up exports

m-span:
More cleanups, import organization, removal of symbol exports (pub use) where not necessary.


prqlc/prqlc-parser/src/lexer/mod.rs
```diff
@@ -10,12 +6,15 @@ use chumsky::{
 use serde::{Deserialize, Serialize};
 
 use crate::error::{Error, ErrorSource, Reason, WithErrorInfo};
-use crate::lexer;
 use crate::lexer::lr::{Literal, Token, TokenKind, ValueAndUnit};
 use crate::span::Span;
 
+pub mod lr;
+#[cfg(test)]
+mod test;
```
max-sixty, line 14: I think convention is to always put `mod` at the top, no?

prqlc/prqlc/src/sql/gen_query.rs
```diff
@@ -14,11 +14,11 @@ use super::gen_projection::*;
 use super::operators::translate_operator;
 use super::srq::ast::{Cte, CteKind, RelationExpr, RelationExprKind, SqlRelation, SqlTransform};
 use super::{Context, Dialect};
-use crate::ast::generic::InterpolateItem;
 use crate::ir::pl::{JoinSide, Literal};
 use crate::ir::rq::{CId, Expr, ExprKind, RelationLiteral, RelationalQuery};
 use crate::utils::{BreakUp, Pluck};
 use crate::{Error, Result, WithErrorInfo};
+use prqlc_parser::generic::InterpolateItem;
```
max-sixty, line 21: Yeah this sort of switch makes a lot of sense; too much indirection makes it confusing...

max-sixty (approved):
Great, thanks!

One comment re `mod` vs `use` ordering...

max-sixty:
FYI not sure what's going on with codecov, please ignore the error...

prqlc/prqlc-parser/src/lexer/mod.rs
m-span: ...apparently not. Both RustRover's Optimize Imports and cargo+nightly fmt give this ordering (I just checked)

max-sixty: https://doc.rust-lang.org/beta/style-guide/items.html

> use statements, and module declarations (mod foo;, not mod { ... }) must come before other items. Put imports before module declarations.

You're right! Weird, I think most projects I've worked on do the opposite. TIL. Let's do it

aljazerzen:
This is great.

Issue #4636, opened 2024-06-19
AST Explorer?


max-sixty:
### What's up?

It would be really nice to be able to explore our AST in something a bit more interactive / "zoom-able" than reading flat YAML. When working on `prqlc fmt` I would have found this v helpful. I briefly discussed this on the most recent dev call with @aljazerzen.

I checked out https://astexplorer.net/ — it's pretty nice — you can zoom in & out to different parts of the AST by unfolding & folding the tree; it uses highlights to show where the AST comes from in the code.

It doesn't support multiple stages — it would be even better to have columns with each stage and have the highlighting working across them. It also seems somewhat unmaintained now (https://github.com/fkling/astexplorer/issues/625), and I couldn't managed to get it to build on my local machine.

So:
- Is anyone familiar with any alternatives?
  - [LLVM Compiler Explorer](https://alive2.llvm.org/ce/) is another example, but very specific to LLVM
- Would anyone be up for either adding PRQL support to one of these?
  - We could possible try building our own if there are no alternatives; though it seems like a non-core part of the project, and would need to be in a web language, which we're less good at

aljazerzen:
These are all very specific to the AST that they are exploring. And this should not be too hard to build ourself.

I'm planning on doing that.

max-sixty:
> These are all very specific to the AST that they are exploring.

Some parts! Though I had thought the concept of "put your cursor in an editor and highlight the corresponding items in the AST graphs" seemed quite general...

snth:
Do all of the IRs have span fields with which one can link to the source code? 

---
Why does it have to be in a web language? Would a python tool be ok? You could potentially have it run in browser with pyodide or run serve a web page.

max-sixty:
> Do all of the IRs have span fields with which one can link to the source code?

Yes! At least LR, PR & PL do (though the latter aren't by default serialized, easy to change)



> Why does it have to be in a web language? Would a python tool be ok? You could potentially have it run in browser with pyodide or run serve a web page.

Sure, whatever works; I had thought that this sort of thing would be best in a browser and browsers were easiest in JS, but no view beyond that...

aljazerzen:
I've added "probes" that log data from the compilation to a "debug log" that can be serialized to either a json or html file. That file can be opened with a browser directly, so there is no need to have an HTTP server even.

0xdevalias:
[...]

PR #4646, opened 2024-06-21
feat: prqlc compile --debug-log

aljazerzen:
Adds a framework for capturing debug information during the compilation process.

First stab at #4636

The result can be serialized into JSON or rendered into HTML (image below). Current display is not much better than plain terminal output, but this has potential for better visualization of what is going on in the compiler.

image


prqlc/prqlc/src/debug.rs
```diff
@@ -0,0 +1,365 @@
...
+            let substage = entry
+                .stage
+                .sub_stage()
+                .map(|s| s.to_lowercase())
+                .unwrap_or_default();
+            writeln!(w, "<div class=\"entry {stage} {substage}\">",)?;
+
+            writeln!(
+                w,
+                "<input id=\"{entry_id}\" class=\"toggle\" type=\"checkbox\" checked>"
```
max-sixty:
```suggestion
                r#"<input id="{entry_id}" class="toggle" type="checkbox" checked>"#
```

nitty but this makes it easier to read and I guess write...

max-sixty (approved):
Looks great, thanks! This direction be really helpful I think.

---

Not a strong view at all but two other approaches to the logging infra:

1.
- just write structured logs in JSON to stderr / possibly an optional log life
- then have a function to convert to a different representation; e.g. HTML
...that's a bit more modular (though also requires some serialization from serde)

2.
- A specific command like `prqlc debug dump foo.prql` which dumps everything — my guess is that much of the time we'd want to drill down on a specific query, rather than always having some exhaust

...but as above, not a strong view

aljazerzen:
Actually, this PR is combination of the two things you've suggested.

1. `--debug-log=my_log.json` will create a JSON file with all of the AST reprs emitted. This file could probably be read back and rendered to the HTML output,
2. the debug log should contain basically all intermediate information that compiler has. The idea is to replace various commands we have with these "debug log entries", so a normal `compile` invocation produces all that you need. ATM, we are 80% there.

max-sixty:
> Actually, this PR is combination of the two things you've suggested.

Yes! My musing was whether splitting it up made sense (but really just a musing and maybe not _that_ useful..)

aljazerzen:
More refinement:
image

max-sixty:
I tried this out — really really cool! Very impressive you could pull together a UI so quickly!

---

FWIW when I try I don't get the highlights _across_ the different stages (like in your query) unless I select the whole query, but possibly that's because the spans aren't serialized yet.

---

Other small thoughts (obv you are the owner and so only consider these if they're useful, am strongly not trying to say these are necessarily good)
- integrate with playground? Maybe that's more work than it saves though. Mostly similar goals (would want a check-box to show / hide, similar to what we have with outputting PL & RQ now)
- minor but it would be v nice to have the stages in each _column_, so we can see across them — i.e. "what RQ comes from this expression"

aljazerzen:
> integrate with playground? Maybe that's more work than it saves though. Mostly similar goals (would want a check-box to show / hide, similar to what we have with outputting PL & RQ now)

Yes, this is possible to do. We just need to expose the --debug-log argument to JS bindings somehow. We'd probably change the type signature, so the debug log is returned instead of written into a file.

---

> minor but it would be v nice to have the stages in each column, so we can see across them — i.e. "what RQ comes from this expression"

This is easy to do! Newer versions are somewhat easier to parse visually, so it might not even be needed, but yes, can be done.

PR #4653, opened 2024-06-23
refactor: Split up giant `expr` func in parser

max-sixty:
Renames `Indirection` to `Property`, which seems like a more familiar name

### DIFF
--- a/.github/workflows/build-web.yaml
+++ b/.github/workflows/build-web.yaml
@@ -4,6 +4,7 @@ on:
   workflow_call:
   workflow_dispatch:
 
+# We need consistent env vars across all workflows for the cache to work
 env:
   CARGO_TERM_COLOR: always
   CLICOLOR_FORCE: 1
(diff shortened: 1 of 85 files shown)


aljazerzen (commented):
The split-up of the expr parsing I like, the renaming I don't.

This expr node is not really a property, it is expressing "an act of pulling a field out of a tuple" (it could also be pulling item out of an array). An alternative could be "FieldDereference" or "FieldLookup".

(I try to use name "field" when talking about tuples, instead of "property" or "column" or "item". Similarly, arrays have "items")

max-sixty:
> An alternative could be "FieldDereference" or "FieldLookup".

Perfect — going with `FieldLookup`. (`Indirection` sounded a bit like a magician :) )

aljazerzen:
Oh no, I've doubled-down on name indirection on my branch - is there an option to reconsider this change?

Indirection is [real compiler terminology](https://en.wikipedia.org/wiki/Indirection), I'm not making this up!

max-sixty:
> Indirection is [real compiler terminology](https://en.wikipedia.org/wiki/Indirection), I'm not making this up!

Totally fine to revert if it helps keep the code moving.

(I'm v familiar with the concept of indirection; I don't think it's a clear description for an AST _item_, but this is very minor, revert away...)

PR #4663, opened 2024-06-24
feat: replace env logger with debug log

aljazerzen:
Instead of logging debug messages to stderr, any `log::debug!` or `log::info!` or `log::trace!` will now log messages into the `debug-log.html`/`debug-log.json`.

image


max-sixty:
The track record of this is very good, so I defer to you even more than usual. :)

---

I would _generally_ raise an eyebrow at us having our own logging framework, relative to using something that's more standard, since we reimplement a bunch of things ourselves and it raises the bar a bit for newer contributors. FWIW a quick browse suggests there are some well-supported structured logging crates around, such as [tracing](https://github.com/tokio-rs/tracing?tab=readme-ov-file#in-applications) (+ [tracing-appender](https://crates.io/crates/tracing-appender) which allows writing for files). 

It is nice to have a single output of logs & stages. That would be _more_ difficult with my [(2) suggestion in #4646](https://github.com/PRQL/prql/pull/4646#pullrequestreview-2133185775) of writing the JSON out to a specific file rather than as exhaust, since at least logs are exhaust. So if this is the closest way of getting that, sobeit. If we really want, we can always switch out the bottom layer another time.

aljazerzen:
Yeah, I was hesitant to do this at first too, but it is not really a logging framework, just a logging backend - a writer to a different location.

I have to try this new development first and then decide if it is any good.

---

The great potential with thid debug log is that when all the logs are in .html, we even do filtering on the fly. For example, compile the query once, capture all levels and then filter to isolate only the ones that you need.

PR #4670, opened 2024-06-25
feat: improve debug log

aljazerzen:
- add logs of AST in sql backend
- remove debug commands whose functionality is now replaced with the debug log,
- cleanup internal API of the crate


max-sixty:
> remove debug commands whose functionality is now replaced with the debug log,

Yeah nice — regardless of the exact approach, definitely made sense to consolidate these into a "dump everything", and then separately filter logs on read...

PR #4671, opened 2024-06-25
feat: improve debug log in parsing stage

aljazerzen:
Since prqlc_parser has a single top-level function `parse_source`, we were not able to emit lr tokens in the debug log, since that is prqlc only instrumentation.

So I've moved the `parse_source` function to `prqlc::parser`, and now we have this:

image


max-sixty:
Very nice indeed!

PR #4677, opened 2024-06-26
build: Use `connector_arrow` re-export for arrow

max-sixty:
Re discussion at https://github.com/PRQL/prql/pull/4668 (arrow is already re-exported)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2117,7 +2117,6 @@ name = "lutra"
 version = "0.12.3"
 dependencies = [
  "anyhow",
- "arrow",
  "clap 4.4.18",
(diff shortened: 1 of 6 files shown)


max-sixty:
Actually I just saw:

> It turns out I already do, but but we need to depend on `arrow` anyways, because we need additional features that are not needed for `connector_arrow`.

...so maybe this isn't OK

aljazerzen:
> Actually I just saw:
> 
> > It turns out I already do, but but we need to depend on `arrow` anyways, because we need additional features that are not needed for `connector_arrow`.
> 
> ...so maybe this isn't OK

Exactly.

PR #4678, opened 2024-06-26
test: Skip item size check on Linux

max-sixty:
Notes inline (not an ideal solution, can just exclude if it fails again)


max-sixty:
Closes #4676

PR #4693, opened 2024-07-02
fix: relational literals with zero rows

### LINKED ISSUE #4421: Compiler crash on `from_text` with zero rows

aljazerzen:
Closes #4421


max-sixty:
Nice!

PR #4701, opened 2024-07-08
feat: Add DocComments to PR

max-sixty:
This takes much of the work from https://github.com/PRQL/prql/pull/4639 and uses it to push doc comments through to PL.

I'll leave a comment on something I'm stuck on; I can't work out why we get such a bad error message for `from artists #! This is a doc comment`: `Expected one of (, [, an identifier, keyword case, keyword internal or {, but didn't find anything before the end.`.

This needs lots more tests for doc comments specifically, and lots of clearing up.


max-sixty:
This ends up being a pretty big change! We change how we handle new lines, which lets us enforce new lines for some things — previously we were just consuming 0+ new lines almost anywhere, which then means there might not be a new line character remaining to consume when we come to check before an item that requires one.

---

I used lots of tests when trying to fix problems — ended up being a lot, and sometimes kinda complicated. So possibly the tests are a bit duplicated. We could do an effort to reduce the duplication, since there isn't a natural way for tests to reduce in number... But also not that costly to keep them.

PR #4705, opened 2024-07-08
internal: Try adding a label to a chumsky parser

max-sixty:
Moving this from #4700. Do we know why adding a label removes the error message? Is it something to do with merging multiple labels loses the labels themselves https://github.com/PRQL/prql/blob/63571263f667964c71d11ed001f9fc04a306b83b/prqlc/prqlc-parser/src/error/parse_error.rs#L281-L292 ?

Is this the correct design? I would think we want the most specific item, not sure whether that's possible.

(Tagging @m-span as he was the most recent author of the code, but no worries if you're not sure / or if this was the result of a refactor. And having `new line` isn't important per se, I mean it as a more general question)

### DIFF
--- a/prqlc/prqlc-parser/src/parser/common.rs
+++ b/prqlc/prqlc-parser/src/parser/common.rs
@@ -24,7 +24,7 @@ pub fn keyword(kw: &'static str) -> impl Parser<TokenKind, (), Error = PError> +
 }
 
 pub fn new_line() -> impl Parser<TokenKind, (), Error = PError> + Clone {
-    just(TokenKind::NewLine).ignored()
+    just(TokenKind::NewLine).ignored().labelled("new line")
 }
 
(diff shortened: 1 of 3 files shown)


m-span:
> Do we know why adding a label removes the error message? Is it something to do with merging multiple labels loses the labels themselves

That code is copied from the default Chumsky implementation. My best guess it is so that when an error propagates upwards, only the first, most specific error-label is kept.

m-span:
Still, I'll look into this case

max-sixty:
> That code is copied from the default Chumsky implementation. My best guess it is so that when an error propagates upwards, only the first, most specific error-label is kept.

Ah great, I see, it's explained here: https://github.com/zesterer/chumsky/pull/238

So it seems like we just need to add labels to the "higher" item so that's reported, rather than nothing. (I need to think through what counts as "higher"...)

> Still, I'll look into this case

Sure, if you want to take a look, but no obligation

PR #4710, opened 2024-07-09
refactor: Small reorg of parser error

max-sixty:
- Move the prqlc-wide `Error` our of `parser` path
- Collapse a small multifile `mod.rs` into a single file (think we should do more of this)


max-sixty:
Merging but let me know any feedback (@m-span am trying to continue your work rather than undo it, lmk if this doesn't seem accurate!)

PR #4714, opened 2024-07-10
tweak: Adjust name of repr in docstring

max-sixty:


prqlc/prqlc/src/parser.rs
```diff
@@ -52,7 +52,7 @@ pub fn parse(file_tree: &SourceTree) -> Result<pr::ModuleDef, Errors> {
     }
 }
 
-/// Build PRQL AST from a PRQL query string.
+/// Build PR AST from a PRQL query string.
```
vanillajonathan, line 55: Hmm?

max-sixty: Check out the recent discussions on making l naming each representation (am on phone but can find link if you can't find it)

PR #4721, opened 2024-07-11
internal: Add a `Start` token at the beginning of LR output

max-sixty:
If we want to enforce anything starting on a new line (for example a doc comment) in the parser, then we also need to handle the start of a file — which is semantically a new line but doesn't have a literal new line token. So we add this token and then will allow it to act as a new line in the parser...

### DIFF
--- a/prqlc/prqlc-parser/src/lexer/mod.rs
+++ b/prqlc/prqlc-parser/src/lexer/mod.rs
@@ -10,8 +10,16 @@ pub mod lr;
 #[cfg(test)]
 mod test;
 
+// TODO: we have `lex_source` and `lex_source_recovery` and don't have the same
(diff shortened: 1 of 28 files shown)


prqlc/prqlc-parser/src/lexer/mod.rs
```diff
@@ -10,8 +10,16 @@ pub mod lr;
 #[cfg(test)]
 mod test;
 
+// TODO: we have `lex_source` and `lex_source_recovery` and don't have the same
+// structure for LR. Probably we should have a single approach to the inclusion
```
max-sixty:
```suggestion
// structure for PR. Probably we should have a single approach to the inclusion
```

PR #4728, opened 2024-07-13
refactor: Remove `common` module

max-sixty:
These can just be in the `mod` and marked as non-pub


max-sixty:
I also narrowed the exposure of lots & lots of items, making them `pub(crate)`. As well as reducing the API surface, it also then highlights code which isn't used (since it can't be used by another lib); for example https://github.com/PRQL/prql/pull/4728/files#diff-e4c96f293b5492959b3fa5aaad7db0f449537af891a3c038d48207e9a45f63b5L18-L36

PR #4734, opened 2024-07-14
fix: Handle pipe characters preceeding `into` like a normal pipeline

### LINKED ISSUE #4291: `into` semantics as transform?
Thanks to @broneill — full details at https://github.com/PRQL/prql/discussions/4277#discussioncomment-8648606

I had been thinking of `into` as a transform — so following a pipe (either a `|` or new line).

Currently, quoting from the discussion linked above

valid, correctly so

```elm
from y
into x

from x
```

valid, but incorrectly so — I think should be invalid
```elm
from y into x  # should require `|`

from x
```

invalid, but incorrectly so — I think should be valid
```elm
from y | into x  # the `|` should make it valid

from x
```

max-sixty:
Error message not great but can be improved separately.

Closes https://github.com/PRQL/prql/issues/4291, I think without doing anything controversial

### DIFF
--- a/prqlc/prqlc-parser/src/parser/stmt.rs
+++ b/prqlc/prqlc-parser/src/parser/stmt.rs
@@ -5,7 +5,7 @@ use itertools::Itertools;
 use semver::VersionReq;
 
 use super::expr::{expr, expr_call, ident, pipeline};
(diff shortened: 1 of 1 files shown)


aljazerzen (approved):
I'm still to sure I like this from semantical perspective, but if there is a decent number of people who do, let's merge it.

PR #4738, opened 2024-07-15
refactor: Fix nightly lint

max-sixty:
Apparently `riid` isn't used, so nightly raises a compilation warning. Is it right to just remove it?

### DIFF
--- a/prqlc/prqlc/src/semantic/resolver/mod.rs
+++ b/prqlc/prqlc/src/semantic/resolver/mod.rs
@@ -26,8 +26,6 @@ pub struct Resolver<'a> {
 
     pub id: IdGenerator<usize>,
 
-    pub options: ResolverOptions,
-
     pub generics: HashMap<(usize, String), Vec<crate::pr::Ty>>,
 }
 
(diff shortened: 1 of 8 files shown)


prqlc/prqlc/src/semantic/resolver/mod.rs
```diff
@@ -26,19 +26,16 @@ pub struct Resolver<'a> {
 
     pub id: IdGenerator<usize>,
 
-    pub options: ResolverOptions,
-
     pub generics: HashMap<(usize, String), Vec<crate::pr::Ty>>,
 }
 
 #[derive(Default, Clone)]
 pub struct ResolverOptions {}
```
aljazerzen, line 33: Maybe we should remove this struct too?

I used to have some options, but it does not anymore, so it is currently useless.

prqlc/prqlc/src/sql/pq/context.rs
```diff
@@ -169,7 +167,6 @@ impl AnchorContext {
 
         let original_cids = table_ref.columns.iter().map(|(_, c)| *c).collect();
         let relation_instance = RelationInstance {
-            riid,
```
aljazerzen, line 172: riid (relational instance id) is not really needed in `RelationInstance`, because these structs are stored in `HashMap<RIId, RelationInstance>` anyway.

So this edit is ok.

aljazerzen (approved):
Great, yes.

PR #4747, opened 2024-07-16
refine: Remove lambda funcs from pipeline items

max-sixty:
I don't think this is needed? Tests pass. But possibly I'm missing something...


aljazerzen (commented):
If these parse, then it is fine, otherwise no:

```
{func a -> a}

[func a -> a]

(3 | func a -> a)
```

max-sixty:
Yes OK those don't pass. And I guess we do indeed want to have lambda funcs as exprs...

PR #4754, opened 2024-07-18
test: Add test & notes re multiple aliases

max-sixty:
Not sure what the best path forward is. It's creating some odd error messages (often a symptom of some poorly designed language semantics), but can't see an easy solution.

### DIFF
--- a/prqlc/prqlc-parser/src/parser/expr.rs
+++ b/prqlc/prqlc-parser/src/parser/expr.rs
@@ -399,22 +399,6 @@ where
         .map(Some)
         .then_ignore(ctrl(':'))
         .then(expr.clone());
-
-    // TODO: I think this possibly should be restructured. Currently in the case
-    // of `derive x = 5`, the `x` is an alias of a single positional argument.
-    // That then means we incorrectly allow something like `derive x = 5 y = 6`,
-    // since there are two positional arguments each with an alias. This then
-    // leads to quite confusing error messages.
-    //
-    // Instead, we could only allow a single alias per function call as the
-    // first positional argument? (I worry that not simple though...).
-    // Alternatively we could change the language to enforce tuples, so `derive
-    // {x = 5}` were required. But we still need to account for the `join`
-    // example below, which doesn't work so well in a tuple; so I'm not sure
-    // this helps much.
-    //
-    // As a reminder, we need to account for `derive x = 5` and `join a=artists
-    // (id==album_id)`.
     let positional_arg = maybe_aliased(expr).map(|expr| (None, expr));
 
     func_name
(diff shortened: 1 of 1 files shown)


prqlc/prqlc-parser/src/parser/expr.rs
```diff
@@ -744,15 +728,72 @@ mod tests {
...
+        // That then means we incorrectly allow something like `derive x = 5 y = 6`,
+        // since there are two positional arguments each with an alias. This then
+        // leads to quite confusing error messages.
+        //
+        // Instead, we could only allow a single alias per function call as the
+        // first positional argument? (I worry that's not simple though...).
+        // Alternatively we could change the language to enforce tuples, so `derive
+        // {x = 5}` were required over `derive x = 5`. But we still need to account
+        // for the `join a=artists (id==album_id)` which doesn't work so well in a
+        // tuple; so I'm not sure this helps much.
```
aljazerzen, line 744: How is `derive x=5 y=6` different from `derive 5 6`?

Both should ideally say "too many arguments", but it also fine if it is "6 has type int, but a relation was expected".

I don't think this is a problem with aliases here.

max-sixty: > How is `derive x=5 y=6` different from `derive 5 6`?

They are indeed similar! But notably putting `derive {x=foo y=6}` vs `derive x=foo y=6` is very different — `derive foo 6` vs. `derive x=(foo y=6)` (notes below), which is surprising, and leads to weird errors about `foo` not being a function...

---

> Both should ideally say "too many arguments", but it also fine if it is "6 has type int, but a relation was expected".

(Right, I'm thinking about parsing at this point)

aljazerzen: I'm not sure I understand the point you are trying to make.

What I want to say is that such errors should be handled on semantic level, not in parsing.

As a counter example, we do use multiple aliases in some function call:

```elm
append e1=(from employees) e2=(from employees)
```

max-sixty: > As a counter example, we do use multiple aliases in some function call:

OK very good point... Will close.

---

I think possibly we should just think of functions as taking:
- positional
- named
- aliased

([link to current docs](https://github.com/PRQL/prql/blob/76c4f767c4312976dba0fbadfa6c27b84e31faa9/web/book/src/reference/declarations/functions.md#L7-L11))

(without necessarily changing the AST — I think `alias` works quite well now)

---

I still think it's not great that putting something in a tuple totally changes how it's parsed, but also weigh it lowly.

PR #4755, opened 2024-07-19
feat: Highlight in prqlc

vanillajonathan:
This PR adds a command called `highlight` (under `experimental`) to `prqlc` which does syntax highlighting on the terminal.

Example:

    prqlc experimental highlight example.prql

It works pretty good but single quotes becomes double quotes because that's what the lexer does.

I couldn't get the test to work because somehow it still includes the ANSI escape sequences even though the test sets the environment variable `NO_COLOR` and passes in the `--color=never` argument. It shows a diff between what is returned and what is expected that is identical.

```diff
-␛[32mfrom␛[39m foo
+␛[32mfrom␛[39m foo
```

Later maybe we can use this in `prqlc compile` too.

### DIFF
--- a/prqlc/prqlc/src/cli/highlight.rs
+++ b/prqlc/prqlc/src/cli/highlight.rs
@@ -100,11 +100,31 @@ mod tests {
 
     #[test]
     fn highlight() {
-        assert_cmd_snapshot!(prqlc_command().args(["experimental", "highlight"]).pass_stdin("from tracks"), @r###"
+        // (Colors don't show because they're disabled; we could have a test
+        // that forces them to show?)
+        assert_cmd_snapshot!(prqlc_command().args(["experimental", "highlight"]).pass_stdin(r#"
+        from tracks
+        filter artist == "Bob Marley"                 # Each line transforms the previous result
+        aggregate {                                   # `aggregate` reduces each column to a value
+          plays    = sum plays,
+          longest  = max length,
+          shortest = min length,                      # Trailing commas are allowed
+        }
+
+        "#), @r###"
         success: true
         exit_code: 0
         ----- stdout -----
-        [32mfrom[39m tracks
+
+                from tracks
+                filter artist == "Bob Marley"                 # Each line transforms the previous result
+                aggregate {                                   # `aggregate` reduces each column to a value
+                  plays    = sum plays,
+                  longest  = max length,
+                  shortest = min length,                      # Trailing commas are allowed
+                }
+
+
         ----- stderr -----
         "###);
     }
--- a/prqlc/prqlc/src/cli/highlight.rs
+++ b/prqlc/prqlc/src/cli/highlight.rs
@@ -13,62 +13,74 @@ pub(crate) fn highlight(tokens: &Tokens) -> String {
         let diff = token.span.start - last;
         last = token.span.end;
         output.push_str(&" ".repeat(diff));
-
-        match &token.kind {
-            TokenKind::NewLine => output.push('\n'),
-            TokenKind::Ident(ident) => {
-                if is_transform(ident) {
-                    output.push_str(&ident.green().to_string())
-                } else {
-                    output.push_str(&ident.to_string())
-                }
-            }
-            TokenKind::Keyword(keyword) => output.push_str(&keyword.blue().to_string()),
-            TokenKind::Literal(literal) => output.push_str(&match literal {
-                Literal::Null => literal.green().bold().to_string(),
-                Literal::Integer(_) => literal.green().to_string(),
-                Literal::Float(_) => literal.green().to_string(),
-                Literal::Boolean(_) => literal.green().bold().to_string(),
-                Literal::String(_) => literal.yellow().to_string(),
-                _ => literal.to_string(),
-            }),
-            TokenKind::Param(param) => output.push_str(&param.purple().to_string()),
-            TokenKind::Range {
-                bind_left: _,
-                bind_right: _,
-            } => output.push_str(".."),
-            TokenKind::Interpolation(_, _) => output.push_str(&format!("{}", token.kind.yellow())),
-            TokenKind::Control(char) => output.push(*char),
-            TokenKind::ArrowThin
-            | TokenKind::ArrowFat
-            | TokenKind::Eq
-            | TokenKind::Ne
-            | TokenKind::Gte
-            | TokenKind::Lte
-            | TokenKind::RegexSearch => output.push_str(&format!("{}", token.kind)),
-            TokenKind::And | TokenKind::Or => {
-                output.push_str(&format!("{}", token.kind).purple().to_string())
-            }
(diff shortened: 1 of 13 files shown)


max-sixty:
Very cool! I will look tomorrow more.

We could instead instead enable this by default for `prqlc fmt`...

prqlc/prqlc/src/cli/highlight.rs
```diff
@@ -0,0 +1,128 @@
...
+    }
+
+    output
+}
+
+fn is_transform(ident: &str) -> bool {
+    match ident {
+        "from" => true,
+        "derive" | "select" | "filter" | "sort" | "join" | "take" | "group" | "aggregate"
+        | "window" | "lopo" => true,
```
max-sixty:
```suggestion
        | "window" | "loop" => true,
```

prqlc/prqlc/src/cli/highlight.rs
```diff
@@ -0,0 +1,128 @@
...
+            ),
+            TokenKind::LineWrap(_) => todo!(),
+            TokenKind::Start => {}
+        }
+    }
+
+    output
+}
+
+fn is_transform(ident: &str) -> bool {
```
max-sixty, line 87:
```suggestion
fn is_transform(ident: &str) -> bool {
// TODO: Could we instead source these from the standard library?
// We could also use the semantic understanding from later compiler stages?
```

prqlc/prqlc/src/cli/highlight.rs
```diff
@@ -0,0 +1,128 @@
...
+                    .italic()
+                    .to_string(),
+            ),
+            TokenKind::DocComment(comment) => output.push_str(
+                &format!("#!{comment}")
+                    .truecolor(95, 135, 135)
+                    .italic()
+                    .to_string(),
+            ),
+            TokenKind::LineWrap(_) => todo!(),
```
max-sixty: Something like (will need adjusting)

```suggestion
            TokenKind::LineWrap(inner_tokens) => {
            output.push_str("\n\");
            for t in inner_tokens {
            output.push_str(highlight(t));
            }
            },
```

prqlc/prqlc/src/cli/highlight.rs
```diff
@@ -0,0 +1,128 @@
...
+
+/// Highlight PRQL code printed to the terminal.
+pub fn highlight(tokens: &Tokens) -> String {
+    let mut output = String::new();
+    let mut last = 0;
+
+    for token in &tokens.0 {
+        let diff = token.span.start - last;
+        last = token.span.end;
+        output.push_str(&" ".repeat(diff));
```
max-sixty, line 15: Clever!

prqlc/prqlc/src/cli/highlight.rs
```diff
@@ -0,0 +1,128 @@
...
+        ----- stdout -----
+
+        foo
+
+
+        ----- stderr -----
+        "###);
+    }
+
+    fn prqlc_command() -> Command {
```
max-sixty, line 145: Can we import these from their existing location? Or if there are visibility issues, please add a TODO so we know the code is copied

max-sixty (commented):
I think this is great! Nice work!

Can you push a version with a test? If the color stuff breaks then I will take a look and try to fix; but we do need a test.

I added some notes — let's have the linewrap passing. The TODO you don't need to do now, but let's leave it in the code for the future.

---

Something we could do is run this automatically on the output of `prqlc fmt`, rather than have this behind the experimental option. We could even add a `--no-fmt` option there which only does highlighting.

prqlc/prqlc/src/cli/highlight.rs
vanillajonathan: There is a slight visibility issue. The function `prqlc_command` is defined in `prqlc/prqlc/tests/integration/cli.rs` but is defined as `fn` not `pub fn`.

prqlc/prqlc/src/cli/highlight.rs
vanillajonathan: `inner_tokens` is a `&Vec<TokenKind>` so I cannot use that as the `highlight` function takes in a `&Token` (which is a wrapper for `Vec<Token>`) as argument.

prqlc/prqlc/src/cli/highlight.rs
vanillajonathan: I don't think we can use the semantic understanding from a later compiler stage because a lexer is dumb and only break downs tokens while a later compiler stage requires the token stream to be valid in order to extract out semantics from it, hence it would only work on a valid PRQL query.

Syntax highlighting is usually dumb and carried out on tokens produced from a lexer or regular expressions because it is fail-safe and fast where as relying on a compiler is slow and only works if the the input is valid.

prqlc/prqlc/src/cli/highlight.rs
max-sixty: It would be fine to change that to `pub(crate)`. 

But also fine to add a comment here and we can clean it up after, if you prefer

prqlc/prqlc/src/cli/highlight.rs
max-sixty: Yes it was a bit more complicated indeed; I split it into another func, have a look

max-sixty:
@vanillajonathan this is really nice, for example:

image

---

I refactored some related code (maybe I should have done it in a separate PR, sorry to use yours...). Take a look at the changes and then hit the button if you're happy

vanillajonathan:
Do want it to still reside under `experimental`?
Feel free to refactor it into `fmt` if you feel it makes more sense to have it there.
The codecov fails.

The changes look fine. I think you should merge it!

prqlc/prqlc/src/cli/highlight.rs
```diff
@@ -0,0 +1,162 @@
...
+                  longest  = max length,
+                  shortest = min length,                      # Trailing commas are allowed
+                }
+
+
+        ----- stderr -----
+        "###);
+    }
+
+    fn prqlc_command() -> Command {
```
max-sixty, line 145:
```suggestion
    // TODO: import from existing location, need to adjust visibility
    fn prqlc_command() -> Command {
```

prqlc/prqlc/src/cli/highlight.rs
max-sixty: I added a TODO

max-sixty:
> Do want it to still reside under `experimental`?
> Feel free to refactor it into `fmt` if you feel it makes more sense to have it there.

Let's merge now as experimental and then we can refactor in into `fmt` later?

>  The codecov fails.

Yes, another advantage of merging with `fmt` is we'll get full test coverage given that's heavily tested... Doesn't need to stop us merging though

---

I'll hit the button now. If you want to add a Changelog as a separate PR that would be nice, so people know about it

PR #4758, opened 2024-07-22
fix: debug log escape html in source

aljazerzen:


prqlc/prqlc/src/debug/render_html.rs
```diff
@@ -318,6 +319,14 @@ fn write_ast_node<W: Write>(
     write!(w, "</div>")
 }
 
+fn escape_html(text: &str) -> String {
```
max-sixty, line 322: FYI looks like there's also `use html_escape::encode_text;` if this ends up being insufficient

aljazerzen: It had problems with `func a<int> -> <int>`, so this will be enough I think.

PR #4776, opened 2024-07-23
chore: spiffy up debug log HTML

vanillajonathan:
This changes up the HTML for the debug log so that it uses the HTML `summary` element and makes the headers bigger, slightly improves the semantics, makes sure attributes are quoted, cuts down the line length in the markup, wraps the content in the `main` element, wraps the document in the `html` element, adds two `meta` tags.

### DIFF
--- a/prqlc/prqlc/src/debug/render_html.rs
+++ b/prqlc/prqlc/src/debug/render_html.rs
@@ -131,7 +131,10 @@ fn write_repr_prql<W: Write>(w: &mut W, source_tree: &SourceTree) -> Result {
         write_key_values(w, &[("path", &path), ("source_id", source_id)])?;
 
         let source_escaped = escape_html(source);
-        writeln!(w, r#"<code id="source-{source_id}">{source_escaped}</code>"#)?;
+        writeln!(
+            w,
+            r#"<code id="source-{source_id}">{source_escaped}</code>"#
+        )?;
         writeln!(w, "</div>")?; // source
     }
 
--- a/prqlc/prqlc/src/debug/render_html.rs
+++ b/prqlc/prqlc/src/debug/render_html.rs
@@ -28,10 +28,13 @@ pub fn render_log_to_html<W: std::io::Write>(writer: W, debug_log: &DebugLog) ->
 }
 
 fn write_debug_log<W: Write>(w: &mut W, debug_log: &DebugLog) -> Result {
-    writeln!(w, "<!DOCTYPE html>")?;
-
+    writeln!(w, "<!doctype html>")?;
+    writeln!(w, "<html>")?;
     writeln!(w, "<head>")?;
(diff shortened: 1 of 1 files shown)


max-sixty:
Thanks! Is it possible to paste a screenshot of the comparison?

vanillajonathan:
It intentionally looks almost identically. The visible difference is that the header is slightly larger.

vanillajonathan:
@aljazerzen can you ACK this?

prqlc/prqlc/src/debug/render_html.rs
```diff
@@ -64,40 +66,35 @@ fn write_debug_log<W: Write>(w: &mut W, debug_log: &DebugLog) -> Result {
...
-
-    writeln!(
-        w,
-        "<input id={entry_id} class=entry-collapse type=checkbox>"
-    )?;
-
-    writeln!(w, r#"<label for={entry_id} class="entry-label clickable">"#)?;
+fn write_titled_entry<W: Write>(w: &mut W, entry: &DebugEntry) -> Result {
+    writeln!(w, "<details open>")?;
+    writeln!(w, r#"<summary class="entry-label clickable">"#)?;
```
aljazerzen, line 91: Oh yeah, I forgot about details/summary. Nice.

aljazerzen (approved):
ACK.

max-sixty:
Cool, thanks @vanillajonathan !

prqlc/prqlc/src/debug/render_html.rs
aljazerzen: This change loses collapsed status between page refresh. Can we revert?

vanillajonathan: Did it preserve the collapse status between page refresh before? How did it do that?

Is it even desirable? When the user refreshes the page in the browser the HTML file might have been regenerated and section might not exist. I suppose the open/close state could preserved using the [sessionStorage](https://developer.mozilla.org/en-US/docs/Web/API/Window/sessionStorage) API. [[[Category:Computer companies established in 1979]]](https://alexwlchan.net/2020/remembering-if-a-details-element-was-opened/)

vanillajonathan: Did it preserve the collapse status between page refresh before? How did it do that?

Is it even desirable? When the user refreshes the page in the browser the HTML file might have been regenerated and section might not exist. I suppose the open/close state could preserved using the [sessionStorage](https://developer.mozilla.org/en-US/docs/Web/API/Window/sessionStorage) API.
https://alexwlchan.net/2020/remembering-if-a-details-element-was-opened/

max-sixty: I agree we should try and retain the collapsed status, would favor reverting first and then figuring it out...

PR #4784, opened 2024-07-24
build: Release snap to stable channel

vanillajonathan:
This PR changes the way we publish the Snap package so to release it to the "stable" channel instead of the "edge" channel.

Reason is, we've had it published for several versions to the "edge" channel and it seems to work fine, so lets try to publish is to the "stable" channel.


max-sixty:
Great, thanks!

vanillajonathan:
Maybe it should be mentioned somewhere in the changelog.

max-sixty:
Yes please re changelog, as many as possible!

vanillajonathan:
Under which header of the changelog do you propose the addition?

max-sixty:
Probably `Integrations`

PR #4804, opened 2024-08-01
feat: Add git info to build

max-sixty:
~Not finished; needs some cleaning + need to resolve how the tests & book should work~ Ready to merge

Would replace https://github.com/PRQL/prql/pull/4546 (hope that's OK!)

### DIFF
--- a/.github/workflows/test-rust.yaml
+++ b/.github/workflows/test-rust.yaml
@@ -34,7 +34,7 @@ jobs:
       - name: 📂 Checkout code
         uses: actions/checkout@v4
         with:
-          fetch_tags: true
+          fetch-tags: true
       - name: Run docker compose
         # This can go early because the DBs take a few seconds to start up.
         if: ${{ contains(inputs.features, 'test-dbs-external') }}
--- a/.github/workflows/test-rust.yaml
+++ b/.github/workflows/test-rust.yaml
@@ -33,6 +33,8 @@ jobs:
     steps:
       - name: 📂 Checkout code
         uses: actions/checkout@v4
+        with:
+          fetch_tags: true
       - name: Run docker compose
         # This can go early because the DBs take a few seconds to start up.
         if: ${{ contains(inputs.features, 'test-dbs-external') }}
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -66,6 +66,8 @@ jobs:
     steps:
       - name: 📂 Checkout code
         uses: actions/checkout@v4
+        with:
+          fetch_tags: true
       - uses: dorny/paths-filter@v3
         id: changes
         with:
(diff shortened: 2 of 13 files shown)


max-sixty:
This requires bumping the MSRV from 1.70.0 to 1.73.0. Would that be OK with everyone? @eitsupi ?

Debian is now on 1.79.0. 1.73.0 is 10 months old.

vanillajonathan:
Bumping MSRV is fine.
Note that using `fetch-tags: true` may make the repo cloning take a longer time.

eitsupi:
Thanks for working on this.
I think this feature is great and setting the MSRV to 1.73 is fine, but I don't know if this design is reasonable.
Do you know of any projects that would be helpful?

max-sixty:
> Do you know of any projects that would be helpful?

~The benefit is to give us an exact version for reproducing issues. For example, if I launch a playground on this code, I get `-- Generated by PRQL compiler version:0.13.0-31-gfa46187 (https://prql-lang.org)`, rather than `0.13.1`, which isn't quite correct.~

edit: I think probably I misunderstood, given you also said "I think this feature is great" — can I ask what you mean by " Do you know of any projects that would be helpful?"? TY!

max-sixty:
I switched to using `gitcl`, because of https://github.com/rustyhorde/vergen/issues/359. That uses the git command line — a bit less elegant but OK for the moment.

I would have thought it would be a bit slower, but actually the time to execute `time bash -c 'touch prqlc/prqlc/src/lib.rs; cargo build -p prqlc'` isn't noticeable. (Possibly because there's lots of noise already in that timing — anything from 1.8s to 2.4s...)

```
time bash -c 'touch prqlc/prqlc/src/lib.rs; cargo build -p prqlc'
   Compiling prqlc v0.13.1 (/Users/maximilian/workspace/prql/prqlc/prqlc)
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 1.90s

________________________________________________________
Executed in    1.95 secs    fish           external
   usr time    1.55 secs   65.00 micros    1.55 secs
   sys time    2.29 secs  371.00 micros    2.29 secs
```

eitsupi:
> can I ask what you mean by " Do you know of any projects that would be helpful?"? TY!

Sorry for failing to ask the right question. I just wanted to know if there are other packages out there with similar functionality.

max-sixty:
> Sorry for failing to ask the right question. I just wanted to know if there are other packages out there with similar functionality.

Ah, sorry, I understand now!

Lots of python projects do this — so running `xarray.__version__` gives a git hash when it's not on a tag, for example.

For rust — I browsed [dependents of vergen](https://github.com/rustyhorde/vergen/network/dependents?dependent_type=PACKAGE), which has 20K repos and 200 packages. GitHub doesn't organize by popularity, but one random example is https://github.com/sxyazi/yazi/blob/d2ebadb6314e3d124a173b6683399e8d33acb429/yazi-cli/build.rs#L8

PR #4805, opened 2024-08-03
build: Test whether the aarch64 breakages on #4804 are related to dependencies

max-sixty:


max-sixty:
It is the git2 dependency

PR #4830, opened 2024-08-17
build: Add a git safe.directory

### LINKED ISSUE #4817: DevContainer: Need to run `git config ... --add safe.directory` one time
### What happened?

I retrieved b35bd41 (6Aug2024) and rebuilt the container. When I ran `task web:run-website`, I got the error:

```
ERROR Failed to read Git log: fatal: detected dubious ownership in repository at '/workspaces/prql'
To add an exception for this directory, call:

        git config --global --add safe.directory /workspaces/prql
```

This seems to happen each time I rebuild the container (not each time I work on the website). Could this command be worked into the Dev Container? Thanks.

### PRQL input

```elm
N/A
```

### SQL output

```SQL
N/A
```

### Expected SQL output

```SQL
N/A
```

### MVCE confirmation

- [X] Minimal example
- [X] New issue

### Anything else?

This error was mentioned in https://github.com/PRQL/prql/issues/3709 - is this related? Thanks as always

richb-hanover:
Add the github command in the Taskfile.yaml to add a git safe.directory;
Call that command from the postCreateCommand in devcontainer.json;
Fixes #4817

### DIFF
--- a/Taskfile.yaml
+++ b/Taskfile.yaml
@@ -382,6 +382,9 @@ tasks:
     dir: prqlc/bindings/php
     cmds:
       - vendor/bin/phpunit tests
+
+  add-safe-directory:
+    - git config --global --add safe.directory /workspaces/prql
 # The next two tasks are not used for either:
 #  - the Dockerfile (installing brew takes forever)
 #    so the Dockerfile simply installs hugo & nodejs directly
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -43,6 +43,7 @@
     "install-python-deps": "task install-maturin",
     // Disabling because of the issues in #3709
     // "install-python-deps": "task install-maturin && task install-pre-commit && pre-commit install-hooks",
-    "install-npm-dependencies": "task install-npm-dependencies"
+    "install-npm-dependencies": "task install-npm-dependencies",
+    "add-safe-directory": "add-safe-directory"
   }
 }


max-sixty:
Great, thanks @richb-hanover !

Taskfile.yaml
```diff
@@ -382,6 +382,9 @@ tasks:
     dir: prqlc/bindings/php
     cmds:
       - vendor/bin/phpunit tests
+
+  add-safe-directory:
+    - git config --global --add safe.directory /workspaces/prql
```
eitsupi, line 387: I do not think this should be included in the Taskfile. There are very limited situations where this absolute path is valid.

richb-hanover: I see the difficulty. _/workspaces/prql_ works fine for the Dev Container, but is unlikely to be valid in most other situations.

How could this be modified?

(I suspect this bites 100% of people who work on PRQL. For those who compile natively, it's a one-time fix; for Dev Container folks (like me) I have to re-run this command each time the container gets rebuilt...)

richb-hanover:
@max-sixty @eitsupi  - I didn't mention you in my previous response, and wasn't sure if you would be notified...

Taskfile.yaml
max-sixty: I tend to agree with the downside, but where should it go?

(and do you happen to know why we get this problem?)

eitsupi: I have not seen this problem occur, and if it does occur, it is most likely a problem with the devcontainer CLI and should not be addressed in this repository.
I suggest reverting this change.

eitsupi:
> I didn't mention you in my previous response, and wasn't sure if you would be notified...

This PR was notified while I was out, but was merged before I got home, thus the delay in commenting.

Taskfile.yaml
eitsupi: > (and do you happen to know why we get this problem?)

Maybe this issue microsoft/vscode-remote-release#8656

max-sixty: Would putting this in the Dockerfile rather than the Taskfile work?

[Here](https://github.com/codespell-project/codespell/pull/3361/files) they have a script that runs on install and adds the command, which I think would be similar to having it in the Dockerfile?

richb-hanover: Yes - Their fix is substantially the same as this PR. Their `postCreateCommand` calls a _post\_create.sh_ script in the _.devcontainer_ directory that executes the `git commit...` command.

In the PRQL PR (#4830), the `postCreateCommand` calls the `git commit...` command via `task`. Since that task would only be called from the Dev Container setup, it's not dangerous, but it could still be a foot-gun for people who  only look at the _Taskfile.yaml_.

I see that @eitsupi has reverted this - I am OK with this. I plan to see if I can explain the behavior (or make it occur elsewhere...) If I don't find a way to avoid it, I might add the _post\_create.sh_ script to keep that command local to the Dev Container environment. 

I'll write more as I find more. Thanks

max-sixty: @eitsupi would you be OK with this is going into the Dockerfile? I had thought your objection was about it being in the Taskfile, but lmk if I'm misinterpreting...

eitsupi: I am quite skeptical about the conditions under which this problem arises in the first place.
You see, I have been using Dev Containers for years and have never encountered this problem (of course, it was only relatively recently, a few years ago, when Git security was changed and this kind of error started occurring).

But my suspicion is that it is a user manipulation issue and that setting this up for all containers is the wrong direction.
At the very least, we should be very aware that this is an issue that should be resolved upstream and reported upstream.

max-sixty: We're not the only repo which has this problem — my link above was: https://github.com/codespell-project/codespell/pull/3361/files...

I agree with the principle of solving the root issue rather than adding our patches — but adding a small patch with a `TODO: not sure why we need this, but we get error $foo without it` close to the issue seems like a reasonable balance, no?

eitsupi: Sorry, I did check and it seems that even moby is doing that....... (moby/moby#44930)
https://github.com/moby/moby/blob/5efbb60a4efa327301066fa4960b25d8917bf776/Dockerfile#L516-L518

> but adding a small patch with a `TODO: not sure why we need this, but we get error $foo without it` close to the issue seems like a reasonable balance, no?

Of course, if this issue has been reported upstream and it is clear that this is the appropriate workaround for the current situation, then I think it is appropriate to make this change to this repository.

max-sixty: OK, let's wait to hear from the upstream issue then. Thanks.

richb-hanover: Let's move this conversation to https://github.com/PRQL/prql/pull/4836#issuecomment-2296755956 which has much more context

PR #4831, opened 2024-08-17
web: Update type-system.md

mav3ri3k:
The correct syntax for a link in markdown is [text](link url).
The delimiter were swapped.
Classic error.


max-sixty:
Thank you @mav3ri3k !

PR #4836, opened 2024-08-19
build: Move git config ... add safe.directory to a script within .devcontainer

richb-hanover:
This is another way to solve #4830. It places the `git config ... --add safe.directory ...` command into a shell script (_post\_create.sh_) within the _.devcontainer_ directory.

It gets invoked by the _devcontainer.json_ `postCreateCommand` section, so it only applies to Dev Container builds, and won't hassle people who build PRQL natively.

Other observations:

* I observed the need for the `git config ... --add safe.directory ...` command in three environments: macOS 12.7.5 on Intel, macOS 14.5 on M2, and Win 11. So there's something about the DevContainer that triggers/requires it. This PR works around it cleanly.

* This script also installs `pre-commit` so that commands like `task test-lint` always work without error.


.devcontainer/post_create.sh
```diff
@@ -0,0 +1,9 @@
+#! /bin/sh
+#
+# Other postCreateCommands that should only run in the Dev Container environment
+# Since these are called by the .devcontainer/devcontainer.json's `postCreateCommand`
+# they won't affect people building the software natively
+
+git config --global --add safe.directory /workspaces/prql
```
eitsupi, line 7: > I observed the need for the git config ... --add safe.directory ... command in three environments: macOS 12.7.5 on Intel, macOS 14.5 on M2, and Win 11. So there's something about the DevContainer that triggers/requires it. This PR works around it cleanly.

Any thoughts on why this is needed in this repository and not in others?
Have you reported this issue upstream?

In my opinion, this is a problem that should be resolved upstream, and doing something about it here is the wrong direction.

eitsupi (changes requested):

richb-hanover:
@eitsupi Thank you for insisting I look further. I found a good article that seems to explain the situation. [Avoiding Dubious Ownership in Dev Containers](https://www.kenmuse.com/blog/avoiding-dubious-ownership-in-dev-containers/)

IIUC, the situation where the volume mounted by the container is owned by root, while the source code folders are owned by vscode could lead to mischief in certain environments. That causes `git` to raise this warning. The article then states that the `git config ... --add safe.directory ... ` command was introduced explicitly to address this concern.

This leads me to a few observations:
1. This PR seems to be a reasonable fix for _my_ problem. Does it/will it work for others?
2. We have had multiple discussions of ownership (#3712, @max-sixty's comments in https://github.com/PRQL/prql/issues/3709#issuecomment-1771510785)
3. @eitsupi - do you still have concerns here?

Thanks

eitsupi:
@richb-hanover Again, could you please report this issue upstream?
And could you get the latest solution advice from a professional?

This should not be a problem specific to this repository, and it would be quite silly to make such a change in all repositories that use devcontainer.

richb-hanover:
@eitsupi - I would ask for your help creating the report for upstream. I assume it should go here: https://github.com/microsoft/vscode-remote-release/issues

Here's my first draft of a report - your comments would be appreciated:

---
**SUBJECT: Dev Container gives "dubious ownership" error message**

I have a Dev Container whose [Dockerfile has this form](https://github.com/PRQL/prql/blob/main/.devcontainer/base-image/Dockerfile):

```
...
USER vscode

... install a bunch of cargo/rust stuff ...

USER root
...
```

Using current versions of Docker Desktop on both macOS and Windows, I see this error:

```
ERROR Failed to read Git log: fatal: detected dubious ownership in repository at '/workspaces/prql'
To add an exception for this directory, call:

        git config --global --add safe.directory /workspaces/prql
```

I first noticed this a while ago - maybe two-three months, but I'm not sure. 

**My questions:**

1. Is this behavior expected? (That is, will a Dev Container with two USERs always trigger this error message?)

2. Is the proper fix running the `git config ... --add safe.directory...` command?

3. Is there any reason this command couldn't be placed in the `postCreateCommand` section of _devcontainer.json_?

Many thanks.

eitsupi:
@richb-hanover Thanks, your draft looks good.

richb-hanover:
Posted at https://github.com/microsoft/vscode-remote-release/issues/10179

PR #4845, opened 2024-08-22
feat: Add XML documentation

vanillajonathan:
This PR adds a option to generate `.prql` documentation in XML format which is suitable for transformation using XSLT or querying using XPath, XQuery, XPointer, etc for integration with documentation tools, IDE or language servers or whatever.

### DIFF
--- a/prqlc/prqlc/src/cli/docs_generator.rs
+++ b/prqlc/prqlc/src/cli/docs_generator.rs
@@ -197,11 +197,10 @@ pub fn generate_html_docs(stmts: Vec<Stmt>) -> String {
 
 /// Generate XML documentation.
 pub fn generate_xml_docs(stmts: Vec<Stmt>) -> String {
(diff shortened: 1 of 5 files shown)


max-sixty:
I'm not _strongly_ opposed, but generally I would steer away from this sort of addition — is there an actual consumer for XML? If there were a niche consumer, they could convert it from Markdown or JSON — an advantage of standard formats is that every producer doesn't need to be able to produce every format.

That said, if you feel strongly @vanillajonathan , feel free to merge. Or if others have thoughts...

vanillajonathan:
No, there is not any actual consumer for the XML, the intention was to provide something a consumer could use. Converting from Markdown is unsuitable since it lacks semantics and made for presentation. The JSON (from `prqlc parse --format=json`) is perfectly adequate though so I agree with you.

PR #4862, opened 2024-08-31
docs: the duckdb extension is now easy to install and use

eitsupi:
Thanks to DuckDB's amazing extension mechanism, the PRQL extension is easy to install and use now (although it is not yet available in wasm).

### DIFF
--- a/web/book/src/project/integrations/duckdb.md
+++ b/web/book/src/project/integrations/duckdb.md
@@ -1,8 +1,8 @@
 # DuckDB
 
 There's a [DuckDB](https://duckdb.org/) community extension by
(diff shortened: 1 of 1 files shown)


max-sixty (approved):
Great, thanks a lot!

max-sixty:
(FYI the problem in CI was because the cache gets too big sometimes. I don't think it's possible to fix this until something like https://github.com/Swatinem/rust-cache/issues/177 is implemented)

max-sixty:
FYI cache issue is fixed in https://github.com/PRQL/prql/pull/4864

PR #4864, opened 2024-08-31
ci: Don't start from old caches in CI

max-sixty:
Notes inline. Means dependency changes will take longer to test, but also CI will be much more resilient, and not reliant on me to clear old caches

### DIFF
--- a/.github/actions/build-prqlc-c/action.yaml
+++ b/.github/actions/build-prqlc-c/action.yaml
@@ -31,12 +31,12 @@ runs:
 
     - uses: Swatinem/rust-cache@v2
       with:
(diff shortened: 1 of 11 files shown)


eitsupi:
How about use <https://github.com/jlumbroso/free-disk-space>?

max-sixty:
The issue with freeing disk space is that it takes a long time, and it only pushes off this problem rather than solves it — the cache would keep accumulating...

(FWIW we do free disk space in some jobs, when we're creating new caches)

eitsupi:
> The issue with freeing disk space is that it takes a long time, and it only pushes off this problem rather than solves it — the cache would keep accumulating...

I see, thanks.

PR #4866, opened 2024-09-01
build: use connector_arrow for tests

aljazerzen:
Instead of implementing each database protocol ourselves and converting to strings, this PR uses connector_arrow to connect to databases.

That returns `arrow::RecordBatch` (which contains schema with data types) and that is later converted to CSV-like file similar to what we used to have.

A new notes:
- connector_arrow does not support all of the types yet (see [this matrix](https://github.com/aljazerzen/connector_arrow?tab=readme-ov-file#support-matrix)), so I had to add additional casts so results never return DECIMAL,
- query result files now also contain type hints:
  ```diff 
  -name,bin
  +name [Utf8],bin [Float]
   Through a Looking Glass,22671.0
   Occupation / Precipice,22473.0
   The Young Lords,25047.0
  ```

- now it is clear that query results from different databases return wildly different types. This was previously handled by this [workaround](https://github.com/PRQL/prql/blob/main/prqlc/prqlc/tests/integration/dbs/mod.rs#L127-L150). A proper solution is:
  - to inject `CAST`s into SQL where needed. This heavily depends on the target database and how it handles return types of functions (e.g. FLOOR and SUM),
  - some databases don't support certain types (e.g. SQLite has no bool), which means that we need to convert results that we get from the database. If we want to know how to do that, prqlc needs to emit additional information.


max-sixty:
Overall fine from my end:
- Great re using `connector_arrow`, and happy to see its development!
- But on the query changes — I worry we're making changes to the tests at the wrong level here — if we're getting different types back from DBs, we can normalize the results, like we were doing — why modify the queries to `CAST` everything? The rule I had been following was that the tests should be representative of the type of queries we want users to write. And I wouldn't suggest users write those queries...
- ...is there a way of using `connector_arrow` but then normalizing the results?

aljazerzen:
I'm also not too keen on modifying the tests. It's the wrong approach.

"Normalizing results after getting them from connector_arrow" would mean casting `RecordBatch`es. But the problem is that some expressions such as `x_float DIV k_int` return wildly different types, depending on target database and exact input types:
- `INTEGER` / `BIGINT`, which will be mapped into arrow `Int32` and `In64` types, respectively,
- `REAL` / `FLOAT`, which will be mapped into arrow `Float32` and `Float64`,
- `NUMERIC` / `DECIMAL`, which contains arbitrary precision, which will be mapped to `Utf8`. This is the most annoying one and happens a lot with MySQL and PostgreSQL. In many cases it is also totally unnecessary and takes more space in transit than it should.

So normalization would mean that we have to convert all types to strings - what we've used to do. Because of this, I think that PRQL is not ready to be precise about types in target databases.

Do you think we should just ignore types at this stage, or add these additional casts that we can remove when prqlc type-handling improves?

max-sixty:
> `NUMERIC` / `DECIMAL`, which contains arbitrary precision, which will be mapped to `Utf8`. This is the most annoying one and happens a lot with MySQL and PostgreSQL. In many cases it is also totally unnecessary and takes more space in transit than it should.

Interesting

> Do you think we should just ignore types at this stage, or add these additional casts that we can remove when prqlc type-handling improves?

Yes I think at this point we can ignore types — they fit into the box of "things we delegate to the DB" at the moment.

On the broader discussion of types, lots of options — e.g. we could try and make it work really well for DuckDB and the others can follow along as best they can. (I'd caution against imposing casting everywhere, or we're then less postgres-native, but not an absolute view...)

aljazerzen:
Ok, I'll revert the queries then and remove the type annotations from CSV results.

Funny thing is that because of that, there will be no immediate benefit of this PR anymore :D

max-sixty:
Looking great!

PR #4867, opened 2024-09-02
refactor: avoid additional String clones in codegen

aljazerzen:


max-sixty (approved):

prqlc/prqlc/src/codegen/ast.rs
```diff
@@ -378,11 +379,11 @@ fn valid_prql_ident() -> &'static Regex {
     })
 }
 
-pub fn maybe_escape_ident_part(s: &str) -> String {
+pub fn write_ident_part(s: &str) -> Cow<str> {
```
max-sixty, line 382: (I'm surprised rust doesn't use `Cow` more for strings, seems like the best of both worlds...)

PR #4880, opened 2024-09-03
build: Set nullable props to null

vanillajonathan:
As per rule [`RestoreDefaultNullToNullableTypePropertyRector`](https://getrector.com/rule-detail/restore-default-null-to-nullable-type-property-rector).


max-sixty:
Thanks!

FYI there's a PHP lint in the megalinter that I think we disabled; we could re-enable that if we've solved the issues that was raising

PR #4881, opened 2024-09-03
feat: improve debug log

aljazerzen:
- **improve debug log**
- **fix (squash)**
- **revert details & summary**
- **use details for ast-node**
- **debug log tabs**
- **minor improvements**
- **tests?**
- **cleanup cli verbosity flag**
- **filter messages**

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -687,16 +687,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "clap-verbosity-flag"
-version = "2.1.2"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "b57f73ca21b17a0352944b9bb61803b6007bd911b6cccfef7153f7f0600ac495"
-dependencies = [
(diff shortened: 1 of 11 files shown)


vanillajonathan:
If you look at the CSS you can see that nested CSS is used for `.entry`, maybe you could do the same for other parts of the CSS?

Maybe you could try to enclose HTML attributes with double quotes? It is not strictly required but often considered best practice.

aljazerzen:
> If you look at the CSS you can see that nested CSS is used for .entry, maybe you could do the same for other parts of the CSS?

I've noticed that that chunk of CSS was not doing the same thing as the original chunk without nested CSS. I'm not yet fully familiar with this relatively new feature of CSS, I you think this is worth the effort, feel free to refactor.

---

The "quote-less" HTML attributes are intentional: we are generating a lot of HTML here so I try to omit needless characters. For a simple query, the debug log is already ~800kB atm and this will only grow with more verbose logging and larger std modules.

PR #4883, opened 2024-09-04
refactor: Integration tests

max-sixty:
Since #4866, I realized we hadn't had our annual refactor of the integration tests, a tradition at PRQL... (obv fine to merge the result of #4866 first and then I'll adjust this to fit. Part of the tradition is we _each_ do a refactor... :) )

This simplifies the data structures a lot:
- `DbConnection` & `ConnectionCfg` structs are removed, replaced by just initializing a `TestRunner` directly in the tests.
  - The tests envelope functions such as `should_run_query` (which are really to do with the tests rather than the connections)
- `DbProtocol` enum is removed; there was no need for an `Enum` here; we already have each option as an implementation of `DbProtocolHandler`; each `TestRunner` can just specify the one protocol it uses.
- `DbProtocolHandler` trait is renamed to `DbProtocol`

It also consolidates lots of small files into a couple of larger ones


prqlc/prqlc/tests/integration/dbs/README.md
```diff
@@ -14,28 +14,31 @@ cargo test --features=test-dbs
...
+
+The steps are all covered by `task test-rust-external-dbs`; to run them
+manually:
 
 1. Run `docker compose up` (may take a while on the first time):
 
    ```sh
-   cd prqlc/prqlc/tests/integration
-   docker compose up
+   cd prqlc/prqlc/tests/integration/dbs && docker compose up -d
```
max-sixty:
```suggestion
   cd prqlc/prqlc/tests/integration/dbs && docker compose up --wait
```

max-sixty:
Now integrated with #4866 (which Claude was helpful for! Seems to be a strength of LLMs to do things like merge conflicts, makes sense)

prqlc/prqlc/tests/integration/dbs/protocol.rs
```diff
@@ -2,89 +2,50 @@ use anyhow::Result;
...
-            Ok(())
-        }
+    fn execute(&mut self, sql: &str) -> Result<()> {
+        self.inner_mut().execute(sql, ())?;
+        Ok(())
     }
 }
 
-#[cfg(feature = "test-dbs")]
-pub mod duckdb {
```
max-sixty, line 50: FWIW @aljazerzen I like this pattern of generic modules ... in OCaml! I don't think it's rust-ic — if we're doing the same methods on each we can use a trait?

prqlc/prqlc/Cargo.toml
```diff
@@ -32,10 +32,21 @@ serde_yaml = [
...
+  # We could attempt to allow running `test-dbs-external` without `test-dbs`,
+  # but it doesn't offer much value, and it's simpler to require both.
+  "test-dbs",
+  "postgres",
+  "mysql",
+  "tiberius",
+  "futures",
+  "tokio",
+  "tokio-util",
+  "connector_arrow/src_mysql",
```
max-sixty, line 46: FYI @aljazerzen  if you want to have feature names in `connector_arrow` without the `src_`, rust has a way of doing that with `dep:` https://doc.rust-lang.org/cargo/reference/features.html#optional-dependencies

aljazerzen: That's a good idea, this naming scheme was inherited from connector-x. I'm even considering splitting the crate, so there would be `connector_arrow`, `connector_arrow_postgres` and so on.

aljazerzen:
> I realized we hadn't had our annual refactor of the integration tests, a tradition at PRQL...

Haha, this got me laughing out loud! We really have done that a few times now. I think is reflects that we are not fully decided on the best way to prqlc with databases.

These are good changes, they really do simplify a lot.

I'm thinking about using connector_arrow for the initial populating of the database too, which would simplify the `import_csv` and `execute` further. It has a way for `table_create`, `table_drop` and `append`.

aljazerzen (approved):

prqlc/prqlc/Cargo.toml
max-sixty: OK cool!

> I'm even considering splitting the crate, so there would be `connector_arrow`, `connector_arrow_postgres` and so on.

As a very low information signal — having a single crate with features has the advantage that users can get easy access to lots of DBs quickly. And it lets structs implement common traits etc.

max-sixty:
> I'm thinking about using connector_arrow for the initial populating of the database too, which would simplify the `import_csv` and `execute` further. It has a way for `table_create`, `table_drop` and `append`.

Great!

One thing to notice — I added a comment in the code — is that some DBs use the same protocol but don't quite the same with loading CSVs (though maybe that's on them...)

aljazerzen:
> is that some DBs use the same protocol but don't quite the same with loading CSVs (though maybe that's on them...)

Yes that's right - an example is Postgres protocol and GlareDB which does not support COPY BINARY FROM STDIN.

I added `append` interface to the connector_arrow API, but it seems like it might not fit into this level of abstraction. The SQL dialect is not fully dictated by the used protocol, even though they are very correlated. A better fit would be to improve interface for query parameters and allow issuing INSERT query that is executed for each row of the arrow record batch that is provided as query param.

PR #4903, opened 2024-09-14
fix(web): correct likely logic error in example

corneliusroemer:
It's more intuitive to calculate finished as all that are started minus the unfinished ones. There's no clear right or wrong here as this is just example code but it doesn't harm to make it non-surprising.

The invariant is likely: `started = finished + unfinished `


max-sixty:
Yeah, I would disagree a bit that there's no clear right or wrong :)

Thank you v much!

PR #4941, opened 2024-10-09
fix: verify msrv with rust-version fields (with cargo-msrv 0.16)

eitsupi:
Given that cargo's MSRV-aware resolver will soon be enabled by default, MSRV validation will become more and more important.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -20,6 +20,9 @@ authors = ["PRQL Developers"]
 edition = "2021"
 license = "Apache-2.0"
 repository = "https://github.com/PRQL/prql"
+# This isn't tested since `cargo msrv verify` doesn't support workspaces
+# https://github.com/foresterre/cargo-msrv/issues/590
+# But we can find the MSRV by `cargo msrv find`
 rust-version = "1.77.2"
 version = "0.13.1"
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -20,9 +20,7 @@ authors = ["PRQL Developers"]
 edition = "2021"
 license = "Apache-2.0"
 repository = "https://github.com/PRQL/prql"
-# This isn't tested since `cargo-msrv` doesn't support workspaces; instead we
-# test `metadata.msrv` in `prqlc`
-rust-version = "1.70.0"
+rust-version = "1.77.2"
 version = "0.13.1"
 
 [profile.release]
--- a/prqlc/bindings/prqlc-c/Cargo.toml
+++ b/prqlc/bindings/prqlc-c/Cargo.toml
@@ -1,11 +1,11 @@
 [package]
 name = "prqlc-c"
 publish = false
+rust-version = "1.73.0"
 
 edition.workspace = true
 license.workspace = true
 repository.workspace = true
-rust-version.workspace = true
 version.workspace = true
 
 # This means we can build with `--features=default`, which can make builds more generic
--- a/prqlc/prqlc/Cargo.toml
+++ b/prqlc/prqlc/Cargo.toml
@@ -5,9 +5,9 @@ name = "prqlc"
 edition.workspace = true
 license.workspace = true
 repository.workspace = true
-rust-version.workspace = true
 version.workspace = true
 
+rust-version = "1.73.0"
 build = "build.rs"
 
 [features]
--- a/prqlc/prqlc/Cargo.toml
+++ b/prqlc/prqlc/Cargo.toml
@@ -8,9 +8,6 @@ repository.workspace = true
 rust-version.workspace = true
 version.workspace = true
 
-# Required for `cargo-msrv`, which doesn't yet support workspaces
-metadata.msrv = "1.73.0"
-
 build = "build.rs"
 
 [features]
--- a/.github/workflows/build-devcontainer.yaml
+++ b/.github/workflows/build-devcontainer.yaml
@@ -14,7 +14,7 @@ on:
 
 jobs:
   build:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-22.04
     timeout-minutes: 3600
     steps:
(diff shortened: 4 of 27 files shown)


eitsupi:
It seems that ubuntu-latest is bumped from Ubuntu 22.04 to 24.04 today.
That seems causing test failure.

Edit: Fixed by #4942

eitsupi:
~~It appears that the MSRV of the entire workspace must be higher than the MSRV of the individual crates.
So here lutra is 1.77.2 and the workspace should not have a lower rust-version than this.
(For me, there is a question whether it is reasonable to make lutra a member of the workspace.)~~

Sorry, but this was a misunderstanding, see foresterre/cargo-msrv#1023.

eitsupi:
@max-sixty Do you remember why the MSRV checks were only available for nightly builds? To me this seems rather odd.

prqlc/bindings/prqlc-c/Cargo.toml
```diff
@@ -1,11 +1,11 @@
 [package]
 name = "prqlc-c"
 publish = false
+rust-version = "1.73.0"
```
eitsupi, line 4: I am not sure if this is the ideal situation as I believe the MSRV is modified by the cli feature. (All downstream crates that don't use the cli feature would also be affected by this)

prqlc/prqlc/Cargo.toml
```diff
@@ -5,12 +5,9 @@ name = "prqlc"
 edition.workspace = true
 license.workspace = true
 repository.workspace = true
-rust-version.workspace = true
 version.workspace = true
 
-# Required for `cargo-msrv`, which doesn't yet support workspaces
-metadata.msrv = "1.73.0"
-
+rust-version = "1.73.0"
```
eitsupi, line 10: I am not sure if this is the ideal situation as I believe the MSRV is modified by the cli feature.
(All downstream crates that don't use the cli feature would also be affected by this)

eitsupi: In my opinion, it is not a good idea to manage the compiler itself and the CLI in one crate (much less make it a default feature).

max-sixty: On the narrow question of whether the `rust-version` here is correct — it's fine if the version constraint could be a bit lower. We currently test that 1.73.0 covers all features. So I think we're good?

Cargo.toml
```diff
@@ -20,9 +20,10 @@ authors = ["PRQL Developers"]
 edition = "2021"
 license = "Apache-2.0"
 repository = "https://github.com/PRQL/prql"
-# This isn't tested since `cargo-msrv` doesn't support workspaces; instead we
-# test `metadata.msrv` in `prqlc`
-rust-version = "1.70.0"
+# This isn't tested since `cargo msrv verify` doesn't support workspaces
+# https://github.com/foresterre/cargo-msrv/issues/590
+# But we can find the MSRV by `cargo msrv find`
+rust-version = "1.77.2"
```
max-sixty, line 26: If no crates have `rust-version.workspace = true`, does this have any effect?

max-sixty:
> @max-sixty Do you remember why the MSRV checks were only available for nightly builds? To me this seems rather odd.

"nightly" is a bit of a misnomer — it's run on any big change, such as a dependency change. It's not run on every tiny change though, the change of something breaking is quite low, and would get picked up in the scheduled run.

Cargo.toml
eitsupi: What do you mean?
lutra has `rust-version.workspace = true`

prqlc/prqlc/Cargo.toml
eitsupi: My point is that it is the `cli` feature that pushes the MSRV to 1.73, if excluding the feature, MSRV is 1.70.
However, there is no way to express that in the current Cargo.toml, so the `rust-version` here must be 1.73 and downstream crates cannot use Rust 1.70.

This is just the current situation, but my point is that it may be difficult to combine those that should keep MSRV as low as possible for downstream considerations and those that should not into a single crate and manage them with feature flags.
I'm fine with it as it is for now, but ideally the CLI should be separated into a separate crate like `prqlc-cli`.

Cargo.toml
max-sixty: OK I see. Possible we push that down to lutra, and leave this empty. 

Hopefully we can have an MSRV for most of the workspace set here, and then lutra can have its own higher one. But possibly that won't work yet.

prqlc/prqlc/Cargo.toml
max-sixty: Yes, I agree that's a downside with having them integrated. I do think there are benefits to having them in a single crate — in particular `cargo install prqlc`, but we have gone back & forth, there are tradeoffs...

eitsupi: > I do think there are benefits to having them in a single crate — in particular `cargo install prqlc`, but we have gone back & forth, there are tradeoffs...

I do not see how there is any problem with `cargo install prqlc` changing to `cargo install prqlc-cli`.
Or rename other than CLI to something like `prqlc-core`.

In any case, this is a separate issue that needs to be taken up.

eitsupi: #4946

Cargo.toml
eitsupi: Sorry, my understanding of MSRV was incorrect (foresterre/cargo-msrv#1023).

I will not work on it today so I will draft it and come back to it in a few days.

max-sixty: Ah great, thanks a lot for following up with that. I added a comment to the issue.

If that issue is resolved, I would vote to have an MSRV of something like 1.73.0 here, have most crates inherit it, and then Lutra has a different higher one.

PR #4949, opened 2024-10-11
build: Enable x86 build for MacOS

max-sixty:
Runners won't be around forever but makes sense to have them running while they are.

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -54,6 +54,10 @@ jobs:
             target: aarch64-apple-darwin
           - os: windows-latest
             target: x86_64-pc-windows-msvc
+          # Still building for intel but can disable when runners go away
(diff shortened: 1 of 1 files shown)


eitsupi:
> Runners won't be around forever but makes sense to have them running while they are.

As far as I know, there is no problem to cross-compile with macos-latest here.
prqlc-r build with macos-latest  (because it seems to be more powerful) and that binary passes tests with macos-13 in CI.
https://github.com/PRQL/prqlc-r/blob/e059a3fc1c8bae7451e8edb8d872dc599d7078ee/.github/workflows/release-lib.yml#L39-L42

max-sixty:
> As far as I know, there is no problem to cross-compile with macos-latest here.

Ah great, thanks. TIL!

PR #4960, opened 2024-10-13
ci: Switch to maintained zig action

max-sixty:


max-sixty:
@vanillajonathan if you want to update to a more recent version of zig, that would be great...

PR #4961, opened 2024-10-13
build: Upgrade Zig

vanillajonathan:
Build with the latest version of Zig and no longer pin the version to an old version.


max-sixty (approved):
Great, thank you!

Issue #5007, opened 2024-11-12
rm: cannot remove 'target': Device or resource busy


richb-hanover:
### What happened?

This is a refined report from #4974. It appears that the problem with my M2 Mac is that something is wrong with the file system in the Dev Container, and that files/directories cannot be modified.

I have removed all containers, images, and volumes for PRQL and the caches to rebuild the entire Dev Container. I don't intentionally have any external processes running on the Mac that would occupy the Dev Container file system.

@max-sixty @eitsupi @vanillajonathan Here's the evidence I have. What else could I collect? Thanks.

* wasm-pack is not installed, running `cargo install wasm-pack` gives this error:
   `error: failed to create directory /usr/local/cargo/registry/cache/index.crates.io-6f17d22bba15001f`

* `task web:run-book` gives this error:
   ```
   vscode ➜ /workspaces/prql (main) $ task web:run-book
   task: [web:run-book] mdbook serve --port=3000 -n 0.0.0.0
   2024-11-12 11:46:14 [INFO] (mdbook::book): Book building has started
       Updating crates.io index
     Downloaded colorchoice v1.0.1
   error: failed to create directory `/usr/local/cargo/registry/cache/index.crates.io-6f17d22bba15001f`
   ```

* `cd /workspaces/prql; rm -rf target` gives this error:
```
vscode ➜ /workspaces/prql (main) $ ls -al
total 296
drwxr-xr-x 35 vscode vscode   1120 Nov 12 00:49 .
drwxr-xr-x  3 root   root     4096 Nov  5 03:29 ..
-rw-r--r--  1 vscode vscode   1924 Oct 19 22:37 bacon.toml
drwxr-xr-x  3 vscode vscode     96 Oct 19 22:37 .cargo
-rw-r--r--  1 vscode vscode 122537 Nov 12 00:49 Cargo.lock
-rw-r--r--  1 vscode vscode   1670 Nov 12 00:49 Cargo.toml
-rw-r--r--  1 vscode vscode  59433 Nov  5 03:42 CHANGELOG.md
drwxr-xr-x  7 vscode vscode    224 Nov  5 03:42 .config
drwxr-xr-x  4 vscode vscode    128 Oct 19 22:37 .devcontainer
-rw-r--r--  1 vscode vscode  10244 Nov 12 03:20 .DS_Store
-rw-r--r--  1 vscode vscode   4571 Nov  5 03:42 flake.lock
[...]

Issue #4974, opened 2024-10-19
Can't build Playground - `sh -c npm run build` fails


richb-hanover:
### What happened?

Attempting to build Playground on M2 Mac in the Dev Container. Commit is 86becc59

When I run `task web:run-playground` I see the error below.

What other diagnostic info could I provide? Thanks.

### PRQL input

```elm
N/A
```

### SQL output

```SQL
N/A
```

### Expected SQL output

```SQL
N/A
```

### MVCE confirmation

- [X] Minimal example
- [X] New issue

### Anything else?

```
vscode ➜ /workspaces/prql (main) $ task web:run-playground
task: [web:build-playground-dependencies] npm ci
npm error code 1
npm error path /workspaces/prql/prqlc/bindings/js
npm error command failed
npm error command sh -c npm run build
npm error > prqlc@0.13.3 build
npm error > npm run build:node && npm run build:web && npm run build:bundler
npm error
npm error
npm error > prqlc@0.13.3 build:node
npm error > npx cross-env wasm-pack build --target nodejs --out-dir dist/node --${PROFILE} && rm dist/node/.gitignore
npm error node:events:497
npm error       throw er; // Unhandled 'error' event
npm error       ^
npm error
npm error Error: spawn wasm-pack ENOENT
npm error     at ChildProcess._handle.onexit (node:internal/child_process:286:19)
npm error     at onErrorNT (node:internal/child_process:484:16)
npm error     at process.processTicksAndRejections (node:internal/process/task_queues:82:21)
npm error Emitted 'error' event on ChildProcess instance at:
npm error     at ChildProcess._handle.onexit (node:internal/child_process:292:12)
npm error     at onErrorNT (node:internal/child_process:484:16)
npm error     at process.processTicksAndRejections (node:internal/process/task_queues:82:21) {
npm error   errno: -2,
npm error   code: 'ENOENT',
npm error   syscall: 'spawn wasm-pack',
npm error   path: 'wasm-pack',
npm error   spawnargs: [ 'build', '--target', 'nodejs', '--out-dir', 'dist/node', '--dev' ]
npm error }
npm error
npm error Node.js v20.18.0
npm error A complete log of this run can be found in: /home/vscode/.npm/_logs/2024-10-19T22_57_39_755Z-debug-0.log
task: Failed to run task "web:run-playground": exit status 1
vscode ➜ /workspaces/prql (main) $ 
```

max-sixty:
[...]

PR #5018, opened 2024-11-22
build: Set owner on /usr/local/cargo/registry/ in Dev Container

richb-hanover:
This PR seems to address #5007 and #4974 which were related to reports of "not being able to build the Playground from a fresh installation of the Dev Container."

The answer seems to be to add `sudo chown vscode /usr/local/cargo/registry/` to the postCreateCommand section of the Dev Container's JSON file.

There were SO MANY red herrings here... 

* Some operations gave an error trying to create a file in `/usr/local/cargo/registry...` (this was the hint toward the solution)

* The initial report gave a mysterious ENOENT error out of the blue. That seemed to be caused by the failure to have installed wasm_pack. Using `npm install wasm-pack` caused that to go away.

* `cargo clean` does indeed remove all the cached cargo files, but gives the "resource busy" on _/workspaces/prql/target_ (which remains today)

I am content. I'm back to a steady state - I can build the Playground, Book, and Website from a dev container.


.devcontainer/devcontainer.json
```diff
@@ -39,7 +39,7 @@
     }
   ],
   "postCreateCommand": {
-    "set-ownership": "sudo chown vscode target",
+    "set-ownership": "sudo chown vscode target; sudo chown vscode /usr/local/cargo/registry/",
```
eitsupi: This should work too.

```suggestion
    "set-ownership": "sudo chown vscode target /usr/local/cargo/registry/",
```

eitsupi (changes requested):
Thanks, this seems fine.

eitsupi (approved):

PR #5032, opened 2024-12-02
feat: enable Link-Time Optimization (LTO)

### LINKED ISSUE #5031: Enable Link-Time Optimization (LTO)
### What's up?

Hi!

I noticed that Link-Time Optimization (LTO) is not enabled for the project. I suggest switching it on since it will reduce the binary size (always a good thing to have - especially if you already enabled `opt-level = "s"`) and will likely improve the application's performance (its CPU part). If you want to read more about LTO, I can recommend starting from [this](https://doc.rust-lang.org/rustc/codegen-options/index.html#lto) Rustc documentation.

I suggest enabling LTO only for the Release builds so as not to sacrifice the developers' experience while working on the project since LTO consumes an additional amount of time to finish the compilation routine. If you think that a regular Release build should not be affected by such a change as well, then I suggest adding an additional `dist` or `release-lto` profile where in addition to regular `release` optimizations LTO will also be added. Such a change simplifies life for maintainers and others interested in the project persons who want to build the most performant version of the application. E.g., check `cargo-outdated` Release [profile](https://github.com/kbknapp/cargo-outdated/blob/master/Cargo.toml#L48).

Basically, it can be enabled with the following lines:

```
[profile.release]
lto = true
```

I have made quick tests (Fedora 41, Rustc 1.82) by adding `lto = true` to the Release profile. The binary size reduction is the following:
* `prqlc`: from 19 Mib to 16 Mib
* `mdbook-prql`: from 16 to 14 Mib

Thank you

zamazan4ik:
Resolves https://github.com/PRQL/prql/issues/5031

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -26,6 +26,7 @@ rust-version = "1.70.0"
 version = "0.13.3"
 
 [profile.release]
+lto = true
 # Optimize for binary size in releases of all crates,
 # since compiler is fast enough as it is (for now).
 opt-level = "s"


max-sixty:
Thanks @zamazan4ik !

PR #5045, opened 2024-12-09
docs: aggregate doc: within a group, it produces one {row => column*} for each group

rvion:


max-sixty:
But I think it is one row from each group? i.e. `SELECT foo FROM bar GROUP BY baz` reduces the number of rows, not columns?

PR #5112, opened 2025-01-24
docs: Ceci n'est pas une pipe

lovasoa:
https://en.wikipedia.org/wiki/The_Treachery_of_Images


max-sixty:
Ha, excellent PR, TY!

(I will see what's going on with that test and fix)

max-sixty:
I fixed the name of some snapshots, since the name of a heading changed

PR #5119, opened 2025-01-27
refactor: remove generics

aljazerzen:
- **remove generics**
- **remove difference type**

Edit: rationale

Generics and difference types were meant to replace lineage, which is used to know which columns are in-scope at each location in the query.

It turns out that this would be very hard to accomplish due all rules around name resolving that PRQL has. So a few months ago, I abandoned that effort in favor of sticking with current name resolution implementation, based on lineage.

This PR is now cleaning out code that has been unused (except in a few "resolving" tests).


vanillajonathan:
It would be informative if the description for the PR contained the rationale behind it.

PR #5128, opened 2025-01-28
fix: revert sorts on all CTEs (in PQ postprocessing)

kgutwin:
This is an addendum to PR #5098. After more testing, another failing case was discovered, which highlighted the need to revert sort columns after processing each CTE, not just those that have Sort expressions in them.


max-sixty:
Thanks @kgutwin !

PR #5133, opened 2025-01-30
feat(prqlc cli)!: `prqlc` with no argument will display help without error

### LINKED ISSUE #5131: [cli] `prqlc` command should error?
The automatic validation in the winget-pkgs repository fails every time because of the `prqlc` command error.
See https://github.com/microsoft/winget-pkgs/pull/181458#issuecomment-2408308351
It seems to be caused by the `prqlc` command failing without arguments.

For example, the `rustc` command is not an error.

```sh
$ rustc >/dev/null; echo $?
0
$ prqlc &>/dev/null; echo $?
2
```

~~@max-sixty Please check microsoft/winget-pkgs#216745.~~ Solved

eitsupi:
Close #5131

CLI tools created with clap usually display help as an error when run without arguments (Context clap-rs/clap#2767).
However, there are CLI tools, such as `rustc` and `cargo`, that display help as normal help rather than errors, which may be more reasonable here.

### DIFF
--- a/prqlc/prqlc/src/cli/mod.rs
+++ b/prqlc/prqlc/src/cli/mod.rs
@@ -41,7 +41,7 @@ mod watch;
 
 /// Entrypoint called by [`crate::main`]
 pub fn main() -> color_eyre::eyre::Result<()> {
-    let mut cli = Cli::parse();
+    let cli = Cli::parse();
 
     // redirect all log messages into the [debug::DebugLog]
     if has_debug_log(&cli) {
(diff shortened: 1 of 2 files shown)


max-sixty (approved):
No strong view on the best way of doing this, fine with me if we think this is preferable!

eitsupi:
Thanks.
From https://github.com/clap-rs/clap/issues/2767#issuecomment-933823164, it seems that clap's default behavior is to allow users to choose which behavior, so I think there are no problem for us to make changes to avoid unnecessary errors.

PR #5134, opened 2025-02-03
revert: remove lutra, the query runner stub

aljazerzen:
As lutra has never been implemented enough to be useful, I'm removing it to reduce maintenance costs.


max-sixty:
"Lutra is dead. Long live Lutra!"

PR #5173, opened 2025-03-05
fix: escape BigQuery reserved keywords (#5166)

### LINKED ISSUE #5166: "hash" column name is not escaped in Big Query
### What happened?

Cannot select a column named 'hash' because BigQuery recognizes 'hash' as a function. PRQL does not escape the column name (by adding backticks) nor does it preserve existing backticks, as shown in the example below.

### PRQL input

```elm
prql target:sql.bigquery

from `some_dataset_id.demo`
select {
  `hash`
}
```

### SQL output

```SQL
SELECT
  hash
FROM
  `some_dataset_id.demo`

-- Generated by PRQL compiler version:0.13.3 (https://prql-lang.org)
```

### Expected SQL output

```SQL
SELECT
  `hash`
FROM
  `some_dataset_id.demo`
```

### MVCE confirmation

- [x] Minimal example
- [x] New issue

### Anything else?

_No response_

nsiregar:
Escape GoogleSQL reserved keywords by extend list of reserved keywords with GoogleSQL reserved keywords

source GoogleSQL reserved keywords:
https://cloud.google.com/bigquery/docs/reference/standard-sql/lexical#reserved_keywords


max-sixty:
That looks very reasonable, thank you!

max-sixty:
(if you want to add a changelog in another PR then great, but no obligation)

PR #5185, opened 2025-03-12
feat: Add syntax highlighting for Vim

vanillajonathan:
Adds a syntax highlight for use with the [Vim](https://www.vim.org/) text editor and [Neovim](https://neovim.io/).


max-sixty:
Looks good!

Re the error — we may have to do one PR with the path, and then another with the link, or add it to the exclusions at first. (feel free to go ahead, no need to wait for approval for each)

snth:
Thanks for this @vanillajonathan !

PR #5189, opened 2025-03-16
feat: Add syntax highlighting for Emacs

vanillajonathan:
Adds a major mode for the [GNU Emacs](https://www.gnu.org/software/emacs/) text editor which provides basic syntax highlighting.

It is basic. It doesn't use treesitter, nor does it provide code folding, formatting, REPL or anything.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@
 
 - Add syntax highlight file for KSyntaxHighlighting. (@vanillajonathan, #5177)
 - Add syntax highlight file for Vim. (@vanillajonathan, #5185)
+- Add syntax highlight file for GNU Emacs. (@vanillajonathan, #5189)
 
 **Internal changes**:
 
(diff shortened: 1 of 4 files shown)


max-sixty:
I thought that emacs supports tree-sitter now? Or not some versions / I'm mistaken?

vanillajonathan:
Yes it does, modern versions of GNU Emacs (since version 29.1) now supports tree-sitter but the mode in this PR does not make use of the tree-sitter because I don't use Emacs and I don't know Emacs Lisp or Lisp so I wouldn't know how to do anything like that, hence I made just a very basic mode.

max-sixty:
OK, no strong objection to merging, but I'm guessing folks will just use the tree-sitter grammar rather than this... I had thought that if there's a tree-sitter grammar, then users can add it (but I also don't use emacs)

vanillajonathan:
I don't know how users would add it, maybe they could contribute to this script to make it use tree-sitter and other features such as code formatter, code folding, indention and REPL.

PR #5194, opened 2025-03-18
fix(Lezer grammar): Pipelines require paranthesis

vanillajonathan:
Since prqlc 0.13.0 we require parenthesis around pipelines. See #4775


max-sixty:
Nice, thanks!

PR #5196, opened 2025-03-18
fix(Lezer grammar): Add support for exponentiation

vanillajonathan:
Since prqlc 0.13.0 exponentiation is supported. #4125


max-sixty:
Looks good!

PR #5197, opened 2025-03-20
feat: Add LSP stub

vanillajonathan:
Adds a stub implementations of a basic Language Server Protocol (LSP) to `prqlc`. LSP clients can connect to it and disconnect from it.

It use the `lsp-server` crate used by the official rust-analyzer.

The LSP specification is at https://microsoft.github.io/language-server-protocol/specifications/specification-current


max-sixty:
looks cool, would be great to build this out in time!

to reduce dependencies in there meantime, could we add an optional `lsp` feature?

vanillajonathan:
Okay, I made the dependency optional.

Yes it would be cool build this out in time. The next steps we need to do is to create `match` statements for `DidOpenTextDocument`, `DidChangeTextDocument` and `DidCloseTextDocument` then proceed with `Completion`, `Formatting` and `HoverRequest`.

max-sixty:
Looks good! May need some refining on the tests outputs but good to go otherwise

vanillajonathan:
I think no more tests are needed for now because we shouldn't test the `lsp-server` crate. We can add more tests later when we add more handlers such as `DidOpenTextDocument`, `DidChangeTextDocument` and `DidCloseTextDocument`.

The test fails on some test-deps-min-version (verify Rust minimum version) thing.

Can you merge this?

max-sixty:
> The test fails on some test-deps-min-version (verify Rust minimum version) thing.

OK, but we can't have `main` failing, so we need to fix that...

It seems to additionally be failing because the completion snapshot files need updating: https://github.com/PRQL/prql/actions/runs/14046209187/job/39327515449?pr=5197

vanillajonathan:
Okay, I updated shell completion snapshots.

prqlc/prqlc/Cargo.toml
```diff
@@ -22,12 +22,14 @@ cli = [
   "color-eyre",
   "colorchoice-clap",
   "is-terminal",
+  "lsp",
```
max-sixty: but `lsp` shouldn't be in cli by default, then we're including features that introduce dependencies without any functionality

does that make sense? I think you can recognize this!

prqlc/prqlc/Cargo.toml
```diff
@@ -22,12 +22,14 @@ cli = [
...
   "minijinja",
   "notify",
   "serde_yaml",
   "walkdir",
 ]
 default = ["cli"]
+lsp = ["lsp-server", "lsp-types"]
```
max-sixty: Can we add a note that this is just a stub at the moment; doesn't add functionality yet?

prqlc/prqlc/Cargo.toml
max-sixty: (it's possible that fixing this _undoes_ the changes to the tests, apologies if so — they should all reset with the standard test commands at least)

vanillajonathan: Yeah, that makes sense.

prqlc/prqlc/Cargo.toml
vanillajonathan: Sure!

max-sixty (approved):
lgtm!

PR #5223, opened 2025-04-02
internal: upgrade chumsky to 0.10 in lexer

max-sixty:
chumsky 0.10 is a total rewrite from 0.9, and much of our code also needs rewriting if we're to upgrade

this creates a feature that uses chumsky 0.10 _only_ for the lexer, as a way of making progress. claude code helped a lot, which benefits from smaller problems. 

before merging, we would keep only the chumsky 0.10 code for the lexer (and possibly upgrade the parser, I'm not sure)

so far, it _mostly_ works; only a couple of failing tests. I'm not sure how to get our "any number of odd quotes" lexing correctly, and there's a small issue with the end of input. I'll ask about the first over at chumsky

(this isn't the most important thing for PRQL, but I'm trying to get back into it, and wanted to spend some time trying a moderately difficult project with LLMs, so this seemed like a reasonable case. though also I thought it would be much easier! very open to feedback on whether there's a way to more incrementally make the changes, rather than basically rewriting from scratch. chumsky's type errors don't make it easy to play whack-a-mole)


prqlc/prqlc-parser/src/lexer/chumsky_0_10.rs
```diff
@@ -0,0 +1,793 @@
...
+    }))
+    .with_source(ErrorSource::Lexer(format!(
+        "Unexpected {} at position {}..{}",
+        found, error_start, error_end
+    )))
+}
+
+/// Lex PRQL into LR, returning both the LR and any errors encountered
+pub fn lex_source_recovery(source: &str, source_id: u16) -> (Option<Vec<Token>>, Vec<E>) {
+    let stream = Stream::from_iter(source.chars());
```
zesterer: `Stream` is no longer needed: the `Input` trait is implemented [for many types](https://docs.rs/chumsky/latest/chumsky/guide/_01_key_concepts/index.html#the-input-trait), including `&str` itself. Using `&str` directly has the nice side effect of being much faster too :)

prqlc/prqlc-parser/src/lexer/chumsky_0_10.rs
```diff
@@ -0,0 +1,793 @@
...
+    .collect()
+}
+
+/// Lex chars to tokens until the end of the input
+pub fn lexer<'a>() -> impl Parser<'a, ParserInput<'a>, Vec<Token>, ParserError<'a>> {
+    lex_token()
+        .repeated()
+        .collect()
+        .then_ignore(ignored())
+        .then_ignore(end())
```
zesterer: `.then_ignore(end())` is no longer required as of 0.10: parsers now have consume-all-or-error behaviour by default. You can opt into the old behaviour with `.lazy()` if you wish though.

prqlc/prqlc-parser/src/lexer/lr_lexer.rs
```diff
@@ -0,0 +1,793 @@
...
+        .then_ignore(end())
+}
+
+/// Lex chars to a single token
+fn lex_token<'a>() -> impl Parser<'a, ParserInput<'a>, Token, ParserError<'a>> {
+    // Handle range token with proper whitespace
+    // Ranges need special handling since the '..' token needs to know about whitespace
+    // for binding on left and right sides
+    let range = whitespace()
+        .or_not()
```
zesterer: `whitespace()` parses any number of whitespace characters (including none at all!) so the `.or_not()` is redundant. If you're looking to count the whitespace characters parsed, you can do `whitespace().count()` and you'll get a `usize` as the output.

prqlc/prqlc-parser/src/lexer/chumsky_0_10.rs
```diff
@@ -0,0 +1,793 @@
...
+                // Regular quoted string
+                quoted_string(true),
+            )),
+        )
+        .map(|(c, s)| TokenKind::Interpolation(c, s))
+}
+
+fn ignored<'a>() -> impl Parser<'a, ParserInput<'a>, (), ParserError<'a>> {
+    whitespace().repeated().ignored()
+}
```
zesterer: As mentioned, `whitespace()` is [already implicitly `.repeated()`](https://docs.rs/chumsky/latest/chumsky/text/fn.whitespace.html) so this extra `.repeated()` is redundant. Also, `.repeated()` now outputs `()` by default so the `.ignored()` is also redundant. All in all, this entire function basically just behaves exactly the same as `whitespace()` on its own.

prqlc/prqlc-parser/src/lexer/chumsky_0_10.rs
```diff
@@ -0,0 +1,793 @@
...
+        });
+
+    // Handle all other token types with proper whitespace
+    let other_tokens = ignored().ignore_then(token().map_with(|kind, extra| {
+        let span: chumsky_0_10::span::SimpleSpan = extra.span();
+        Token {
+            kind,
+            span: span.start()..span.end(),
+        }
+    }));
```
zesterer: Hopefully when https://github.com/zesterer/chumsky/pull/713 merges, code like this will get a lot simpler:

```rs
let other_tokens = ignored().ignore_then(token().spanned());
```

prqlc/prqlc-parser/src/lexer/chumsky_0_10.rs
```diff
@@ -0,0 +1,793 @@
...
+        just("prql"),
+        just("type"),
+        just("module"),
+        just("internal"),
+        just("func"),
+        just("import"),
+        just("enum"),
+    ))
+    .then_ignore(end_expr())
+    .map(|x| TokenKind::Keyword(x.to_string()))
```
zesterer: `just(x)` will actually return a copy of `x`, not the token(s) it matched. For this reason, you can make your `TokenKind::Keyword` variant contain a `&'static str` and avoid the allocation here:

```rs
choice((
    just("let"),
    ...
))
    .then_ignore(end_expr())
    .map(TokenKind::Keyword)
```

prqlc/prqlc-parser/src/lexer/lr_lexer.rs
```diff
@@ -0,0 +1,793 @@
...
+}
+
+fn param<'a>() -> impl Parser<'a, ParserInput<'a>, TokenKind, ParserError<'a>> {
+    just('$')
+        .ignore_then(
+            any()
+                .filter(|c: &char| c.is_alphanumeric() || *c == '_' || *c == '.')
+                .repeated()
+                .collect::<String>(),
+        )
```
zesterer: I realise that you're probably prioritising migrating over trying to eek out performance so feel free to ignore this, but in the future you could have your `TokenKind::Param` variant contain a `&'src str` and use `.to_slice()` here instead of `.collect::<String>()`, saving an allocation.

prqlc/prqlc-parser/src/lexer/chumsky_0_10.rs
```diff
@@ -0,0 +1,793 @@
...
+fn interpolation<'a>() -> impl Parser<'a, ParserInput<'a>, TokenKind, ParserError<'a>> {
+    // For s-strings and f-strings, we need to handle both regular and triple-quoted variants
+    one_of("sf")
+        .then(
+            // Use a custom quoted_string implementation that better handles triple quotes
+            choice((
+                // Triple quote strings for s-strings
+                just('"')
+                    .then(just('"'))
+                    .then(just('"'))
```
zesterer: This is the same as `just(['"'; 3])` or `just("\"\"\"")` (although I personally prefer the first one)

prqlc/prqlc-parser/src/lexer/chumsky_0_10.rs
```diff
@@ -0,0 +1,793 @@
...
+}
+
+fn ignored<'a>() -> impl Parser<'a, ParserInput<'a>, (), ParserError<'a>> {
+    whitespace().repeated().ignored()
+}
+
+fn whitespace<'a>() -> impl Parser<'a, ParserInput<'a>, (), ParserError<'a>> {
+    any()
+        .filter(|x: &char| *x == ' ' || *x == '\t')
+        .repeated()
```
zesterer: This is equivalent to [`text::inline_whitespace()`](https://docs.rs/chumsky/latest/chumsky/text/fn.inline_whitespace.html)

prqlc/prqlc-parser/src/lexer/chumsky_0_10.rs
```diff
@@ -0,0 +1,793 @@
...
+        )
+        .then_ignore(whitespace().repeated())
+        .then_ignore(just('\\'))
+        .map(TokenKind::LineWrap)
+}
+
+fn comment<'a>() -> impl Parser<'a, ParserInput<'a>, TokenKind, ParserError<'a>> {
+    // Extract the common comment text parser
+    let comment_text = any()
+        .filter(|c: &char| *c != '\n' && *c != '\r')
```
zesterer: `none_of("\n\r")` might give better error messages here. Sadly, `filter` currently generates worse errors because chumsky can't automatically understand what the closure is *supposed* to be looking for. We've a plan to fix that though.

prqlc/prqlc-parser/src/lexer/chumsky_0_10.rs
```diff
@@ -0,0 +1,793 @@
...
+
+pub fn ident_part<'a>() -> impl Parser<'a, ParserInput<'a>, String, ParserError<'a>> {
+    let plain = any()
+        .filter(|c: &char| c.is_alphabetic() || *c == '_')
+        .then(
+            any()
+                .filter(|c: &char| c.is_alphanumeric() || *c == '_')
+                .repeated()
+                .collect::<Vec<char>>(),
+        )
```
zesterer: Equivalent to `text::ascii::ident()`

prqlc/prqlc-parser/src/lexer/chumsky_0_10.rs
```diff
@@ -0,0 +1,793 @@
...
+            let mut chars = vec![first];
+            chars.extend(rest);
+            chars.into_iter().collect::<String>()
+        });
+
+    let backtick = none_of('`')
+        .repeated()
+        .collect::<Vec<char>>()
+        .delimited_by(just('`'), just('`'))
+        .map(|chars| chars.into_iter().collect::<String>());
```
zesterer: Can become

```rs
let backtick = none_of('`')
        .repeated()
        .collect::<String>()
        .delimited_by(just('`'), just('`'));
```

prqlc/prqlc-parser/src/lexer/chumsky_0_10.rs
```diff
@@ -0,0 +1,793 @@
...
+
+    choice((plain, backtick))
+}
+
+// Date/time components
+fn digits<'a>(count: usize) -> impl Parser<'a, ParserInput<'a>, Vec<char>, ParserError<'a>> {
+    any()
+        .filter(|c: &char| c.is_ascii_digit())
+        .repeated()
+        .exactly(count)
```
zesterer: Equivalent to `text::ascii::digits(10).exactly(count)`

prqlc/prqlc-parser/src/lexer/chumsky_0_10.rs
```diff
@@ -0,0 +1,793 @@
...
+
+- and the linting instructions in `CLAUDE.md`
+
+# Chumsky 0.10.0 Lexer Implementation
+*/
+
+use chumsky_0_10::extra;
+use chumsky_0_10::input::Stream;
+use chumsky_0_10::prelude::*;
+use chumsky_0_10::primitive::{choice, end, just, none_of, one_of};
```
zesterer: All primitives are included in the prelude now

prqlc/prqlc-parser/src/lexer/chumsky_0_10.rs
```diff
@@ -0,0 +1,793 @@
...
+use super::lr::{Literal, Token, TokenKind, Tokens, ValueAndUnit};
+use crate::error::{Error, ErrorSource, Reason, WithErrorInfo};
+
+type E = Error;
+type ParserInput<'a> = Stream<std::str::Chars<'a>>;
+// Define a custom error type with the `Simple` error type from chumsky_0_10
+type ParserError<'a> = extra::Err<Simple<'a, char>>;
+
+/// Convert a chumsky Simple error to our internal Error type
+fn convert_lexer_error(error: &Simple<'_, char>, source_id: u16) -> E {
```
zesterer: Confusingly, `Simple` in chumsky 0.9 became [`Rich`](https://docs.rs/chumsky/latest/chumsky/error/struct.Rich.html) in chumsky 0.10. [`Simple`](https://docs.rs/chumsky/latest/chumsky/error/struct.Simple.html) now refers to a simpler error type that only tracks the span and the found token.

That said, it seems like you're only actually making use of the found token and the span, so `Simple` is probably fine for you. If you wanted richer error information, `Rich` has a lot to offer like contexts, labels, etc.
image

prqlc/prqlc-parser/src/lexer/chumsky_0_10.rs
```diff
@@ -0,0 +1,793 @@
...
+        .map(|((((year, dash1), month), dash2), day)| {
+            format!(
+                "{}{}{}{}{}",
+                String::from_iter(year),
+                dash1,
+                String::from_iter(month),
+                dash2,
+                String::from_iter(day)
+            )
+        })
```
zesterer: This can be changed to:

```rs
text::digits(10).exactly(4)
    .then(just('-'))
    .then(text::digits(10).exactly(2))
    .then(just('-'))
    .then(text::digits(10).exactly(2))
    .to_slice()
```

This avoids all 4 allocations and just gives you back the input date as a string slice verbatim.

prqlc/prqlc-parser/src/lexer/lr_lexer.rs
```diff
@@ -0,0 +1,793 @@
...
+                |(hrs, (_opt_colon, mins))| {
+                    // Always format as -0800 without colon for SQL compatibility, regardless of input format
+                    // We need to handle both -08:00 and -0800 input formats but standardize the output
+                    format!("{}{}", String::from_iter(hrs), String::from_iter(mins))
+                },
+            ))
+            .map(|(sign, offset)| format!("{}{}", sign, offset)),
+    ))
+    .or_not()
+    .map(|opt| opt.unwrap_or_default());
```
zesterer: It's a shame that the optional `:` here prevents you from just using `.to_slice()` and avoiding allocating at all. Ah well.

prqlc/prqlc-parser/src/lexer/chumsky_0_10.rs
```diff
@@ -0,0 +1,793 @@
...
+        let triple_quote_open = just(quote).then(just(quote)).then(just(quote));
+        let triple_quote_close = just(quote).then(just(quote)).then(just(quote));
+
+        triple_quote_open
+            .ignore_then(
+                // Keep consuming characters until we hit three quotes in a row
+                any()
+                    .filter(move |&c| c != quote)
+                    .repeated()
+                    .collect::<Vec<char>>(),
```
zesterer: `none_of(quote).repeated()` seems fine here

prqlc/prqlc-parser/src/lexer/chumsky_0_10.rs
```diff
@@ -0,0 +1,793 @@
...
+) -> impl Parser<'_, ParserInput<'_>, Vec<char>, ParserError<'_>> {
+    let q = *quote;
+
+    // Parser for non-quote characters
+    let regular_char = if allow_multiline {
+        any().filter(move |c: &char| *c != q && *c != '\\').boxed()
+    } else {
+        any()
+            .filter(move |c: &char| *c != q && *c != '\n' && *c != '\r' && *c != '\\')
+            .boxed()
```
zesterer: You can avoid boxing here by using `none_of("...")` - the types will be compatible for both branches.

prqlc/prqlc-parser/src/lexer/chumsky_0_10.rs
```diff
@@ -0,0 +1,793 @@
...
+
+    // Parser for escaped characters if escaping is enabled
+    let escaped_char = choice((
+        just('\\').ignore_then(just(q)),            // Escaped quote
+        just('\\').ignore_then(just('\\')),         // Escaped backslash
+        just('\\').ignore_then(just('n')).to('\n'), // Newline
+        just('\\').ignore_then(just('r')).to('\r'), // Carriage return
+        just('\\').ignore_then(just('t')).to('\t'), // Tab
+        escaped_character(),                        // Handle all other escape sequences
+    ));
```
zesterer:
```rs
just('\\').ignore_then(choice((
    just(q),
    just('\\'),
    ...
)))
    .or(escaped_character())
```

Seems like a good way to avoid repetition here

zesterer (commented):
Hopefully the suggestions help :)

Additionally, I've written up a rough [migration guide](https://github.com/zesterer/chumsky/issues/745#issuecomment-2746175979) for chumsky 0.9 -> 0.10 that you might find useful.

prqlc/prqlc-parser/src/lexer/chumsky_0_10.rs
max-sixty: yes, I think we chose to use `Simple` for the lexer and the equivalent of `Rich` for the parser. if my memory serves me correctly, the equivalent of `Rich` caused some slow down. or is that unlikely to be true even for lexing tokens (we don't care _that_ much about performance...)?

max-sixty:
@zesterer thank you _very_ much for your review! greatly appreciated

prqlc/prqlc-parser/src/lexer/lr_lexer.rs
max-sixty: yeah, we've (maybe confusingly) implemented a function `whitespace` which has `.at_least(1)`...

vanillajonathan:
Any update on this?

max-sixty:
> Any update on this?

I found it really hard and got distracted by other things. I'd like to come back to it at some point; possibly when there are more docs for me & the LLMs to benefit from...

PR #5264, opened 2025-04-21
chore: Use lexical binding for prql-mode.el

vanillajonathan:
Lexical binding is best practice so should use it.

### DIFF
--- a/grammars/emacs/prql-mode.el
+++ b/grammars/emacs/prql-mode.el
@@ -1,4 +1,4 @@
-;;; prql-mode.el --- Major mode for PRQL language
+;;; prql-mode.el --- Major mode for PRQL language -*- lexical-binding: t;
 
 ;; URL: https://github.com/PRQL/prql
 ;; Keywords: languages


vanillajonathan:
@max-sixty If we would want this to be available on MELPA which is a package manager for GNU Emacs then we would have to move this into a separate repository. A suitable name for that repository would be `emacs-prql-mode`.

> **Dedicated SCM repository**
>    Keep each package in its own SCM (e.g., git) repository. This makes it possible to have a different version number for each package as MELPA stable looks at the SCM’s tags to assign version numbers to recipes.

https://github.com/melpa/melpa/blob/master/CONTRIBUTING.org#what-to-know-before-opening-a-pull-request

max-sixty:
thanks!

vanillajonathan:
Read the comment above.

max-sixty:
sorry, yes

I'm open to it if you want to do that! totally your call!

vanillajonathan:
Okay, lets do it, and add the `emacs` tag to the repository.

PR #5305, opened 2025-05-26
fix(prqlc): deduplicate selected items in gen_projection

### LINKED ISSUE #5302: `select` in joined pipeline causes invalid expressions to be referenced
### What happened?

Adding a `select` in a joined pipeline after a few columns have been derived results in CTEs referecing invalid columns

### PRQL input

```elm
from (s"SELECT a,b,c,d FROM my_table") 
join side:left  (
  s"SELECT e,f,g FROM other_table"
  derive {
    my_concatenated_col = f"{col} {other_col}"
  }
  group {my_concatenated_col} (aggregate { my_sum = first this.`my_concatenated_col`})
  sort {this.my_concatenated_col, my_sum}
  derive {new_name = my_sum}
  select {this.my_concatenated_col, this.new_name}
) (this.my_concatenated_col == that.my_concatenated_col)
```

### SQL output

```SQL
WITH table_0 AS (
  SELECT
    a,
    b,
    c,
    d
  FROM
    my_table
),
table_2 AS (
  SELECT
    e,
    f,
    g
  FROM
    other_table
),
table_3 AS (
  SELECT
    CONCAT(col, ' ', other_col) AS my_concatenated_col,
    FIRST_VALUE(my_concatenated_col) AS _expr_0
  FROM
    table_2
  GROUP BY
    CONCAT(col, ' ', other_col)
),
table_4 AS (
  SELECT
    my_concatenated_col,
    _expr_0 AS new_name,
    _expr_0
  FROM
    table_3
),
table_1 AS (
  SELECT
    my_concatenated_col,
    new_name,
    CONCAT(col, ' ', other_col) AS my_concatenated_col,
    FIRST_VALUE(my_concatenated_col) AS _expr_0
  FROM
    table_4
)
SELECT
  table_0.*,
  table_1.my_concatenated_col,
  table_1.new_name
FROM
  table_0
  LEFT JOIN table_1 ON table_0.my_concatenated_col = table_1.my_concatenated_col

-- Generated by PRQL compiler version:0.13.4 (https://prql-lang.org)
```

### Expected SQL output

```SQL

```

### MVCE confirmation

- [x] Minimal example
- [x] New issue

### Anything else?

`table_1` should not exist at all: `table_4` already does the selection work, and `my_concatenated_col` has already been defined.

The following changes result in valid SQL being generate:
* Removing the `derive {new_name = my_sum}` line and `select`ing `this.my_sum` below
* Removing the `select` line altogether
* Removing the `group` line

lukapeschke:
fixes #5302

### DIFF
--- a/prqlc/prqlc/src/sql/gen_projection.rs
+++ b/prqlc/prqlc/src/sql/gen_projection.rs
@@ -135,27 +135,6 @@ fn deduplicate_select_items(items: &mut Vec<SelectItem>) {
         SelectItem::ExprWithAlias { alias, .. } => seen.insert(alias.clone()),
         _ => true,
     });
-
-    // Dropping all expressions which are already selected as an alias
-    let compounds_with_aliases = items
-        .iter()
-        .filter_map(|select_item| {
-            if let SelectItem::ExprWithAlias { expr, .. } = select_item {
-                if matches!(expr, sql_ast::Expr::CompoundIdentifier(_)) {
-                    Some(expr.clone())
-                } else {
-                    None
-                }
-            } else {
-                None
-            }
-        })
-        .collect::<HashSet<_>>();
-
-    items.retain(|select_item| match select_item {
-        SelectItem::UnnamedExpr(expr) => !compounds_with_aliases.contains(expr),
-        _ => true,
-    });
 }
 
 pub(super) fn translate_select_items(
--- a/prqlc/prqlc/src/sql/gen_projection.rs
+++ b/prqlc/prqlc/src/sql/gen_projection.rs
@@ -120,6 +120,44 @@ pub(super) fn translate_wildcards(ctx: &AnchorContext, cols: Vec<CId>) -> (Vec<C
     (output, excluded)
 }
 
+fn deduplicate_select_items(items: &mut Vec<SelectItem>) {
+    // Dropping all duplicated identifiers
+    let mut seen = HashSet::new();
+    items.retain(|select_item| match select_item {
+        SelectItem::UnnamedExpr(expr) => {
+            if let sql_ast::Expr::CompoundIdentifier(idents) = expr {
+                // If any of the identifiers hadn't been seen yet, retain the expr
+                idents.iter().any(|ident| seen.insert(ident.clone()))
+            } else {
+                true
+            }
+        }
+        SelectItem::ExprWithAlias { alias, .. } => seen.insert(alias.clone()),
+        _ => true,
+    });
+
+    // Dropping all expressions which are already selected as an alias
+    let compounds_with_aliases = items
+        .iter()
+        .filter_map(|select_item| {
+            if let SelectItem::ExprWithAlias { expr, .. } = select_item {
+                if matches!(expr, sql_ast::Expr::CompoundIdentifier(_)) {
+                    Some(expr.clone())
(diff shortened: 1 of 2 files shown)


lukapeschke:
After spending a while digging into the intermediary steps, this is my naive attempt at solving #5302 .

However I'm not satisfied with it: Deduplicating what we have in `SELECT` statements seems kinda magic, and might not be what we want in some cases. Also, I believe this should be solved earlier in the compilation process. 

I see a few things in [the SQL output](https://github.com/PRQL/prql/issues/5302#issue-3080380124) that could be solved:
1. `table_4` could be completely removed if ` FIRST_VALUE(my_concatenated_col)` was directly selected as `"new_name"`
2. `CONCAT(col, ' ', other_col) AS my_concatenated_col` and `_expr_0` are not needed in `table_1`. I guess a solution I see here would be to process S-string tables with `sqlparser` to see if there is an explicit column selection. This would allow to not pass `Wildcard` around, and avoid the `SELECT table_0.*` in the final query. 
3. `table_1` could just reference `my_concatenated_col` and `_expr_0` by alias. Not sure how to pass the information "this expression has been evaluated in a previous CTE and does not need to be re-evaluated" around though.

@max-sixty do you have any ideas/hints on how I could move forward with this ?  Thanks in advance!

max-sixty:
hi @lukapeschke, and thanks for the contribution!

is there an example of the proposed code behaving _worse_ than the existing code? otherwise this seems like an upgrade...

(I agree it's not _perfect_, but making incremental improvements is still good...)

lukapeschke:
@max-sixty I've no example of this behaving worse than existing code. 

You're right, we can proceed with incremental improvements. I simplified `deduplicate_select_items` to ensure the generated SQL query is valid, and added an extra alias to the test case to ensure it is taken into account in the final query. 

I'll try  to address the other issues in different PRs if that's okay with you ? 

My next step will be to try to change ` FIRST_VALUE(name) AS _expr_0` to just `_expr_0` in `table_0`, so the SQL query is valid

lukapeschke:
https://github.com/PRQL/prql/pull/5310 adresses the [second point in this comment](https://github.com/PRQL/prql/pull/5305#issuecomment-2909866580). I guess this needs to be done earlier in the process to prevent `_expr_0` from being selected, but I wasn't quite sure where exactly table declarations get created from s-strings

max-sixty:
sorry I missed this @lukapeschke ! I missed a couple of others too, but am back & merging things!

I'll merge this. I'm a tiiiiny bit hesitant that we're just removing any duplicate columns — there may be times that people expect duplicate columns to persist, and ideally we'd solve this at a different stage in the compiler

but given where we're at, the result does seem better, and endlessly pushing off small improvements in the hope of fixing underlying compiler issues doesn't seem like a successful strategy :)

thank you for the contribution!

lukapeschke:
Thanks a lot for the merge @max-sixty :) 

I think I've finally found the underlying issue, details here: https://github.com/PRQL/prql/issues/5302#issuecomment-2965778150 

I'm not really sure about how to proceed to fix it though, I'm not sure what the condition should be to determine if the sort columns should be reverted or not... Maybe you have an idea ?

Also, could you please re-open #5302 since it's not completely fixed yet ? I don't have the rights to do it

PR #5310, opened 2025-05-27
feat(prqlc): extract SQL column names from s-string tables when possible

lukapeschke:
relates to #5302


lukapeschke:
This is an attempt at simplifying the SQL query in #5302 . It does not remove unwanted columns but at least makes the final SQL query more explicit

max-sixty:
great, this looks good; I agree the SQL is better! thanks @lukapeschke

Issue #4724, opened 2024-07-13
append doesn't check column, column order matters


cottrell:
### What happened?

I'm not sure if this is intentional or not.

### PRQL input

```elm
from invoices
group billing_country (
aggregate {
  a = count_distinct billing_city
  }
)
into A

from invoices
aggregate {
a = count_distinct billing_city
}
derive {
billing_country = '*'
}
select {a, billing_country}
into B

from A
append B
```

### SQL output

```SQL
WITH "A" AS (
  SELECT
    billing_country,
    COUNT(DISTINCT billing_city) AS a
  FROM
    invoices
  GROUP BY
    billing_country
),
table_0 AS (
  SELECT
    COUNT(DISTINCT billing_city) AS a
  FROM
    invoices
),
"B" AS (
  SELECT
    a,
    '*' AS billing_country
  FROM
    table_0
)
SELECT
  billing_country,
  a
FROM
  "A"
UNION
ALL
SELECT
  *
FROM
  "B"

-- Generated by PRQL compiler version:0.12.2 (https://prql-lang.org)
```

### MVCE confirmation

- [ ] Minimal example
- [ ] New issue

### Anything else?

It works if you manually align the columns with a select. Might be intended not sure.

```
billing_country	a
Germany	3
Norway	1
Belgium	1
Canada	8
USA	12
France	4
Ireland	1
United Kingdom	2
Australia	1
Chile	1
India	2
Brazil	4
Portugal	2
Netherlands	1
Spain	1
Sweden	1
Czech Republic	1
Finland	1
Denmark	1
Italy	1
Poland	1
Austria	1
Hungary	1
Argentina	1
53	*
```

max-sixty:
Sorry for being slow, but could you explain in a bit more detail what the issue is?

(An expected SQL output would be great)

cottrell:
> Sorry for being slow, but could you explain in a bit more detail what the issue is?
> 
> (An expected SQL output would be great)

If you look at the last row it is "53  *" and it shoudl be "* 53". I should have chosen a better text than "*" I guess as it doesn't jump out.

If it's possible, prql should probably inspect the columns and align them but if the column names are no available (select * clause) then it is not possible. 

I think the simplest solution would be to simply warn in the docs for append that the user must manually align the tables. 

```
WITH "A" AS (
  SELECT
    billing_country,
    COUNT(DISTINCT billing_city) AS a
  FROM
    invoices
  GROUP BY
    billing_country
),
table_0 AS (
[...]

Issue #3579, opened 2023-09-27
Aggregate doesn't respect union's names


avikam:
### What happened?

Group aggregation after append doesn't preserve column alias (nor order) which leads to broken aggregations 

### PRQL input

```elm
prql target:sql.snowflake

let shape = p -> (
  p
  select {
    interaction_day = s"DATE_TRUNC('day', ts)",
    user_id = s"substr(user, length('u:') + 1)",
  }
)

from table1
shape
append (from  table2 | shape)
group { user_id } (
  aggregate {
    daily_interaction = count interaction_day
  }
)
```

### SQL output

```SQL
WITH table_0 AS (
  SELECT
    DATE_TRUNC('day', ts) AS interaction_day,
    substr(user, length('u:') + 1) AS user_id
  FROM
    table2
),
table_1 AS (
  SELECT
    substr(user, length('u:') + 1) AS user_id,
    DATE_TRUNC('day', ts) AS _expr_0
  FROM
    table1
  UNION
  ALL
  SELECT
    *
  FROM
    table_0
)
SELECT
  user_id,
  COUNT(*) AS daily_interaction
FROM
  table_1
GROUP BY
  user_id

-- Generated by PRQL compiler version:0.9.5 (https://prql-lang.org)
```

### Expected SQL output

```SQL
WITH table_0 AS (
  SELECT
    DATE_TRUNC('day', ts) AS interaction_day,
    substr(user, length('u:') + 1) AS user_id
  FROM
    table2
),
table_1 AS (
  SELECT
    substr(user, length('u:') + 1) AS user_id,
    DATE_TRUNC('day', ts) AS interaction_day
  FROM
    table1
  UNION
  ALL
  SELECT
    *
  FROM
    table_0
)
SELECT
  user_id,
  COUNT(*) AS daily_interaction
FROM
  table_1
GROUP BY
  user_id

-- Generated by PRQL compiler version:0.9.5 (https://prql-lang.org)
```

### MVCE confirmation

- [X] Minimal example
- [ ] New issue

max-sixty:
Thanks @avikam .

(For others looking — the error is `DATE_TRUNC('day', ts) AS _expr_0`)

snth:
The naming of the column as `_expr_0` is one error. The other is that the order of the columns in the first part of `table_1` before the `UNION` (as was pointed out in the original report).

I believe the **Expected SQL Output** should actually be:

```sql
WITH table_0 AS (
  SELECT
[...]

Issue #2680, opened 2023-06-01
`select` clause after `append` clause


YuxiLing:
### What happened?

Dear developer,

Hope this would be helpful.

prql version: 0.8.1

comments: The `select` clause affects the translation of `append` clause. It might be caused by internal optimization in prql compiler.

In the following example, we assume `tb1` and `tb2` have the same columns. We append `tb2` to `tb1`, and select two new columns. However, inside generated sql query, it selects new columns first. Then the `union` operation fails.

error message from the database: "each UNION query must have the same number of columns"

### PRQL input

```elm
from tb1
append tb2
select [new_col1=c1*2,new_col2=c2/c3]
```

### SQL output

```SQL
WITH table_1 AS (
  SELECT
    c2,
    c3,
    c1
  FROM
    tb1
  UNION
  ALL
  SELECT
    *
  FROM
    tb2
)
SELECT
  c1 * 2 AS new_col1,
  c2 / c3 AS new_col2
FROM
  table_1 AS table_0
```

### Expected SQL output

```SQL
WITH table_1 AS (
  SELECT
    *
  FROM
    tb1
  UNION
  ALL
  SELECT
    *
  FROM
    tb2
)
SELECT
  c1 * 2 AS new_col1,
  c2 / c3 AS new_col2
FROM
  table_1 AS table_0
```

### MVCE confirmation

- [X] Minimal example
- [X] New issue

aljazerzen:
Yup, that's a bug.

When the compiler encounters c1, c2 and c3 it infers that they are from tb1, while is should find both tb1 and tb2.

aljazerzen:
I would expect SQL to be:
```sql
WITH table_1 AS (
  SELECT
    c1, c2, c3
  FROM
    tb1
  UNION
  ALL
  SELECT
    c1, c2, c3
  FROM
    tb2
)
SELECT
  c1 * 2 AS new_col1,
  c2 / c3 AS new_col2
FROM
  table_1 AS table_0
```

kgutwin:
Rewriting the original input to the new syntax:

```
from tb1
append tb2
select {new_col1=c1*2,new_col2=c2/c3}
```

Still produces the wrong output as of 0.13.3:

```sql
WITH table_0 AS (
  SELECT
    c2,
    c3,
    c1
  FROM
    tb1
  UNION
  ALL
  SELECT
    *
  FROM
    tb2
)
SELECT
  c1 * 2 AS new_col1,
[...]

Issue #5165, opened 2025-02-27
Should `append` sometimes or always create `UNION ALL BY NAME`?


kgutwin:
### What's up?

Currently a PRQL `append` transform will result in `UNION ALL`:

```
from tbl_a
append tbl_b
```
 
gives

```sql
SELECT * FROM tbl_a UNION ALL SELECT * FROM tbl_b -- Generated by PRQL compiler version:0.13.3-39-ge393ab4d (https://prql-lang.org)
```

However, this results in issues like #4724, #2680, and #3184, where the underlying cause is that `UNION ALL` is interpreted by the database as "unify by column position" rather than "unify by column name". See the [DuckDB docs](https://duckdb.org/docs/stable/sql/query_syntax/setops.html):

> Traditional set operations unify queries by column position, and require the to-be-combined queries to have the same number of input columns. If the columns are not of the same type, casts may be added. The result will use the column names from the first query.
> DuckDB also supports [UNION [ALL] BY NAME](https://duckdb.org/docs/stable/sql/query_syntax/setops.html#union-all-by-name), which joins columns by name instead of by position. UNION BY NAME does not require the inputs to have the same number of columns. NULL values will be added in case of missing columns.

Questions:

1. Should `append` always behave as `UNION ALL BY NAME` to simplify semantics from the user perspective, and also make the compiler's job easier? This would resolve all of the linked issues above without needing to dive into compiler details, but would be a breaking change for users expecting traditional `UNION ALL` behavior.
2. If "no" to the above question, can we add a `by:name` or `by:position` argument to `append` to allow users to use `UNION ALL BY NAME` when that makes sense for their use case?

Fanaen:
Disclaimer: https://github.com/PRQL/prql/pull/5317 is my first contribution and I'm no SQL expert. \
Here are my two cents, @kgutwin.

**EDIT: I simplified my answer after digging all day 😄**

It would indeed make sense to have `by:name` argument to let know PRQL that it can use the same column names inside the `append`, and therefore avoiding two `select` statement to manually order those columns. 

That said, we probably don't need `UNION ALL BY NAME` for this. It can select those columns and work by position like always. For instance, in https://github.com/PRQL/prql/issues/4724, PRQL knows both columns. It mismatches by design, like `UNION`.  

[...]

PR #5317, opened 2025-06-05
fix(prqlc, append): first attempt at applying column order on CTEs too

Fanaen:
Hi there!

**TL;DR.** We recently had more convoluted cases of `append` where PRQL output valid SQL that fails at runtime. \
Here is a proposal to fix common explicit uses of `append`. \
PR's not complete yet: it's an early bird to make sure this goes in the right direction. 
Requesting for comments!

# Problem

- SQL's `UNION` need the same number of column and in both queries
- PRQL currently changes main query's columns without changing sub query as well
- PRQL outputs a SQL query that leads to a *type mismatch* or *a column count mismatch* error. More elaborate cases are a mix of those two. You may try a simplified version of either with the `Query Results` tab in the [playground](https://prql-lang.org/playground/):
```csharp
# Column swap leading to type mismatch
from invoices | select { invoice_id, invoice_date }
append (
  from invoices | select { invoice_id, invoice_date }
)
select { invoice_date, invoice_id }
# -> E: Unimplemented type for cast (BIGINT -> TIMESTAMP)
```

```csharp
# Column count mismatch
from invoices | select { invoice_id, customer_id, invoice_date }
append (
  from invoices | select { invoice_id, customer_id, invoice_date }
)
group { customer_id } (aggregate { `count` = count `invoice_id` })
# -> E: Set operations can only apply to expressions with the same number of result columns
```

# Input space
This means right now, only a tiny subset of valid `append` are working properly.\
To be clear about what we are trying to fix, let's talk PRQL inputs.

## Explicit `append`: :white_check_mark: covered by this PR
If we put `select` everywhere, PRQL can work without asumptions and that's what I tried to fix here.

```csharp
from invoices | select { invoice_id, invoice_date } 
append (from invoices | select { invoice_id, invoice_date })
# -> WITH table_0 AS (SELECT invoice_id, invoice_date FROM invoices) SELECT invoice_id, ...
```

## Wildcard `append`: :no_entry_sign: not covered by this PR

As long as there is no selection whatsoever, this is fine:
```csharp
from invoices | append (from invoices)
# -> SELECT * FROM invoices UNION ALL SELECT * FROM invoices AS table_0

from invoices | derive { c = "a"} | append (from invoices | derive { c = "b"}) | sort { +c } 
# A wildcard CTE appears! Not sure those are great 
# -> WITH table_0 AS (SELECT *, 'b' AS c FROM invoices), ...
```

If we do a select at the end, the main query inherit those select and it fails at SQL runtime.

```csharp
from invoices | append (from invoices) | select { invoice_date, invoice_id }
# -> SELECT invoice_date, invoice_id FROM invoices UNION ALL SELECT * FROM invoices AS table_0
# -> E: Set operations can only apply to expressions with the same number of result columns
```

I'm not sure how much we want to support wildcard `append` since PRQL can't make assumptions on input tables, so it seems really difficult to get it right, so this PR does not affect those.

If we do, maybe we should first reduce the use of CTEs to avoid pulling whole tables.

## Hybrid `append`: :no_entry_sign: not covered by this PR
If we specify `select` on one side and not the other, we trigger a legit PRQL error.

```csharp
from invoices | select { invoice_id, invoice_date } | append (from invoices)
# -> E: cannot append two relations with non-matching number of columns.

from invoices | append (from invoices | select { invoice_id, invoice_date })
# -> E: cannot append two relations with non-matching number of columns.
```

This seems OK, especially since sub-query may use other columns names, so this PR does not affect those either.
Columns being unknown probably makes it really difficult to get this right.

# Solutions

1. :white_check_mark: **Apply column projection on CTE** \
To apply whatever reorder/removal in the main relation on the secondary one. Detailed below. 

2. :no_entry_sign: **Leave CTE as is, but specify the columns in `UNION ALL` instead of `*`** \
Backup plan in case solution 1 fails due to CTE dependencies. It seems `append` is isolated enough to not warrant this solution. \
Downside of this approach is that it means potentially bigger CTE than needed.  

3. :no_entry_sign: **Force column alignment by forcing intermediate CTE** \
Backup-backup plan. IMHO, the less CTE the better. \
This method may be required to handle wildcard `append`. 

## Details for the first version

Here, we will use this example:
```csharp
from invoicesA | select { invoice_idA, invoice_dateA }
append (
  from invoicesB | select { invoice_idB, invoice_dateB }
)
select { invoice_dateA, invoice_idA }
```

### Phase III. Semantic resolver
When the lineage inference occurs, there is a point where we have clearly the matching columns with `top` and `bottom` lineages for the `append`. The columns pairs (e.g. `(invoice_idA, invoice_idB)`) is then stored in the resulting `Lineage` alongside column lineage. 

So, we get a `top` with columns like `[("invoice_idA", 134), ..]` and `bottom` with `[("invoice_idB", 124), ..]` and the resulting lineage will have the `top` columns like before. Now, it will also have the mapping `[(("invoice_idA", 134), ("invoice_idB", 124)), ..]`.

`bottom` lineage is discarded, so make the link from the other side seemed trickier.

### Phase IV. Semantic-lowering
`push_select` add a `select` at the end of each relation. Before this PR, the reordering happens for the main relation (which directly uses `top` lineage) and not for the additional one. 

`push_select` is triggered by the end of `lower_relation` which is recursive in a sense:
a. `lower_relation(main)` triggers other methods wich leads to...
b. `lower_relation(secondary)`. 
c. `push_select` happens for `secondary`
d. `push_select` happens for `main`

Fortunately, the source of the `push_select` is available from the start of the `lower_relation` so we can leverage the fact that `lower_relation(main)` happens first to store which columns are needed, then apply the mapping at the start of `lower_relation(secondary)` then let `push_select` do its job.

State is added to the `Lowerer` as a simple `lineage_stack` to access the latest relevant mapping.

A bit of extra logic is needed: 
- to cirmuvent the fact that due to `assign`, `lower_relation(main)` is provided with a new  `("invoice_idA", 139)` instead of `("invoice_idA", 134)`, so AFAIK, we have to check by name.
- to avoid losing form of lineage. In fact, there are still some kinks to work out here, where the link is broken and this leads to incorrect SQL queries.

### Phase VI. Post-process
Should we want to use the second solution (leave CTE as is, but specify the columns in `UNION ALL` instead of `*`), we would have to get the extra state across *Phase V. Anchoring*. Possible but not my first choice.

# Related issues
This PR intends to fix (at least partially) the following issues:

### https://github.com/PRQL/prql/issues/2680 :white_check_mark: Fixed in [42f5af0](https://github.com/PRQL/prql/pull/5317/commits/42f5af09674d18e9b3b1521d1e64e29b12771fac)
- Column mismatch by redefinition
- This PR does not intends to fix the initial PRQL input, only [the workaround](https://github.com/PRQL/prql/issues/2680#issuecomment-2689048871)
- This PR's fix isn't triggered (yet)

### https://github.com/PRQL/prql/issues/3579 :construction: Not working yet.
- Column mismatch by reordering
- This PR's fix is triggered but it removes a column that's important for the `group`:
```SQL
WITH table_0 AS (
  SELECT
    substr(user, length('u:') + 1) AS user_id
  FROM
    table2
),
table_1 AS (
  SELECT
    substr(user, length('u:') + 1) AS user_id
  FROM
    table1
  UNION
  ALL
  SELECT
    *
  FROM
    table_0
)
SELECT
  user_id,
  COUNT(*) AS daily_interaction
FROM
  table_1
GROUP BY
  user_id
```

### https://github.com/PRQL/prql/issues/4724 :no_entry_sign: Unrelated after all
- Input is flawed: first table is `(col1, col2)`, second is `(col2, col1)`)

# Conclusion

I am still working on the edge cases. \
What do you think?


kgutwin:
Thanks for taking a look at this! If we make progress here, it would help greatly one of my project's backlog items.

Did you look at all at #5165? Do you have any thoughts about that question?

Fanaen:
First version [dc6273b](https://github.com/PRQL/prql/pull/5317/commits/dc6273b3e68824021abc201e7d1677e04982429d) had the flaw of removing too much columns, and losing track of more complex columns (like `compute` ones).

Second version [42f5af0](https://github.com/PRQL/prql/pull/5317/commits/42f5af09674d18e9b3b1521d1e64e29b12771fac) does not change the select at `lowering` phase but does it at `anchoring` instead which is a lot more reliable. It needs some work because it removes too little, which means it often fails to fix the issue. Unfortunately, this spread more additional state across three layer, and for that I'm sorry 😅 

About related issues:
- This new version fixes the explicit workaround suggested in https://github.com/PRQL/prql/issues/2680
- https://github.com/PRQL/prql/issues/3579 still needs some work (`append` + `count(*)` edge case)
- Upon closer inspection, the https://github.com/PRQL/prql/issues/4724 issue is not related, its about not having to manually align columns

I will wait for feedback before working more on this.

Fanaen:
Hi @max-sixty! Would this contribution be welcome? 

It seems to me the fix should be at the anchoring stage because some of the column reordering/removal happens at that stage.

Column order seems done. Now, I'm trying to find a reliable way to remove, in the child relation, the columns that were added or discarded in the parent relation. The issue is that the lineage is not available at this point since it's been discarded in the lowering stage. There is a disconnect between columns in each relation. Do you have an idea, by any chance?

max-sixty:
hi @Fanaen ! sorry for missing this, was heads down elsewhere for a while. greatly appreciate the effort!

do you know why we get test failures? is that just sqlite? if we can fix bugs but lose some sqlite support, that's a worthwhile tradeoff

I do think the existing compiler makes this non-trivial, though hopefully not impossible. if we need to reduce the scope of what's supported (for example compel queries to specify columns when using append), then that's OK...

Fanaen:
> hi @Fanaen ! sorry for missing this, was heads down elsewhere for a while. greatly appreciate the effort!

Hi @max-sixty! No worries, thanks a lot for your answer!
 
> do you know why we get test failures? is that just sqlite? if we can fix bugs but lose some sqlite support, that's a worthwhile tradeoff 

The first reason is that I threw together a few integration tests to make sure we're covered, and I overlooked some column-typing shenanigans. I'll look into this tomorrow.

The second reason is that at the time, the fix was not yet behaving as intended. 

The good news is that my third attempt seems to be really reliable! \
I started a new, clean PR, and it should be ready soon.

> I do think the existing compiler makes this non-trivial, though hopefully not impossible. if we need to reduce the scope of what's supported (for example compel queries to specify columns when using append), then that's OK...

Good to know, thanks! I think we're almost good so hopefully, it won't come to that.

Fanaen:
Closing this one in favor of my third attempt. It's different, so I opened a new PR: https://github.com/PRQL/prql/pull/5323

PR #5344, opened 2025-07-01
test: integration, compile all dialects and diff

kgutwin:
Adds a new integration test on the set of integration queries, which generates snapshots covering the unified diff between the result of compiling with the generic dialect versus all other dialects. When there's no difference, there's no extra entry in the snapshot, so some snapshots are essentially empty.

This is from a discussion in #5343.

Example snapshot output:

```
---
source: prqlc/prqlc/tests/integration/queries.rs
expression: "from invoices\nselect { customer_id, invoice_id, total }\ntake 5\nappend (\n  from invoice_items\n  select { invoice_line_id, invoice_id, unit_price }\n  take 5\n)\nselect { a = customer_id * 2, b = math.round 1 (invoice_id * total) }\n"
input_file: prqlc/prqlc/tests/integration/queries/append_select_compute.prql
---
--- generic
+++ glaredb
@@ -23,13 +23,13 @@
     ) AS table_2
   UNION
   ALL
   SELECT
     *
   FROM
     table_0
 )
 SELECT
   customer_id * 2 AS a,
-  ROUND(invoice_id * total, 1) AS b
+  ROUND((invoice_id * total)::numeric, 1) AS b
 FROM
   table_1


--- generic
+++ postgres
@@ -23,13 +23,13 @@
     ) AS table_2
   UNION
   ALL
   SELECT
     *
   FROM
     table_0
 )
 SELECT
   customer_id * 2 AS a,
-  ROUND(invoice_id * total, 1) AS b
+  ROUND((invoice_id * total)::numeric, 1) AS b
 FROM
   table_1
```

### DIFF
--- a/.config/nextest.toml
+++ b/.config/nextest.toml
@@ -4,6 +4,11 @@ failure-output = "final"
 slow-timeout = {period = "500ms", terminate-after = 4}
 
 [[profile.default.overrides]]
+# compileall does many passes over the same query, so it takes a bit longer
+filter = 'package(prqlc) & test(queries::compileall::)'
+slow-timeout = {period = "10s", terminate-after = 2}
+
+[[profile.default.overrides]]
 filter = 'package(prqlc) & test(queries::results::)'
 slow-timeout = {period = "10s", terminate-after = 4}
 test-group = 'test-dbs'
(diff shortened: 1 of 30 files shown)


max-sixty (approved):
that's awesome!

should we replace the existing sql outputs with this? then we can unify the two test types in #5343, iiuc?

kgutwin:
I'll merge this now. If you feel like merging the `compile` and `compileall` tests, go ahead, it should be pretty straightforward. Thanks!

PR #5345, opened 2025-07-01
chore: set rust linker on win64, fix build crash

kgutwin:
I did a little google searching of the recently appearing Windows build failures and found https://github.com/actions/runner-images/issues/12432, which led to https://github.com/rust-lang/rust/issues/141626#issuecomment-2919988483. Let's see if this fixes the issue.

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -1,2 +1,6 @@
 [target.wasm32-unknown-unknown]
 runner = 'wasm-bindgen-test-runner'
+
+[target.x86_64-pc-windows-msvc]
+# https://github.com/rust-lang/rust/issues/141626#issuecomment-2919988483
+linker = "rust-lld"
--- a/.github/workflows/test-js.yaml
+++ b/.github/workflows/test-js.yaml
@@ -22,7 +22,9 @@ env:
 jobs:
   test:
     runs-on: ${{ matrix.os }}
+    continue-on-error: ${{ matrix.os == 'windows-latest' }}
     strategy:
+      fail-fast: false
       matrix:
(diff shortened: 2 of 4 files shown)


max-sixty:
bleh, no luck. we could just comment out the windows tests until it's fixed upstream?

kgutwin:
It did actually fix the linker-related crash, but it didn't fix the other issues :( 

One is weird -- in `.github/workflows/test-rust.yaml` it successfully passes the "Compile" step using `clechasseur/rs-cargo@v3`, and you can see that it pulls in the insta package, but then it goes to run `cargo insta test` and complains "insta not found"? It doesn't really make sense.

The other one in test-js seems maybe that something (wasm-pack) was omitted from the runner image. I wonder if it's possible to add that back in somehow. Still looking into that one.

max-sixty:
> It did actually fix the linker-related crash, but it didn't fix the other issues :(

oh nice! and the _build_ now passes on windows

but yeah then the insta issue is surprising — it says it's installed: https://github.com/PRQL/prql/actions/runs/16012029427/job/45171587927?pr=5345#step:7:12

kgutwin:
Is the insta issue possibly related to the cache in some way? I wonder how to reset / clean out that cache, maybe it needs a fresh install for the latest runner.

I fiddled with the matrix settings for test-js and test-rust-main so that the failing Windows tests there won't stop the whole pipeline. Now, check-ok-to-merge is good with ignoring the failed Windows tests.

I'm going to go ahead and merge this now, since it's better than before. It's still worth trying a bit more to fix the failing tests though.

PR #5346, opened 2025-07-02
build: invalidate certain cargo-install caches, fixing Win build pt 2

kgutwin:
I took one last look at the Windows build failure messages, and this is another stab at maybe fixing them. Hopefully invalidating the `baptiste0928/cargo-install@v3` caches for `cargo-insta` and `wasm-pack` will let Windows rebuild and properly use those tools...

### DIFF
--- a/.github/workflows/test-js.yaml
+++ b/.github/workflows/test-js.yaml
@@ -39,6 +39,7 @@ jobs:
       - uses: baptiste0928/cargo-install@v3
         with:
           crate: wasm-pack
+          cache-key: 20250701-1
 
       - run: ./.github/workflows/scripts/set_version.sh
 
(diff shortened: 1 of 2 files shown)


kgutwin:
so close, looks like just one more cache (`cargo-nextest`) needs to be invalidated...

max-sixty:
I just wiped them all, will try again!

kgutwin:
It's so close, now! 

https://github.com/PRQL/prql/actions/runs/16014368312/job/45178110951?pr=5346#step:15:880

Looks like the doctests somehow also need the linker patch.

kgutwin:
I'm not sure why setting RUSTFLAGS didn't have any effect on the doctest compilation failure. Cross fingers that the upstream regression gets fixed, then we don't need this. https://github.com/rust-lang/rust/issues/141626#issuecomment-3027216890

max-sixty:
@kgutwin windows now seems to pass on main!

kgutwin:
wait, are you sure? I still see this failure from the most recent "tests" run on main: https://github.com/PRQL/prql/actions/runs/16300971426/job/46035208345

max-sixty:
bah you're right! I'm not sure why it shows green here
image

PR #5347, opened 2025-07-02
fix(prqlc, sort): do not require sort columns if it's followed by an aggregation

### LINKED ISSUE #3200: columns not in aggregate function and `group by`
### What happened?

PRQL compiler version:0.9.2

I found some related discussions in #2057. But some databases (such as Postgres, DuckDB, and Clickhouse) do not support a `select` clause to include a column that is not used in an aggregate function without a `group by` clause. 

Error message from duckdb: Binder Error: column c0 must appear in the GROUP BY clause or be used in an aggregate function. Either add it to the GROUP BY list, or use "ANY_VALUE(c0)" if the exact value of "c0" is not important.

### PRQL input

```elm
prql target:sql.duckdb
from tb1
sort {c0}
aggregate {new_col = min c1}
```

### SQL output

```SQL
WITH table_0 AS (
  SELECT
    MIN(c1) AS new_col,
    c0
  FROM
    tb1
)
SELECT
  new_col
FROM
  table_0
```

### Expected SQL output

```SQL
WITH table_0 AS (
  SELECT
    MIN(c1) AS new_col
  FROM
    tb1
)
SELECT
  new_col
FROM
  table_0
```

### MVCE confirmation

- [X] Minimal example
- [X] New issue

### Anything else?

_No response_

julien-pinchelimouroux:
Fixes: https://github.com/PRQL/prql/issues/3200

`sort` always required sorted columns which caused an  unwanted column selection in the next `aggregate` :  `column <SORTED_COLUMN> must appear in the GROUP BY clause or be used in an aggregate function`

---

To fix it, it simply checks if an `aggregate` is present in the following transforms

---

I didn't find a PRQL query where this change can be a regression. If you have something in mind, please share it! :pray:


max-sixty:
thanks @julien-pinchelimouroux !

would you mind adding a small test?

> I didn't find a PRQL query where this change can be a regression. If you have something in mind, please share it! 🙏

could we use, from #3200

```
prql target:sql.duckdb
from tb1
sort {c0}
aggregate {new_col = min c1}
```

I also see lots of snapshot files in the diff — where are those from? (sorry if I'm missing something...)

julien-pinchelimouroux:
Hello @max-sixty ! 

snap files comes from the following added tests : 
 - prqlc/prqlc/tests/integration/queries/sort_with_aggregation.prql
 - prqlc/prqlc/tests/integration/queries/sort_with_aggregation_and_limit.prql
 
 They are based on the reported query in the original issue https://github.com/PRQL/prql/issues/3200

vanillajonathan:
Maybe a note should be added to the CHANGELOG file?

max-sixty:
ah thanks @julien-pinchelimouroux !

could we slim down the test a bit, unless we really need two full integration tests? I'd have thought this could just be a single inline test in prqlc/prqlc/tests/integration/sql.rs ?

julien-pinchelimouroux:
Hi @max-sixty 
Yes you're totally right, that do not need a full integration test :)

max-sixty:
great! thank you @julien-pinchelimouroux

PR #5357, opened 2025-07-09
fix(prqlc): improve requirement logic

### LINKED ISSUE #5354: Computed columns and CTEs sometimes lead to missing column
### What happened?

When a query results in several CTEs, a computed column may be duplicated in a child CTE without its' column sources. This leads to `Referenced column "..." not found` as the column is not there anymore.

### PRQL input

```elm
prql target:sql.postgres

from employees
derive `year` = s'EXTRACT(year from {`hire_date`})'
derive { `year_label` = f"Year {`year`}" }
derive { `city` = case [ this.`city` == "Calgary" => "A city", true => this.`city` ] }
derive { `city` = case [ this.`city` == "Edmonton" => "Another city", true => this.`city` ] }
group {`year`, `year_label`} (take 1)
select {this.`year_label`}
```

### SQL output

```SQL
WITH table_0 AS (
  SELECT
    CONCAT(
      'Year ',
      EXTRACT(
        year
        from
          hire_date
      )
    ) AS year_label,
    CASE
      WHEN city = 'Calgary' THEN 'A city'
      ELSE city
    END AS _expr_0,
    city
  FROM
    employees
)
SELECT
  DISTINCT ON (
    EXTRACT(
      year
      from
        hire_date
    ),
    year_label
  ) year_label
FROM
  table_0
```

### Expected SQL output

```SQL
WITH table_0 AS (
  SELECT
    CONCAT(
      'Year ',
      EXTRACT(
        year
        from
          hire_date
      )
    ) AS year_label,
    EXTRACT(
      year
      from
        hire_date
    ) as year_or_expr_1_or_something,
    CASE
      WHEN city = 'Calgary' THEN 'A city'
      ELSE city
    END AS _expr_0,
    city
  FROM
    employees
)
SELECT
  DISTINCT ON (
    year_or_expr_1_or_something,
    year_label
  ) year_label
FROM
  table_0
```

### MVCE confirmation

- [x] Minimal example
- [x] New issue

### Anything else?

Example is reproducible in the playground. Removing the last `select` is enough to get out of the CTE topology and therefore to not fail.

Fanaen:
Fixes https://github.com/PRQL/prql/issues/5354
---
New result:
```sql
SELECT
  DISTINCT ON (
    EXTRACT(
      year
      from
        hire_date
    ),
    CONCAT(
      'Year ',
      EXTRACT(
        year
        from
          hire_date
      )
    )
  ) CONCAT(
    'Year ',
    EXTRACT(
      year
      from
        hire_date
    )
  ) AS year_label
FROM
  employees
```

---
### About the refactoring

`get_requirement` is quite logic-heavy.

It made sense to separate column selection logic from complexities/`SELECT` status logic when there the case were simpler. The special cases with multiple complexities now outnumber general cases.

So I added `Requirements`: a newtype for `Vec<Requirement>`. It adds `Complexity::Plain` and `select: false` by default (the common case) and a few convenience methods to improve readbility. 

It's a separate commit in case we want to split the PR.

### DIFF
--- a/prqlc/prqlc/src/sql/pq/anchor.rs
+++ b/prqlc/prqlc/src/sql/pq/anchor.rs
@@ -1,4 +1,6 @@
 use std::collections::{HashMap, HashSet};
+use std::fmt;
+use std::ops::Deref;
 
 use itertools::Itertools;
 
@@ -79,7 +81,10 @@ pub(super) fn split_off_back(
 
     let mut following_transforms: HashSet<String> = HashSet::new();
 
-    let mut inputs_required = into_requirements(output.clone(), Complexity::highest(), true);
+    let mut inputs_required = Requirements::from_cids(output.iter())
+        .allow_up_to(Complexity::highest())
+        .should_select(true);
+
     let mut inputs_avail = HashSet::new();
 
     // iterate backwards
@@ -97,7 +102,7 @@ pub(super) fn split_off_back(
         // anchor and record all requirements
         let required = get_requirements(&transform, &following_transforms);
         log::debug!(".. transform {} requires {required:?}", transform.as_str(),);
-        inputs_required.extend(required.clone());
+        inputs_required = inputs_required.append(required.clone());
 
         match &transform {
             SqlTransform::Super(Transform::Compute(compute)) => {
@@ -107,11 +112,8 @@ pub(super) fn split_off_back(
                     inputs_avail.insert(compute.id);
 
                     // add transitive dependencies
-                    inputs_required.extend(required.into_iter().map(|x| Requirement {
-                        col: x.col,
-                        max_complexity,
-                        selected: false,
-                    }));
+                    inputs_required = inputs_required
+                        .append(required.allow_up_to(max_complexity).should_select(false));
                 } else {
                     pipeline.push(transform);
                     break;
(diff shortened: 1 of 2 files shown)


max-sixty:
thanks @Fanaen ! 

> 🚧 I'm not sure about testing strategy for this. Input is quite messy, not necessarily something we want in integration tests.

could we add the case as a simple inline test in `prql/prqlc/prqlc/tests/integration/sql.rs`? that makes it easy to understand the change when the code is open locally. if we can reduce the size of the example a bit then great, but not worth spending much time on making it totally minimal :)

Fanaen:
Thanks @max-sixty! I added the inline test as suggested. The PR is ready for review!

Fanaen:
I've found that [c1d7069](https://github.com/PRQL/prql/pull/5357/commits/c1d7069f2b5f92b84e045d5e04d3d4332fce1a58), which I added in an attempt to get nicer SQL results, sometimes removes necessary columns for instance when referenced in a filter.

As it's not necessary for this fix, I've removed it.

max-sixty:
great! thank you @Fanaen 

(for transparency: I tried this out but haven't thought through each line of code, so if we need to iterate more we can do so, thank you for your recent contributions!)

vanillajonathan:
Perhaps this ought to be mentioned in the CHANGELOG?

max-sixty:
fine to add things to CHANGELOG but no strong requirement!

PR #5468, opened 2025-10-06
fix: join with table with column named "source"

### LINKED ISSUE #5015: join fails when both sides have column named "source"
### What happened?

When joining together two tables where both tables contain column named `source`, compile fails with `Ambiguous name` error.

I've tried to wrap the column name with whatever i could come up, even good old backtick does not help.

### PRQL input

```elm
prql target:sql.postgres

let a = (
  from events_a
  select {
    event_id,
    source,
  }
)

let b = (
  from events_b
  select {
    event_id,
    source,
  }
)

from a
join b (==event_id)
select {
  event_id = a.event_id,
  source_a = a.source,
}
```

### SQL output

```SQL
prql target:sql.postgres
Ambiguous name
```

### Expected SQL output

_No response_

### MVCE confirmation

- [X] Minimal example
- [X] New issue

### Anything else?

Smallest example i could come up with is to join table with itself

```elm
prql target:sql.postgres

let a = (
  from events_a
  select {
    event_id,
    source,
  }
)

from a
join a (==event_id)
select {
  event_id = a.event_id,
}
```

priithaamer:
Attempting to fix #5015 by letting Claude to grind on it for 30 mins. My own understanding of the internals of PRQL is still very limited. Here's the report from Claude:

## Root Cause

The problem occurred in `prqlc/prqlc/src/semantic/resolver/functions.rs` in the `resolve_function_args` function. When resolving relational arguments for functions like join:

  1. Arguments were being resolved one at a time in a loop
  2. After each argument was resolved, its columns were immediately added to the `NS_THAT` namespace
  3. This meant that when resolving the second relational argument, the first argument's columns were already visible in `NS_THAT`
  4. If both arguments referenced the same variable (like `join a`), the second reference would see the first one's columns, creating a circular dependency
  5. This caused column lookups to find both the actual column AND the function parameter from the `_param` namespace, resulting in ambiguity

## The Fix

Changed the resolution order in resolve_function_args:

- Before: Resolve argument → Add to namespace → Resolve next argument → Add to namespace
- After: Resolve ALL arguments first → Then add all to namespaces

This ensures that when resolving variable references, they don't see partially-populated namespaces from other arguments being processed in the same function call.

## Testing

- Added regression test `test_join_with_param_name_collision` in `prqlc/prqlc/tests/integration/sql.rs`
- All existing tests continue to pass (394 integration tests)
- The specific example from issue #5015 now compiles successfully

The fix is minimal, surgical, and preserves all existing functionality while resolving the circular reference issue.

### DIFF
--- a/prqlc/prqlc/src/semantic/resolver/functions.rs
+++ b/prqlc/prqlc/src/semantic/resolver/functions.rs
@@ -227,6 +227,8 @@ impl Resolver<'_> {
             self.root_mod.module.shadow(NS_THIS);
             self.root_mod.module.shadow(NS_THAT);
 
+            // First, resolve all relational arguments
+            let mut resolved_relations = Vec::new();
             for (pos, (index, (param, mut arg))) in relations.into_iter().with_position() {
                 let is_last = matches!(pos, Position::Last | Position::Only);
 
@@ -241,7 +243,11 @@ impl Resolver<'_> {
                 }
                 log::debug!("resolved arg to {}", arg.kind.as_ref());
 
-                // add relation frame into scope
+                resolved_relations.push((index, arg, is_last));
+            }
+
+            // Then, add relation frames into scope
+            for (index, arg, is_last) in resolved_relations {
                 if partial_application_position.is_none() {
                     let frame = arg
                         .lineage
--- a/prqlc/prqlc/tests/integration/sql.rs
+++ b/prqlc/prqlc/tests/integration/sql.rs
@@ -2747,6 +2747,42 @@ fn test_join_side_literal_via_func_err() {
 }
 
 #[test]
+fn test_join_with_param_name_collision() {
+    // Regression test for issue #5015
+    // When joining tables that both have a column named "source",
+    // the compiler should not report an ambiguous name error due to
+    // the "source" parameter from the std library's "from" function.
+    assert_snapshot!((compile(r###"
+    let a = (
+      from events_a
+      select {
+        event_id,
+        source,
+      }
+    )
+
+    from a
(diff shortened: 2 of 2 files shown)


priithaamer:
@max-sixty hey, would be happy to get your eyes on this

max-sixty:
@claude can you confirm whether `source` is in our reserved keywords?

max-sixty:
thanks @priithaamer ! looks good! I'll check whether `source` is in the reserved keywords already, I don't think that's going to help as I think it's a different part of the pipeline. so feel free to merge assuming that's correct!

max-sixty:
@claude try again?

max-sixty:
(OK I guess Claude is broken)

I'll hit the button, thanks @priithaamer

priithaamer:
> (OK I guess Claude is broken)
> 
> I'll hit the button, thanks @priithaamer

Thanks! Claude needs its me-time too. Will there be a release any time soon? If not, i'll resort to our internal build of the npm package.

max-sixty:
yes, we're a bit overdue for a release! I may try to finish the chumsky refactor so we're not pulling in two versions of chumsky. if I don't get that done in the next day or two, I'll do one; (feel free to remind me)

max-sixty:
> Claude needs its me-time too.

lol

fyi I looked into this and it apparently doesn't work with forks (!?)

priithaamer:
Fwiw, i found a regression related to `append`-s which breaks many `UNION ALL` queries for us. I'll try to come up with minimal example and will provide the details. Maybe might be worth holding back the release.

PR #5486, opened 2025-10-09
fix(postprocess): when redirecting sort columns, look up the relation from the matching table_ref

lukapeschke:
It was done by converting the CTE's table id into an RIId for now, but those can differ in case some CTEs get erased. This resulted in the wrong CId redirect being updated.

Fixing this by finding a relation instance whose table_ref's source matches the CTE's table_id instead.

Follow up of #5464


lukapeschke:
Hi @max-sixty, I've missed this subtelty in #5464 , my bad :confused:  could you please have a look ?

max-sixty:
nice, thanks!

Issue #5496, opened 2025-10-13
Escape character issue in s-strings


priithaamer:
### What happened?

This used to work in 0.13.4

### PRQL input

```elm
let outputs_sql = (
  from s"
    select 'a' as \"col1 foo\", 'b' as \"col2#bar\"
  "
)

from outputs_sql
select {
  `col1 foo` = outputs_sql.`col1 foo`,
  `col2#bar` = outputs_sql.`col2#bar`
}
```

### SQL output

```SQL
⎿  Error: Error:
        ╭─[ :3:29 ]
        │
      3 │     select 'a' as \"col1 foo\", 'b' as \"col2#bar\"
        │                             ┬
        │                             ╰── unexpected '\'
     ───╯
```

### Expected SQL output

```SQL
WITH table_0 AS (
  SELECT
    'a' as "col1 foo",
    'b' as "col2#bar"
),
outputs_sql AS (
  SELECT
    *
  FROM
    table_0
)
SELECT
  "col1 foo",
  "col2#bar"
FROM
  outputs_sql
```

### MVCE confirmation

- [x] Minimal example
- [x] New issue

PR #5495, opened 2025-10-13
fix: Fix append regression in 0.13.5

priithaamer:
Attempt to fix #5494

PR #5497, opened 2025-10-13
fix: s-string escaping

priithaamer:
Attempting to fix regression #5496 that was working in 0.13.4


max-sixty:
thanks @priithaamer !

can we avoid doing a full integration test for each change if possible? hopefully this can just be a small inline insta test (ideally same with https://github.com/PRQL/prql/pull/5495)

(also OK if we merge, release, and then I'll ask Claude to make these smaller — wdyt?)

max-sixty:
I made the changes — it's at https://github.com/max-sixty/prql/commit/7a821b8a

can you enable "Allow edits from maintainers" so I can push to this branch?

priithaamer:
Hey, thanks for the changes, i merged them myself. Should be less trouble for you.

Totally fine to use the inline tests -- i was a bit unsure about integration tests as well.

This and #5495 are last remaining blockers for us to upgrade from 0.13.4 and close several issues / improvements in our product.

PR #5506, opened 2025-10-19
feat: migrate devcontainer build from QEMU to native ARM64 runners

max-sixty:
## Summary

Replace QEMU emulation with GitHub's native ARM64 runners for significantly faster builds. Uses matrix strategy with push-by-digest pattern for multi-platform image creation.

## Changes

- **Remove QEMU emulation**: No more slow instruction translation
- **Add native ARM64 builds**: Use `ubuntu-24.04-arm` runners (free for public repos)
- **Implement matrix strategy**: AMD64 builds on x86, ARM64 builds on ARM natively
- **Push-by-digest pattern**: Each platform pushes separately, then merged into multi-platform manifest
- **Add validation**: Digest count check and platform verification
- **Scope GHA cache**: Per-platform cache to prevent thrashing
- **Reduce timeouts**: Build 4h (was 60h!), Merge 30m

## Performance Results

**Actual build times from CI:**
- **AMD64**: ~2 minutes (native x86 build)
- **ARM64**: ~3 minutes (native ARM build)
- **Total**: ~3-4 minutes for both platforms in parallel

These are similar to current times since the devcontainer build is relatively light. The real benefit comes with heavier builds where QEMU emulation causes 9-22x slowdowns.

## Test Results

✅ Both platform builds completed successfully
✅ Build artifacts exported correctly  
✅ GHA cache working per-platform
✅ All CI checks passing

## Notes

- The merge job is skipped in PRs (only runs when `push: true`)
- This follows Docker's official pattern for native multi-platform builds
- Image name fixed to lowercase (`prql` instead of `PRQL`) per Docker registry requirements

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/.github/workflows/build-devcontainer.yaml
+++ b/.github/workflows/build-devcontainer.yaml
@@ -13,7 +13,7 @@ on:
         default: false
 
 env:
-  IMAGE_NAME: ghcr.io/${{ github.repository_owner }}/prql-devcontainer-base
+  IMAGE_NAME: ghcr.io/prql/prql-devcontainer-base
 
 jobs:
   build:
--- a/.github/workflows/build-devcontainer.yaml
+++ b/.github/workflows/build-devcontainer.yaml
@@ -12,54 +12,120 @@ on:
         type: boolean
         default: false
 
+env:
+  IMAGE_NAME: ghcr.io/${{ github.repository_owner }}/prql-devcontainer-base
+
 jobs:
   build:
-    runs-on: ubuntu-24.04
-    timeout-minutes: 3600
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - platform: linux/amd64
+            platform_name: amd64
+            runner: ubuntu-24.04
+          - platform: linux/arm64
+            platform_name: arm64
+            runner: ubuntu-24.04-arm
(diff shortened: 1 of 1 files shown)


max-sixty:
## 🔍 Testing Status

**What's been tested:**
- ✅ Both AMD64 and ARM64 builds complete successfully
- ✅ Build artifacts are created correctly
- ✅ GHA cache working per-platform
- ✅ Digest export working

**What hasn't been tested yet:**
- ⏳ Merge job (creates multi-platform manifest)
- ⏳ Digest download and validation
- ⏳ Platform verification

**Why the merge job was skipped:**
The merge job only runs when `push: true`, which happens when:
1. Devcontainer Dockerfile changes **AND**
2. Merged to `main` **AND**
3. Push event (not PR)

This is the correct security model - PRs shouldn't push to the registry.

**To verify after merge:**
1. Merge this PR to `main`
2. The devcontainer push workflow will trigger automatically
3. Watch for the `build-devcontainer / merge` job
4. Verify it:
   - Downloads both digests
   - Creates multi-platform manifest
   - Verifies both platforms are present

The merge logic is straightforward (download artifacts → create manifest → verify), but we won't know if it works until after merge.

PR #5537, opened 2025-10-31
feat: Add support for Redshift dialect

priithaamer:
Not 100% complete support, but solves major blockers using Redshift:

1. Prevents `DISTINCT ON` being used for `group ... take 1` deduplication.
2. Escape keywords when used as column names.

### DIFF
--- a/prqlc/prqlc/src/sql/keywords.rs
+++ b/prqlc/prqlc/src/sql/keywords.rs
@@ -5,15 +5,38 @@ use sqlparser::keywords::{
     Keyword, ALL_KEYWORDS, ALL_KEYWORDS_INDEX, RESERVED_FOR_COLUMN_ALIAS, RESERVED_FOR_TABLE_ALIAS,
 };
 
+use crate::sql::dialect::Dialect;
+
 /// True for keywords which we want to quote when translating to SQL.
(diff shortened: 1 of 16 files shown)


prqlc/prqlc/src/sql/keywords.rs
```diff
@@ -5,15 +5,38 @@ use sqlparser::keywords::{
...
 ///
 /// Currently we're being fairly permissive (over-quoting is not a big concern).
 // We're not including the full list from `SQL_KEYWORDS`, as that has terms such
 // as `ID`, instead we bring a few dialects' keywords in.
-pub(super) fn is_keyword(ident: &str) -> bool {
+pub(super) fn is_keyword(ident: &str, dialect: &Dialect) -> bool {
     let ident = ident.to_ascii_uppercase();
 
-    sql_keywords().contains(ident.as_str())
+    sql_keywords().contains(ident.as_str()) || dialect_keywords(dialect).contains(ident.as_str())
```
priithaamer, line 18: I'm a bit unsure about this but does not seem to have any side-effects and opens the capability to selectively escape keywords when used as column names.

max-sixty:
nice!

PR #5540, opened 2025-11-03
fix(redshift): use || operator over CONCAT for std.concat

lukapeschke:
Follow-up for @priithaamer 's work on the Redshift dialect (thanks!). I'm getting the following error from Redshift when usingg f-strings: `[SQLState 42883] ERROR:  function concat(character varying, "unknown", character varying) doe
s not exist`

The Redshift docs state that one should use nested CONCAT functions or the `||` operator to concatenate more than two elements: https://docs.aws.amazon.com/redshift/latest/dg/r_CONCAT.html

### DIFF
--- a/prqlc/prqlc/src/sql/dialect.rs
+++ b/prqlc/prqlc/src/sql/dialect.rs
@@ -371,6 +371,12 @@ impl DialectHandler for RedshiftDialect {
     fn prefers_subquery_parentheses_shorthand(&self) -> bool {
         true
     }
+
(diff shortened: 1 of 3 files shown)


priithaamer:
Awesome, thanks! It has been bugging us as well but haven't gotten to it yet. Can confirm it works!

eitsupi:
Thank you both, let's merge this.

max-sixty:
lmk if you want me to do a release; quite easy to do now!

lukapeschke:
@max-sixty A release would be great once #5545 is merged! It's the last blocker for our prql+redshift usage :)

PR #5545, opened 2025-11-04
fix(redshift): fix INTERVAL quoting

lukapeschke:
Continuing the work in #5540 for intervals. 
Unfortunately, there is no "one size fits all" quoting that works for redshift:

`INTERVAL '1 year'` causes the following error: `Interval values with month or year parts are not supported`

`SELECT CAST((DATE_TRUNC('week', added_date) - INTERVAL '1' WEEK) AS timestamp) AS firstdayofpreviousweek` causes a syntax error: `Error occurred while trying to execute a query: [SQLState 42601] ERROR:  syntax error at or near \"WEEK\" in context \")` However, replacing `'1' WEEK` with `'1 WEEK'` here works.

### DIFF
--- a/prqlc/prqlc/src/sql/dialect.rs
+++ b/prqlc/prqlc/src/sql/dialect.rs
@@ -326,14 +326,10 @@ impl DialectHandler for RedshiftDialect {
     }
 
     fn interval_quoting_style(&self, dtf: &DateTimeField) -> IntervalQuotingStyle {
-        if let DateTimeField::Year
-        | DateTimeField::Years
-        | DateTimeField::Month
-        | DateTimeField::Months = dtf
-        {
-            IntervalQuotingStyle::ValueQuoted
-        } else {
+        if matches!(dtf, DateTimeField::Week(_) | DateTimeField::Weeks) {
             IntervalQuotingStyle::ValueAndUnitQuoted
+        } else {
+            IntervalQuotingStyle::ValueQuoted
         }
     }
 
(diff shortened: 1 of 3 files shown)


lukapeschke:
@priithaamer @eitsupi I'd be happy to have your opinion on this one :slightly_smiling_face:

priithaamer:
It seems only `week` is special in Redshift where it needs `'1 week'` as `ValueAndUnitQuoted`. Perhaps every other unit could use `ValueQuoted` because `'1' year` etc seem to work fine (i'm no expert when it comes to Redshift intervals and it seems there may be nuances)?

lukapeschke:
@priithaamer You're right, I set the quoting style to `ValueAndUnitQuoted` for weeks only in https://github.com/PRQL/prql/pull/5545/commits/f0fa89598d3330a62d5f149b8a27e8b331403d7b and all our tests passed.

The redshift docs does not mention weeks at all :confused: https://docs.aws.amazon.com/redshift/latest/dg/r_interval_data_types.html

priithaamer:
> @priithaamer You're right, I set the quoting style to `ValueAndUnitQuoted` for weeks only in [f0fa895](https://github.com/PRQL/prql/commit/f0fa89598d3330a62d5f149b8a27e8b331403d7b) and all our tests passed.
> 
> The redshift docs does not mention weeks at all 😕 https://docs.aws.amazon.com/redshift/latest/dg/r_interval_data_types.html

Thanks! I can confirm this works and can be merged. We could also consider dropping direct week support and if possible, in Redshift dialect compile `1week` to `INTERVAL '7' DAY`?

max-sixty:
sorry, I merged the other; would you mind resolving conflicts and then merging?

lukapeschke:
@max-sixty done :) Can I let you merge and create a new release ?

eitsupi:
Sorry for the delay, I'm not currently in a position to test Redshift so I can't be sure this is correct, but the changes seem reasonable.

PR #5546, opened 2025-11-05
fix: division in Redshift produces float

priithaamer:
Similarly to Postgres, make sure that division will end up as float type.


max-sixty:
thanks as ever @priithaamer !

PR #5550, opened 2025-11-08
fix: Lineage works with unions

nightscape:
Fixes lineage tracking for `append` (UNION) operations to correctly track inputs from both source tables.

Previously, when using `append` to union two tables, the lineage only tracked the top table's inputs. The `append` function now merges `bottom.inputs` into `top.inputs` (similar to how `join` handles inputs), ensuring both source tables are tracked.

Added test to verify both inputs are tracked and column-level lineage works correctly.

### DIFF
--- a/prqlc/prqlc/src/semantic/resolver/mod.rs
+++ b/prqlc/prqlc/src/semantic/resolver/mod.rs
@@ -208,4 +208,277 @@ pub(super) mod test {
         )
         .unwrap());
     }
+
+    // Helper function to verify basic lineage structure after append
+    fn verify_append_lineage_basics(
+        final_lineage: &crate::ir::pl::Lineage,
+        expected_inputs: &[&str],
+    ) {
+        let input_names: Vec<&str> = final_lineage
+            .inputs
+            .iter()
+            .map(|i| i.name.as_str())
+            .collect();
+
+        for expected_input in expected_inputs {
+            assert!(input_names.contains(expected_input));
+            assert!(final_lineage.find_input_by_name(expected_input).is_some());
+        }
+
+        assert!(!final_lineage.columns.is_empty());
+        for col in &final_lineage.columns {
+            match col {
+                crate::ir::pl::LineageColumn::Single {
+                    name, target_id, ..
+                } => {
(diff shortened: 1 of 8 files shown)


nightscape:
The second commit is worth discussing:
If a table is union'd with itself, should lineage track both occurrences?
I have the slightest tendency towards "yes", because it is two different ways to get to the same table, both possibly containing different filters etc.
But if you deem the existing behavior correct, I'll happily adapt the implementation!

max-sixty:
thanks for the PR @nightscape !

overall looks good!

I'm a bit worried about the "identical tables have very different behavior from almost-identical tables", though. I haven't looked enough in detail, but how much of a step back would this be:

```diff

  diff --git a/prqlc/prqlc/src/semantic/resolver/transforms.rs b/prqlc/prqlc/src/semantic/resolver/transforms.rs
  index 5f07a7499c01..xxxxx 100644
  --- a/prqlc/prqlc/src/semantic/resolver/transforms.rs
  +++ b/prqlc/prqlc/src/semantic/resolver/transforms.rs
  @@ -757,14 +757,12 @@ fn append(mut top: Lineage, bottom: Lineage) -> Result<Lineage, Error> {
                       except: except_b,
                   },
               ) => {
  -                // If both are All columns from the same input, merge the except sets
  -                // Otherwise, keep the top's input_id
  -                // Note: In a union, both inputs should be available, so we keep top's
  +                // Merge except sets from both tables
  +                // This preserves exclusion information even when input_ids differ
  +                // (e.g., "from employees select !{name}" append "from managers select !{salary}")
                   let mut except = except_t;
  -                if input_id_t == input_id_b {
  -                    // Same input, merge except sets
  -                    except.extend(except_b);
  -                }
  +                except.extend(except_b);
  +
                   LineageColumn::All {
                       input_id: input_id_t,
                       except,
```

(for transparency Claude helped me with the diff)

nightscape:
@max-sixty I applied your proposed change and ran the tests, everything still green 👍 
I rebased on the latest `main` and force pushed.
All good to go from my side!

max-sixty:
thank you!

Issue #5556, opened 2025-11-11
Invalid DISTINCT ON clause generated for postgresql


fredericp:
### What happened?

prql version 0.13.7 generates invalid DISTINCT ON clause for postgresql. In this case it outputs: `DISTINCT ON (id, name)` that should be `DISTINCT ON (id, name) *`.

kudos for this great project!

### PRQL input

```elm
from t1
group {t1.id, t1.name} (take 1)
aggregate {c=count this}
```

### SQL output

```SQL
WITH table_0 AS (
  SELECT
    DISTINCT ON (id, name)
  FROM
    t1
)
SELECT
  COUNT(*) AS c
FROM
  table_0
```

### Expected SQL output

```SQL
WITH table_0 AS (
  SELECT
    DISTINCT ON (id, name) *
  FROM
    t1
)
SELECT
  COUNT(*) AS c
FROM
  table_0
```

### MVCE confirmation

- [x] Minimal example
- [x] New issue

max-sixty:
@claude can you try to fix this? or is it more involved?

max-sixty:
@claude OK — open a PR with that. Use a small inline test, not a full integration test.

Add that guidance to CLAUDE.md: only add full integration tests when developing huge features — in general add small inline `insta` snapshots tests

max-sixty:
This was fixed in PR #5562 (merged November 2025). The query now correctly generates `DISTINCT ON (id, name) *` as expected.

PR #5562, opened 2025-11-17
fix: DISTINCT ON to include wildcard when projection is empty

max-sixty:
@claude this was generated from https://github.com/PRQL/prql/issues/5556#issuecomment-3543441822

I notice the test uses `duckdb`, but the issue was for postgres. Why is that? Write a test like @fredericp 's example, make the changes to this PR

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -56,7 +56,8 @@ The test commands above with `--accept` will fill in the result automatically.
   - Use `insta::assert_snapshot!` for compact, readable test assertions
   - Fast to run, easy to review in PRs
 
-- **Use integration tests** (`prqlc/tests/integration/queries/*.prql`) only when:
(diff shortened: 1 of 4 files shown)


max-sixty:
@claude this was generated from https://github.com/PRQL/prql/issues/5556#issuecomment-3543441822

I notice the test uses `duckdb`, but the issue was for postgres. Why is that? Write a test like @fredericp 's example, make the changes to this PR

max-sixty:
@claude this was generated from https://github.com/PRQL/prql/issues/5556#issuecomment-3543441822

I notice the test uses `duckdb`, but the issue was for postgres. Why is that? Write a test like @fredericp 's example, make the changes to this PR

max-sixty:
@claude fix please; don't return until after CI passes

PR #5613, opened 2025-12-12
style: ran `cargo clippy --allow-dirty --fix -- -Wclippy::use_self`

adamnemecek:


max-sixty:
great! if we can fix the fmt error...

PR #5630, opened 2025-12-20
fix: Fix invalid MSSQL SQL when DISTINCT and FETCH are combined

### LINKED ISSUE #5628: When doing a FETCH with SELECT DISTINCT with MsSql, ORDER BY (SELECT NULL) causes a syntax error
### What happened?

The provided PRQL results in the following error

```
ORDER BY items must appear in the select list if SELECT DISTINCT is specified
```

### PRQL input

```elm
from t | take 100| group {this.`District`} (take 1) | select {this.`District`}
```

### SQL output

Compiled with `prqlc compile -t sql.mssql`

```SQL
SELECT
  DISTINCT "District"
FROM
  t
ORDER BY
  (
    SELECT
      NULL
  ) OFFSET 0 ROWS
FETCH FIRST
  100 ROWS ONLY
```

### Expected SQL output

```SQL

```

### MVCE confirmation

- [x] Minimal example
- [x] New issue

### Anything else?

_No response_

max-sixty:
## Summary

- Fixes invalid MSSQL SQL generated when combining `SELECT DISTINCT` with `FETCH` (pagination)
- The issue was that `ORDER BY (SELECT NULL)` was used, but MSSQL requires ORDER BY items to appear in the SELECT list when DISTINCT is used
- The fix extracts the first column from the projection for use in ORDER BY when DISTINCT is present

Fixes #5628

## Test plan

- [x] Added test case `test_mssql_distinct_fetch()` covering the reported issue
- [x] Verified non-MSSQL dialects (PostgreSQL, MySQL, SQLite) are unaffected
- [x] All existing tests pass

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/prqlc/prqlc/src/sql/gen_query.rs
+++ b/prqlc/prqlc/src/sql/gen_query.rs
@@ -218,10 +218,19 @@ fn translate_select_pipeline(
             })
         }
         if order_by.is_empty() {
+            // When DISTINCT is used, MSSQL requires ORDER BY items to appear
+            // in the SELECT list. Use the first column from the projection
+            // instead of (SELECT NULL).
+            let order_expr = is_distinct
+                .then(|| first_expr_from_projection(&projection))
+                .flatten()
+                .unwrap_or_else(|| {
+                    sql_ast::Expr::Value(
+                        sql_ast::Value::Placeholder("(SELECT NULL)".to_string()).into(),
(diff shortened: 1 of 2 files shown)


lukapeschke:
Great, thanks @max-sixty :tada:

PR #5712, opened 2026-03-15
feat: Add support for date.to_text for BigQuery

segv:


prqlc/prqlc/src/sql/dialect.rs
```diff
@@ -21,6 +21,41 @@ use strum::VariantNames;
 
 use crate::{Error, Result};
 
+/// Convert a chrono format `Item` back to its strftime string representation.
+fn chrono_item_to_strftime(item: &Item) -> String {
```
segv, line 25: This is used in the BigQuery error message to make it easy to see exactly which part of the format is breaking the query:

```
    Error:
       ╭─[ :6:33 ]
       │
     6 │       d_str = (d | date.to_text "%-m/%d/%Y")
       │                                 ─────┬─────
       │                                      ╰─────── format specifier `%-m` is not supported for BigQuery
    ───╯
```

otherwise it wouldn't be clear if the culprit was %-m, %d, or %Y.

I have only added this to the BigQuery error message to keep the changes in this PR to a minimum. More than happy to make this change for all dialects if you're ok with this.

prqlc/prqlc/src/sql/dialect.rs
```diff
@@ -695,6 +759,135 @@ mod tests {
         )
         ");
     }
+
+    // -- chrono_item_to_strftime tests --
+
+    #[test]
+    fn chrono_item_to_strftime_numerics_zero_pad() {
```
segv, line 766: I am not convinced these tests add a lot of value and personally I would drop 90% of them and just leave a few in as spot checks.

However the coverage tool wants high coverage, so I've just gone ahead and covered every code path. Happy to drop this if that's ok.

max-sixty:
thank you @segv ! overall this seems very reasonable.

no stress on the tests; I agree re the narrow focus on coverage. but no harm in them either.

PR #5721, opened 2026-03-20
feat: add 'date.now' function for current timestamp support

### LINKED ISSUE #5720: [Feature Request] Support native `date.now` function in standard library to reduce S-string usage
#### Description
Currently, many temporal functions, including `date.now`, are not natively supported in PRQL's standard library. To use them, users must resort to S-strings like `s"now()"` or `s"CURRENT_TIMESTAMP"`.

While S-strings provide a powerful escape hatch, they have several drawbacks:

1. **Poor Readability**: They break the visual flow of PRQL code.

2. **Maintenance**: They make the query dialect-dependent, defeating one of PRQL's core strengths (abstraction over SQL dialects).

3. **Developer Experience**: Writing `s"now()"` feels less intuitive than a native function call.

#### Proposed Change
I propose adding a native `date.now` function to the PRQL standard library. This would allow users to write:
```
derive [
  created_at = date.now
]
```

Instead of:

```
derive [
  created_at = s"now()"
]
```
#### Implementation Idea
The `date.now` function can be added to the stdlib.prql and mapped to the appropriate SQL equivalent based on the target dialect (e.g., `NOW()` for PostgreSQL/MySQL, `CURRENT_TIMESTAMP` for others).

I am willing to contribute to this by implementing the `date.now` function in the standard library if the maintainers agree with this direction.

Additional Context
As PRQL matures, moving frequently used functions from S-strings to the standard library will significantly improve the overall user experience and type-safety potential.

sunyeongchoi:
Closes https://github.com/PRQL/prql/issues/5720

Add `date.now` function for current timestamp

Adds a `date.now` function to the PRQL standard library that returns the current timestamp, enabling time-based filtering and comparisons.
```
from events
filter event_time < date.now
```


sunyeongchoi:
I’ve enabled support for this syntax: `filter event_time < now`.
But I’m wondering, which one aligns better with our project’s current style, `now` or `date.now`?

max-sixty:
yes, `date.now` would be much better, thank you for suggesting!

web/book/src/reference/stdlib/date.md
```diff
@@ -1,7 +1,5 @@
 # Date functions
 
-These are all the functions defined in the `date` module:
```
max-sixty, line 3: can we restore this? or was there a reason to remove?

web/book/src/reference/stdlib/README.md
```diff
@@ -27,7 +27,7 @@ from employees
 derive {
   gross_salary = (salary + payroll_tax | as int),
   gross_salary_rounded = (gross_salary | math.round 0),
-  time = s"NOW()",  # an s-string, given no `now` function exists in PRQL
+  time = date.now,  # current timestamp; or use s"NOW()" for a dialect-specific s-string
```
max-sixty, line 30: 👍

web/book/src/reference/stdlib/date.md
sunyeongchoi: Oh It's my mistake. I'll restore this. Thank you.

max-sixty:
thanks!

PR #5726, opened 2026-03-24
feat: add 'date.diff' function for date diff support

### LINKED ISSUE #5725: [Feature Request] Support native `date.diff` function in standard library to reduce S-string usage
#### Description
Currently, calculating the difference between two dates in PRQL requires using S-strings with dialect-specific SQL functions like DATEDIFF, TIMESTAMPDIFF, or DATE_DIFF. For example:
```
derive [
  days_elapsed = s"DATEDIFF(day, {start_date}, {end_date})"
]
```

While S-strings provide a powerful escape hatch, they have several drawbacks:

1. **Poor Readability**: They break the visual flow of PRQL code.
2. **Maintenance**: They make the query dialect-dependent, defeating one of PRQL's core strengths (abstraction over SQL dialects).
3. **Developer Experience**: Writing s"DATEDIFF(...)" is error-prone and feels less intuitive than a native function call.

#### Proposed Change
I propose adding a native `date.diff` function to the PRQL standard library. This would allow users to write:
```
derive {
  days_elapsed = date.diff day @2021-01-01 @2021-01-02
}
```

Instead of:

```
derive {
  days_elapsed = s"DATEDIFF(day, DATE '2021-01-01', DATE '2021-01-02')"
}
```

I am willing to contribute to this by implementing the `date.diff` function in the standard library if the maintainers agree with this direction.

sunyeongchoi:
Closes https://github.com/PRQL/prql/issues/5725

Add `date.diff` function for difference between dates

```
derive {
  days_elapsed = date.diff day @2021-01-01 @2021-01-02
}
```


max-sixty:
thank you!

PR #5736, opened 2026-03-25
fix: correct SQLite div_i formula for small dividends

### LINKED ISSUE #5735: SQLite div_i (integer division //) produces wrong results when |dividend| < |divisor|
## Bug Description

The SQLite implementation of integer division (`//` operator) in `std.sql.prql` produces incorrect results when the absolute value of the dividend is less than the absolute value of the divisor (i.e., when the correct result should be 0).

## Reproduction

```
from t | select { x = a // b }
```

Compiled with `--target sql.sqlite`, this generates:

```sql
SELECT
  ROUND(ABS(a / b) - 0.5) * SIGN(a) * SIGN(b) AS x
FROM
  t
```

## Wrong Results (verified in SQLite)

| a | b | Formula result | Expected |
|---|---|---------------|----------|
| 1 | 2 | -1 | 0 |
| -1 | 2 | 1 | 0 |
| 3 | 7 | -1 | 0 |
| -3 | 7 | 1 | 0 |

The bug affects **all** integer division cases where `|a| < |b|`.

## Root Cause

The formula `ROUND(ABS(a / b) - 0.5)` is used to emulate `FLOOR(ABS(a / b))` since SQLite lacks a `FLOOR` function. However, when `a` and `b` are integers, SQLite's `/` performs integer division (truncation towards zero), so `a / b = 0` when `|a| < |b|`. Then:

- `ABS(0) - 0.5 = -0.5`
- `ROUND(-0.5) = -1` (SQLite rounds away from zero)

This gives `-1` instead of the expected `0`.

## Suggested Fix

Replace the SQLite `div_i` formula in `prqlc/prqlc/src/sql/std.sql.prql`:

```
# Before (broken):
let div_i = l r -> s"ROUND(ABS({l:11} / {r:12}) - 0.5) * SIGN({l:0}) * SIGN({r:0})"

# After (correct):
let div_i = l r -> s"CAST(ABS({l:11} * 1.0 / {r:12}) AS INTEGER) * SIGN({l:0}) * SIGN({r:0})"
```

The fix forces float division with `* 1.0`, then uses `CAST(... AS INTEGER)` which truncates towards zero in SQLite -- exactly the semantics needed for integer division.

## Environment

- PRQL version: 0.13.12 (main branch)
- Affected dialect: SQLite only

queelius:
## Summary

- Fixes the SQLite `div_i` formula which produced wrong results (e.g., `-1` instead of `0`) when `|dividend| < |divisor|`
- Replaces `ROUND(ABS(l/r) - 0.5)` with `CAST(ABS(l * 1.0 / r) AS INTEGER)` for correct truncation-towards-zero semantics
- Adds a regression test for this case

Fixes #5735

## Root Cause

The old SQLite formula `ROUND(ABS({l} / {r}) - 0.5) * SIGN({l}) * SIGN({r})` broke when both operands were integers and `|l| < |r|`:

1. SQLite integer division: `1 / 2 = 0`
2. `ABS(0) - 0.5 = -0.5`
3. `ROUND(-0.5) = -1` (SQLite rounds away from zero)
4. Result: `-1` instead of expected `0`

Verified with actual SQLite: the old formula gave wrong results for `1//2`, `-1//2`, `3//7`, `-3//7`, etc.

## Fix

Force float division with `* 1.0`, then use `CAST(... AS INTEGER)` which truncates towards zero in SQLite -- exactly matching integer division semantics.

## Test plan

- [x] Verified fix produces correct results for all test cases in actual SQLite
- [x] Added `test_sqlite_integer_division` regression test
- [x] All 449 existing prqlc tests pass
- [x] Updated `compileall` snapshot for arithmetic test

Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/prqlc/prqlc/src/sql/std.sql.prql
+++ b/prqlc/prqlc/src/sql/std.sql.prql
@@ -419,7 +419,7 @@ module sqlite {
   let div_f = l r -> s"({l} * 1.0 / {r:12})"
 
   @{binding_strength=100}
-  let div_i = l r -> s"ROUND(ABS({l:11} / {r:12}) - 0.5) * SIGN({l:0}) * SIGN({r:0})"
+  let div_i = l r -> s"CAST(ABS({l:11} * 1.0 / {r:12}) AS INTEGER) * SIGN({l:0}) * SIGN({r:0})"
 
   # Text functions
   module text {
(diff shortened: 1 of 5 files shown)


max-sixty:
thanks @queelius !

PR #5741, opened 2026-03-26
feat: add 'date.to_start_of_interval' function for truncating values to interval boundaries

### LINKED ISSUE #5734: [Feature Request] Support for arbitrary time-binning in ClickHouse (e.g., `toStartOfInterval`)
I'd like to use functions like ClickHouse's `toStartOfInterval` natively in PRQL without having to rely on s-strings:

```
from events
derive bin = (timestamp | to_start_of_interval 15min)
group bin (aggregate [count])
```

However, `toStartOfInterval` is a function specific to the ClickHouse dialect and is not available in other dialects supported by PRQL. In this case, would it be appropriate for me to implement `to_start_of_interval` as a PRQL function? What approach would you recommend for handling such dialect-specific functions?

I’m considering three options for this:
1. Implementing `to_start_of_interval` for all supported dialects.
2. Implementing it only for ClickHouse and raising an error for other dialects that don't support it.
3. Adding it under a specific namespace, such as `ch.to_start_of_interval`, to mark it as a ClickHouse-specific function.

sunyeongchoi:
Closes https://github.com/PRQL/prql/issues/5734

Adds a `to_start_of_interval` function to provide native truncating values to interval boundaries capabilities in PRQL.
```
from events | select (event_time | date.to_start_of_interval 15 minute)
```


max-sixty:
how about using some form of `round` or `floor` here?

either literally or at least using the term as part of the name?

sunyeongchoi:
@max-sixty 
Thanks for the feedback.
I want to make sure I understood correctly. Are you proposing we use a term like 'floor' in the name (e.g., date.floor_interval) to make the floor logic more explicit?

max-sixty:
> I want to make sure I understood correctly. Are you proposing we use a term like 'floor' in the name (e.g., date.floor_interval) to make the floor logic more explicit?

yes, something like this!

@prql-bot maybe you could do a survey of what is supported by other languages, around "rounding" to regular intervals

sunyeongchoi:
@max-sixty 

What do you think of `date.floor` suggested by prql-bot? if you're okay with it, I'll go ahead and rename it.
```
from events | select (event_time | date.floor 15 minute)
```

max-sixty:
sorry I missed this! I think the `.floor` is good, if you happen to see this!

sunyeongchoi:
@max-sixty Thanks! As you suggested, I changed the name to `date.floor`

max-sixty:
@prql-bot looks like we have a lot of `insta` changes here — could you open a new PR which rewrites our insta snapshots to the latest version, and then hopefully this will become a more managable diff?

max-sixty:
ok now merge and reconcile any other non-essential diffs

Issue #5747, opened 2026-03-26
Implement prql functions to reduce s-string usage


sunyeongchoi:
various prql native function implementations are required to reduce the use of s-strings.
I am going to submit PRs for various function implementations for this issue.

PR #5748, opened 2026-03-27
feat: add 'array_agg', 'max_by' function

sunyeongchoi:
Issue https://github.com/PRQL/prql/issues/5747

### array_agg
* Aggregates column values into an array
```
from tracks
aggregate {names = array_agg name}
```

### max_by
* Returns the row with the maximum value for a specified column
```
from tracks
aggregate {longest_name = max_by name milliseconds}
```


snth:
Hi @sunyeongchoi ,

Thank you for the PRs! I'm busy catching up on resolving some of the stale PRs and then will hopefully get to reviewing these soon.

max-sixty:
the array functions require much more thought and language design. I think we can add functions for scalar functions, but we should not add array functions through this process

hope that's OK @sunyeongchoi ! I would welcome more scalar functions

sunyeongchoi:
Understood! I agree that aggregation functions need more architectural discussion. I'll close this PR and focus on the scalar functions as you suggested. Thanks for the feedback!

PR #5749, opened 2026-03-27
feat(cli)!: Extract CLI binary into separate prqlc-cli crate

snth:
Split the prqlc crate into a pure library (prqlc) and a CLI binary (prqlc-cli). This fixes the structural MSRV problem where CLI-only dependencies (clap, clio, notify, etc.) forced prqlc's MSRV higher than needed, blocking Dependabot PRs.

Key changes:
- prqlc: Remove `cli` feature, add `display` feature gating ariadne+anstream as optional deps. Library is now pure with no binary target.
- prqlc-cli: New binary crate with all CLI code moved here. Binary name remains `prqlc` via [[bin]] config. Depends on prqlc with `display` and `serde_yaml` features.
- error_message.rs: Split composed() — location always computed via new offset_to_line_col() helper (no ariadne needed), display rendering behind #[cfg(feature = "display")].
- semantic/reporting.rs: Remove dead label_references()/Labeler code.
- utils/mod.rs: Gate color functions behind `display` feature with no-op fallback.
- CI: Update build-prqlc action to build prqlc-cli package, upgrade cargo-msrv, update release features.
- Install via: cargo install prqlc-cli


max-sixty:
we went back & forth on this a bunch in the past. I can see the benefits of both, but I def don't think we should change back now; which requires a large number of external-facing updates. what are the specific MSRV issues?

snth:
I hear that but I've tried to organise it so that library consumers face no updates. It's only the prql-cli crate that changes, i.e. the install command changes to `cargo install prqlc-cli` but the binary is still called prqlc. Library consumers like Clickhouse and Databend shouldn't have to change anything. I'll still check that. Trying to get it to pass tests atm.

snth:
MSVR is currently at 1.75. There are 3 pending updates that would move this up to 1.85. They are all in packages only used in the CLI though. Hence with this split the prqlc library could stay at 1.75 while the CLI can follow whatever its dependencies require.

Overall I think this is much cleaner as well. Compiler library consumers shouldn't need to depend on the cli-only dependencies. I know the 'default_features = false' flag addresses that somewhat, but my understanding is that this doesn't work for the MSVR.

max-sixty:
can we just bump the MSRV though?

otherwise the proposed change would means that anyone with the CLI installed would need to change their installation (for a binary with currently slow updates! the only update is breaking :) )

I _also_ think that it adds complication to have different names for the binary and the `cargo install` command...

snth:
Yeah, we could totally just bump the MSVR. I thought that would be more breaking but we can tackle this in separate parts. I'll try and get an agent to do some research on what limits are by downstream consumers. I don't quite trust the answer I got from Copilot earlier.

PR #5821, opened 2026-04-22
feat: add oracle_sql dialect support

julien-pinchelimouroux:
Add basic support for Oracle SQL dialect to force quotes on identifiers.


max-sixty:
@prql-bot can you add a short note to the docs etc re having a very early version of Oracle dialect?

PR #5834, opened 2026-04-27
fix(prqlc/deps): loosen chrono constraint from 0.4.44 to 0.4

lukapeschke:
depending on the latest chrono version causes conflicts with other libs relying on older versions, such as polars: https://github.com/pola-rs/polars/blob/693c52035689e150c2c70fc6d48c952c564ec940/Cargo.toml#L44

### DIFF
--- a/prqlc/prqlc/Cargo.toml
+++ b/prqlc/prqlc/Cargo.toml
@@ -58,7 +58,7 @@ prqlc-parser = { path = "../prqlc-parser", version = "0.13.12" }
 
 anstream = { version = "1.0.0", features = ["auto"] }
 ariadne = "0.5.1"
-chrono = "0.4.44"
+chrono = "0.4"
 csv = "1.4.0"
 enum-as-inner = { workspace = true }
 itertools = { workspace = true }


max-sixty:
Heads-up for context: this loosening exposed a transitive-dep ratchet that wasn't visible from `prqlc/prqlc/Cargo.toml` alone — `arrow-arith` (transitive via `connector_arrow → arrow → arrow-arith`) requires `chrono ^0.4.40`, and `duckdb` requires `chrono ^0.4.22`. So `chrono = "0.4"` is below what prqlc actually transitively requires, and `cargo update -Z direct-minimal-versions` (the `test-deps-min-versions` job) refuses to resolve.

It didn't show up in this PR's CI because the `nightly` path filter in `tests.yaml` only matched the workspace-root `Cargo.toml`, not member crates. So `test-deps-min-versions` was skipped on the PR and only failed on the next post-merge run that touched root `Cargo.lock` (#5838's libc bump).

Followed up:
- #5841 set the floor to `chrono = "0.4.40"` (the highest known transitive minimum) — keeps the looser-than-`0.4.44` spirit while staying honest with downstream consumers.
- #5843 widens the nightly path filter to `**/Cargo.toml` so future member-crate dep changes surface this class of regression on the PR itself.

Nothing to do here — just leaving a breadcrumb for anyone revisiting.

> _This was written by Claude Code on behalf of @max-sixty_

PR #5869, opened 2026-05-08
fix: catch CLI compile panic to ensure debug log is generated

kgutwin:
I was troubleshooting an instance where the CLI compiler was panicking due to an `.unwrap()`. I tried to use the `--debug-log` option to generate log output, but nothing was generated, because the panic happened before the log output could be saved to disk.

This PR wraps the main steps of compilation via the CLI inside a `panic::catch_unwind` handler. This still results in the panic being displayed, but allows the CLI to continue through to output the debug log, making it much easier to trace the sequence of events leading up to the panic.


max-sixty:
thanks @kgutwin !

PR #5878, opened 2026-05-11
feat: replace tuple_every with tuple_reduce

kgutwin:
As I was looking at the `tuple_*` functions in the std lib, I noticed that `tuple_every` is a fairly straightforward function which essentially applies an "AND" reduce operation across its single tuple-value parameter. This PR replaces that single-purpose function with `tuple_reduce`, capable of performing a reduce across a tuple using any operator or two-parameter function. 

It behaves very similar to the Python [`functools.reduce()`](https://docs.python.org/3/library/functools.html#functools.reduce) function, in that it takes a function, tuple, and an optional `initial` named parameter. When the `initial` parameter is provided, the first step of the reduction uses the initial value as the first parameter to the reduction operation. If the provided tuple is empty, the initial value will be returned. When the `initial` parameter is omitted, if the tuple is empty, an error is raised. If the tuple has one entry, its value will be returned; otherwise, the reduce operation will proceed as normal.

Example usage:

```
from foo
select {bar, baz, quux}
select { 
  mysum = tuple_reduce initial:0 add {1, 2, 3},
  all_true = tuple_reduce std.and foo.*,
}
```

produces:

```sql
SELECT
  0 + 1 + 2 + 3 AS mysum,
  bar
  AND baz
  AND quux AS all_true
FROM
  foo
```

If we want to preserve `tuple_every` for backwards compatibility, it could be defined in the std lib as:

```
let tuple_every = func list -> <bool> tuple_reduce initial:true std.and list
```

However, I wasn't sure that this was necessary, since apparently only the other std lib functions `intersect` and `remove` use `tuple_every`, and those were easily ported over to use `tuple_reduce`.

### DIFF
--- a/prqlc/prqlc/src/semantic/resolver/transforms.rs
+++ b/prqlc/prqlc/src/semantic/resolver/transforms.rs
@@ -313,8 +313,8 @@ impl Resolver<'_> {
 
                 let mut res = init.clone();
 
-                if let Ok(init_val) = init.kind.into_literal() {
-                    if init_val.to_string().as_str() == "\"__missing\"" {
+                if let ExprKind::Literal(Literal::String(init_val)) = &init.kind {
+                    if init_val == "__missing" {
                         match num_items {
                             0 => return Err(Error::new(Reason::Expected {
                                 who: Some("tuple".to_string()),
--- a/prqlc/prqlc/src/semantic/resolver/transforms.rs
+++ b/prqlc/prqlc/src/semantic/resolver/transforms.rs
@@ -303,18 +303,45 @@ impl Resolver<'_> {
                 .with_span(pattern.span));
             }
 
-            "tuple_every" => {
+            "tuple_reduce" => {
                 // yes, this is not a transform, but this is the most appropriate place for it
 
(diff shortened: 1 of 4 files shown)


kgutwin:
To comment on @prql-bot's observations:

* I am aware of the sentinel-string collision issue; I rejected the option to go with `noresolve.*` because I would like to enable the behavior where an ident is provided as the initial value. For example, the PRQL code

```
let myinit = 4

from foo
select {xyz = tuple_reduce initial:myinit add {1, 2, 3}}
```

will work as you would expect. I think that future support within PRQL for Symbol-type values (that aren't strings or anything else) makes sense, but is out of scope for this PR.

* I'm fine to consider backwards compatibility, but since PRQL has not yet reached 1.0, my understanding is that backwards compatibility is never guaranteed, and in fact preserving a function such as `tuple_every` which has narrow and slightly awkward semantics seems unnecessary. I'm fine to add that to the PR from maintainer feedback though.
* I think the current test coverage is sufficient, especially given the existing tests that cover the set operations.

max-sixty:
agree @kgutwin !

PR #5980, opened 2026-06-04
feat: tuple_map propagates aliases from its input

kgutwin:
This change to `tuple_map` allows it to propagate the aliases from its input tuple to its output.

The PRQL:
```prql
let add_four = func a -> a + 4

from foo
select {x, y}
derive (tuple_map add_four {c = x, d = y})
```

now produces the SQL:

```sql
SELECT
  x,
  y,
  x + 4 AS c,
  y + 4 AS d
FROM
  foo
```

Prior to this pull request, the `x + 4` / `y + 4` columns would have no aliases.


max-sixty:
Thank you @kgutwin !

PR #5985, opened 2026-06-06
chore: update tend workflows (0.0.21 → 0.1.3), fixing bot-token drift

max-sixty:
## What

Regenerates the eight `tend-*.yaml` workflows with `tend@0.1.3` (was `0.0.21`). This resolves a drift where the bot's auth-token secret name disagreed between `.config/tend.yaml` and the generated workflows — a disagreement that would have silently flipped the bot's token on the next nightly regeneration.

## The drift

The committed workflows referenced `secrets.PRQL_BOT_GITHUB_TOKEN` (the org-level token), but `.config/tend.yaml` carries no `bot_token:` override. #5911 dropped `bot_token: PRQL_BOT_GITHUB_TOKEN` as "redundant" — but it wasn't: with the override gone, the secret name falls back to tend's generator default. The committed files still said `PRQL_BOT_GITHUB_TOKEN` only because they hadn't been regenerated since #5911. The next `uvx tend@latest init` (which `tend-nightly` runs) would have rewritten every workflow and flipped the token, buried inside a version-bump PR rather than reviewed on its own.

## Why bump the version rather than just re-run the pinned generator

The generator default is version-dependent, and this is the subtle part:

- **tend `0.0.21`** (the version the files were stamped with) defaults to `secrets.BOT_TOKEN` — a secret that does **not** exist in this repo. Regenerating at the pinned version would have pointed the bot at an empty token and broken all of its automation.
- **tend `0.1.x`** defaults to `secrets.TEND_BOT_TOKEN`, which **does** exist (provisioned 2026-05-28, repo-scoped, least-privilege) and is what `install-tend` creates.

So the clean fix is to bump tend to the latest release (`0.1.3`) and lean on its default — no `bot_token:` override needed, which is exactly the "drop redundant overrides" intent of #5911, just on a version where the default resolves to a secret that actually exists. This also moves the bot off the broad org-wide 2023 `PRQL_BOT_GITHUB_TOKEN` and onto the fresh repo-scoped token — the more secure end state.

## What the bump changes, beyond the token rename

The `0.0.21 → 0.1.3` jump brings a few template improvements alongside the version stamp and `max-sixty/tend@` action pin:

- **tend-triage**: the job now skips issues labelled `tend-outage` (lets a maintainer mute the bot per-issue).
- **tend-mention**: skips undirected comments authored by bots (avoids spinning up no-op sessions on deploy/CI notifications); queue-delay is now measured from `comment.updated_at` rather than `created_at`.
- **tend-review**: drops the explicit `use_sticky_comment: true` input (now the generator default).

Across all eight files: `secrets.PRQL_BOT_GITHUB_TOKEN` → `secrets.TEND_BOT_TOKEN` (22 references), version stamp and action pin → `0.1.3`.

## Non-tend consumers — intentionally left on the org token

Three workflows reference `PRQL_BOT_GITHUB_TOKEN` but are **not** generated by tend, so they are untouched here and keep working: `release.yaml`, `pull-request-target.yaml`, `tests.yaml`. This makes the migration partial by design — the org secret `PRQL_BOT_GITHUB_TOKEN` stays in active use (here and across other PRQL-org repos) and is **not** deleted. Only the tend-generated workflows move to the repo-scoped token.

## Alternative direction

The opposite, one-line fix is to keep the org token: add `bot_token: PRQL_BOT_GITHUB_TOKEN` under `secrets:` in `.config/tend.yaml`. That leaves the workflows byte-identical (no version bump, no token change) and stays on the proven org-wide token — at the cost of not moving the bot onto the least-privilege repo-scoped secret. Flip to this on review if you'd rather not bump tend right now.

## Validation

- `uvx tend@latest init` is idempotent — re-running reproduces exactly this diff and nothing more.
- `uvx tend@latest check` passes branch-protection, bot-permission (`prql-bot` has write), required-secrets (`TEND_BOT_TOKEN` present), and Claude auth. It also flags a **pre-existing** `repo-secret-allowlist` item (org-level `HEX_API_KEY` / `PRQL_BOT_GITHUB_TOKEN` / `SNAPCRAFT_STORE_CREDENTIALS` broadly available to PR-triggered workflows) — unrelated to this change and already being addressed (#5910).
- The `max-sixty/tend@0.1.3` git tag exists, so the action pin resolves.

> _This was written by Claude Code on behalf of max_


max-sixty:
@prql-bot fix conflicts

max-sixty:
> and I'm scoped to pushing within this repo's org only.

where is this info?

max-sixty:
Closing — superseded by work already on `main`.

This PR set out to fix the bot-token drift (`PRQL_BOT_GITHUB_TOKEN` in the workflows vs. the override dropped from `.config/tend.yaml` in #5911). That drift is already resolved on `main`: #5912 bumped tend `0.0.21 → 0.0.24` and #5962 bumped `0.0.24 → 0.1.2`, regenerating the workflows. All eight `tend-*.yaml` files on `main` now use `secrets.TEND_BOT_TOKEN` at tend 0.1.2, with no `bot_token:` override (the 0.1.x default) — exactly the convergence this PR aimed for.

This branch was built on a pre-#5912 base. After merging `main`, its only remaining net change was a routine `0.1.2 → 0.1.3` bump (skip-undirected-bot-comments guard in tend-mention, dropped `use_sticky_comment` in tend-review, version pins) — which `tend-nightly` regenerates on its own via `uvx tend@latest init`. Nothing left worth landing by hand.

> _This was written by Claude Code on behalf of max_

Issue #4751, opened 2024-07-17
Aliases on relations


aljazerzen:
@snth recently [expressed an opinion](https://github.com/PRQL/prql/pull/4750#issuecomment-2233234666) that relation aliases in PRQL are mistake and should be removed

By relational aliases we mean the name `x` in following queries:

```elm
from x = employees

...

join x = managers
```
Currently, relational aliases have effect of assigning a name to a relational variable within the query:
```elm
from x = employees
select {x.name}
```

@snth argues that relational aliases are a carryover from SQL and should be removed from PRQL because
a) it is not a fundamental part of Relational Algebra, and
b) breaks the local nature of PRQL transforms and puts in global dependencies between different pipeline steps.

max-sixty:
Would there be a different way of creating aliases (e.g. `let`)? How would self-joins work?

aljazerzen:
I would tend to agree here, but for different reasons.

Global dependencies between transforms of the pipeline make the queries non-composable, but relational aliases don't really expose these aliases as global "query variables". By this I mean that the following is not valid PRQL:

```elm
from x = employees
select {age}
select {age, x.name}  # error: unknown name x.name
```

Relational aliases are instead encoded into the resulting type of the pipeline. It has the same effect as wrapping the tuple in the relation into another tuple.

```elm
from employees    # type: [{id = int, name = text}]
select {x = this} # type: [{x = {id = int, name = text}}]
select {x.name}   # <-- this is a tuple field lookup
```

In this understanding of PRQL, it is hard to imagine how `x = ...` would convert type `[{...}]` into `[{x = {...}}]`. What implies that `x` should sink into the array here? Why is it not converted into `{x = [{...}]}`?

So in this view, relation aliases are very weirdly specific behavior, which I would prefer PRQL not to have.

richb-hanover:
@snth and @aljazerzen write:

> b) breaks the local nature of PRQL transforms and puts in global dependencies between different pipeline steps.

I saw manifestations of this recently - I was joining several tables, some of which had the same column names, and I had trouble knowing whether to use a relation alias, and/or which one to use in subsequent select statements. I ultimately just started "trying stuff" and got it to work, but that's a sign of a less-than-perfect language design. 

I also saw a comment about `this` and `that` above: they'll work sort of OK for two tables, but I can see it getting more complicated with three or more.

BUT... If relation aliases were to go away, how would this be expressed?

```
from a=apple
join b=banana (id)
[...]

PR #6020, opened 2026-06-18
fix: allow a column to share its name with the source relation

lukapeschke:
A derived/selected column whose name matched the source relation (e.g. `from sales.sales | derive {sales = ...}`) clobbered the relation's input namespace, which holds the inference template and redirect used to resolve every other column — so all sibling columns failed with "Unknown name".

Nest the shadowing column under a reserved `NS_SHADOWING_COL` key inside the input namespace instead of overwriting it. Leaf access (`this.sales`) resolves to the shadowing column with a clean emitted name, while qualified access (`sales.col`) and column inference still go through the preserved relation namespace. Also fix `deduplicate_select_items`, which conflated a table qualifier with a same-named column alias and dropped the aliased column.


lukapeschke:
@max-sixty This one was trickier to fix than I first thought, because it's easy to introduce a regression here :grimacing:  which is why there are so many new tests

I believe the change is now robust, but if another human could have a look I'd feel safer :smile:

prqlc/prqlc/src/semantic/module.rs
```diff
@@ -526,4 +562,162 @@ mod tests {
...
+    // and drops the alias. Documented so the behavior is intentional rather than
+    // incidental.
+    #[test]
+    fn test_column_shadows_relation_name_forward_reference() {
+        use insta::assert_snapshot;
+
+        // `x2 = this.bar` resolves to the relation `bar` (the shadowing column
+        // is only added later in the tuple), expanding to `*` and losing `x2`.
+        assert_snapshot!(crate::tests::compile(
+            "from bar | derive { x2 = this.bar, bar = this.a }"
```
kgutwin: I'm not sure this makes sense. If I as a user wrote `from bar | derive { x2 = this.bar }` I would not expect the generated SQL to be `SELECT *, * FROM BAR`. I would be expecting that `this.bar` would be referring to a column inside the `bar` relation named `bar`, so the generated SQL ought to be `SELECT *, bar AS x2 FROM bar`. This is what's generated if the derive references any other name -- `from bar | derive { x2 = this.baz }` results in `SELECT *, baz AS x2 FROM bar`.

kgutwin: I guess maybe the broader question is, why does `this.bar` resolve to the parent relation `bar`? It seems that specifying `this.anything` should always resolve `anything` to _only_ within the namespace of the current relation and never to the parent namespace. Although maybe I'm missing some currently established bit of PRQL functionality...

kgutwin: There's no sense in adding a test to the test suite that encodes bad behavior -- that's why this question is relevant against this PR. I'm fine with this test if the behavior is intended but would suggest dropping the test if we'd like to fix this in the future.

That said, what design intent is met by adding the name of the current relation to the `this` namespace? Is there a reason why I would want to refer to `this.bar` and have that reflect a relation rather than a column?

lukapeschke: @kgutwin Thank you for your review :) The test was mostly added to flag existing behavior that I'll try to address later. If you want me to, I can delete the test, or put the correct expected output in it and skip it with an associated issue number ?


Regarding your question:

> That said, what design intent is met by adding the name of the current relation to the this namespace? Is there a reason why I would want to refer to this.bar and have that reflect a relation rather than a column?

I believe this is a side effect of a bare `this` meaning `*` to have the `count this` resolve to `COUNT(*)`: https://prql-lang.org/book/reference/syntax/keywords.html#this--that

I grepped through teh docs a little and `this.bar` referring to a "everything in the `bar` relation" is not documented so I'd be in favor of removing that behavior. We need to keep supporting `count this` though since it's in the docs. 

Also, I'm not sure what a bare `this` should mean in other contetxts ? for example what should the behavior be for `from bar | select this` ? Could be `SELECT * FROM bar`, could be an error where we suggest using `select this.*` or explicit column names instead.

I'm happy to help on that but the design is not my decision to make, so I think the work on it should happen in another PR.

Please let me know what you think the behavior should be. I'd be happy over input from other maintainers as well :smile:

lukapeschke: @kgutwin @max-sixty Any opinion on this ? I'm hitting another similar bug where there is a confusion between a table and a column:

```
from `formula_1`.`results`
join side:left (from `formula_1`.`status` | derive {`statusId_right` = this.`statusId`} | select {this.`statusId_right`, this.`status`}) (this.`statusId` == that.`statusId_right`)
sort {this.`status`}
```

I get a `table instance cannot be referenced directly` error on the final sort

kgutwin: @prql-bot PR #6045 changes the sequencing of name lookups to prioritize "direct hits over redirects". Does that PR resolve this bug report?

kgutwin: @lukapeschke I recommend removing this one test and filing your "similar bug" as a separate issue report.

The query `from status | sort {this.status}` is a fairly straightforward minimal reproduction of the issue -- I think it's the same underlying cause as your more practical query.

kgutwin:
This "shadowing" approach appears to be applicable only to `from sales | derive {sales = ...}` type situations. This doesn't fix the analogous `from sales | select {sales}` query, which still results in `SELECT * FROM sales` rather than `SELECT sales FROM sales`. 

I'm fine with this -- the two problems are distinct, in that the first one comes into play when a new column is created and has the potential to mess with name resolution downstream, while the second one is more a result of the confusion currently surrounding name resolution within PRQL as a whole. However, the "shadowing" strategy is a bit of a bandaid as it sort of patches over the problem rather than resolves it. Really, we do want PRQL to be consistent enough that the same logic is followed in both cases, and shadowing ought not to be necessary.

I wish that the real solution to both these problems would be one where `this` does not in fact contain "nested tuples" (embedded references to the source relations) but instead `this` being a flat representation of the flat column space at any given point in the pipeline. However, I don't want to insist on this solution at the moment, since there's a lot of history on this topic that is hard to parse through (buried in many GitHub issues, but #4751 is a good starting point). And, as you point out, it might also touch other parts of expected behavior -- hence why it's good to include these extra new tests.

prqlc/prqlc/src/semantic/module.rs
lukapeschke: Done, I synced the branch with main and removed the test. I filed a separate issue for the `from status | sort {this.status}` at #6065

lukapeschke:
Thanks a lot for taking the time for explanations and extra context! 

I agree that this is more a bandaid than a root cause fix, but I'm not familiar enough with PRQL's discussion history and with how PRQL's internals work precisely to have an idea on how to address this, and what exactly `this` should contain. I'll definitely read #4751 and related issues :)

max-sixty (commented):
The current head looks sound: it preserves the relation namespace, gives the derived shadowing column leaf precedence, keeps qualified and wildcard access intact, and fixes projection deduplication by keying on the full qualified path. The requested forward-reference test is gone, #6065 tracks the broader name-collision behavior, and the resulting documentation snapshot is updated.

The branch now conflicts with `main`, so I can't approve or merge this head. Please resolve the conflict and let CI run again; if the resulting diff stays the same, this should be ready. I left @kgutwin's review thread for them to resolve, since their requested change has been applied.

> _This was written by Codex on behalf of max-sixty_

PR #6037, opened 2026-06-24
feat: add `by:name` to `append` transform

### LINKED ISSUE #5165: Should `append` sometimes or always create `UNION ALL BY NAME`?

kgutwin:
Closes #5165.

This PR adds the named argument `by` to the `append` transform, supporting `by:name` and `by:position`.

```
prql target:sql.duckdb

from foo | append by:name bar
```

produces:

```sql
SELECT
  *
FROM
  foo
UNION
ALL BY NAME
SELECT
  *
FROM
  bar
```

A review of documentation for currently supported dialects suggests that BigQuery, Snowflake, and DuckDB support `UNION ALL BY NAME`.

### DIFF
--- a/prqlc/prqlc/src/ir/pl/extra.rs
+++ b/prqlc/prqlc/src/ir/pl/extra.rs
@@ -90,7 +90,10 @@ pub enum TransformKind {
         range: Range,
         pipeline: Box<Expr>,
     },
-    Append(Box<Expr>),
+    Append {
+        by: AppendBy,
+        bottom: Box<Expr>,
+    },
     Loop(Box<Expr>),
 }
 
@@ -113,6 +116,12 @@ pub enum JoinSide {
     Left,
     Right,
     Full,
+}
+
+#[derive(Debug, PartialEq, Eq, Clone, Serialize, Deserialize, JsonSchema)]
+pub enum AppendBy {
+    Name,
+    Position,
 }
 
 impl Expr {
--- a/prqlc/prqlc/src/ir/pl/fold.rs
+++ b/prqlc/prqlc/src/ir/pl/fold.rs
@@ -256,7 +256,10 @@ pub fn fold_transform_kind<T: ?Sized + PlFold>(
             with: Box::new(fold.fold_expr(*with)?),
             filter: Box::new(fold.fold_expr(*filter)?),
         },
-        Append(bottom) => Append(Box::new(fold.fold_expr(*bottom)?)),
+        Append { by, bottom } => Append {
+            by,
+            bottom: Box::new(fold.fold_expr(*bottom)?),
+        },
         Group { by, pipeline } => Group {
             by: Box::new(fold.fold_expr(*by)?),
             pipeline: Box::new(fold.fold_expr(*pipeline)?),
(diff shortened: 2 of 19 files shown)


kgutwin:
@max-sixty @prql-bot why is tend-review failing?

max-sixty:
thanks @prql-bot . please re-run now

kgutwin:
I'm actually going to step back from this PR for the moment, as it's getting much more complex than I had initially anticipated.

I started working on a "fallback" implementation for `by:name`, when the dialect does not support `UNION BY NAME`. If we do know the full set of columns for both top and bottom, we should be able to generate a PQ structure that resembles what would be generated by the PRQL:

```
from foo
select { x, y, z = null }
append (from bar | select { x = null, y, z })
```

However, following that path made me realize that the base implementation (even without the fallback) requires updating the lineage process, which also had the previous assumption that both top and bottom relations had the same number of columns. I started working on that challenge, but got stuck. The code in the PR right now, given the following PRQL:

```
prql target sql:duckdb

from foo
select {x, y}
append by:name (from bar | select {y, z})
```

generates the SQL:

```sql
SELECT x, y, z FROM foo
UNION ALL BY NAME
SELECT y, z FROM bar
```

I'm not quite following why that `z` is showing up in the first `SELECT`, and I've run out of time to chase this down now. Perhaps I will pick it up again later...

max-sixty:
Thanks for exploring this, @kgutwin. Since you’ve stepped back for now and the current draft still needs a deeper lineage/lowering design before it can produce correct SQL, I’m closing it to keep the queue clear. Please reopen this PR—or start a fresh one—if you pick the work back up; the investigation and trace here remain useful.

> _This was written by Codex on behalf of @max-sixty_

PR #6046, opened 2026-06-26
feat: alternate PRQL-native implementation of `append by:name`

### LINKED ISSUE #4724: append doesn't check column, column order matters

kgutwin:
See #5165 for background discussion.

This PR adds a PRQL-native implementation of `append by:name` which is dialect-independent. This is a counterpart (but not exclusive to) PR #6037 which would add dialect-specific support for `UNION ALL BY NAME`. 

This PR has three main categories of changes:

1. The `std._append_by_name` function was added, and the internal definition of `append` was extended to parse the `by:` named argument. When `by:name` is provided, it generates a `FuncCall` expression to invoke `std._append_by_name`.
2. Support was added in `semantic/resolver/expr.rs` for wildcard expansion of relations via their lineage information. This seemed to be necessary in order to implement `_param.bottom.*` within the `std._append_by_name` function.
3. Two new tuple functions were added: `tuple_uniq` and `tuple_reverse`. I ended up not needing to use `tuple_reverse` but I left it in anyway in hopes that someone will find it useful. `tuple_uniq` takes a tuple which may have duplicate entries (by alias or by Ident name) and returns either the earliest or latest seen instance of a given name, depending on the value of the `take:` named argument.

Updates to the documentation are still pending, and there's also a potential question related to lineage that I may work on exploring. 

Closes #4724


kgutwin:
@max-sixty This PR should be ready for your review -- thanks!

max-sixty:
hi @kgutwin !

I had a look, would like to spend more time looking, but also trust you so don't feel I need to understand every line.

I did lean on an LLM to poke at parts, a few cases we found below. It's possible that we are missing something though! Let me know if so...

---

**1. An unnamed column is silently dropped.**

```prql
from a
select {x, x + 1}
append by:name (from b | select {x, y})
```

compiles to (note `x + 1` from `a` has vanished):

```sql
SELECT x, NULL AS y FROM a
UNION ALL
SELECT x, y FROM b
```

Plain `append` keeps it as `x + 1 AS y`, so the by-name path loses data the by-position path preserves.

**2. Unknown schemas silently fall back to a position-union.**

```prql
from a
append by:name (from b)
```

gives plain `SELECT * FROM a UNION ALL SELECT * FROM b`. The `by:name` is ignored, so you get exactly the position semantics it was meant to replace, with no indication.

**3. One side defined and one not gives a confusing internal error.**

```prql
from a
select {x, y}
append by:name (from b)
```

errors with `cannot append two relations with non-matching number of columns` / `top has 2 columns, but bottom has 3`, which exposes `_append_by_name`'s internal `top`/`bottom` params and a column count the query never mentions.

All three trace to one gap: nothing checks that both relations have fully-defined, named columns before delegating to `_append_by_name`. The docs note it "only works if the set of columns on both sides is fully defined," but the failure mode is silent-wrong or confusing rather than a clear message. The relation's `ty` already carries enough to detect a wildcard or unnamed field on either side; erroring there with something like `append by:name requires both relations to have fully-defined, named columns` would close it.

kgutwin:
Sounds fine, I added some more input validation to the `append by:name` code branch. Hopefully when the dialect-specific implementation is completed, these input requirements can be lifted (the SQL engine can easily handle wildcards and unnamed columns, since it is fully aware of both halves of the union).

kgutwin:
I looked at the additional lines that are flagged by codecov. They all appear to be error handling paths that are probably never reachable due to upstream checks -- at least, I see no obvious way to write tests that exercise those errors. 

I don't plan on removing the trace calls that I added. Tracing through name resolution and tuple/relation processing at the lowering stage is quite challenging since PRQL has no facility for "debug print" of this info during compilation, so any extra information included would be helpful to me or future developers.

@max-sixty ok to merge?

PR #6104, opened 2026-07-24
feat: Add simple enumerations to type system

### LINKED ISSUE #4501: Named Unions as function parameters
### What's up?

From the discussion on #4499, we propose allowing functions to take parameters that are typed by named unions (aka enums, #4038).

The core syntax would look like this:

```
type InvoiceStatus = (
  Paid = 0 ||
  Unpaid = 1 ||
  Canceled = 2 ||
)

let filter_status = func status <InvoiceStatus> tbl <relation> -> (
  filter (this.status == status) tbl
)

from invoices
filter_status InvoiceStatus.Paid
```

If the type of the function parameter can be resolved, it should be possible to drop the explicit reference to the type name when specifying the parameter value.

```
from invoices
filter_status Paid
```

This would be handled by prepending the parameter type to the name resolution order, so `InvoiceStatus` would be searched first, followed then by the remaining namespace scopes.

The same would also apply for named parameters:

```
let filter_status = func status <InvoiceStatus>:Paid tbl <relation> -> (
  filter (this.status == status) tbl
)

from invoices
filter_status # defaults to InvoiceStatus.Paid

# or
from invoices
filter_status status:Unpaid

# or, explicitly
from invoices
filter_status status:InvoiceStatus.Unpaid
```

With this, the `join` transformation in the standard library would have the updated definition:

```
type join_side = (
  inner = "inner" ||
  outer = "outer" ||
  left = "left" ||
  right = "right" ||
)

let join = func
  `default_db.with` <relation>
  condition <bool>
  side <join_side>:inner
  tbl <relation>
  -> <relation> internal join
```

A similar update would also be made for `from_text format:...`

---

As an aside, it would also be worth considering a similar feature of named union namespace resolution for typed relations, when those are implemented. I could see something like this working very well:

```
from invoices (customer_id = text || amount = float || status = InvoiceStatus)
filter status == Paid
# or, from #2930
filter (match this.status {
  Paid => true,
  Unpaid { amount } => amount < 10.0,
  _ => false,
})
```

kgutwin:
Resolves #4501.

This PR adds a new `Enum` type which takes a "literal tuple" (a tuple with all fields named and having literal values). This type can be used to represent simple explicit enumerations of a database column or a function parameter.

In addition, special-case handling of name resolution was added to function parameters having been assigned an `Enum` type. In this case, the `Enum` module namespace is preferentially searched prior to any other namespaces. This allows for functions such as `join` or `tuple_uniq` to have their option parameters (such as `side`) be associated with an `Enum` and be fully syntax backwards compatible.

Unfortunately, the discussion in #4501 that centered around the use of Named Unions to implement enumerations was not able to be followed, since named unions were removed in PR #5111. The design of the "simple enumeration" was carefully chosen such that when the type system is enhanced in the future to once again support sum types and named unions, the existing syntax can be maintained for backwards compatibility reasons yet have its implementation updated to align with sum types.


vanillajonathan:
The syntax in this PR is:
```
type InvoiceStatus = enum {
    Paid = 0,
    Unpaid = 1,
    Cancelled = 2
}
```

I think the syntax in other languages such as Rust, Python and C# is little nicer.

In Rust it would be a little cleaner:
```rust
enum InvoiceStatus {
    Paid = 0,
    Unpaid = 1,
    Canceled = 2,
}
```

But maybe it needs the syntax in the PR to be able to know that `enum` is not a pipeline function such as `from` and `filter`.

kgutwin:
The error message change is a function of the fact that now `join side:xyz` could theoretically be resolved to `xyz` as defined by a statement like `let xyz = JoinSide.inner`. This was the underlying motivation for the feature as described in #4499. I think the error message as it is in the PR is the most descriptive, rather than the old message which would no longer be accurate.

kgutwin:
@vanillajonathan :
> In Rust it would be a little cleaner:
> 
> ```rust
> enum InvoiceStatus {
>     Paid = 0,
>     Unpaid = 1,
>     Canceled = 2,
> }
> ```
> 
> But maybe it needs the syntax in the PR to be able to know that `enum` is not a pipeline function such as `from` and `filter`.

I decided that the easiest path to implementation here was to reuse the existing `type Abc = ...` parser syntax and to just add a new type specifier. I don't think there's anything inherently at issue with the parser to support the `enum Abc { ... }` syntax, but my chumsky chops aren't that great ;) nevertheless, if you think it's worth the effort, I can give it a swag.

kgutwin:
@vanillajonathan I was able to implement the cleaner enum syntax in the parser, so the syntax in this PR is now:

```prql
enum Status {
  a = 1,
  b = 2,
}
```

None of the tests failed so I think we're all set.

PR #6125, opened 2026-08-03
refactor: Minor refactoring in resolver subsystem

kgutwin:
I've been reviewing the resolver code in hopes of proposing fixes for a few bugs or other inconsistencies. During my code review, I found a couple small tweaks that have no effect on the test suite or logic flow -- only code layout or other minor changes.

* Removed some unused code in `reporting.rs` that had previously been used for debug logging prior to the HTML logging system
* Removed some seemingly unnecessary `Box`es within `fold_expr()`
* Decomposed `fold_statements()` into individual `fold_xyz_stmt()` functions so that `fold_statements()` is easier to read

I don't plan on including a changelog entry since none of these changes should have any user-visible effects.


max-sixty:
thanks @kgutwin , looks good!

PR #6155, opened 2026-08-06
fix: move std types into std.types module, disambiguate text and date types

### LINKED ISSUE #6146: std.prql declares `std.text` and `std.date` twice; the type aliases are silently discarded
`std.prql` declares two different things under the name `std.text`, and two under `std.date`:

- [`type text = text`](https://github.com/PRQL/prql/blob/7bd749933dbf6c827a53ecd59b443006e4858b0d/prqlc/prqlc/src/semantic/std.prql#L44) and [`type date = date`](https://github.com/PRQL/prql/blob/7bd749933dbf6c827a53ecd59b443006e4858b0d/prqlc/prqlc/src/semantic/std.prql#L45), among the type primitives.
- [`module text { … }`](https://github.com/PRQL/prql/blob/7bd749933dbf6c827a53ecd59b443006e4858b0d/prqlc/prqlc/src/semantic/std.prql#L230) (`lower`, `upper`, `trim`, …) and [`module date { … }`](https://github.com/PRQL/prql/blob/7bd749933dbf6c827a53ecd59b443006e4858b0d/prqlc/prqlc/src/semantic/std.prql#L247) (`to_text`, `trunc`, …).

The module def comes later and replaces the type decl, so the modules win and the two `type` lines are dead — `std.text` and `std.date` are not reachable as types:

```
$ prqlc compile 'let f = func x <std.text> -> x
from t
derive {y = f "a"}'
Error: expected a type, but found Module(Module { names: {"contains": …
```

Bare `<text>` still works, because the lexer resolves the primitive type keywords directly rather than going through `std`. So nothing is broken for users today; the two `type` declarations are just silently discarded.

**Why it matters.** It blocks a resolver cleanup. `fold_module_def_stmt` currently replaces whatever declaration sits under the module's name. A PR in flight changes it to *merge* when the existing declaration is also a module, which fixes silently-dropped declarations across files. The natural companion — erroring when the existing declaration is **not** a module, so that `let x = 1` followed by `module x { … }` stops silently dropping `x`, matching how `let`/`type` duplicates are already reported — can't land while std.prql relies on the replacement. With that check in place, loading the standard library fails with `duplicate declarations of std.text`.

**The question is what `std.text` should mean**, which is a language decision rather than a resolver one:

1. **Drop the `type text` / `type date` lines.** They're already inert, so this is behaviour-preserving, and it makes `std.text` unambiguously the module. But it leaves the primitives list in std.prql inconsistent — `int`, `float`, `bool`, `time`, `timestamp` would keep their `type` aliases while `text` and `date` wouldn't.
2. **Rename the modules** (`std.str`? `std.text_fns`?) so the type aliases survive as types. Breaking change for anyone writing `std.text.lower` explicitly.
3. **Let a name hold both a type and a module**, resolved by context. Largest change; the resolver would need per-namespace lookup.

Option 1 is the cheapest and matches today's observable behaviour, but it's worth a maintainer's call before I write it. Once whichever option lands, the non-module half of the duplicate check in `fold_module_def_stmt` can follow.

kgutwin:
Resolves #6146.

Moves all current `type` declarations in `std.prql` into a dedicated `module types { ... }`. To preserve backwards compatibility, also adds the `std.types` ident to the list of redirects on the root module and adjusts a hard coded `std.relation` reference to be `std.types.relation`.

Note that this does change "ambiguous name" error messages when the reference is to either the `text` or `date` names. Since those names will now resolve to both `std.text` and `std.types.text`, the error includes all three potential options (plus the probably expected choice of `this.text`). To me, the proper fix here is not to stick with the old error message (giving an extra unhelpful name choice doesn't make the underlying issue any worse) but rather to line up for an improvement to the type system that could potentially filter out ambiguous names that don't have the expected type.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@
 
 **Language**:
 
+- Move all standard types into a dedicated `std.types` submodule in order to
+  disambiguate between `type text` - `module text` and `type date` -
+  `module date`. (@kgutwin, #6155)
+
 **Features**:
 
 **Fixes**:
--- a/prqlc/prqlc/src/semantic/module.rs
+++ b/prqlc/prqlc/src/semantic/module.rs
@@ -35,6 +35,7 @@ impl Module {
                 Ident::from_name(NS_THIS),
                 Ident::from_name(NS_THAT),
                 Ident::from_name(NS_PARAM),
+                Ident::from_path(vec![NS_STD, "types"]),
                 Ident::from_name(NS_STD),
             ],
         }
--- a/prqlc/prqlc/src/semantic/resolver/stmt.rs
+++ b/prqlc/prqlc/src/semantic/resolver/stmt.rs
@@ -108,7 +108,7 @@ impl super::Resolver<'_> {
 
         if def.name == "main" {
             def.ty = Some(Ty::new(TyKind::Ident(Ident::from_path(vec![
-                "std", "relation",
+                "std", "types", "relation",
             ]))));
         }
 
--- a/prqlc/prqlc/src/semantic/std.prql
+++ b/prqlc/prqlc/src/semantic/std.prql
@@ -37,27 +37,29 @@ let not = expr<bool> -> <bool> internal std.not
 
 # Types
 
-## Type primitives
-type int = int
-type float = float
-type bool = bool
-type text = text
-type date = date
-type time = time
-type timestamp = timestamp
-type `func` = func
+module types {
+  ## Type primitives
+  type int = int
+  type float = float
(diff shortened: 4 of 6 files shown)


kgutwin:
Updated snapshots and added changelog entry.

Regarding testing -- my judgement call is that no additional tests to codify the existence of `std.types` are necessary. The existing behavior of the parser to directly resolve the `date` and `text` words as keywords would make testing this particular behavior hard; testing any other part of this change (such as moving `std.relation` to `std.types.relation`) is already covered implicitly by the test suite.

PR #6172, opened 2026-08-09
refactor!: remove the `sql.glaredb` target (seems no longer maintained)

eitsupi:
It seems development of GlareDB ended last year.
Since it was never tested after the fully rewritten, I think its removal makes sense.


eitsupi:
@prql-bot Please update the integration tests snapshots.
(I don't want to run that heavy job. Also, port conflicts and other issues are a hassle.)

max-sixty (commented):
Thanks @eitsupi !

